PR #10, opened 2018-02-02
Prepro

erlendcognite:
- Add preprocessing module
- Add unit tests for preprocessing module
- Add preprocessing module to restructured text file for docstring docs generation

### DIFF
--- a/cognite/preprocessing.py
+++ b/cognite/preprocessing.py
@@ -19,8 +19,8 @@ def _merge_dataframes(df1, df2):
     if df2 is None:
         return df1
 
-    return pd.merge(df1, df2, on='timestamp', how='outer')\
-        .sort_values(by='timestamp')\
+    return pd.merge(df1, df2, on='timestamp', how='outer') \
+        .sort_values(by='timestamp') \
         .reset_index(drop=True)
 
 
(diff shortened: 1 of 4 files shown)


cognite/preprocessing.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Adds NaN to timestamps with missing values.
+
+    Args:
+        dataframes (list(pandas.DataFrame)):  Input dataframes.
+
+    Returns:
+        pandas.DataFrame: Dataframes merged into one with evenly spaced intervals.
+    """
+    df_merged = reduce(lambda x, y: _merge_dataframes(x, y), dataframes)
```
johhat: Pandas Concat is an alternative here I think. Might be faster.
https://pandas.pydata.org/pandas-docs/stable/generated/pandas.concat.html

cognite/preprocessing.py
```diff
@@ -0,0 +1,161 @@
...
+        tuple: tuple containing:
+            pandas.DataFrame: Dataframe with columns containing NaN values removed.
+            numpy.array: Array of bools indicating which columns were kept.
+    """
+    df_copy = dataframe.set_index('timestamp')
+    selected_columns_mask = df_copy.notnull().all().values
+    return dataframe.dropna(axis=1, how='any'), selected_columns_mask
+
+
+def normalize(dataframe):
```
johhat: This will fail if one of the columns is constant. Something we should let the user know? Only matters if zero-variance columns have not been removed.

PR #20, opened 2018-03-20
Speedup

erlendcognite:
Parallelize get_datapoints and get_datapoints_frame to increase download speed.

### DIFF
--- a/cognite/_utils.py
+++ b/cognite/_utils.py
@@ -72,6 +72,8 @@ def datetime_to_ms(dt):
     epoch = datetime.utcfromtimestamp(0)
     return int((dt - epoch).total_seconds() * 1000)
 
+def round_to_nearest(x, base):
+    return int(base * round(float(x)/base))
 
 def granularity_to_ms(time_string):
     '''Returns millisecond representation of granularity time string'''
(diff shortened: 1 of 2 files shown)


cognite/_utils.py
```diff
@@ -123,6 +126,30 @@ def _get_first_datapoint(tag, start, end, api_key, project):
...
+    elif start is None:
+        start = time_now - _time_ago_to_ms('1w-ago')
+
+    if isinstance(end, datetime):
+        end = datetime_to_ms(end)
+    elif isinstance(end, str):
+        end = time_now - _time_ago_to_ms(end)
+    elif end is None:
+        end = time_now
+
```
queCognite: do we have default value for 'start' as well? Afaik, it is 1w-ago for the public API.

cognite/_utils.py
```diff
@@ -157,13 +187,12 @@ def _print_progress(self):
...
     def _get_start_end(self, tag_ids, start, end):
-        from cognite.timeseries import get_latest
         first_timestamp = float("inf")
         for tag in tag_ids:
             if isinstance(tag, str):
-                first_datapoint = _get_first_datapoint(tag, start, end, self.api_key, self.project)
+                first_datapoint = get_first_datapoint_ts(tag, start, end, self.api_key, self.project)
             else:
-                first_datapoint = _get_first_datapoint(tag['tagId'], start, end, self.api_key, self.project)
+                first_datapoint = get_first_datapoint_ts(tag['tagId'], start, end, self.api_key, self.project)
```
queCognite: maybe make 'tagId' as a constant? It will be changed in the new public API version. They are trying to get away from 'tagId' concept.

PR #26, opened 2018-03-24
Add code coverage

psalaberria002:
coverage.py is used for creating reports.
Cobertura reports are uploaded to Jenkins and codecov.io.

Changes in the Jenkinsfile:
- Formatting
- Add configmap and secret for codecov
- Make use of pipenv so coverage and yolk run successfully.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,21 +2,25 @@
 
 podTemplate(
     label: 'jnlp-cognite-sdk-python',
-    containers: [containerTemplate(name: 'python',
-                                   image: 'python:3.6.4',
-                                   command: '/bin/cat -',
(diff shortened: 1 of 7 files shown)


Jenkinsfile
```diff
@@ -30,18 +34,24 @@ podTemplate(
                 sh("pip3 install pipenv")
             }
             stage('Install dependencies') {
-                sh("pipenv install --system")
+                sh("pipenv install -d")
```
erlendcognite: We had some issues with Jenkins and installing dependencies in a virtualenv, thats why we used the system flag. Seems to work now, so I think we can leave it, but I will come to you if we experience that this fails again.

PR #38, opened 2018-04-09
py2.7 compatibility: Add script removing typehints from source code upon build in Jenkins.

erlendcognite:


erlendcognite:
Yeah, but I prefer this solution to just removing the typehints completely.

PR #42, opened 2018-04-13
Use autopaging flag for relevant queries and return cursor if False.

erlendcognite:
Also update tests.

### DIFF
--- a/cognite/timeseries.py
+++ b/cognite/timeseries.py
@@ -541,8 +541,8 @@ def get_timeseries(prefix=None, description=None, include_metadata=False, asset_
 
         project (str):          Project name.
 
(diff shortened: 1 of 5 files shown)


cognite/timeseries.py
```diff
@@ -544,6 +541,9 @@ def get_timeseries(prefix=None, description=None, include_metadata=False, asset_
 
         project (str):          Project name.
 
+        autopaging (bool):      Whether or to automatically page through results. If set to true, limit will be
```
psalaberria002: Defaults to?

erlendcognite: False ;)

PR #48, opened 2018-04-24
Update get_base_url() to require specifying api version.

erlendcognite:


cognite/v04/assets.py
```diff
@@ -44,7 +44,7 @@ def get_assets(name=None, path=None, description=None, metadata=None, depth=None
     '''
     api_key, project = config.get_config_variables(
         kwargs.get('api_key'), kwargs.get('project'))
-    url = config.get_base_url() + '/projects/{}/assets'.format(project)
+    url = config.get_base_url(api_version=0.4) + '/projects/{}/assets'.format(project)
```
stianlagstad: Nitpick: Maybe `0.4` should be stored in a variable somewhere in the root of `cognite/v04` instead of appearing in  the code multiple times?

(Ignore this if you want)

PR #51, opened 2018-04-25
Adapt to API v0.5 changes

erlendcognite:


cognite/v05/assets.py
```diff
@@ -10,7 +10,7 @@
 import cognite._constants as constants
 import cognite._utils as utils
 import cognite.config as config
-from cognite.v05.data_objects import AssetResponse, AssetDTO
+from cognite.v05.data_objects import AssetResponse, AssetListResponse, AssetDTO
```
psalaberria002: AssetResponse does not exists anymore right?

psalaberria002: nevermind you are right

PR #54, opened 2018-05-09
Change naming and fix bugs in DTO module, add live data generator to timeseries module.

erlendcognite:
- Bug in DTOs: to_pandas() would mutate internal representation of data for certain response objects.
- Add live data generator.
- Breaking Changes:
1. rename data_objects.py -> dto.py
2. rename timeseries paremeter to 'time_series' when input is list of time series and 'name' when input is single time series.
3. rename AssetDTO -> Asset
4. rename EventDTO -> Event
5. rename TimeSeriesDTO -> TimeSeries
6. rename RawRowDTO -> RawRow
7. rename DatapointDTO -> Datapoint

### DIFF
--- a/cognite/v04/dto.py
+++ b/cognite/v04/dto.py
@@ -54,7 +54,7 @@ class CogniteDataObject():
             return self.internal_representation.get('data').get('previousCursor')
 
 
-class RawRowDTO(object):
+class RawRow(object):
     """DTO for a row in a raw database.
 
     The Raw API is a simple key/value-store. Each row in a table in a raw database consists of a
@@ -222,7 +222,7 @@ class TimeseriesResponse(CogniteDataObject):
         return pd.DataFrame(items)
 
 
-class TimeSeriesDTO(object):
+class TimeSeries(object):
     """Data Transfer Object for a timeseries.
 
     Attributes:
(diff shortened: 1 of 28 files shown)


cognite/v05/timeseries.py
```diff
@@ -150,8 +151,8 @@ def _get_datapoints_helper(timeseries, aggregates=None, granularity=None, start=
         list of datapoints: A list containing datapoint dicts.
     '''
     api_key, project = kwargs.get('api_key'), kwargs.get('project')
-    timeseries = timeseries.replace('/', '%2F')
-    url = config.get_base_url(api_version=0.5) + '/projects/{}/timeseries/data/{}'.format(project, timeseries)
+    name = name.replace('/', '%2F')
```
stianlagstad: Is there some sort of urlencode-function you could use here instead? and elsewhere

cognite/v04/dto.py
```diff
@@ -212,7 +213,7 @@ def to_json(self):
 
     def to_pandas(self):
         '''Returns data as a pandas dataframe'''
-        items = self.internal_representation['data']['items']
+        items = deepcopy(self.internal_representation['data']['items'])
```
stianlagstad: Does it make sense to write a test that proves that the mutation bug is fixed?

cognite/v05/timeseries.py
erlendcognite: Yes, there is. Will use.

cognite/v04/dto.py
erlendcognite: Good idea - makes perfect sense.

PR #56, opened 2018-05-11
Implement post multitag datapoints

stianlagstad:
Would like some feedback on this before I do the same with v0.5.

### DIFF
--- a/cognite/_utils.py
+++ b/cognite/_utils.py
@@ -205,7 +205,7 @@ def first_fit(
     '''Returns list of bins with input items inside.'''
 
     # Sort the input list in decreasing order
-    list_items = sorted(list_items, key=lambda x: len(x.datapoints), reverse=True)
+    list_items = sorted(list_items, key=get_count, reverse=True)
 
     list_bins = [Bin(get_count=get_count)]
 
--- a/cognite/_utils.py
+++ b/cognite/_utils.py
@@ -177,7 +177,9 @@ class Bin:
 
     def __init__(self, get_count):
         '''
-        :param get_count: A function that will take an element and get the count of something in it.
+
+        Args:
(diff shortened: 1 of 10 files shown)


cognite/v04/timeseries.py
```diff
@@ -191,6 +191,81 @@ def _get_datapoints_helper(tag_id, aggregates=None, granularity=None, start=None
...
+        list_items=timeseries_with_datapoints_limited,
+        max_size=ul_dps_limit,
+        get_count=lambda x: len(x.datapoints)
+    )
+
+    for bin in timeseries_to_upload_binned:
+        body = {
+            'items': [ts_with_data.__dict__ for ts_with_data in bin]
+        }
+        res = _utils.post_request(url, body=body, headers=headers)
```
KennethSkaar: Shouldn't you check that the result here is OK for each request? Or will it throw an exception on failure?

stianlagstad: I definitely should! Will look at that now.

stianlagstad: @KennethSkaar: Shouldn't the logic within `_utils.post_request` be enough? If an apierror is thrown from there, then the loop here won't continue

KennethSkaar: true, looks like it does the right thing

cognite/_utils.py
```diff
@@ -168,3 +169,64 @@ def __init__(self, tag_ids):
...
+
+
+class Bin:
+    entries: []
+    get_count: Callable
+
+    def __init__(self, get_count):
+        '''
+        :param get_count: A function that will take an element and get the count of something in it.
+        '''
```
erlendcognite: Google style docstrings

PR #57, opened 2018-05-11
Add a depth series interface for use with witsml extractor

KennethSkaar:


cognite/v05/data_objects.py
```diff
@@ -186,6 +186,16 @@ def __next__(self):
             self.counter += 1
             return self.datapoints_objects[self.counter - 1]
 
+class DatapointDepthDTO(object):
```
erlendcognite: Rename this to DepthDatapoint

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,310 @@
...
+from cognite.v05 import timeseries
+from cognite.v05.dto import  LatestDatapointResponse, \
+    DatapointDepth, \
+    TimeSeries, TimeSeriesResponse, Datapoint
+
+
+
+
+
+def post_datapoints(depthseries, depthdatapoints: List[DatapointDepth], **kwargs):
```
erlendcognite: depthseries -> name

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,310 @@
...
+    '''Insert a list of datapoints.
+
+    Args:
+        timeseries (str):       Name of timeseries to insert to.
+
+        datapoints (list[v05.data_objects.Datapoint): List of datapoint data transfer objects to insert.
+
+        offset(int): Offset of index. -1 means append to existsing series, 0 means start at first index
```
erlendcognite: Documentation is generated from docstrings so they should be correct.

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,310 @@
...
+        body = {'items': [dp.__dict__ for dp in datapoints[i:i + ul_dps_limit]]}
+        _utils.post_request(url, body=body, headers=headers)
+        body = {'items': [dp.__dict__ for dp in depthpoints[i:i + ul_dps_limit]]}
+        res = _utils.post_request(urldepth, body=body, headers=headers)
+        i += ul_dps_limit
+
+    return res.json()
+
+
+def get_latest(depthseries, **kwargs):
```
erlendcognite: depthseries -> name

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,310 @@
...
+    '''Returns a LatestDatapointObject containing the latest datapoint for the given timeseries.
+
+    Args:
+        timeseries (str):       The name of the timeseries to retrieve data for.
```
erlendcognite: docstring +1

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,310 @@
...
+                  'items': timeseries}})
+
+def _generateIndexName(depthSeriesName):
+    return depthSeriesName+"_DepthIndex"
+
+def post_depth_series(depth_series: List[TimeSeries], **kwargs):
+    '''Create a new depth series.
+
+    Args:
+        depthseries (list[v05.data_objects.TimeSeries]):   List of time series data transfer objects to create.
```
erlendcognite: depth_series*

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,310 @@
...
+    return res.json()
+
+
+def update_depth_series(depth_series: List[TimeSeries], **kwargs):
+    '''Update an existing time series.
+
+    For each field that can be updated, a null value indicates that nothing should be done.
+
+    Args:
+        timeseries (list[v05.data_objects.TimeSeries]):   List of time series data transfer objects to update.
```
erlendcognite: depth_series*

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,310 @@
...
+            dsdto.isString = None
+            dsdto.unit = None
+        body = {
+            'items': [ts.__dict__ for ts in depth_series]
+        }
+        res = _utils.put_request(url, body=body, headers=headers)
+    return res.json()
+
+
+def delete_depth_series(depthseries, **kwargs):
```
erlendcognite: depth_series

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    res = _utils.delete_request(url, headers=headers)
+    if res == {}:
+        url = config.get_base_url(api_version=0.5) + '/projects/{}/timeseries/{}'.format(project, _generateIndexName(depthseries))
+        res = _utils.delete_request(url, headers=headers)
+
+    return res.json()
+
+def reset_depth_series(depthseries, **kwargs):
```
erlendcognite: depth_series

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,310 @@
+# -*- coding: utf-8 -*-
+"""Depthseries Module
+
+This module mirrors the Timeseries API, but handles pairs of timeseries to emulate depth(or other indexed) data series. It allows you to fetch data from the api and output it in various formats.
```
stianlagstad: Could/should it be called something other than depthseries, since it indeed can be something else? TwoDimensionalTimeseries or something?

stianlagstad: I'm not sure what I think here.. Could/should it be extended so that it can handle timeseries with N different values (not just two)?

KennethSkaar: As it is now it is a depth series only, as we assume a unit of m and name the index accordingly. Its not really a timeseries with 2 different values though either, it has 1 value and 1 index, but the index in this case is not time. 

The trajectory work we will do next will be 2 or more data series, and AFAIK still depth as an index. Not sure yet how much that will just be an extension of this though.

stianlagstad: Let's keep this as depthseries then, and refactor if that makes sense when we look at trajectories. Pretty sure we'll be the only ones using this anyway :)

cognite/v05/depthseries.py
```diff
@@ -41,17 +40,17 @@ def post_datapoints(depthseries, depthdatapoints: List[DatapointDepth], **kwargs
     Returns:
         An empty response.
     '''
-    depthseries = depthseries.replace('/', '%2F')
+    name = name.replace('/', '%2F')
```
erlendcognite: You can use urllib.parse.quoteplus for url encoding

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,316 @@
...
+
+    Keyword Args:
+        api_key (str): Your api-key.
+
+        project (str): Project name.
+
+    Returns:
+        An empty response.
+    '''
+    name = name.replace('/', '%2F')
```
erlendcognite: just minor nitpick: you should use urllib.parse.quoteplus to url encode these names

cognite/v05/depthseries.py
```diff
@@ -0,0 +1,316 @@
...
+        api_key (str):          Your api-key.
+
+        project (str):          Project name.
+
+    Returns:
+        v05.data_objects.LatestDatapointsResponse: A data object containing the requested data with several getter methods with different
+        output formats.
+    '''
+    api_key, project = config.get_config_variables(kwargs.get('api_key'), kwargs.get('project'))
+    name = name.replace('/', '%2F')
```
erlendcognite: +1 urllib.parse.quote_plus

PR #85, opened 2018-07-19
Add project and api_key as argument to  _get_datapoints_user_defined_limit

henrikCognite:
I also made _get_datapoints_helper defauts to get api_key and project from configuration if not spesifed

### DIFF
--- a/cognite/v05/timeseries.py
+++ b/cognite/v05/timeseries.py
@@ -66,7 +66,8 @@ def get_datapoints(name, aggregates=None, granularity=None, start=None, end=None
 
     if kwargs.get("limit"):
         return _get_datapoints_user_defined_limit(
-            name, aggregates, granularity, start, end, limit=kwargs.get("limit"), protobuf=kwargs.get("protobuf")
+            name, aggregates, granularity, start, end, limit=kwargs.get("limit"), protobuf=kwargs.get("protobuf"),
+            api_key=api_key, project=project
         )
 
     diff = end - start
@@ -155,7 +156,7 @@ def _get_datapoints_helper(name, aggregates=None, granularity=None, start=None, 
     Returns:
         list of datapoints: A list containing datapoint dicts.
     """
-    api_key, project = kwargs.get("api_key"), kwargs.get("project")
+    api_key, project = config.get_config_variables(kwargs.get("api_key"), kwargs.get("project"))
     url = config.get_base_url(api_version=0.5) + "/projects/{}/timeseries/data/{}".format(project, quote(name, safe=""))
 
     use_protobuf = kwargs.get("protobuf", True) and aggregates is None
--- a/cognite/v05/timeseries.py
+++ b/cognite/v05/timeseries.py
@@ -149,14 +149,14 @@ def _get_datapoints_helper(name, aggregates=None, granularity=None, start=None, 
         protobuf (bool):        Download the data using the binary protobuf format. Only applicable when getting raw data.
                                 Defaults to True.
 
-        api_key (str):          Your api-key.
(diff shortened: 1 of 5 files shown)


cognite/v05/timeseries.py
```diff
@@ -155,7 +156,7 @@ def _get_datapoints_helper(name, aggregates=None, granularity=None, start=None,
     Returns:
         list of datapoints: A list containing datapoint dicts.
     """
-    api_key, project = kwargs.get("api_key"), kwargs.get("project")
+    api_key, project = config.get_config_variables(kwargs.get("api_key"), kwargs.get("project"))
```
erlendcognite: Is this necessary/correct? User should not use this method directly, and its caller method has already checked for global config.

erlendcognite:
Would be good to have a test making sure this works correctly :)

cognite/v05/timeseries.py
erlendcognite: I mean get_datapoints as the caller method. The user should use configure_session to set client-wide configuration. Should not have to pass it in every request.

henrikCognite: Yes, I also agree that the end user should use `configure_session` if they only work on one project at the time and then it is not necessary to pass `api_key` and `project` in every request such at `get_datapoints`. But if you for some reason work with more that one project and/or api_key in one session, you need to have the ability to pass these argument in methods.

My objection was just that almost all the helper methods/private methods calls `api_key, project = config.get_config_variables(kwargs.get("api_key"), kwargs.get("project"))` even though the caller methods already do this. I changed this method to be more consistent with the other ones.

But I actually think the better approach would be to call `api_key, project = config.get_config_variables(kwargs.get("api_key"), kwargs.get("project"))` in the caller methods, then pass `api_key` and `project` to the helper methods and then call `api_key, project = kwargs.get("api_key"), kwargs.get("project")`. But then I think that the docstring in the helper methods should be 
```
        api_key (str):          Your api-key. Obligatory in this helper method.

        project (str):          Project name. Obligatory in this helper method.
```

erlendcognite: Agree. Great if you make the necessary changes and write tests!

cognite-cicd:
[Tests failed](https://jenkins.cognite.ai/job/github-builds/job/cognite-sdk-python/job/PR-85/2/testReport/).

#### TestDatapointsFrame.test_get_dps_frame_with_config_values_from_argument
```
self = <tests.v05.test_timeseries.TestDatapointsFrame object at 0x7f26230f15c0>
unset_config_variables = ('ZDU1NDM1MTAtMzUxZC00ODA5LWE2YmUtYTk3YzBkNDg5YTc5', 'mltest')

    def test_get_dps_frame_with_config_values_from_argument(self, unset_config_variables):
        res = timeseries.get_datapoints_frame(
            time_series=["constant"],
            start=1522188000000,
            end=1522620000000,
            aggregates=["avg"],
            granularity="1m",
            api_key=unset_config_variables[0],
            project=unset_config_variables[1]
        )
>       assert res

tests/v05/test_timeseries.py:186: 

self = Empty DataFrame
Columns: [timestamp, constant|average]
Index: []

    def __nonzero__(self):
        raise ValueError("The truth value of a {0} is ambiguous. "
                         "Use a.empty, a.bool(), a.item(), a.any() or a.all()."
>                        .format(self.__class__.__name__))
E       ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().

../../.local/share/virtualenvs/ds_cognite-sdk-python_PR-85-W4WGZU75QPY4SU-Ud5vtOnu/lib/python3.6/site-packages/pandas/core/generic.py:1576: ValueError
```

PR #87, opened 2018-07-23
Use context manager to better handle process lifetime.

erlendcognite:


gablank:
Just a quick comment: I believe the calls to Pool.close() and Pool.join() are superfluous, as closing the processes are handled by the context manager.

gablank:
I'm pretty sure the calls to join() are redundant as well. The very first example of using multiprocessing.Pool in the Python documentation does not use it. See [this](https://docs.python.org/3/library/multiprocessing.html) link.

gablank:
Also, if you look at the source code of the Pool class, the context manager calls _terminate() which in turn calls _terminate_pool(). _terminate_pool() joins all spawned processes and threads.

PR #91, opened 2018-08-13
Add data transfer service

erlendcognite:


setup.py
```diff
@@ -1,14 +1,17 @@
-import cognite
+import re
+
 from setuptools import find_packages, setup
 
 packages = find_packages(exclude=["tests*"])
 
+version = version = re.search('^__version__\s*=\s*"(.*)"', open("cognite/__init__.py").read(), re.M).group(1)
```
NilsBarlaug-cognite: Should probably find out what the "correct" way to do this is.

Issue #102, opened 2018-09-07
timeseries.get_datapoints does not support end='now'


oholsen:
'now' is the default for end (undocumented in the SDK, documented in the REST API). 

_utils.interval_to_ms() should support 'now'.

erlendcognite:
Omitting the argument all together will default it to current timestamp. But this is noted, and will be implemented.

oholsen:
Have a PR in progress for this.

PR #105, opened 2018-09-10
#102: Support end="now" in get_datapoints()

oholsen:
BTW, why all the start/end fuss (string -> int) in the SDK? AFIK the REST API supports the same string arguments. It does makes sense to support DateTime objects, but there is no need to map "now", "2w-ago", etc to ms since the epoch.

Let me know if the test could be improved to properly test end="now". As is, it pretty much just checks that the code completes without failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Python SDK to ensure excellent user experience for developers and data scientist
 
 ## Prerequisites
 In order to start using the Python SDK, you need
-- Python3 and pip
+- Python3 (>= 3.3) and pip
 - An API key. Never include the API key directly in the code or upload the key to github. Instead, set the API key as an environment variable. See the usage example for how to authenticate with the API key.
 
 This is how you set the API key as an environment variable on Mac OS and Linux:
--- a/cognite/_utils.py
+++ b/cognite/_utils.py
@@ -143,8 +143,7 @@ def _log_request(method, url, **kwargs):
 
 
 def datetime_to_ms(dt):
-    epoch = datetime.utcfromtimestamp(0)
-    return int((dt - epoch).total_seconds() * 1000)
+    return int(dt.timestamp() * 1000)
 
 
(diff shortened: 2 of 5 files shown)


cognite-cicd:
[Tests failed](https://jenkins.cognite.ai/job/github-builds/job/cognite-sdk-python/job/PR-105/1/testReport/).

#### TestConversions.test_datetime_to_ms
```
self = <tests.test_utils.TestConversions object at 0x7f4b6a3de240>

    def test_datetime_to_ms(self):
        from datetime import datetime
    
        assert utils.datetime_to_ms(datetime(2018, 1, 31)) == 1517356800000
        assert utils.datetime_to_ms(datetime(2018, 1, 31, 11, 11, 11)) == 1517397071000
        assert utils.datetime_to_ms(datetime(100, 1, 31)) == -59008867200000
        with pytest.raises(TypeError):
>           utils.datetime_to_ms(None)

tests/test_utils.py:281: 

dt = None

    def datetime_to_ms(dt):
>       return int(dt.timestamp() * 1000)
E       AttributeError: 'NoneType' object has no attribute 'timestamp'

cognite/_utils.py:146: AttributeError
```

README.md
```diff
@@ -12,7 +12,7 @@ Python SDK to ensure excellent user experience for developers and data scientist
 
 ## Prerequisites
 In order to start using the Python SDK, you need
-- Python3 and pip
+- Python3 (>= 3.3) and pip
```
erlendcognite: Could you add `python_requires='>=3.3'` to setup.py?

erlendcognite:
@oholsen Because we parallelize the downloading of data, so we need to map the string representation to ms in order to split the data into chunks correctly.

README.md
oholsen: Sure!

oholsen: Done!

Issue #117, opened 2018-09-24
Add `includeOutsidePoints`-parameter to get_datapoints


gablank:
The parameter `includeOutsidePoints` of the `/api/0.5/projects/{project}/timeseries/{id}/data`-endpoint seems to be missing in the Python API.

erlendcognite:
Fixed in #118 . Will be available in next release (0.10.4)

Issue #111, opened 2018-09-17
Should handle pagination


f1cognite:
When for example using `get_assets` then maybe the function should not only return the first 1000 elements but automatically return all assets.

erlendcognite:
Fixed in #118 . Will be available in next release (0.10.4)

PR #118, opened 2018-09-28
Implement feature requests #117 and #111

erlendcognite:


erlendcognite:
Implements features requested in #117 and #111

PR #125, opened 2018-10-10
Fix #124

### LINKED ISSUE #124: Bug in the not-documentation of TimeSeries `isStep` parameter
The parameter is documented as `isStep`, but it's actually `is_step`. Since [the DTO](https://doc.cognitedata.com/0.5/#schemaposttimeseriesv2dto) is documented as `isStep`, I don't know what the right solution is.

But I do know it's [not a bug](https://cognitedata.slack.com/archives/C85QG3UBU/p1539089928000100?thread_ts=1539089156.000100&cid=C85QG3UBU) in the documentation!

oholsen:


oholsen:
lost type, will add

erlendcognite:
Have these fixes in #126. Also we use google style docstrings.

PR #147, opened 2018-11-10
Retry logic

erlendcognite:


cognite/_utils.py
```diff
@@ -20,13 +20,17 @@
 log = logging.getLogger("cognite-sdk")
 
 
+def _status_is_valid(status_code: int):
+    return status_code < 400
```
wjoel: There are some others that (for our API) should be retried, like 503 and possibly 500. Maybe 502 as well, if that can happen due to nginx or the Google Load Balancer (GLB) returning it if a container gets killed in the middle of a request.

Here's a useful list: https://stackoverflow.com/a/47841642

erlendcognite: So now it will retry anything >= 400. The list you reference there is a list of status codes that should _not_ be retried (including 500, 502, 503), so should I exclude all of those?

wjoel: It's a good question, probably all 4xx responses should not be retried. A bad request will be a bad request no matter how many times we retry it, so the retries just add traffic on our API and increases the response time for the user if it retries several times in the background before giving an error back.

We have the same issue in the Spark CDP data source, would be nice to get some guidance on this.

I've asked about it in #api: https://cognitedata.slack.com/archives/C85QG3UBU/p1541937580216600

PR #186, opened 2019-01-08
Fixed typo in docs

hermord:
Just a single typo: saw this during the hackathon and it bugged me.

### DIFF
--- a/cognite/client/stable/datapoints.py
+++ b/cognite/client/stable/datapoints.py
@@ -441,7 +441,7 @@ class DatapointsClient(APIClient):
         Args:
             name (str):       Name of timeseries to insert to.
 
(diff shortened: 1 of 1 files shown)


cognite/client/stable/datapoints.py
```diff
@@ -441,7 +441,7 @@ def post_datapoints(self, name, datapoints: List[Datapoint]) -> None:
         Args:
             name (str):       Name of timeseries to insert to.
 
-            datapoints (list[stable.datapoints.Datapoint): List of datapoint data transfer objects to insert.
+            datapoints (list[stable.datapoints.Datapoint]): List of datapoint data transfer objects to insert.
```
erlendcognite: should actually be List with capital L as well. https://docs.python.org/3/library/typing.html#typing.List

PR #188, opened 2019-01-08
Added delete_time_series_by_id

hermord:
Rationale: starting with API `v0.6`, time series names may be non-unique.

There isn't necessarily a huge need for this change right now (afaik posting a time series with a duplicate name is not possible), but I wanted to try this simple change first to get a feel of implementation/testing in `cognite_sdk`.

### DIFF
--- a/cognite/client/experimental/time_series.py
+++ b/cognite/client/experimental/time_series.py
@@ -43,8 +43,7 @@ class TimeSeriesClient(APIClient):
         """
         url = "/timeseries/delete"
         body = {"items": ids}
-        params = {}
-        self._post(url, body=body, params=params)
+        self._post(url, body=body)
 
     def get_time_series_by_id(self, id: int) -> TimeSeriesResponse:
         """Returns a TimeseriesResponse object containing the requested timeseries.
--- a/cognite/client/experimental/time_series.py
+++ b/cognite/client/experimental/time_series.py
@@ -25,6 +25,27 @@ class TimeSeriesClient(APIClient):
     def __init__(self, **kwargs):
         super().__init__(version="0.6", **kwargs)
 
+    def delete_time_series_by_id(self, ids: List[int]) -> None:
+        """Delete multiple time series by id.
+
+        Args:
+            ids (List[int]):   IDs of time series to delete.
+
+        Returns:
+            None
+
+        Examples:
+            Delete a single time series by id::
+
+                client = CogniteClient()
+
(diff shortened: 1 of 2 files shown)


hermord:
Also I didn't add the `mutliple_` prefix, breaking the symmetry with `get_time_series_by_id`. My rationale was that for get-by-id the multiple and singular cases are non-trivially different (hit different endpoints) whereas here the singular case would be just `delete_multiple_time_series_by_id([id])`.

cognite-cicd:
[Tests failed](https://jenkins.cognite.ai/job/github-builds/job/cognite-sdk-python/job/PR-188/3/testReport/).

#### TestTimeseries.test_delete_time_series_by_id
```
self = <test_time_series.TestTimeseries object at 0x7f592cdeb2b0>

    @pytest.fixture(scope="class")
    def created_ts_id(self):
        tso = TimeSeries(TS_NAME)
        res = stable_time_series.post_time_series([tso])
>       yield time_series.search_for_time_series(name=TS_NAME).to_json()[0]["id"]
E       IndexError: list index out of range

tests/test_experimental/test_time_series.py:28: IndexError
```

tests/test_experimental/test_time_series.py
```diff
@@ -19,6 +21,16 @@ def ts_name():
 
 
 class TestTimeseries:
+    @pytest.fixture(scope="class")
+    def created_ts_id(self):
+        tso = TimeSeries(TS_NAME)
```
erlendcognite: This looks good, but there is really no need for the ts_name fixture. Seems it was just lying around from some old code. Just generate the random name in this fixture, and move it outside the test class. Also, then it will not need the class scope.

cognite/client/experimental/time_series.py
```diff
@@ -25,6 +25,27 @@ class TimeSeriesClient(APIClient):
...
+        Examples:
+            Delete a single time series by id::
+
+                client = CogniteClient()
+
+                client.time_series.delete_time_series_by_id(ids=[my_ts_id])
+        """
+        url = "/timeseries/delete"
+        body = {"items": ids}
+        params = {}
```
erlendcognite: Nitpick: params can be left out. No need to pass the empty dict as the underlying requests lib will interpret None as  no parameters.

hermord: Fair, I think I just carried that over from `get_ts_by_id`.

PR #195, opened 2019-01-14
CogniteClient: Assert that api_key matches project

matiashf-cognite:
API keys are project specific, so specifying an API key for a different project will always result in an error at a later stage.

This patch throws an exception as early as possible if the client is badly configured, to help users identify which line of code they should change.

### DIFF
--- a/cognite/client/cognite_client.py
+++ b/cognite/client/cognite_client.py
@@ -39,8 +39,8 @@ class CogniteClient:
     All services are made available through this object. See examples below.
 
     Args:
-        api_key (str): API key
-        project (str): Project. Defaults to project of given API key.
+        api_key (str): API key. If not given, uses the COGNITE_API_KEY environment variable.
+        project (str): Project. If given, verify that the API key belongs to this project.
         base_url (str): Base url to send requests to. Defaults to "https://api.cognitedata.com"
         num_of_retries (int): Number of times to retry failed requests. Defaults to 5.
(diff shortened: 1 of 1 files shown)


matiashf-cognite:
Sorry, bad formatting. I'll reformat with black first.

cognite/client/cognite_client.py
```diff
@@ -112,9 +112,12 @@ def __init__(
 
         self._requests_session = self._requests_retry_session()
 
-        self._project = project
-        if project is None:
-            self._project = self.login.status().project
+        self._project = None  # Allow the LoginClient object to be created
```
erlendcognite: This line is redundant

cognite/client/cognite_client.py
```diff
@@ -112,9 +112,12 @@ def __init__(
...
+        self._project = self.login.status().project
+        if project is not None and project != self._project:
```
erlendcognite: If we choose to check login.status() no matter if they pass project or not, we may as well remove the project parameter and always fetch it using the api-key. The reasoning behind letting the user pass the project parameter was to not render the client useless in case the login endpoint is down.

erlendcognite: (Also that test is failing for this very reason. It passes a dummy base_url to the client, and passes project to avoid invoking login.status(). You can fix it by patching that api call.)

Issue #198, opened 2019-01-15
post_datapoints_frame is undocumented


matiashf-cognite:
`post_datapoints_frame(dataframe)` does not take a time series name as an argument, and does not have an example of how to use the function. @trygvekk mentioned that the name of the data frame is accepted as a data frame header, but he did not have an example.

PR #199, opened 2019-01-15
Expanded post_datapoints_frame docs

dshynkev:
As requested in #198

### DIFF
--- a/cognite/client/stable/datapoints.py
+++ b/cognite/client/stable/datapoints.py
@@ -798,8 +798,7 @@ class DatapointsClient(APIClient):
     def post_datapoints_frame(self, dataframe) -> None:
         """Write a dataframe.
         The dataframe must have a 'timestamp' column with timestamps in milliseconds.
-        The names of the remaining columns should be precisely the names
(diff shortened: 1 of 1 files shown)


cognite/client/stable/datapoints.py
```diff
@@ -796,13 +796,31 @@ def _get_datapoints_frame_user_defined_limit(self, time_series, aggregates, gran
         return df
 
     def post_datapoints_frame(self, dataframe) -> None:
-        """Write a dataframe
+        """Write a dataframe.
+        The dataframe must have a 'timestamp' column with timestamps in milliseconds.
```
erlendcognite: milliseconds since epoch.

cognite/client/stable/datapoints.py
```diff
@@ -796,13 +796,31 @@ def _get_datapoints_frame_user_defined_limit(self, time_series, aggregates, gran
...
+        The names of the remaining columns specify the names of the time series to which column contents will be written.
```
erlendcognite: Maybe also include that the time series to which the data is being posted must already have been created.

Issue #210, opened 2019-01-24
Pretty print for DatapointsQuery


johhat:
Quite useful indeed.

Issue #208, opened 2019-01-24
get_multi_time_series_datapoints unexpectedly mutates items in datapoint_queries - suprising results ensues


johhat:
The items in datapoint_queries are mutated within get_multi_time_series_datapoints. In turn, reusing the same function arguments will lead to unexpected results where only subsets of the requested data is returned.

Example:
![image](https://user-images.githubusercontent.com/8568753/51670001-f7ec7d00-1fc5-11e9-9a44-b8408e729e45.png)

Note: Changing the get_it function to the following makes the behaviour predictable again since the DatapointsQuery-objects are reconstructed at each function call.

![image](https://user-images.githubusercontent.com/8568753/51670160-5ade1400-1fc6-11e9-9b8c-078048219bdf.png)

Issue #207, opened 2019-01-24
get_multi_time_series_datapoints returns an iterator rather than a list


johhat:
The documentation says
https://github.com/cognitedata/cognite-sdk-python/blob/54350e0df653c57594b19fe7d63c23b94c5fb6e7/cognite/client/stable/datapoints.py#L523-L526
But the actual return value is
https://github.com/cognitedata/cognite-sdk-python/blob/54350e0df653c57594b19fe7d63c23b94c5fb6e7/cognite/client/stable/datapoints.py#L576

PR #212, opened 2019-01-24
Fix get_multi_timeseries_datapoints

erlendcognite:
No longer mutates DatapointsQuery object #208 

Pretty print for DatapointsQuery object #210 

Fix return type in documentation #207


erlendcognite:
> Are you going to bump the version?

@f1cognite I was intending to only bump patch version. The only breaking change is removing the to_ndarray() method, but this was completely removed for all other responses in 0.12, and was just forgotten here. So I figured it did not merit a minor bump just for that change.

Issue #226, opened 2019-02-04
Timeseries documentation typo, get_time_series()


datakristy:
![image](https://user-images.githubusercontent.com/46494163/52201416-a072ca80-286b-11e9-8564-bc4f51dbf32c.png)
This example in the newest docs show the function `get_timeseries`, but this function doesn't exist anymore! `get_time_series` is the new function.

The screenshot is the first code example in https://cognite-sdk-python.readthedocs-hosted.com/en/latest/cognite.html

erlendcognite:
Addressed in #227

PR #227, opened 2019-02-04
Update limits

erlendcognite:


cognite/client/cognite_client.py
```diff
@@ -52,7 +52,7 @@ class CogniteClient:
 
                 from cognite import CogniteClient
                 client = CogniteClient()
-                res = client.time_series.get_timeseries()
```
f1cognite: This is a breaking change...? Should we revert this change?

erlendcognite: it's in a docstring. fixes outdated documentation.

erlendcognite: see #226

PR #231, opened 2019-02-08
Add a demo to be run in Google Collab @ Uppsala Univeristy

bradleywhitlock:
The notebook is to be run using google collab rather than jupyter notebook.


erlendvollset:
@bradleywhitlock Is this to be merged in or not?

PR #234, opened 2019-02-11
Fix rawrow argdoc

stianlagstad:


stianlagstad:
FYI: The test failure is due to a change introduced in https://github.com/cognitedata/api/pull/1655, leading to the deleteTables-endpoint responding with 200 even though the table to be deleted doesn't exist. I'll rebuild once that behaviour is back to the way it used to be (Satyajit is on it).

stianlagstad:
Should be fixed in https://github.com/cognitedata/api/pull/1668 . Triggering new build now.

cognite/__init__.py
```diff
@@ -2,4 +2,4 @@
 from cognite.client.exceptions import APIError
 
 __all__ = ["data_transfer_service"]
-__version__ = "0.12.14"
+__version__ = "0.12.15"
```
erlendcognite:
```suggestion
__version__ = "0.12.14"
```

PR #237, opened 2019-02-15
0.13.0

erlendvollset:
Non-Breaking Changes:
- Adds support for getting credentials from thread-local storage (Requirement for model hosting)
- Moves all autopaging functionality to common method on ApiClient class

Breaking Changes:
- Moves all modules under `cognite.client` namespace so that we can have other packages under `cognite` namespace
- Removes data_transfer_service
- Rewrites model hosting client


cognite/client/cognite_client.py
```diff
@@ -83,13 +84,15 @@ def __init__(
         timeout: int = None,
         debug: bool = None,
     ):
+        thread_local_api_key, thread_local_project = self._get_thread_local_credentials()
+
         environment_api_key = os.getenv("COGNITE_API_KEY")
```
NilsBarlaug-cognite: Should there be an `environment_project` as well?

cognite/client/cognite_client.py
```diff
@@ -112,8 +115,8 @@ def __init__(
 
         self._requests_session = self._requests_retry_session()
 
-        self._project = project
-        if project is None:
+        self._project = project or thread_local_project
```
NilsBarlaug-cognite: Maybe this should follow the decision of which api-key to use? Might be annoying (or nice feature?) if you have project in thread local or environment, but then specify api-key as argument (but not project). Then it will not auto fetch project, but instead use thread local or environment - which is not what you expect and a 403 will be puzzling.

cognite/client/experimental/model_hosting/models.py
```diff
@@ -0,0 +1,397 @@
+# -*- coding: utf-8 -*-
```
NilsBarlaug-cognite:
```suggestion
```

cognite/client/experimental/model_hosting/models.py
```diff
@@ -0,0 +1,397 @@
...
+
+        Args:
+            name (str):             Name of model
+            description (str):      Description
+            metadata (Dict[str, Any]):          Metadata about model
+            input_fields (List[str]):   List of input fields the model accepts
+            output_fields (List[str]:   List of output fields the model produces
+
+        Returns:
+            Dict: The created model.
```
NilsBarlaug-cognite:
```suggestion
            experimental.model_hosting.models.ModelResponse: The created Model
```

cognite/client/experimental/model_hosting/models.py
```diff
@@ -0,0 +1,397 @@
...
+        res = self._post(url, body=model_body)
+        return ModelResponse(res.json())
+
+    def list_models(self, limit: int = None, cursor: int = None, autopaging: bool = False) -> ModelCollectionResponse:
+        """List all models.
+
+        Args:
+            limit (int): Maximum number of models to return. Defaults to 250.
+            cursor (str): Cursor to use to fetch next set of results.
+            autopaging (bool): Whether or not to automatically page through all results. Will disregard limit.
```
NilsBarlaug-cognite: This is more of a question for the future moving towards 1.0 and @f1cognite: Maybe pagination should be completely hidden for the user and auto pagination always on? Resulting in just a limit parameter specifying how long the auto pagination will go. And then if the user would like to stream, support generators as well or something.

cognite/client/experimental/model_hosting/models.py
```diff
@@ -0,0 +1,397 @@
...
+            id (int): Create a new version under the model with this id
+            name (str): Name of model version. Must be unique on the model.
+            source_package_id (int):    Use the source package with this id
+            train_source_package_id (int):  Use this source package for training. If omitted, will default to
+                                            source_package_id.
+            metadata (Dict[str, Any]):  Metadata about model version
+            description (str):  Description of model version
+            args (Dict[str, Any]):   Dictionary of arguments to pass to the training job.
+            scale_tier (str):   Which scale tier to use. Must be either "BASIC" or "CUSTOM"
+            machine_type (str): Specify a machiene type Applies only if scale_tier is "CUSTOM".
```
NilsBarlaug-cognite:
```suggestion
            machine_type (str): Specify a machine type Applies only if scale_tier is "CUSTOM".
```

cognite/client/cognite_client.py
erlendvollset: No, I don't think this is necessary since we autofetch the project in case it is not specified. I think it makes it easier to make mistakes which can result in 401s if we have that.

cognite/client/cognite_client.py
erlendvollset: If we were to use environment project, then I sort of agree with you. But since thread-local config will only be used by users who really know what they are doing, normal users will only override the autofetching by specifying the project as a parameter. So I think it is difficult to get in a situation where you get an unexpected 401.

cognite/client/experimental/model_hosting/models.py
erlendvollset: Agree, that would be neat.

cognite/client/__init__.py
```diff
@@ -0,0 +1,4 @@
+from cognite.client.cognite_client import CogniteClient
+from cognite.client.exceptions import APIError
+
+__version__ = "0.13.0"
```
NilsBarlaug-cognite:
```suggestion
__version__ = "0.13.0a1"
```
Or something?

f1cognite:
LGTM but I don't like breaking changes 😄

PR #264, opened 2019-03-03
Fix update timeseries PUT format

f1cognite:


f1cognite:
@erlendvollset Why wasn't this caught by this test? https://github.com/cognitedata/cognite-sdk-python/blob/master/tests/test_stable/test_timeseries.py#L27

erlendvollset:
Because it's not updating any values which are different between snake_case and camelCase. It's updating `unit`.

PR #271, opened 2019-03-05
Expose autopaging in CogniteClient.get

dshynkev:
Not sure if this is acceptable, but it would certainly make our lives easier.

The specific use case that would benefit from this is asset classification where we fetch an entire asset hierarchy for training. We want to fetch types (which are not in v0.5), so we are forced to use `CogniteClient.get`, but we would also like not to reimplement autopaging.


dshynkev:
[Extend CDP helper with types](https://trello.com/c/pngOnnmb/122-extend-cdp-helper-with-types)

PR #287, opened 2019-03-11
Upgrade and lock cognite_logger

NilsBarlaug-cognite:


cognite/client/cognite_client.py
```diff
@@ -16,6 +14,7 @@
 from cognite.client.stable.raw import RawClient
 from cognite.client.stable.tagmatching import TagMatchingClient
 from cognite.client.stable.time_series import TimeSeriesClient
+from cognite.logger import logger
```
erlendvollset: import configure_logger

PR #290, opened 2019-03-12
Fix link to SDK docs in README.md

dragly:


erlendvollset:
The url is correct. Only the page was accidentaly made private. it's fixed now.

dragly:
Ah, okay. Then perhaps the link on doc.cognitedata.com should be updated? Or is there a good reason to have two separate URLs to the same docs?

https://doc.cognitedata.com/quickstart/#the-python-software-development-kit-sdk

The above page links to
https://cognite-sdk-python.readthedocs-hosted.com/en/latest/
While README.md in this repo links to
https://cognite-docs.readthedocs-hosted.com/en/latest/

PR #310, opened 2019-03-25
Use os.path for Windows compatibility

wjoel:


cognite/client/experimental/model_hosting/models.py
```diff
@@ -494,11 +494,11 @@ def upload_artifacts_from_directory(self, model_id: int, version_id: int, direct
             None
         """
         upload_tasks = []
-        root_directory_name = directory.split("/")[-1]
+        root_directory_name = os.path.basename(directory)
         for root, dirs, files in os.walk(directory):
             for file_name in files:
                 file_path = os.path.join(root, file_name)
-                full_file_name = file_path.split("{}/".format(root_directory_name))[-1]
+                full_file_name = os.path.relpath(file_path, root_directory_name)
```
erlendvollset: I think you should remove the root_directory_namr variable and just find the rleative path to `directory`

erlendvollset:
Could you also bump the version to 0.13.2? We will probably have windows users in stavanger today.

PR #320, opened 2019-04-06
Add support for posting and visualizing asset hierarchy

erlendvollset:


cognite/client/api/assets.py
```diff
@@ -411,3 +454,146 @@ def search(
...
+        return unblocked_children
+
+    def get_unblocked_assets(self, limit) -> List[List[Asset]]:
+        unblocked_assets_lists = []
+        unblocked_assets = []
+
+        for asset in self.remaining_assets:
+            if asset.parent_ref_id is None or asset.parent_id is not None:
+                unblocked_assets.append(asset)
+            elif self.ref_id_to_id.get(asset.parent_ref_id) is not None:
```
NilsBarlaug-cognite:
```suggestion
            elif asset.parent_ref_id in self.ref_id_to_id:
```

cognite/client/api/assets.py
```diff
@@ -411,3 +454,146 @@ def search(
...
+        if len(unblocked_assets) < limit:
+            max_descendents = limit - len(unblocked_assets)
+            unblocked_assets.extend(self.get_descendants_unblocked_by_ref_id(unblocked_assets, max_descendents))
+
+        if len(unblocked_assets) > 0:
+            unblocked_assets_lists.append(unblocked_assets)
+
+        for unblocked_assets in unblocked_assets_lists:
+            for unblocked_asset in unblocked_assets:
+                self.remaining_assets.remove(unblocked_asset)
```
NilsBarlaug-cognite: This will probably become too expensive when you have many assets (e.g. doing linear search and linear shift over a million items several thousand times). Maybe make `self.remaining_assets` a set?

cognite/client/api/assets.py
```diff
@@ -411,3 +454,146 @@ def search(
...
+        self.ref_id_to_remaining_children = self.initialize_ref_id_to_remaining_children_map(self.remaining_assets)
+        self.ref_id_to_id = {}
+
+        self.request_queue = queue.Queue()
+        self.response_queue = queue.Queue()
+
+        self.created_assets = AssetList([])
+
+    @staticmethod
+    def validate_asset_hierarchy(assets: List[Asset]) -> None:
```
NilsBarlaug-cognite: Should this method also check for cycles (verify it's a forrest)?

cognite/client/api/assets.py
```diff
@@ -411,3 +454,146 @@ def search(
...
+        for asset in assets:
+            if asset.parent_ref_id in ref_id_to_children:
+                ref_id_to_children[asset.parent_ref_id].append(asset)
+        return ref_id_to_children
+
+    def get_descendants_unblocked_by_ref_id(self, assets: List[Asset], limit):
+        unblocked_children = []
+        for asset in assets:
+            if asset.ref_id in self.ref_id_to_remaining_children:
+                asset_children = self.ref_id_to_remaining_children[asset.ref_id].copy()
```
NilsBarlaug-cognite: Why do we need to copy here?

cognite/client/api/assets.py
```diff
@@ -411,3 +454,146 @@ def search(
...
+                for child_asset in asset_children:
+                    unblocked_children.append(child_asset)
+                    self.ref_id_to_remaining_children[asset.ref_id].remove(child_asset)
```
NilsBarlaug-cognite: Should this rather be moved to the end of `get_unblocked_assets()`?

cognite/client/api/assets.py
NilsBarlaug-cognite: Should `self.ref_id_to_remaining_children` also be updated?

cognite/client/api/assets.py
```diff
@@ -411,3 +454,146 @@ def search(
...
+                    if len(unblocked_children) == limit:
+                        return unblocked_children
+
+                    max_descendents = limit - len(unblocked_children)
+                    unblocked_children.extend(self.get_descendants_unblocked_by_ref_id([child_asset], max_descendents))
```
NilsBarlaug-cognite: Would it be better to add them breadth first? What to we expect to unblock the most assets?

cognite/client/api/assets.py
```diff
@@ -411,3 +454,146 @@ def search(
...
+
+    def get_unblocked_assets(self, limit) -> List[List[Asset]]:
+        unblocked_assets_lists = []
+        unblocked_assets = []
+
+        for asset in self.remaining_assets:
+            if asset.parent_ref_id is None or asset.parent_id is not None:
+                unblocked_assets.append(asset)
+            elif self.ref_id_to_id.get(asset.parent_ref_id) is not None:
+                asset.parent_id = self.ref_id_to_id[asset.parent_ref_id]
```
NilsBarlaug-cognite: Does this violate the constraint of not having `parent_id` and `parent_ref_id` at the same time?

PR #362, opened 2019-05-23
adding type and subtype to event data type

sanderland:
enabling creating event which have a type/subtype (recently added to the API)


erlendvollset:
Data classes are generated from the api spec, so they should not be updated manually. They were just added back into the spec, so I'll update them tomorrow.

PR #364, opened 2019-05-27
Improve handling and documentation of lookups by ids not found.

erlendvollset:


cognite/client/_api/time_series.py
```diff
@@ -37,17 +37,48 @@ def __iter__(self) -> Generator[TimeSeries, None, None]:
         """
         return self.__call__()
 
+    @overload
     def retrieve(
         self, id: Union[int, List[int]] = None, external_id: Union[int, List[int]] = None
     ) -> Union[TimeSeries, TimeSeriesList]:
```
oholsen: Also a None here?

cognite/client/_api/time_series.py
```diff
@@ -37,17 +37,48 @@ def __iter__(self) -> Generator[TimeSeries, None, None]:
...
+    def retrieve(self, id: int = None, external_id: str = None) -> Union[TimeSeries, None]:
+        ...
+
+    @overload
+    def retrieve(self, id: List[int] = None, external_id: List[str] = None) -> TimeSeriesList:
+        ...
+
+    def retrieve(
+        self, id: Union[int, List[int]] = None, external_id: Union[int, List[int]] = None
+    ) -> Union[TimeSeries, TimeSeriesList]:
```
oholsen: Also a None in the return type?

cognite/client/_api/time_series.py
```diff
@@ -37,17 +37,48 @@ def __iter__(self) -> Generator[TimeSeries, None, None]:
...
+            Retrieving time series by id::
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.time_series.retrieve(id=[1,2])
+        """
+        ...
+
+    @overload
+    def retrieve(self, id: int = None, external_id: str = None) -> Union[TimeSeries, None]:
```
dbrattli: Optional[Timeseries]

cognite/client/_api/assets.py
```diff
@@ -59,17 +59,54 @@ def __iter__(self) -> Generator[Asset, None, None]:
...
+            Get assets by external id::
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.assets.retrieve(external_id=["1", "abc"])
+        """
+        ...
+
+    @overload
+    def retrieve(self, id: int = None, external_id: str = None) -> Union[Asset, None]:
```
dbrattli: Optional[int] and Optional[Asset]

cognite/client/_api/raw.py
```diff
@@ -364,7 +364,7 @@ def delete(self, db_name: str, table_name: str, key: Union[str, List[str]]) -> N
             task_unwrap_fn=lambda task: task["json"]["items"], task_list_element_unwrap_fn=lambda el: el["key"]
         )
 
-    def retrieve(self, db_name: str, table_name: str, key: str) -> Row:
+    def retrieve(self, db_name: str, table_name: str, key: str) -> Union[Row, None]:
```
dbrattli: Optional[Row]

cognite/client/_api/events.py
```diff
@@ -67,17 +67,54 @@ def __iter__(self) -> Generator[Event, None, None]:
...
+            Get an event by external id::
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.events.retrieve(external_id="abc")
+        """
+        ...
+
+    @overload
+    def retrieve(self, id: int = None, external_id: str = None) -> Union[Event, None]:
```
dbrattli: Optional[Event]

cognite/client/_api/files.py
```diff
@@ -71,17 +71,54 @@ def __iter__(self) -> Generator[FileMetadata, None, None]:
...
+            Get file meta data by external id::
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.files.retrieve(external_id=["1", "abc"])
+        """
+        ...
+
+    @overload
+    def retrieve(self, id: int = None, external_id: str = None) -> Union[FileMetadata, None]:
```
dbrattli: Optional[FileMetaData]

cognite/client/_api/assets.py
```diff
@@ -59,17 +59,54 @@ def __iter__(self) -> Generator[Asset, None, None]:
         """
         return self.__call__()
 
+    @overload
     def retrieve(
         self, id: Union[int, List[int]] = None, external_id: Union[str, List[str]] = None
```
dbrattli: Must be Union[int, List[int], None]. Anything that can be None must have type None, Optional[T] or Union[T1, T2, None]

erlendvollset:
Ah, tests for py35 are failing. Seems overloading is not supported there. In that case I think it makes more sense to always raise `CogniteNotFoundError`, instead of changing the behaviour to return `None` when only a single id is provided. What do you think, @dbrattli @oholsen ?

erlendvollset:
Alternatively, we could just split it into two methods.

dbrattli:
Yes, I think that is a good idea, since the first overload basically shortcuts the typing so you will not get a type warning if you would happen to use the method in a wrong way. Python is not good at method overloading and the more Pythonic way would be to have separate methods.

PR #385, opened 2019-06-04
Implement .count(), .first(), .latest() on TimeSeries object

erlendvollset:


cognite/client/data_classes/time_series.py
```diff
@@ -73,6 +73,44 @@ def plot(
...
+        """Returns the number of datapoints in this time series.
+
+        This result may not be completely accurate, as it is based on aggregates which may be occasionally out of date.
+
+        Returns:
+            int: The number of datapoints in this time series.
+        """
+        identifier = utils._auxiliary.assert_at_least_one_of_id_or_external_id(self.id, self.external_id)
+        dps = self._cognite_client.datapoints.retrieve(
+            start=0, end="now", aggregates=["count"], granularity="10d", **identifier
```
f1cognite: Why 10?

cognite/client/data_classes/time_series.py
```diff
@@ -73,6 +73,44 @@ def plot(
...
+        return None
+
+    def first(self) -> Optional["Datapoint"]:
+        """Returns the first datapoint in this time series.
+
+        Returns:
+            Datapoint: A datapoint object containing the value and timestamp of the first datapoint.
+        """
+        identifier = utils._auxiliary.assert_at_least_one_of_id_or_external_id(self.id, self.external_id)
+        dps = self._cognite_client.datapoints.retrieve(**identifier, start=0, end="now", limit=1)
```
f1cognite: Are you guaranteed that they are sorted by time 🤔 ?

erlendvollset: Yes.

cognite/client/data_classes/time_series.py
erlendvollset: I guess I just had to pick something. 10d * 10000 dps limit = 273 years, so then we will always get the counts in a single request.

PR #396, opened 2019-06-07
Optimize Datapoints._load()

sanderland:
This is a result of tracking down why a relatively big script was CPU bound with low network. 
The _load function shows up as taking 60-70% of the time. 
This commit does not fully resolve the issue, _load is still the function taking most of the time and will likely require further optimization. Still, it results in a ~33% speed up of the code in question.

An alternative is:
```
			for key in expected_fields:
				data = []
				for dp in dps_object["datapoints"]:            
					data.append(dp.get(key))
				snake_key = utils._auxiliary.to_snake_case(key)
				setattr(instance,snake_key,data)

```
This was found to be slightly slower for len(expected_fields)==2, but the difference is small.

Profiling results with yappi

image

### DIFF
--- a/cognite/client/data_classes/datapoints.py
+++ b/cognite/client/data_classes/datapoints.py
@@ -217,13 +217,12 @@ class Datapoints:
                 snake_key = utils._auxiliary.to_snake_case(key)
                 setattr(instance, snake_key, [])
         else:
-            for dp in dps_object["datapoints"]:
-                for key in expected_fields:
-                    snake_key = utils._auxiliary.to_snake_case(key)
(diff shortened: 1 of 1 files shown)


sanderland:
Using list comprehensions with [] instead of .get results in another significant speedup.
This makes ._load drop out of the top 100 and JSONDecoder.raw_decode the next slowest function at around 10% time taken

sanderland:
Just adding this to clarify the screenshot above is outdated after the other commits.
Profiling results after the additional changes show the _load function is no longer first - the list comprehension is second, taking around 3% total time.
image

PR #407, opened 2019-06-17
New implementation for retrieve_dataframe and addition of query_dataframe

sanderland:
This work was started from trying to get around some of the limitations in query(), as well as the performance of DataPointsList for larger queries. Particularly, calling retrieve_dataframe several times in threads causes major problems with the multiple threadpools getting in each other's way (either in the GIL or the connection pool)

This re-implements retrieve_dataframe to be completely pandas based, and not requiring multiple successive thread pools or getting count aggregates. The strategy is probably applicable to the existing datapoints fetcher, but for now I've stayed away from touching that.

Advantages:
* For single dataframes, it can be significantly faster (usually 25% faster, but up to 8x, see below)
* Allows retrieval of multiple dataframes with same id (as in quality filtering use cases etc)
* Strategy for getting datapoints is somewhat simpler, could also simplify existing code when merged.

Disadvantages:
* Some code duplication for now.
* Concurrency doesn't use existing SDK tools due to adding jobs on the fly

Performance tests (SDK here is a30):
retrieving dataframe of shape (42538, 4) check =  3 / 3
	SDK 1.38s +/- 0.21s
	New 1.08s +/- 0.07s
retrieving dataframe of shape (235584, 4) check =  3 / 3
	SDK 5.25s +/- 0.09s
	New 4.62s +/- 0.11s
retrieving dataframe of shape (2750724, 4) check =  3 / 3
	**SDK 161.30s +/- 9.15s
	New 67.67s +/- 2.14s**
retrieving dataframe of shape (35, 1) check =  3 / 3
	SDK 0.18s +/- 0.01s
	New 0.18s +/- 0.01s
retrieving dataframe of shape (50400, 171) check =  3 / 3
	SDK 16.04s +/- 0.24s
	New 13.93s +/- 0.17s
retrieving dataframe of shape (8, 684) check =  3 / 3
	SDK 5.96s +/- 0.03s
	New 6.41s +/- 0.31s <- slightly worse for many very shallow queries
retrieving dataframe of shape (15946081, 1) check =  3 / 3
	**SDK 914.30s +/- 5.93s
	New 110.46s +/- 12.32s**

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -1,6 +1,7 @@
 import copy
 import math
 import threading
+import time
 from collections import defaultdict, namedtuple
 from concurrent.futures.thread import ThreadPoolExecutor
 from datetime import datetime
@@ -374,8 +375,8 @@ class DatapointsAPI(APIClient):
         end: Union[int, str, datetime],
         aggregates: List[str],
         granularity: str,
-        id: Union[int, List[int]] = None,
-        external_id: Union[str, List[str]] = None,
+        id: Union[int, List[int], Dict[str, Any], List[Dict[str, Any]]] = None,
+        external_id: Union[str, List[str], Dict[str, Any], List[Dict[str, Any]]] = None,
     ):
         """Get a pandas dataframe describing the requested data.
 
@@ -425,7 +426,7 @@ class DatapointsAPI(APIClient):
             query (Union[DataFrameQuery, List[DataFrameQuery]): List of dataframe queries.
 
         Returns:
-            Union[Datapoints, DatapointsList]: A Datapoints object containing the requested data, or a list of such objects.
+            Union[pd.DataFrame, List[pd.DataFrame]]: A Datapoints object containing the requested data, or a list of such objects.
 
         Examples:
 
@@ -906,18 +907,17 @@ class _DFColQuery:
 
 class DataFrameFetcher:
     START_JOBS_PER_THREAD = 2  # how many jobs per worker thread to aim for at the start
-    SLEEP_WAIT_NEW_JOBS = 0.010  # how long to sleep while waiting for new jobs
+    SLEEP_WAIT_NEW_JOBS = 0.01
 
     def __init__(self, client):
         self.np, self.pd = utils._auxiliary.local_import("numpy", "pandas")
         self.client = client
-        self.jobs = []
         self.lock = threading.Lock()
-
-    def fetch(self, dps_queries, max_workers=None):
-        self.nworkers = max_workers or self.client._config.max_workers
-        self._preprocess_queries(dps_queries)
+        self.nworkers = self.client._config.max_workers
         self.idle = [False] * self.nworkers
+
+    def fetch(self, dps_queries):
+        self._preprocess_queries(dps_queries)
 
         with ThreadPoolExecutor(self.nworkers) as p:
             for i in range(self.nworkers):
@@ -952,7 +952,7 @@ class DataFrameFetcher:
                 with self.lock:  # lock here to prevent idle from being inconsistent
                     q = self.jobs.pop()
                     self.idle[tid] = False
-                self.get_datapoints(q, tid)
+                self.get_datapoints(q)
             except IndexError:  # no more jobs - empty not thread safe
                 self.idle[tid] = True
                 if all(self.idle):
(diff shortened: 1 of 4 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -856,3 +890,125 @@ def _get_datapoints(
...
+    SLEEP_WAIT_NEW_JOBS = 0.010  # how long to sleep while waiting for new jobs
+
+    def __init__(self, client):
+        self.np, self.pd = utils._auxiliary.local_import("numpy", "pandas")
+        self.client = client
+        self.jobs = []
+        self.lock = threading.Lock()
+
+    def fetch(self, dps_queries, max_workers=None):
+        self.nworkers = max_workers or self.client._config.max_workers
```
erlendvollset: self.nworkers and self.idle should be instantiated in the constructor.

cognite/client/_api/datapoints.py
```diff
@@ -856,3 +890,125 @@ def _get_datapoints(
...
+            except IndexError:  # no more jobs - empty not thread safe
+                self.idle[tid] = True
+                if all(self.idle):
+                    break
+                time.sleep(self.SLEEP_WAIT_NEW_JOBS)  # wait for other threads to finish, or new jobs to appear
+            except Exception as e:
+                q.exception = e
+
+    def _preprocess_queries(self, dps_queries):
+        self.jobs = []
```
erlendvollset: redundant?

cognite/client/_api/datapoints.py
```diff
@@ -856,3 +890,125 @@ def _get_datapoints(
...
+        new_queries = []
+        for i in range(nparts):
+            qc = copy.copy(query)
+            qc.start = query.start + i * chunk_size
+            qc.end = query.start + (i + 1) * chunk_size
+            new_queries.append(qc)
+        new_queries[-1].end = query.end
+        return new_queries
+
+    def get_datapoints(self, query, tid):
```
erlendvollset: tid is unused

cognite/client/_api/datapoints.py
```diff
@@ -856,3 +890,125 @@ def _get_datapoints(
...
+
+
+class DataFrameFetcher:
+    START_JOBS_PER_THREAD = 2  # how many jobs per worker thread to aim for at the start
+    SLEEP_WAIT_NEW_JOBS = 0.010  # how long to sleep while waiting for new jobs
+
+    def __init__(self, client):
+        self.np, self.pd = utils._auxiliary.local_import("numpy", "pandas")
+        self.client = client
+        self.jobs = []
```
erlendvollset: I would prefer if self.jobs was a Queue, and you add jobs to the queue as you progress. That way you don't have to worry about locking and marking workers as idle.

cognite/client/data_classes/datapoints.py
```diff
@@ -339,3 +339,21 @@ def __init__(
...
+    def __init__(
+        self,
+        start: Union[str, int, datetime],
+        end: Union[str, int, datetime],
+        aggregates: List[str],
+        granularity: str,
+        id: Union[int, List[int]] = None,
+        external_id: Union[str, List[str]] = None,
+    ):
+        self.start = cognite.client.utils._time.timestamp_to_ms(start)
```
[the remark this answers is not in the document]
sanderland: not necessary either, changed.

cognite/client/_api/datapoints.py
sanderland: yes

cognite/client/_api/datapoints.py

cognite/client/_api/datapoints.py
sanderland: Queue vs stack doesn't matter much here, but if we keep this, but the idle has a use, let's discuss.
Replacing the idle logic with something more proper (semaphore?) is an idea, but something for later I think

cognite/client/_api/datapoints.py
```diff
@@ -856,3 +890,125 @@ def _get_datapoints(
...
+            try:
+                with self.lock:  # lock here to prevent idle from being inconsistent
+                    q = self.jobs.pop()
+                    self.idle[tid] = False
+                self.get_datapoints(q, tid)
+            except IndexError:  # no more jobs - empty not thread safe
+                self.idle[tid] = True
+                if all(self.idle):
+                    break
+                time.sleep(self.SLEEP_WAIT_NEW_JOBS)  # wait for other threads to finish, or new jobs to appear
```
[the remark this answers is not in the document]

cognite/client/_api/datapoints.py

cognite/client/_api/datapoints.py
sanderland: see what you mean now regarding queue, using that now and got rid of the job queue-in-a-job queue logic

PR #414, opened 2019-06-21
Add Schedule log.

HMEiding:


cognite/client/experimental/model_hosting/schedules.py
```diff
@@ -22,6 +22,33 @@ class ScheduleCollectionResponse(CogniteCollectionResponse):
     _RESPONSE_CLASS = ScheduleResponse
 
 
+class LogEntry(CogniteResponse):
+    def __init__(self, internal_representation):
+        super().__init__(internal_representation)
+        item = self.to_json()
+        print("hey hey")
+        print(item)
```
erlendvollset: remove these print statements

cognite/client/experimental/model_hosting/schedules.py
```diff
@@ -120,3 +147,17 @@ def delete_schedule(self, id: int) -> None:
         """
         url = "/analytics/models/schedules/{}".format(id)
         self._delete(url=url)
+
+    def get_log(self, id: int) -> ScheduleLog:
+        """Return schedule log by id.
+
```
erlendvollset: Add some more description here about the logs the user can expect to have returned. i.e. "Two logs, one for failed scheduled predictions and one for successful."

cognite/client/experimental/model_hosting/schedules.py
HMEiding: Wow, that was sloppy...

erlendvollset: I do it all the time. We should actually add a check in Jenkins which scans the library for print()s.

PR #423, opened 2019-06-28
Add possibility to disable ssl verifiction.

HMEiding:


cognite/client/utils/_auxiliary.py
```diff
@@ -140,14 +141,18 @@ def get_user_agent():
...
-        warnings.warn(
-            "You are using version {} of the SDK, however version {} is available. "
-            "Upgrade or set the environment variable 'COGNITE_DISABLE_PYPI_VERSION_CHECK' to suppress this "
-            "warning.".format(this_version, newest_version),
-            stacklevel=3,
-        )
+    newest_version = None
+    try:
+        newest_version = utils._version_checker.get_newest_version_in_major_release("cognite-sdk", this_version)
+        if newest_version != this_version:
```
erlendvollset: move the if statement to after the except.

cognite/client/utils/_auxiliary.py
```diff
@@ -14,6 +14,7 @@
 import string
 import warnings
 from decimal import Decimal
+from ssl import SSLError
```
erlendvollset: I'm not sure if this is the correct exception. It may be `from requests.exceptions import SSLError`. Try to find a way to reproduce the error and find where the exception stems from.

cognite/client/utils/_version_checker.py
```diff
@@ -9,17 +9,19 @@ def check_if_version_exists(package_name: str, version: str):
...
     for v in versions:
         if _is_newer_major(v, version):
             major, minor, micro, pr_cycle, pr_version = _parse_version(v)
     return _format_version(major, minor, micro, pr_cycle, pr_version)
 
 
-def get_all_versions(package_name: str):
-    res = requests.get("https://pypi.python.org/simple/{}/#history".format(package_name))
+def get_all_versions(package_name: str, verify_ssl: bool):
+    session = requests.Session()
```
HMEiding: I don't configure the session with HTTPAdapter as in _api_client.py. Is this necessary?

wjoel: I think you can skip the session and just pass the verify flag directly to `requests.get`. Either that or use an existing session - no need to create a session just for this one request.

cognite/client/_api_client.py
```diff
@@ -52,19 +52,23 @@ def _init_requests_session():
...
 
 
 class APIClient:
     _RESOURCE_PATH = None
     _LIST_CLASS = None
 
     def __init__(self, config: utils._client_config.ClientConfig, api_version: str = None, cognite_client=None):
-        self._request_session = _REQUESTS_SESSION
-        self._request_session_with_retry = _REQUESTS_SESSION_WITH_RETRY
+        (self._request_session, self._request_session_with_retry) = _init_requests_session(config.verify_ssl)
```
wjoel: This changes the behavior a bit, instead of having two global requests sessions each client will now have its own two request sessions. Unless there's a reason to, we probably shouldn't change this.

cognite/client/utils/_version_checker.py
HMEiding: Agreed, I'll change this.

cognite/client/_api_client.py
HMEiding: I did it this way because the _init_requests_session() now takes the argument verify_ssl which is not known before the constructor of APIClient is called.

wjoel: @NilsBarlaug-cognite / @erlendvollset: could you think of any reason why having two sessions per client, instead of two sessions globally, might be an issue?

PR #427, opened 2019-07-04
Fix exclusive_to argument in sequences

youwucognite:


cognite/client/experimental/sequences.py
```diff
@@ -177,14 +177,14 @@ class SequenceDataRequest:
 
     Args:
         inclusive_from (int):    Row number to get from (inclusive).
-        inclusive_to (int):      Row number to get to (inclusive).
+        exclusive_to (int):      Row number to get to (exclusive).
```
wjoel: Should probably update https://doc.cognitedata.com/api/0.6/#operation/getSequenceData where it says `Highest row number included. Default: No limit` which sounds inclusive to me, but the argument named `exclusiveTo` so it is exclusive.

eventh: I will make sure that happens, tracked down the location of the description here:
https://github.com/cognitedata/api/blob/master/src/main/java/com/cognite/api/dto/sequences/SequenceDataRequestDTO.java#L13

PR #431, opened 2019-07-04
Updating classes to the new version of the api spec, and temporarily skip failing tests.

HMEiding:


wjoel:
Mention that you're also updating the classes to the new version of the API spec.

PR #432, opened 2019-07-04
Configure SSL verification via environment variable

HMEiding:
Implement new solution after input from Erlend.

### DIFF
--- a/tests/tests_unit/test_cognite_client.py
+++ b/tests/tests_unit/test_cognite_client.py
@@ -5,9 +5,8 @@ import sys
 import threading
 import types
 from multiprocessing.pool import ThreadPool
-from ssl import SSLError
 from time import sleep
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 import pytest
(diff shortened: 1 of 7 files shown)


tests/tests_unit/test_cognite_client.py
```diff
@@ -198,15 +198,23 @@ def test_verify_ssl_enabled_by_default(self, mock_requests, mock_findall):
         assert c._api_client._request_session.verify == True
         assert c._api_client._request_session_with_retry.verify == True
 
+    @pytest.fixture
+    def set_env_disable_ssl(self):
```
wjoel: Did we get this to work?

HMEiding: It works for the assert which tests the verify on line 25 of `cognite.client.utils._version_checker`, which is not part of the two global sessions. For this one, the environment variable in the fixture is set before the call. The two global sessions, however, are set before the environment variable, and I therefore removed those two asserts

PR #437, opened 2019-07-09
with fix to get asset subtree, no longer sorts by depth

cmm16:
Fixes issue where subtree for asset lists would sort on depth which is no longer an attribute of assets

### DIFF
--- a/cognite/client/_api/assets.py
+++ b/cognite/client/_api/assets.py
@@ -280,7 +280,7 @@ class AssetsAPI(APIClient):
                 subtree.
 
         Returns:
-            AssetList: The requested assets sorted topologically.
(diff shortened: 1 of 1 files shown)


cognite/client/_api/assets.py
```diff
@@ -285,7 +285,7 @@ def retrieve_subtree(self, id: int = None, external_id: str = None, depth: int =
         utils._auxiliary.assert_exactly_one_of_id_or_external_id(id, external_id)
```
erlendvollset: Update the return statement on the docstring. It is no longer sorted topologically.

erlendvollset:
Closing this as I have addressed this particular issue in #442

PR #438, opened 2019-07-09
Copyright to Cognite

eventh:
Signed-off-by: Even Wiik Thomassen <even.thomassen@cognite.com>

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -44,7 +44,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "cognite-sdk"
-copyright = "2019, Cognite"
+copyright = "2019, Cognite AS"
 author = "Erlend Vollset"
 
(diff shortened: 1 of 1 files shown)


wjoel:
Should it be "Cognite" or "Cognite AS" for those things?

eventh:
I used the same as the homepage, just Cognite :)

docs/source/conf.py
```diff
@@ -44,7 +44,7 @@
 
 # General information about the project.
 project = "cognite-sdk"
-copyright = "2018, Erlend Vollset"
+copyright = "2019, Cognite"
```
wjoel: According to https://cognitedata.atlassian.net/wiki/spaces/COG/pages/7143447/Open+Source+Policy it should be "Cognite AS"

eventh: Fixed

PR #452, opened 2019-07-18
Implement initial SequencesAPI

sanderland:
This is an initial SequencesAPI implementation, with all the api calls covered and unit tests
- Needs some thought on whether/how we want a datatype returned instead of lists/dict
- No integration tests yet.

### DIFF
--- a/cognite/client/_api/sequences.py
+++ b/cognite/client/_api/sequences.py
@@ -1,10 +1,11 @@
 import math
 from typing import *
-from cognite.client.utils._experimental_warning import experimental_api
 
 from cognite.client import utils
 from cognite.client._api_client import APIClient
 from cognite.client.data_classes import Sequence, SequenceFilter, SequenceList, SequenceUpdate
+from cognite.client.utils._experimental_warning import experimental_api
+
 
 @experimental_api(api_name="Sequences")
 class SequencesAPI(APIClient):
@@ -364,7 +365,9 @@ class SequencesDataAPI(APIClient):
         pd = utils._auxiliary.local_import("pandas")
         col, data = self._retrieve_with_columns(start, end, columns, external_id, id)
 
-        transposed_data = zip(*[[d or math.nan for d in row["values"]] for row in data]) # make sure string columns don't have 'None'
+        transposed_data = zip(
+            *[[d or math.nan for d in row["values"]] for row in data]
+        )  # make sure string columns don't have 'None'
         data_dict = {k: list(v) for k, v in zip(col, transposed_data)}
         return pd.DataFrame(index=[d["rowNumber"] for d in data], data=data_dict)
 
--- a/cognite/client/_api/sequences.py
+++ b/cognite/client/_api/sequences.py
@@ -17,7 +17,7 @@ class SequencesAPI(APIClient):
         self.data = SequencesDataAPI(*args, **kwargs)
 
     def __call__(
-        self, chunk_size: int = None, asset_ids: List[int] = None, limit: int = None
+        self, chunk_size: int = None, limit: int = None
     ) -> Generator[Union[Sequence, SequenceList], None, None]:
         """Iterate over sequences
 
@@ -25,15 +25,13 @@ class SequencesAPI(APIClient):
 
         Args:
             chunk_size (int, optional): Number of sequences to return in each chunk. Defaults to yielding one event a time.
-            asset_id (int, optional): List sequences related to this asset.
             limit (int, optional): Maximum number of assets to return. Defaults to 25. Set to -1, float("inf") or None
                 to return all items.
 
         Yields:
             Union[Sequence, SequenceList]: yields Sequence one by one if chunk is not specified, else SequenceList objects.
         """
-        filter = {"assetIds": str(asset_ids) if asset_ids else None}
-        return self._list_generator(method="GET", chunk_size=chunk_size, filter=filter, limit=limit)
+        return self._list_generator(method="GET", chunk_size=chunk_size, filter=None, limit=limit)
 
     def __iter__(self) -> Generator[Sequence, None, None]:
         """Iterate over sequences
@@ -102,14 +100,12 @@ class SequencesAPI(APIClient):
         utils._auxiliary.assert_type(external_ids, "external_id", [List], allow_none=True)
         return self._retrieve_multiple(ids=ids, external_ids=external_ids, wrap_ids=True)
 
-    def list(self, asset_ids: Optional[List[int]] = None, limit: int = 25) -> SequenceList:
+    def list(self, limit: int = 25) -> SequenceList:
         """Iterate over sequences
 
         Fetches sequences as they are iterated over, so you keep a limited number of objects in memory.
 
         Args:
-            include_metadata (bool, optional): Whether or not to include metadata
-            asset_ids (List[int], optional): List sequences related to these assets.
             limit (int, optional): Max number of sequences to return. Defaults to 25. Set to -1, float("inf") or None
                 to return all items.
 
@@ -138,8 +134,7 @@ class SequencesAPI(APIClient):
                 >>> for seq_list in c.sequences(chunk_size=2500):
                 ...     seq_list # do something with the sequences
         """
-        filter = {"assetIds": str(asset_ids) if asset_ids else None}
-        return self._list(method="GET", filter=filter, limit=limit)
+        return self._list(method="GET", filter=None, limit=limit)
 
     def create(self, sequences: Union[Sequence, List[Sequence]]) -> Union[Sequence, SequenceList]:
         """Create one or more sequences.
@@ -187,7 +182,7 @@ class SequencesAPI(APIClient):
         """Update one or more sequences.
 
         Args:
-            item (Union[Sequence, SequenceUpdate, List[Union[Sequence, SequenceUpdate]]]): sequences to update
+            item (Union[Sequence, SequenceUpdate, List[Union[Sequence, SequenceUpdate]]]): Sequences to update
 
         Returns:
             Union[Sequence, SequenceList]: Updated sequences.
@@ -263,8 +258,8 @@ class SequencesDataAPI(APIClient):
         """Insert rows into a sequence
 
         Args:
-            columns: list of id or external id for the columns of the sequence
-            rows: dictionary of row number => data, where data is a list corresponding to the given columns
+            columns (List[Union[int, str]]): List of id or external id for the columns of the sequence
+            rows (Dict[int, List[Union[int, float, str]]]): Dictionary of row number => data, where data is a list corresponding to the given columns
             id (int): Id of sequence to insert rows into.
             external_id (str): External id of sequence to insert rows into.
 
@@ -300,7 +295,7 @@ class SequencesDataAPI(APIClient):
         """Delete rows from a sequence
 
         Args:
-            rows: list of row numbers
+            rows (List[int]): List of row numbers
             id (int): Id of sequence to delete rows from
             external_id (str): External id of sequence to delete rows from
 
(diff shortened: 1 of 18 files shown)


wjoel:
@f1cognite, did we agree that experimental APIs should be added only to separate SDKs (ie. `cognite-sdk-python-experimental` or something in this case)? If so, this should become part of that instead of being merged in here.

sanderland:
@wjoel commits earlier today add some warning when using experimental api classes. I think the concern for separate SDKs is the maintenance burden of keeping them in sync.

tests/tests_unit/test_api/test_sequences.py
```diff
@@ -0,0 +1,256 @@
...
+        assert df.shape[0] > 0
+        assert df.index == [0]
+
+    def test_retrieve_dataframe_convert_null(self, mock_get_datapoints_with_null):
+        df = SEQ_API.data.retrieve_dataframe(external_id="foo")
+        assert {"strcol", "intcol"} == set(df.columns)
+        assert "None" not in df.strcol
+        assert df.strcol.isna().any()
+        assert df.intcol.isna().any()
+        assert 2 == df.shape[0]
```
erlendvollset: Should test `.to_pandas()` on the Sequence and SequenceList objects aswell.

wjoel:
@sanderland yes, I saw that, but we had a meeting just a few weeks ago where it was agreed that we shouldn't add experimental features to the standard SDK, but rather distribute them as separate dependencies.

Another suggestion was to have them in an experimental namespace, but it was decided against that.

This seems to be another alternative (just printing a warning), so I'm wondering if there has been a new decision that we should also follow in other SDKs.

It's not mentioned in the meeting notes but if I remember correctly it was discussed in this meeting: https://cognitedata.atlassian.net/wiki/spaces/API/pages/992444610/2019-06-27+SDK+Prioritisation

erlendvollset:
> @f1cognite, did we agree that experimental APIs should be added only to separate SDKs (ie. `cognite-sdk-python-experimental` or something in this case)? If so, this should become part of that instead of being merged in here.

Now it is done so that if you do `from cognite.client import CogniteClient` that client will have noe experimental features. But if you do `from cognite.client.experimental import CogniteClient` it will have the experimental APIs available and a warning will be issued the first time you call an experimental method. The warning states that the API may be subject to breaking changes even between patch versions and that it should not be used in any production code.

erlendvollset:
Have not discussed it with @f1cognite though. We will hold off on releasing this until we have done so, but we HAVE to expose this functionality somehow and I feel that this is a good way of doing it.

wjoel:
I'm personally fine with it, but @f1cognite and @maciejmrowiec seemed quite adamant. ;)

erlendvollset:
@sanderland Should also add a new section in the docs named `Experimental API` or something and include the docstrings for your methods there. See docs/source/cognite.rst.

cognite/client/_cognite_client.py
```diff
@@ -67,6 +68,7 @@ def __init__(
         self.time_series = TimeSeriesAPI(self._config, api_version=self.API_VERSION, cognite_client=self)
         self.raw = RawAPI(self._config, api_version=self.API_VERSION, cognite_client=self)
         self.three_d = ThreeDAPI(self._config, api_version=self.API_VERSION, cognite_client=self)
+        self.sequences = SequencesAPI(self._config, api_version=self.API_VERSION, cognite_client=self)
```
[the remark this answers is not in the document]
sanderland: like client.experimental.sequences ? that's a bit overkill in addition to the warnings, no?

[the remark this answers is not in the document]
sanderland: done re experimental.

the docstring stuff has too much magic going on to see why they fail.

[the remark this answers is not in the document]
sanderland: signatures likewise, and time series is not tested here either, so not sure why sequences needs to be

sanderland:
> @sanderland Should also add a new section in the docs named `Experimental API` or something and include the docstrings for your methods there. See docs/source/cognite.rst.

done

cognite/client/_cognite_client.py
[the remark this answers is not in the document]
sanderland: done

tests/tests_unit/test_api/test_sequences.py
sanderland: done

sanderland:
The test_docstring_examples is what's breaking tests, will need your help on this @erlendvollset since there's too much magic mocking going on

sanderland:
@erlendvollset regarding tests_unit/test_api/test_signatures.py, this fails when adding sequences and I can't see why. Time series is also not included here, so maybe the magic isn't suitable for this API.

cognite/client/_api/sequences.py
```diff
@@ -0,0 +1,403 @@
...
+class SequencesAPI(APIClient):
+    _RESOURCE_PATH = "/sequences"
+    _LIST_CLASS = SequenceList
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.data = SequencesDataAPI(*args, **kwargs)
+
+    def __call__(
+        self, chunk_size: int = None, asset_ids: List[int] = None, limit: int = None
```
erlendvollset: According to @MariuszCognite this should take an `externalId` parameter, but there is no `assetIds` filtering.

cognite/client/_api/sequences.py
```diff
@@ -0,0 +1,403 @@
...
+
+                >>> from cognite.client.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.sequences.retrieve_multiple(external_ids=["abc", "def"])
+        """
+        utils._auxiliary.assert_type(ids, "id", [List], allow_none=True)
+        utils._auxiliary.assert_type(external_ids, "external_id", [List], allow_none=True)
+        return self._retrieve_multiple(ids=ids, external_ids=external_ids, wrap_ids=True)
+
+    def list(self, asset_ids: Optional[List[int]] = None, limit: int = 25) -> SequenceList:
```
erlendvollset: same as for the __call__ signature.

cognite/client/_api/sequences.py
```diff
@@ -0,0 +1,403 @@
...
+        """Iterate over sequences
+
+        Fetches sequences as they are iterated over, so you keep a limited number of objects in memory.
+
+        Args:
+            include_metadata (bool, optional): Whether or not to include metadata
```
erlendvollset: This should be removed

cognite/client/_api/sequences.py
```diff
@@ -0,0 +1,403 @@
...
+        """
+        self._delete_multiple(wrap_ids=True, ids=id, external_ids=external_id)
+
+    def update(
+        self, item: Union[Sequence, SequenceUpdate, List[Union[Sequence, SequenceUpdate]]]
+    ) -> Union[Sequence, SequenceList]:
+        """Update one or more sequences.
+
+        Args:
+            item (Union[Sequence, SequenceUpdate, List[Union[Sequence, SequenceUpdate]]]): sequences to update
```
erlendvollset:
```suggestion
            item (Union[Sequence, SequenceUpdate, List[Union[Sequence, SequenceUpdate]]]): Sequences to update
```

cognite/client/_api/sequences.py
```diff
@@ -0,0 +1,403 @@
...
+        columns: List[Union[int, str]],
+        rows: Dict[int, List[Union[int, float, str]]],
+        id: int = None,
+        external_id: str = None,
+    ) -> None:
+        """Insert rows into a sequence
+
+        Args:
+            columns: list of id or external id for the columns of the sequence
+            rows: dictionary of row number => data, where data is a list corresponding to the given columns
```
erlendvollset: Missing type for columns and rows. First letter of description should be capitalized.

cognite/client/_api/sequences.py
```diff
@@ -0,0 +1,403 @@
...
+        summary.raise_compound_exception_if_failed_tasks()
+
+    def _insert_data(self, task):
+        self._post(url_path=self._DATA_PATH, json={"items": [task]})
+
+    def delete(self, rows: List[int], id: int = None, external_id: str = None) -> None:
+        """Delete rows from a sequence
+
+        Args:
+            rows: list of row numbers
```
erlendvollset: Missing type and capitalize first letter of description.

cognite/client/_api/sequences.py
```diff
@@ -0,0 +1,403 @@
...
+        columns: List[Union[int, str]] = None,
+        external_id: str = None,
+        id: int = None,
+    ) -> List[dict]:
+        """Retrieve data from a sequence
+
+        Args:
+            start: (inclusive) row number to start from
+            end: (exclusive) upper limit on the row number
+            columns: list of id or external id for the columns of the sequence. Defaults to all if None is passed.
```
erlendvollset: Types and capitalization.

cognite/client/_api/sequences.py
```diff
@@ -0,0 +1,403 @@
...
+        Returns:
+            List of sequence data
+        """
+        col, data = self._retrieve_with_columns(start, end, columns, external_id, id)
+        return data
+
+    def retrieve_dataframe(
+        self,
+        start: int = 0,
+        end: int = None,
```
erlendvollset: start and end should not have default values, makes the code easier to read when used. Also nice for consistency with ts api.

docs/source/cognite.rst
```diff
@@ -768,3 +768,62 @@ Convert timestamp to milliseconds since epoch
 Convert milliseconds since epoch to datetime
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 .. autofunction:: cognite.client.utils.ms_to_datetime
+
+
+Experimental features
+=====================
+These features are not part of the core v1.0.0 API, and subject to breaking changes.
```
erlendvollset: This could cause some confusion with the `cognite-sdk-core` distribution. Maybe something like "These features are subject to breaking changes and should not be used in any production code." instead.

cognite/client/_api/sequences.py
sanderland: API has default limits like this (0 lower, no upper)
Sequences are not time series, and getting the entire (small, static) sequence is likely a common usecase

erlendvollset: I'm sure it is common usecase. Just as it is common to request all the data in a time series (start=0, end="now"). 
But if i read the two following lines

`retrieve_dataframe(id=1)`
`retrieve_dataframe(id=1, start=0, end=None)`

It is a lot more clear in the second one what is being fetched. Explicit > implicit.

sanderland: or sensible defaults > long copy-pasted argument list, but if you insist.

erlendvollset: We hade this as defaults for time series datapoints before. Discussions were had - it was agreed that fetching all the data is not always a sensible default. We want to make things intuitive yes, but we also want to avoid abuse of our APIs.

f1cognite:
Sorry for being late to the discussion here (I haven't had internet connection for a few days).

The solution here is fine for me, and yes there is an overhead with having two SDKs.
My requirement is that it should be perfectly clear for the user of the SDK that it is experimental and the user can't use the features by mistake (which was solved perfectly in this scenario).

My first standpoint was to not include experimental features at all in any of the stable SDKs but @wjoel changed my mind with the argument that it should also be easy for our users to test out new APIs in playground.

@maciejmrowiec Please shout out if you disagree.

And @sanderland - thank you for the PR! 🚀

erlendvollset:
@f1cognite Thanks. Sounds like we are in agreement then!

PR #458, opened 2019-07-23
Sequences integration tests

sanderland:
* sequences integration tests that cover mostly several backend bugs
* minor docstring fix


sanderland:
fixed backend so integration test on search passes :)

PR #464, opened 2019-07-25
fix for handling case of missing count aggregates

### LINKED ISSUE #421: Retrieving datapoints will fail if aggregation jobs are queued up.
The current method for retrieving datapoints uses count aggregates to get the distribution of the data. If aggregates have not been created for some newly posted datapoints yet, the current implementation of the DatapointsFetcher will fail.

sanderland:
Fixes issue reported in https://cognitedata.slack.com/archives/CG10VQPFX/p1563803171004700
Also fixes #421

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -799,7 +799,7 @@ class DatapointsFetcher:
             )
         except CogniteAPIError:
             res = []
-        if not res:  # string based series or aggregates not yet calculates
+        if not res:  # string based series or aggregates not yet calculated
             return [_DPWindow(start, end)]
         counts = list(zip(res.timestamp, res.count))
         windows = []
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -8,6 +8,7 @@ import cognite.client.utils._time
 from cognite.client import utils
 from cognite.client._api_client import APIClient
 from cognite.client.data_classes import Datapoints, DatapointsList, DatapointsQuery
+from cognite.client.exceptions import CogniteAPIError
 
 
 class DatapointsAPI(APIClient):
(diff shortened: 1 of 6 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -786,15 +787,20 @@ def _get_windows(self, id, start, end, granularity, request_limit, user_limit):
...
+                end=end,
+                ts_item={"id": id},
+                aggregates=["count"],
+                granularity=count_granularity,
+                include_outside_points=False,
+                limit=None,
+            )
+        except CogniteAPIError:
+            res = []
+        if not res:  # string based series or aggregates not yet calculates
```
wjoel: "yet calculated'

cognite/client/_api/datapoints.py
```diff
@@ -786,15 +787,20 @@ def _get_windows(self, id, start, end, granularity, request_limit, user_limit):
...
+            res = self._get_datapoints_with_paging(
+                start=start,
+                end=end,
+                ts_item={"id": id},
+                aggregates=["count"],
+                granularity=count_granularity,
+                include_outside_points=False,
+                limit=None,
+            )
+        except CogniteAPIError:
```
wjoel: Could we be more specific about what type of CogniteAPIError we catch here?

sanderland: not really. in a sense failure here already gives some specifics in that (1) a normal datapoint request succeeded and (2) the type of error is not worth retrying, or retries failed.
we could check the text or error code after catching, but in any case, the user didn't ask us to get count aggregate, it's not strictly needed, and our response is to simply get datapoints with a different (non-paralellized) strategy, so we should avoid passing the exception up to the user.

cognite/client/_api/datapoints.py
sanderland: fixed

cognite/client/_api/datapoints.py
wjoel: It could fail for other interesting reasons, which could be relevant. Not sure how we feel about logging but it might be worth doing to not hide this completely. On the other hand we don't seem to have a logger available so maybe that's not the way, @erlendvollset would know. :)

Since we're doing `if not res:` immediately after it seems this could be simplified to
```
        except CogniteAPIError:
            # assume it's a string based series or aggregates not yet calculated
            return [_DPWindow(start, end)]
```

sanderland: @wjoel for floating point timeseries with heavy ingestion / out of date aggregates we could have res=[] without exception (see issue #421), this is why I wrote it like that.

Issue #425, opened 2019-07-03
parent_external_id should not be validated by sdk


nodegard:
**Describe the bug**
Cannot create asset hierarchy using parent_external_id when more than 1000 are posted, when the parent asset is not part of the request.

**To Reproduce**
Runnable code reproducing the error.
```
from cognite.client import CogniteClient
from cognite.client.data_classes.assets import Asset

client = CogniteClient()

# Make root
asset = Asset(external_id="root",
               name="name",
               description="description",
               metadata=None,
               source=None)

created = client.assets.create(asset)

# Make lots of children
assets = []
for i in range(1,1005):
  assets.append(Asset(external_id="child_{0}".format(i),
               name="name",
               description="description",
               metadata=None,
               source=None,
               parent_external_id="root"))
try:
  created = client.assets.create(assets)
except Exception as e:
  print(e)
```

**Expected behavior**
Request using parent external id should be allowed even when parent is not a part of the request.

**System information (please complete the following information):**
 - Python Version: CPython/3.5.2
 - SDK Version: 1.0.0a33

**Additional context**
Copying asset hierarchy from one tenant to another.

eventh:
The current validating logic is flawed, it requires that the batch of nodes you are posting includes a root node. It is not possible to create more than 1000 assets in a hierarchy attached to an existing asset, when the relationship is defined by parent_external_id.

The only possible workaround is not using parent_external_id.
It would be better to remove this faulty validation as it doesn't provide any value except limit use cases that should be possible.

erlendvollset:
If the parent_external_id set on an asset is not part of the batch of nodes you are posting, it should be treated as a "root" node on the client side, meaning that it should be subject to the same sorting of its descendants (in the same batch), and posted in that order. @HMEiding will address this issue and the other related to #469 in this sprint.

PR #471, opened 2019-07-28
WIP: Removed parent_external_id validation in _validate_asset_hierarchy.

eventh:
The validation checked that all assets that are being created, if they
use parent_external_id, points to an external_id that are now being created.
This doesn't work if you are creating assets pointing to existing assets,
and want to use parent_external_id.

This commit removes the parent_external_id validation completely.
Also removed the unittest.

Signed-off-by: Even Wiik Thomassen <even.thomassen@cognite.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Changes are grouped as follows
 - Separate read/write fields on data classes
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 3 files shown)


eventh:
This fixes #425

eventh:
Will do, I discovered other parts of the code relying on this validation.

Issue #453, opened 2019-07-19
Review sequences client before unmarking as experimental


erlendvollset:
Once Sequences API has been released in v1 we should do a review of the current client implementation and settle on an interface. Thinking in particular about the following:

- implementing a custom data class for sequence data. 
- .insert_dataframe() method
- .rows() method on the Sequence object.

sanderland:
- one thing that keeps catching me off-guard is that create takes a column object with camelCase names, and returns a column list of dicts where [c.id for c in seq.columns] doesn't work (since it's c['id']), which clashes with seq.id being fine.
- delete_range is not supported in the api, but could be in the sdk

sanderland:
All of the above are done in 1.0.4, but will leave this open for further comments

PR #472, opened 2019-07-30
improvements and additions to the sequences SDK

sanderland:
implements most of the suggestions from #453

- sequencedata object
- helper functions on sequence object (rows, column_ids, etc)
- Sequences insert takes its parameters row data first, and no longer requires columns to be passed.
- Sequences create now clears invalid fields such as 'id' in columns specification, so sequences can more easily re-use existing specifications.
- delete_range funtion
- insert function accepts {rowNumber: .., values: ...} dict
- insert_dataframe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@ Changes are grouped as follows
 - Separate read/write fields on data classes
 
 ## [Unreleased]
+### Changed
+- Sequences data retrieval now returns a SequenceData object
+
+### Added
+- Variety of useful helper functions for Sequence and SequenceData objects
+- Sequences delete_range function
 
 ## [1.0.3] - 2019-07-26
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,6 @@ Changes are grouped as follows
 - Variety of useful helper functions for Sequence and SequenceData objects, including .column_ids and .column_external_ids properties, iterators and slice operators.
 - Sequences insert_dataframe function.
 - Sequences delete_range function.
-- Support for external id column headers in datapoints.insert_dataframe().
 
 ## [1.0.3] - 2019-07-26
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,14 @@ Changes are grouped as follows
 
 ## [Unreleased]
 ### Changed
-- Sequences data retrieval now returns a SequenceData object
+- Sequences data retrieval now returns a SequenceData object.
+- Sequences insert takes its parameters row data first, and no longer requires columns to be passed.
+- Sequences create now clears invalid fields such as 'id' in columns specification, so sequences can more easily re-use existing specifications.
 
 ### Added
-- Variety of useful helper functions for Sequence and SequenceData objects
-- Sequences delete_range function
+- Variety of useful helper functions for Sequence and SequenceData objects.
+- Sequences insert_dataframe function
+- Sequences delete_range function.
 
 ## [1.0.3] - 2019-07-26
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,14 +24,15 @@ Changes are grouped as follows
 ### Changed
 - Sequences data retrieval now returns a SequenceData object.
 - Sequences insert takes its parameters row data first, and no longer requires columns to be passed.
-- Sequences insert now accepts tuples and raw-style data input
+- Sequences insert now accepts tuples and raw-style data input.
 - Sequences create now clears invalid fields such as 'id' in columns specification, so sequences can more easily re-use existing specifications.
+- Sequence data function now require column_ids or column_external_ids to be explicitly set, rather than both being passed through a single columns field
 
 ### Added
-- Variety of useful helper functions for Sequence and SequenceData objects, particularly .column_ids() and .column_external_ids()
-- Sequences insert_dataframe function
+- Variety of useful helper functions for Sequence and SequenceData objects, including .column_ids and .column_external_ids properties, iterators and slice operators.
+- Sequences insert_dataframe function.
 - Sequences delete_range function.
-- Support for external id column headers in datapoints.insert_dataframe()
+- Support for external id column headers in datapoints.insert_dataframe().
 
 ## [1.0.3] - 2019-07-26
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,10 +23,11 @@ Changes are grouped as follows
 ### Changed
 - Sequences data retrieval now returns a SequenceData object.
 - Sequences insert takes its parameters row data first, and no longer requires columns to be passed.
+- Sequences insert now accepts tuples and raw-style data input
 - Sequences create now clears invalid fields such as 'id' in columns specification, so sequences can more easily re-use existing specifications.
 
 ### Added
-- Variety of useful helper functions for Sequence and SequenceData objects.
+- Variety of useful helper functions for Sequence and SequenceData objects, particularly .column_ids() and .column_external_ids()
 - Sequences insert_dataframe function
 - Sequences delete_range function.
 
--- a/cognite/client/_api/sequences.py
+++ b/cognite/client/_api/sequences.py
@@ -137,11 +137,11 @@ class SequencesAPI(APIClient):
         """
         return self._list(method="GET", filter=None, limit=limit)
 
-    def create(self, sequences: Union[Sequence, List[Sequence]]) -> Union[Sequence, SequenceList]:
+    def create(self, sequence: Union[Sequence, List[Sequence]]) -> Union[Sequence, SequenceList]:
         """Create one or more sequences.
 
         Args:
-            sequences (Union[Sequence, List[Sequence]]): Sequence or list of Sequence to create.
+            sequence (Union[Sequence, List[Sequence]]): Sequence or list of Sequence to create.
 
         Returns:
             Union[Sequence, SequenceList]: The created sequences.
@@ -160,11 +160,12 @@ class SequencesAPI(APIClient):
                 >>> seq2 = c.sequences.create(Sequence(external_id="my_copied_sequence", columns=seq.columns))
 
         """
-        if isinstance(sequences, list):
-            sequences = [self._clean_columns(seq) for seq in sequences]
+        utils._auxiliary.assert_type(sequence, "sequences", [List, Sequence])
+        if isinstance(sequence, list):
+            sequence = [self._clean_columns(seq) for seq in sequence]
         else:
-            sequences = self._clean_columns(sequences)
-        return self._create_multiple(items=sequences)
+            sequence = self._clean_columns(sequence)
+        return self._create_multiple(items=sequence)
 
     def _clean_columns(self, sequence):
         sequence = copy.copy(sequence)
@@ -273,14 +274,17 @@ class SequencesDataAPI(APIClient):
         rows: Union[
             Dict[int, List[Union[int, float, str]]], List[Tuple[int, Union[int, float, str]]], List[Dict[str, Any]]
         ],
-        columns: List[Union[int, str]] = None,
+        column_ids: Optional[List[int]] = None,
+        column_external_ids: Optional[List[str]] = None,
         id: int = None,
         external_id: str = None,
     ) -> None:
         """Insert rows into a sequence
 
         Args:
-            columns (List[Union[int, str]]): List of id or external id for the columns of the sequence. If 'None' is passed, all columns ids will be retrieved from metadata and used in that order.
+            column_ids (Optional[List[int]]): List of ids for the columns of the sequence.
+                If 'None' is passed to both column_ids and columns_external_ids, all columns ids will be retrieved from metadata and used in that order.
+            column_external_ids (Optional[List[str]]): List of external id for the columns of the sequence.
             rows (Union[ Dict[int, List[Union[int, float, str]]], List[Tuple[int,Union[int, float, str]]], List[Dict[str,Any]]]):  The rows you wish to insert.
                 Can either be a list of tuples, a list of {"rowNumber":... ,"values": ...} objects or a dictionary of rowNumber: data. See examples below.
             id (int): Id of sequence to insert rows into.
@@ -296,7 +300,7 @@ class SequencesDataAPI(APIClient):
                 >>> c = CogniteClient()
                 >>> seq = c.sequences.create(Sequence(columns=[{"valueType": "STRING","valueType": "DOUBLE"}]))
                 >>> data = [(1, ['pi',3.14]), (2, ['e',2.72]) ]
-                >>> res1 = c.sequences.data.insert(columns=seq.column_ids(), rows=data, id=1)
+                >>> res1 = c.sequences.data.insert(column_ids=seq.column_ids, rows=data, id=1)
 
             They can also be provided as a list of API-style objects with a rowNumber and values field::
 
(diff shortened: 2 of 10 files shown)


sanderland:
+ Sequences insert takes its parameters row data first, and no longer requires columns to be passed.
+ Sequences create now clears invalid fields such as 'id' in columns specification, so sequences can more easily re-use existing specifications.
+ insert dataframe

cognite/client/_api/sequences.py
```diff
@@ -253,35 +270,63 @@ def __init__(self, sequences_api, *args, **kwargs):
...
+
+        if isinstance(rows, dict):
+            all_rows = [{"rowNumber": k, "values": v} for k, v in rows.items()]
+        elif isinstance(rows, list) and len(rows) > 0 and isinstance(rows[0], dict):
+            all_rows = rows
+        elif isinstance(rows, list) and (len(rows) == 0 or isinstance(rows[0], tuple)):
+            all_rows = [{"rowNumber": k, "values": v} for k, v in rows]
+        else:
+            raise ValueError("Invalid format for 'rows', expected a list of tuples, list of dict or dict")
+
```
erlendvollset: Should we do some checks here to make sure the number of values matches the number of columns? Or maybe the error messages from the API are sufficient?

cognite/client/_api/sequences.py
```diff
@@ -291,6 +336,32 @@ def insert(
         )
         summary.raise_compound_exception_if_failed_tasks()
 
+    def insert_dataframe(self, dataframe, id=None, external_id=None):
```
erlendvollset: type hints!

cognite/client/_api/sequences.py
```diff
@@ -317,9 +388,37 @@ def delete(self, rows: List[int], id: int = None, external_id: str = None) -> No
 
         self._post(url_path=self._DATA_PATH + "/delete", json={"items": [post_obj]})
 
+    def delete_range(self, start, end, id: int = None, external_id: str = None) -> None:
```
erlendvollset: type hints!

cognite/client/_api/sequences.py
```diff
@@ -350,38 +454,22 @@ def retrieve_dataframe(
...
-            seqdata.extend(data)
+            columns = columns or items[0]["columns"]
+            callback(data)
             if len(data) < task["limit"]:
                 break
             task["inclusiveFrom"] = data[-1]["rowNumber"] + 1
             if task["exclusiveTo"] and task["inclusiveFrom"] >= task["exclusiveTo"]:
                 break
-        return columns, seqdata
+        return columns
```
erlendvollset: I don't really like this callback pattern here, the method is called `fetch_data` but does not return the data which is a little confusing. It makes the code less readable. I would prefer if you return columns and seqdata as before and handle the resulting seqdata in the calling method instead of with the callback.

cognite/client/data_classes/sequences.py
```diff
@@ -47,6 +48,39 @@ def __init__(
 
     # GenStop
 
+    def rows(self, start, end) -> List[dict]:
```
erlendvollset: type hints!

cognite/client/data_classes/sequences.py
```diff
@@ -145,3 +179,136 @@ def remove(self, value: List) -> SequenceUpdate:
...
+            type(self) == type(other)
+            and self.id == other.id
+            and self.external_id == other.external_id
+            and self.row_numbers == other.row_numbers
+            and self.values == other.values
+        )
+
+    def __getitem__(self, item) -> List[Union[int, str, float]]:
+        # slow, should be replaced by dict cache if it sees more than incidental use
+        return self.values[self.row_numbers.index(item)]
```
erlendvollset: May be slow, but I like this idea!

I think it should behave slightly differently though. If a slice is passed I think it should return a new SequenceData object containing only the requested values, instead of a list of lists. This should also be reflected in the type hint.

cognite/client/data_classes/sequences.py
```diff
@@ -145,3 +179,136 @@ def remove(self, value: List) -> SequenceUpdate:
...
+
+    def column_external_ids(self):
+        """Retrieves list of column external ids for the sequence, for use in e.g. data retrieve or insert methods
+
+        Returns:
+            List of sequence column external ids
+        """
+        return [c.get("externalId") for c in self.columns]
+
+    def column_value_types(self):
```
erlendvollset: Missing return type hints on these helper methods. Also I think these should be properties.

cognite/client/data_classes/sequences.py
```diff
@@ -145,3 +179,136 @@ def remove(self, value: List) -> SequenceUpdate:
...
+            and self.external_id == other.external_id
+            and self.row_numbers == other.row_numbers
+            and self.values == other.values
+        )
+
+    def __getitem__(self, item) -> List[Union[int, str, float]]:
+        # slow, should be replaced by dict cache if it sees more than incidental use
+        return self.values[self.row_numbers.index(item)]
+
+    def __getattr__(self, item) -> List[Union[int, str, float]]:
```
erlendvollset: I don't really like this. If I just mistyped one of the actual attributes on the object, and I get this custom ValueError instead of an AttributeError I would be confused. Can we move this into a `def get_column(id=None, external_id=None)` method instead?

cognite/client/data_classes/sequences.py
```diff
@@ -145,3 +179,136 @@ def remove(self, value: List) -> SequenceUpdate:
...
     _UPDATE = SequenceUpdate
+
+
+class SequenceData:
+    """An object representing a list of rows from a sequence.
+
+    Args:
+        id (int): Id of the sequence the data belong to
+        external_id (str): External id of the sequence the data belong to
+        rows (List[dict]): combined row numbers and row data object from the API. If you pass this, row_numbers/values are ignored.
```
erlendvollset:
```suggestion
        rows (List[dict]): Combined row numbers and row data object from the API. If you pass this, row_numbers/values are ignored.
```

cognite/client/data_classes/sequences.py
```diff
@@ -145,3 +179,136 @@ def remove(self, value: List) -> SequenceUpdate:
...
+        row_numbers (List[int]): The data row numbers.
+        values (List[List[ Union[int, str, float]]]): The data values, one row at a time.
+        columns: List[dict]: The column information, in the format returned by the API
```
erlendvollset: Inconsistent line endings period/no period. Sorry, OCD 😅

cognite/client/data_classes/sequences.py
```diff
@@ -145,3 +179,136 @@ def remove(self, value: List) -> SequenceUpdate:
...
+            ix = self.column_external_ids().index(item)
+        except ValueError as e:
+            raise ValueError(
+                "Column {} not found, Sequence column external ids are {}".format(item, self.column_external_ids())
+            )
+        return [r[ix] for r in self.values]
+
+    def __iter__(self) -> Generator[Tuple[int, List[Union[int, str, float]]], None, None]:
+        for row, values in zip(self.row_numbers, self.values):
+            yield row, values
```
erlendvollset: Feel free to disagree here, I'm not very adamant about this. But I feel that this should have the same return type as `__getitem__` per yielded item, so not return the row number. But it may be useful, I don't know.

CHANGELOG.md
```diff
@@ -20,7 +20,17 @@ Changes are grouped as follows
...
 ## [Unreleased]
+
+### Changed
+- Sequences data retrieval now returns a SequenceData object.
+- Sequences insert takes its parameters row data first, and no longer requires columns to be passed.
+- Sequences insert now accepts tuples and raw-style data input
+- Sequences create now clears invalid fields such as 'id' in columns specification, so sequences can more easily re-use existing specifications.
+
 ### Added
+- Variety of useful helper functions for Sequence and SequenceData objects, particularly .column_ids() and .column_external_ids()
```
erlendvollset: Feel free to include the names of all the helper functions you made. I got some feedback on our changelog that it should be more detailed.

cognite/client/_api/sequences.py
```diff
@@ -332,8 +431,13 @@ def retrieve(
...
+        utils._auxiliary.assert_exactly_one_of_id_or_external_id(id, external_id)
+        post_obj = self._process_ids(id, external_id, wrap_ids=True)[0]
+        post_obj.update(self._process_columns(columns))
+        post_obj.update({"inclusiveFrom": start, "exclusiveTo": end})
+        seqdata = []
+        column_response = self._fetch_data(post_obj, lambda data: seqdata.extend(data))
+        return SequenceData(id=id, external_id=external_id, rows=seqdata, columns=column_response)
 
     def retrieve_dataframe(
         self, start: int, end: int, columns: List[Union[int, str]] = None, external_id: str = None, id: int = None
```
erlendvollset: Missing return type hint.

erlendvollset:
This is starting to look very good BTW, great work!!! 🏆 🥇

cognite/client/_api/sequences.py
```diff
@@ -253,35 +270,63 @@ def __init__(self, sequences_api, *args, **kwargs):
...
+        columns: List[Union[int, str]] = None,
         id: int = None,
         external_id: str = None,
     ) -> None:
         """Insert rows into a sequence
 
         Args:
-            columns (List[Union[int, str]]): List of id or external id for the columns of the sequence
-            rows (Dict[int, List[Union[int, float, str]]]): Dictionary of row number => data, where data is a list corresponding to the given columns
+            columns (List[Union[int, str]]): List of id or external id for the columns of the sequence. If 'None' is passed, all columns ids will be retrieved from metadata and used in that order.
```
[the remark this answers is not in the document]
sanderland: changing this to two parameters everywhere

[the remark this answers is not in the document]
sanderland: not sure how to resolve a similar implicit dependency in the insert_dataframe though

cognite/client/_api/sequences.py
sanderland: I think the API has clear enough error messages.

cognite/client/_api/sequences.py
sanderland: yessir!

cognite/client/_api/sequences.py
sanderland: this is normal for dataframe since it may not be imported, no?

cognite/client/_api/sequences.py
sanderland: needed for not keeping all data in memory, as discussed

cognite/client/data_classes/sequences.py
sanderland: It's hard to implement efficiently, and I'm not sure it will see use. Maybe something to revisit later.

cognite/client/data_classes/sequences.py
sanderland: if pandas can do it, why can't we ;)
but very well, changed to get_column

cognite/client/data_classes/sequences.py
sanderland: I think you give getitem a row number already, so getting it back is strange. But iterating over pairs is natural for this data type, since it's not a list.

cognite/client/_api/sequences.py
```diff
@@ -336,10 +340,10 @@ def insert(
         )
         summary.raise_compound_exception_if_failed_tasks()
 
-    def insert_dataframe(self, dataframe, id=None, external_id=None):
+    def insert_dataframe(self, dataframe, id: int = None, external_id: str = None):
```
erlendvollset: return typehint `-> None`

cognite/client/_api/sequences.py
```diff
@@ -310,12 +314,12 @@ def insert(
...
-                >>> res1 = c.sequences.data.insert(columns=['stringColumn','intColumn'], rows=data, id=1)
+                >>> res1 = c.sequences.data.insert(column_external_ids=['stringColumn','intColumn'], rows=data, id=1)
         """
         utils._auxiliary.assert_exactly_one_of_id_or_external_id(id, external_id)
-        if columns is None:
-            columns = self._sequences_api.retrieve(id=id, external_id=external_id).column_ids()
-
+        if column_ids is None and column_external_ids is None:
+            column_ids = self._sequences_api.retrieve(id=id, external_id=external_id).column_ids
+        print("ids", column_ids, "ext", column_external_ids)
```
erlendvollset: remove print

cognite/client/_api/sequences.py
erlendvollset: Could just make it into a generator instead?

cognite/client/data_classes/sequences.py
erlendvollset: Why is that hard to implement efficiently? `__getitem__` should support accepting a slice, but if you want to revisit this later you should at the very least raise a TypeError if a slice is passed, now it will raise a more cryptic ValueError saying that the slice cannot be found in the list. I would not say that slicing a list is something that will not see use.

cognite/client/data_classes/sequences.py
```diff
@@ -229,18 +232,32 @@ def __eq__(self, other):
...
         )
 
-    def __getitem__(self, item) -> List[Union[int, str, float]]:
+    def __getitem__(self, item: int) -> List[Union[int, str, float]]:
         # slow, should be replaced by dict cache if it sees more than incidental use
         return self.values[self.row_numbers.index(item)]
 
-    def __getattr__(self, item) -> List[Union[int, str, float]]:
-        """Get a column by externalId
+    def get_column(self, external_id: str = None, id: int = None) -> List[Union[int, str, float]]:
```
erlendvollset: Do you have a unit test for this one?

cognite/client/_api/sequences.py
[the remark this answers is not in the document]
erlendvollset: Should use the same approach as `.insert_dataframe()` in time series datapoints, where we have a `external_id_headers: bool = False` parameter.

cognite/client/data_classes/sequences.py
erlendvollset: But you are implementing all the list magic methods, so it should behave like a list.

cognite/client/data_classes/sequences.py
erlendvollset: Still missing these.

cognite/client/_api/sequences.py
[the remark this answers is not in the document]
sanderland: ok

cognite/client/data_classes/sequences.py
sanderland: it's kind of behaving like a dict. we could call this iteritems instead if that makes you happy

cognite/client/data_classes/sequences.py
sanderland: it gets tested in the integration tests

cognite/client/data_classes/sequences.py
erlendvollset: .items() in that case, which is the method you have on a dict.

cognite/client/data_classes/sequences.py
erlendvollset: That is not the purpose of the integration tests. It's ok to use it in the integration test, but it should have its own unit test, which fails only if this method breaks.

sanderland: raising a valueerror now, we could potentially support step!=1 slices in the future, but it's a bit complicated since the row numbers could be all over the place

cognite/client/data_classes/sequences.py
sanderland: dict has items (returns the entire object) and iteritems (returns an iterator), so it's the latter I think

cognite/client/data_classes/sequences.py
sanderland: now is also test in unit test

cognite/client/data_classes/sequences.py
```diff
@@ -234,6 +234,8 @@ def __eq__(self, other):
 
     def __getitem__(self, item: int) -> List[Union[int, str, float]]:
         # slow, should be replaced by dict cache if it sees more than incidental use
+        if isinstance(item, slice):
+            raise ValueError("Slicing SequenceData not supported")
```
erlendvollset: Should be a TypeError

cognite/client/data_classes/sequences.py
erlendvollset: You are correct.

cognite/client/data_classes/sequences.py
sanderland: fixed

Issue #460, opened 2019-07-24
Implement .create() method on files API to allow creating file(s) without uploading


eventh:
**Describe the solution you'd like**
The CDF API allows you to create a file with metadata, and not upload anything. The SDK should enable the same work flow, by making the path argument optional.

**Additional context**
The use case for this feature is that it enables us to create files with metadata, and later upload the content, so we can separate these two steps.

greenbech:
This is sort of blocking me, so I can try to come up with a PR for this

erlendvollset:
@HMEiding We have this on our backlog for the upcoming sprimt, @HMEiding will submit a PR.

eventh:
Great, I'm on vacation so didn't have time to fix it myself :)

eventh:
@erlendvollset @HMEiding did you guys create this?

PR #478, opened 2019-08-06
Allow file creation without uploading a file

greenbech:
This PR implements what’s described in #460 

The interface for creating a file without uploading content is done by setting this first argument in `FilesAPI.upload` to `None`. I first thought it would be better/cleaner to make `path` an optional argument (as described in the issue), but that would be a breaking change.

The test only checks if no exception occurs. Should maybe add a more thorough test?

### DIFF
--- a/cognite/client/_api/files.py
+++ b/cognite/client/_api/files.py
@@ -287,7 +287,9 @@ class FilesAPI(APIClient):
         """Upload a file
 
         Args:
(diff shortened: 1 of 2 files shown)


erlendvollset:
This is not how we want to solve this, we want to have a .create() method which will be more consistent with the other APIs. There is already another PR open on this, #478. We have this task on our backlog for the upcoming sprint, so I will close this PR.

PR #479, opened 2019-08-06
3d integration tests

HMEiding:


tests/tests_integration/test_api/test_three_d.py
```diff
@@ -0,0 +1,109 @@
...
+@pytest.fixture(scope="class")
+def new_asset_mapping(test_revision):
+    revision, model_id = test_revision
+
+    node_id = COGNITE_CLIENT.three_d.revisions.list_nodes(model_id=model_id, revision_id=revision.id)[0].id
+    asset_id = COGNITE_CLIENT.assets.list(limit=1)[0].id
+    asset_mapping = ThreeDAssetMapping(node_id=node_id, asset_id=asset_id)
+    res = COGNITE_CLIENT.three_d.asset_mappings.create(
+        model_id=model_id, revision_id=revision.id, asset_mapping=asset_mapping
+    )
```
erlendvollset: You have a race condition here. Since we are running these integration tests in parallell for three different python versions, you will have this mapping be created and deleted 3 times. One of two things will happen:

1) API complains that mapping already exists
2) Test using this fixture will fail since another integration test will have deleted the mapping

Either drop this integration test or find a more robust of handling it. Nothing of the latter comes to mind immediately.

tests/tests_integration/test_api/test_three_d.py
```diff
@@ -0,0 +1,96 @@
...
+        revision, model_id = test_revision
+        added_metadata = {"key": "value"}
+        revision_update = ThreeDModelRevisionUpdate(id=revision.id).metadata.add(added_metadata)
+        COGNITE_CLIENT.three_d.revisions.update(model_id=model_id, item=revision_update)
+
+
+class TestThreeDFilesAPI:
+    def test_retrieve(self, test_revision):
+        revision, model_id = test_revision
+        base_url = COGNITE_CLIENT._config.base_url
```
erlendvollset: not used

PR #481, opened 2019-08-07
Add .create() method to files API.

HMEiding:


cognite/client/_api/files.py
```diff
@@ -72,6 +72,54 @@ def __iter__(self) -> Generator[FileMetadata, None, None]:
...
         return self.__call__()
 
+    def create(
+        self,
+        name: str,
+        external_id: str = None,
+        source: str = None,
+        metadata: Dict[str, Any] = None,
+        asset_ids: List[int] = None,
+        overwrite: bool = False,
```
erlendvollset: You can just take a FileMetadata object as parameter here, that way you won't have to worry about keeping this signature in sync with the auto-generated FileMetadata constructor signature. You will still need to include the overwrite parameter of course.

You are also missing the return type hint here which should be Tuple[FileMetadata, str].

cognite/client/_api/files.py
```diff
@@ -72,6 +72,54 @@ def __iter__(self) -> Generator[FileMetadata, None, None]:
...
+            metadata (Dict[str, Any]): Customizable extra data about the file. String key -> String value.
+            asset_ids (List[int]): No description.
+            overwrite (bool): If 'overwrite' is set to true, and the POST body content specifies a 'externalId' field,
+                fields for the file found for externalId can be overwritten. The default setting is false.
+                If metadata is included in the request body, all of the original metadata will be overwritten.
+                File-Asset mappings only change if explicitly stated in the assetIds field of the POST json body.
+                Do not set assetIds in request body if you want to keep the current file-asset mappings.
+
+        Returns:
+            (FileMetaData, str): Tuple containing the file metadata and upload url of the created file.
```
erlendvollset: Tuple[FileMetadata, str]

cognite/client/_api/files.py
```diff
@@ -72,6 +72,54 @@ def __iter__(self) -> Generator[FileMetadata, None, None]:
...
+
+        Examples:
+
+            Create a file::
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.files.create(name="my_file", metadata={"key": "value"})
```
erlendvollset: file_metadata, upload_url = c.files.create(...)

tests/tests_integration/test_api/test_files.py
```diff
@@ -30,6 +30,10 @@ def test_files():
 
 
 class TestFilesAPI:
+    def test_create(self):
+        file_metadata, upload_url = COGNITE_CLIENT.files.create(name="mytestfile")
```
erlendvollset: Maybe just add an `assert False == file_metadata.uploaded` before deleting the file?

tests/tests_unit/test_api/test_files.py
```diff
@@ -126,6 +146,12 @@ def download_link_callback(request):
 
 
 class TestFilesAPI:
+    def test_create(self, mock_file_create_response):
+        file_metadata, upload_url = FILES_API.create(name="bla")
+        response_body = mock_file_create_response.calls[0].response.json()
+        assert FileMetadata._load(response_body) == file_metadata
+        assert response_body["uploadUrl"] == upload_url
```
erlendvollset: Set expected values on the left side of the operator.

cognite/client/_api/files.py
HMEiding: For some reason, doing this causes an error in test_docstring_examples.py. It complains that there are not enough values to unpack (expected 2, got 0). Can you see why?

tests/tests_unit/test_api/test_files.py
HMEiding: That is what I'm doing here.

erlendvollset: ah of course 🤦‍♂

cognite/client/_api/files.py
erlendvollset: Yes, it's because we mock the client so in the test .create() actually returns as MagickMock object. Just leave it like this!

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Changes are grouped as follows
 - Separate read/write fields on data classes
 
 ## [Unreleased]
+- The files API can now create a file without uploading content.
```
erlendvollset:
```suggestion
### Added
- The files API can now create a file without uploading content.
```

PR #484, opened 2019-08-09
Improved logic for parent_external_id in AssetPoster

HMEiding:


tests/tests_unit/test_api/test_assets.py
```diff
@@ -243,8 +243,7 @@ def generate_asset_tree(root_external_id: str, depth: int, children_per_node: in
 class TestAssetPoster:
     def test_validate_asset_hierarchy_parent_ref_null_pointer(self):
```
erlendvollset: You can just remove this test alltogether.

tests/tests_integration/test_api/test_assets.py
```diff
@@ -99,3 +99,16 @@ def test_post_asset_hierarchy(self, new_asset_hierarchy):
...
+    def test_create_asset_hierarchy_parent_external_id_not_in_request(self):
+        root_external_id = "my_root_{}".format(utils._auxiliary.random_string(5))
+        root = Asset(external_id=root_external_id, name="my_root")
+        root = COGNITE_CLIENT.assets.create(root)
+
+        children = generate_asset_tree(
+            root_external_id=root_external_id, depth=5, children_per_node=10, current_depth=2
+        )
+        children = COGNITE_CLIENT.assets.create(children)
+        assert COGNITE_CLIENT.assets._CREATE_LIMIT < len(children)
```
erlendvollset: Why do make this assertion?

I think we should also test that the resulting hierarchy has the intended structure.

cognite/client/_api/assets.py
```diff
@@ -417,8 +417,6 @@ def _validate_asset_hierarchy(assets) -> None:
 
             parent_ref = asset.parent_external_id
             if parent_ref:
-                if parent_ref not in external_ids:
```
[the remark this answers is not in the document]
HMEiding: I'll check if there is some way to show unused variables. Hopefully, I don't have to use linting to achieve this...

tests/tests_integration/test_api/test_assets.py
HMEiding: It's just to make sure that the _AssetPoster is tested. I agree that we should test the structure as well.

tests/tests_integration/test_api/test_assets.py
```diff
@@ -101,14 +101,26 @@ def test_get_subtree(self, root_test_asset):
...
             root_external_id=root_external_id, depth=5, children_per_node=10, current_depth=2
         )
-        children = COGNITE_CLIENT.assets.create(children)
-        assert COGNITE_CLIENT.assets._CREATE_LIMIT < len(children)
 
-        COGNITE_CLIENT.assets.delete(external_id=root_external_id, recursive=True)
+        try:
+            COGNITE_CLIENT.assets.create(root)
+            COGNITE_CLIENT.assets.create(children)
+
```
HMEiding: The setup could be put in a fixture, but since the main purpose of the test is to test the create() functions, I felt it was more natural to include it in the test. What do you think?

erlendvollset: Up to you really, putting it in a fixture would make it more similar to the above test, so maybe easier to understand. But I agree with your reasoning.

PR #489, opened 2019-08-12
Retrieve dataframe improvements

sanderland:
implements the suggested improvements to retrieve_dataframe
the dictionary of {id: .. aggregates: ...} make for some finicky edge cases

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -1,4 +1,5 @@
 import math
+import re as regexp
 import threading
 from collections import defaultdict, namedtuple
 from datetime import datetime
@@ -96,9 +97,6 @@ class DatapointsAPI(APIClient):
             limit=limit,
         )
         dps_list = fetcher.fetch(query)
-        for dp in dps_list:
-            dp._granularity = granularity
-            dp._cognite_client = self._cognite_client
         if is_single_id:
             return dps_list[0]
         return dps_list
@@ -423,25 +421,87 @@ class DatapointsAPI(APIClient):
                 ...         aggregates=["interpolation"], granularity="1m", include_aggregate_name=False, complete="fill,dropna")
         """
         pd = utils._auxiliary.local_import("pandas")
-        id_df = pd.DataFrame()
-        external_id_df = pd.DataFrame()
-        if id is not None and external_id is not None and complete:
-            raise ValueError("Can not complete dataframes with mixed id and externalId")
 
         if id is not None:
-            id_df = self.retrieve(
+            id_dpl = self.retrieve(
                 id=id, start=start, end=end, aggregates=aggregates, granularity=granularity, limit=limit
-            ).to_pandas(column_names="id", include_aggregate_name=include_aggregate_name, complete=complete)
+            )
+            id_df = id_dpl.to_pandas(column_names="id")
+        else:
+            id_df = pd.DataFrame()
+            id_dpl = DatapointsList([])
+
         if external_id is not None:
-            external_id_df = self.retrieve(
+            external_id_dpl = self.retrieve(
                 external_id=external_id,
                 start=start,
                 end=end,
                 aggregates=aggregates,
                 granularity=granularity,
                 limit=limit,
-            ).to_pandas(include_aggregate_name=include_aggregate_name, complete=complete)
-        return pd.concat([id_df, external_id_df], axis="columns")
+            )
+            external_id_df = external_id_dpl.to_pandas()
+        else:
+            external_id_df = pd.DataFrame()
+            external_id_dpl = DatapointsList([])
+
+        df = pd.concat([id_df, external_id_df], axis="columns")
+
+        complete = [s.strip() for s in (complete or "").split(",")]
+        if set(complete) - set(["fill", "dropna", ""]):
+            raise ValueError("complete should be 'fill', 'fill,dropna' or Falsy")
+
+        if "fill" in complete and df.shape[0] > 1:
+            ag_used_by_id = {
+                dp.id: [attr for attr, _ in dp._get_non_empty_data_fields(get_empty_lists=True)]
+                for dpl in [id_dpl, external_id_dpl]
+                for dp in (dpl.data if isinstance(dpl, DatapointsList) else [dpl])
+            }
+            ts_meta = self._cognite_client.time_series.retrieve(
+                ids=[id for id, aggs_used in ag_used_by_id.items() if "interpolation" in aggs_used]
+            )
+            is_step_dict = {str(ts.get(field)): bool(ts.is_step) for ts in ts_meta for field in ["id", "external_id"]}
+            df = self._dataframe_fill(self, df, granularity, is_step_dict)
+
+            if "dropna" in complete:
+                self._dataframe_safe_dropna(df, set([ag for id, ags in ag_used_by_id for ag in ags]))
+
+        if not include_aggregate_name:
+            Datapoints._strip_aggregate_names(df)
+
+        return df
+
+    def _dataframe_fill(self, df, granularity, is_step_dict):
+        np, pd = utils._auxiliary.local_import("numpy", "pandas")
+        df = df.reindex(
+            np.arange(
+                df.index[0],
+                df.index[-1] + pd.Timedelta(microseconds=1),
+                pd.Timedelta(microseconds=cognite.client.utils._time.granularity_to_ms(granularity) * 1000),
+            ),
+            copy=False,
+        )
+        df.fillna({c: 0 for c in df.columns if regexp.search(c, r"\|(sum|totalVariance|count)$")}, inplace=True)
+        int_cols = [c for c in df.columns if regexp.search(c, r"\|interpolation$")]
+        lin_int_cols = [c for c in int_cols if not is_step_dict[regexp.match(r"(.*)\|\w+$", c)[1]]]
+        step_int_cols = [c for c in df.columns if regexp.search(c, r"\|stepInterpolation$")] + list(
+            set(int_cols) - set(lin_int_cols)
+        )
+        df[lin_int_cols] = df[lin_int_cols].interpolate(limit_area="inside")
+        df[step_int_cols] = df[step_int_cols].ffill()
+        return df
+
+    def _dataframe_safe_dropna(self, df, aggregates_used):
+        supported_aggregates = ["sum", "count", "total_variance", "interpolation", "step_interpolation"]
+        not_supported = set(aggregates_used) - set(supported_aggregates + ["timestamp"])
+        if not_supported:
+            raise ValueError(
+                "The aggregate(s) {} is not supported for dataframe completion with dropna, only {} are".format(
+                    [utils._auxiliary.to_camel_case(a) for a in not_supported],
+                    [utils._auxiliary.to_camel_case(a) for a in supported_aggregates],
+                )
+            )
+        df.dropna(inplace=True)
 
     def retrieve_dataframe_dict(
         self,
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -485,7 +485,7 @@ class DatapointsAPI(APIClient):
         )
         df.fillna({c: 0 for c in df.columns if regexp.search(c, r"\|(sum|totalVariance|count)$")}, inplace=True)
         int_cols = [c for c in df.columns if regexp.search(c, r"\|interpolation$")]
-        lin_int_cols = [c for c in int_cols if not is_step_dict[regexp.match(r"(.*)\|\w+$", c)[1]]]
+        lin_int_cols = [c for c in int_cols if not is_step_dict[regexp.match(r"(.*)\|\w+$", c).group(1)]]
         step_int_cols = [c for c in df.columns if regexp.search(c, r"\|stepInterpolation$")] + list(
             set(int_cols) - set(lin_int_cols)
         )
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -448,7 +448,7 @@ class DatapointsAPI(APIClient):
         df = pd.concat([id_df, external_id_df], axis="columns")
 
         complete = [s.strip() for s in (complete or "").split(",")]
-        if set(complete) - set(["fill", "dropna", ""]):
+        if set(complete) - {"fill", "dropna", ""}:
             raise ValueError("complete should be 'fill', 'fill,dropna' or Falsy")
 
         if "fill" in complete and df.shape[0] > 1:
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -385,8 +385,8 @@ class DatapointsAPI(APIClient):
                 ids. Can also be object specifying aggregates. See example below.
             limit (int): Maximum number of datapoints to return for each time series.
             complete (str): post-processing of the dataframe.
-                Pass "fill" to insert missing entries into the index, and complete data where possible (supports interpolation, stepInterpolation, count, sum, totalVariation).
-                Pass "fill,dropna" to additionally drop rows in which any aggregate for any time series has missing values (typically rows at the start and end for interpolation aggregates).
+                Pass 'fill' to insert missing entries into the index, and complete data where possible (supports interpolation, stepInterpolation, count, sum, totalVariation).
+                Pass 'fill,dropna' to additionally drop rows in which any aggregate for any time series has missing values (typically rows at the start and end for interpolation aggregates).
                 This guarantees that all returned dataframes have the exact same shape and no missing values anywhere, and is only supported for aggregates using any of sum, count, totalVariance, interpolation and stepInterpolation
 
         Returns:
@@ -477,9 +477,12 @@ class DatapointsAPI(APIClient):
                 ids. Can also be object specifying aggregates. See example below.
             limit (int): Maximum number of datapoints to return for each time series.
             complete (str): post-processing of the dataframe.
-                Pass "fill" to insert missing entries into the index, and complete data where possible (supports interpolation, stepInterpolation, count, sum, totalVariation).
-                Pass "fill,dropna" to additionally drop rows in which any aggregate for any time series has missing values (typically rows at the start and end for interpolation aggregates).
+                Pass 'fill' to insert missing entries into the index, and complete data where possible (supports interpolation, stepInterpolation, count, sum, totalVariation).
+                Pass 'fill,dropna' to additionally drop rows in which any aggregate for any time series has missing values (typically rows at the start and end for interpolation aggregates).
                 This guarantees that all returned dataframes have the exact same shape and no missing values anywhere, and is only supported for aggregates using any of sum, count, totalVariance, interpolation and stepInterpolation
+
+        Returns:
+            Union[pandas.DataFrame,Dict[pandas.DataFrame]]: The requested dataframe if one aggregate was requested, or a dictionary of aggregate: dataframe if multiple aggregates were requested.
 
         Examples:
 
@@ -496,9 +499,6 @@ class DatapointsAPI(APIClient):
                 >>> c = CogniteClient()
                 >>> dfs = c.datapoints.retrieve_dataframe_dict(id=[1,2,3], start="2w-ago", end="now",
                 ...          aggregates=["interpolation","count"], granularity="1h", complete="fill,dropna")
-
-        Returns:
-            Union[pandas.DataFrame,Dict[pandas.DataFrame]]: The requested dataframe if one aggregate was requested, or a dictionary of aggregate: dataframe if multiple aggregates were requested.
         """
         all_aggregates = aggregates
         for queries in [id, external_id]:
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -384,7 +384,7 @@ class DatapointsAPI(APIClient):
             external_id (Union[str, List[str], Dict[str, Any], List[Dict[str, Any]]]): External id or list of external
                 ids. Can also be object specifying aggregates. See example below.
             limit (int): Maximum number of datapoints to return for each time series.
-            include_aggregate_name (bool): Include '|aggregate' in the column name. Defaults to True and should only be set to False when only a single aggregate is requested.
+            include_aggregate_name (bool): Include 'aggregate' in the column name. Defaults to True and should only be set to False when only a single aggregate is requested per id/externalId.
             complete (str): post-processing of the dataframe.
                 Pass 'fill' to insert missing entries into the index, and complete data where possible (supports interpolation, stepInterpolation, count, sum, totalVariation).
                 Pass 'fill,dropna' to additionally drop rows in which any aggregate for any time series has missing values (typically rows at the start and end for interpolation aggregates).
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -385,9 +385,9 @@ class DatapointsAPI(APIClient):
                 ids. Can also be object specifying aggregates. See example below.
             limit (int): Maximum number of datapoints to return for each time series.
             complete (str): post-processing of the dataframe.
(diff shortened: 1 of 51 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -411,7 +417,117 @@ def retrieve_dataframe(
...
+            str, List[str], Dict[str, Union[int, List[str]]], List[Dict[str, Union[int, List[str]]]]
+        ] = None,
+        limit: int = None,
+        complete=None,
+    ):
+        """Get a dictionary of aggregate: pandas dataframe describing the requested data.
+
+        Args:
+
+            see retrieve_dataframe
```
erlendvollset: This needs to be completed.

docs/source/cognite.rst
```diff
@@ -429,6 +429,10 @@ Retrieve pandas dataframe
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 .. automethod:: cognite.client._api.datapoints.DatapointsAPI.retrieve_dataframe
 
+Retrieve pandas dataframes indexed by aggregate
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. automethod:: cognite.client._api.datapoints.DatapointsAPI.retrieve_dataframe
```
erlendvollset: .retrieve_dataframe_dict

erlendvollset:
> I considered the two parameter option, but you don't really want dropping without filling (since the api is quite aggressive about not returning rows, and a single nan would drop the entire row).

Is it necessary to support the dropna option at all? isn't it a one liner in pandas? Or i see that you disallow dropna for some aggregates, what is the reasoning for that?

sanderland:
if you don't fill the aggregates and then dropna, it will potentially drop rows in the middle, which is almost certainly not intended.
Essentially it's a way of doing this safely, all in one place.
In addition, for the _dict version, if you dropna after, your dataframes may be differently sized, which is also potentially problematic.

docs/source/cognite.rst
sanderland: fixed

cognite/client/_api/datapoints.py
sanderland: fixed

cognite/client/_api/datapoints.py
```diff
@@ -1,8 +1,9 @@
 import math
+import re
 import threading
 from collections import defaultdict, namedtuple
 from datetime import datetime
-from typing import *
+from typing import Any, Dict, List, Tuple, Union  # typing.re was clashing with re
```
erlendvollset: Remove this comment.

cognite/client/_api/datapoints.py
```diff
@@ -411,7 +417,128 @@ def retrieve_dataframe(
...
+        for queries in [id, external_id]:
+            if isinstance(queries, list) and queries and isinstance(queries[0], dict):
+                for it in queries:
+                    for ag in it.get("aggregates") or []:
+                        if ag not in all_aggregates:
+                            all_aggregates.append(ag)
+
+        df = self.retrieve_dataframe(start, end, aggregates, granularity, id, external_id, limit, complete)
+        df_dict = {ag: df.filter(like="|" + ag).rename(columns=lambda s: s[: -len(ag) - 1]) for ag in all_aggregates}
+        return df_dict[aggregates[0]] if len(aggregates) == 1 else df_dict
```
erlendvollset: This should just always return the dict. The name of the method is retrieve_dataframe_dict, so it feels unnatural that it does this unpacking when there is only value in that dict.

cognite/client/_api/datapoints.py
```diff
@@ -411,7 +417,128 @@ def retrieve_dataframe(
...
+        start: Union[int, str, datetime],
+        end: Union[int, str, datetime],
+        aggregates: List[str],
+        granularity: str,
+        id: Union[int, List[int], Dict[str, Union[int, List[str]]], List[Dict[str, Union[int, List[str]]]]] = None,
+        external_id: Union[
+            str, List[str], Dict[str, Union[int, List[str]]], List[Dict[str, Union[int, List[str]]]]
+        ] = None,
+        limit: int = None,
+        complete=None,
```
erlendvollset: typehint

cognite/client/_api/datapoints.py
```diff
@@ -367,6 +368,7 @@ def retrieve_dataframe(
             str, List[str], Dict[str, Union[int, List[str]]], List[Dict[str, Union[int, List[str]]]]
         ] = None,
         limit: int = None,
+        complete=None,
```
erlendvollset: typehint

cognite/client/_api/datapoints.py
```diff
@@ -382,6 +384,10 @@ def retrieve_dataframe(
             external_id (Union[str, List[str], Dict[str, Any], List[Dict[str, Any]]]): External id or list of external
                 ids. Can also be object specifying aggregates. See example below.
             limit (int): Maximum number of datapoints to return for each time series.
+            complete (str): post-processing of the dataframe.
+                Pass 'fill' to insert missing entries into the index, and complete data where possible (supports interpolation, stepInterpolation, count, sum, totalVariation).
+                Pass 'fill,dropna' to additionally drop rows in which any aggregate for any time series has missing values (typically rows at the start and end for interpolation aggregates).
+                This guarantees that all returned dataframes have the exact same shape and no missing values anywhere, and is only supported for aggregates using any of sum, count, totalVariance, interpolation and stepInterpolation
```
erlendvollset: Does "This" refer to 'fill' or 'fill,dropna', or both?

cognite/client/_api/datapoints.py
```diff
@@ -411,7 +417,128 @@ def retrieve_dataframe(
                 granularity=granularity,
                 limit=limit,
             ).to_pandas()
-        return pd.concat([id_df, external_id_df], axis="columns")
+        df = pd.concat([id_df, external_id_df], axis="columns")
+        complete = complete or ""
```
erlendvollset: maybe `complete = [s.strip() for s in complete.split(",")] if complete else []` is better. Then it won't matter if the user passes "fill,dropna", "fill, dropna", or "dropna, fill" etc.

cognite/client/_api/datapoints.py
```diff
@@ -382,6 +384,10 @@ def retrieve_dataframe(
             external_id (Union[str, List[str], Dict[str, Any], List[Dict[str, Any]]]): External id or list of external
                 ids. Can also be object specifying aggregates. See example below.
             limit (int): Maximum number of datapoints to return for each time series.
+            complete (str): post-processing of the dataframe.
```
erlendvollset:
```suggestion
            complete (str): Post-processing of the dataframe.
```

cognite/client/_api/datapoints.py
```diff
@@ -411,7 +417,128 @@ def retrieve_dataframe(
...
+            else:
+                is_step_dict = None
+            df = self._complete_dataframe(df, granularity, is_step_dict)
+
+            if "dropna" in complete:
+                completed_aggregates = ["sum", "count", "totalVariance", "interpolation", "stepInterpolation"]
+                if set(aggregates) - set(completed_aggregates):
+                    raise ValueError(
+                        "One of the aggregates {} not in the supported aggregates {} for dropna".format(
+                            complete, completed_aggregates
```
erlendvollset: aggregates, completed_aggregates?

cognite/client/_api/datapoints.py
sanderland: happy to rename the method while keeping this functionality - suggestions?

cognite/client/_api/datapoints.py

cognite/client/_api/datapoints.py

cognite/client/_api/datapoints.py
sanderland: only the latter. I was trying to make a list but don't know the syntax

cognite/client/_api/datapoints.py
sanderland: currently works anyway since it checks substring by in, but probably list is cleaner/safer. added this and made the value check stronger

cognite/client/_api/datapoints.py
sanderland: yep, + cleaned up

sanderland:
moved logic to to_pandas as suggested. This results in quite a lot of parameters to Datapoints.to_pandas (since there are some awkward cross-dependencies), but they have reasonable defaults and it's mostly an internal method anyway.

erlendvollset:
Just set those as private attributes on the Datapoints object before returning it from .retrieve(). Then the user won't have to pass the parameters twice when using to_pandas().

cognite/client/_api/datapoints.py
```diff
@@ -410,9 +429,64 @@ def retrieve_dataframe(
...
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> dfs = c.datapoints.retrieve_dataframe_dict(id=[1,2,3], start="2w-ago", end="now",
+                ...          aggregates=["interpolation","count"], granularity="1h", complete="fill,dropna")
+        """
+        all_aggregates = aggregates
+        for queries in [id, external_id]:
+            if isinstance(queries, list) and queries and isinstance(queries[0], dict):
+                for it in queries:
+                    for ag in it.get("aggregates") or []:
```
erlendvollset: `it.get("aggregates",[])`

cognite/client/_api/datapoints.py
```diff
@@ -393,15 +400,27 @@ def retrieve_dataframe(
                 >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
                 >>> df = c.datapoints.retrieve_dataframe(id=[1,2,3], start="2w-ago", end="now",
-                ...         aggregates=["average"], granularity="1h")
+                ...         aggregates=["average","sum"], granularity="1h")
+
+            Get a pandas dataframe with the index regularly spaced at 1 minute intervals, missing values completed and without the aggregate name in the columns::
```
erlendvollset: It is confusing to me as a reader that "missing values are completed" but at the same time I am passing a parameter "dropna".

cognite/client/_api/datapoints.py
erlendvollset: Just drop the newline then.

cognite/client/_api/datapoints.py
erlendvollset: still no typehint?

cognite/client/data_classes/datapoints.py
```diff
@@ -271,19 +333,69 @@ def __str__(self):
             i["datapoints"] = utils._time.convert_time_attributes_to_datetime(i["datapoints"])
         return json.dumps(item, default=lambda x: x.__dict__, indent=4)
 
-    def to_pandas(self, column_names="externalId") -> "pandas.DataFrame":
+    def to_pandas(
+        self, column_names="externalId", include_aggregate_name=True, complete=None, complete_granularity=None
```
erlendvollset: typehints.

cognite/client/data_classes/datapoints.py
```diff
@@ -171,12 +172,26 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
             dumped = {utils._auxiliary.to_camel_case(key): value for key, value in dumped.items()}
         return {key: value for key, value in dumped.items() if value is not None}
 
-    def to_pandas(self, column_names="externalId") -> "pandas.DataFrame":
+    def to_pandas(
+        self,
+        column_names: str = "externalId",
+        include_aggregate_name: bool = True,
+        complete: bool = None,
```
erlendvollset: This `complete` parameter should behave the same way as in retrieve_dataframe(). Or is there a good reason why it is a bool here?

cognite/client/data_classes/datapoints.py
```diff
@@ -189,15 +204,54 @@ def to_pandas(self, column_names="externalId") -> "pandas.DataFrame":
...
+        if not include_aggregate_name:
+            if len(fields) > 1:
+                raise ValueError(
+                    "include_aggregate_name=False can not be used here as it would lead to duplicate column names for multiple aggregates {}".format(
+                        fields
+                    )
+                )
+            df.rename(columns=lambda s: re.sub(s, r"\|\w+$", ""), inplace=True)
+
+        return df
```
erlendvollset: This is a very long method. Maybe try splitting it up so it is a little more readable?

cognite/client/data_classes/datapoints.py
```diff
@@ -235,6 +289,14 @@ def _extend(self, other_dps):
             else:
                 value.extend(other_value)
 
+    def _get_non_empty_data_keys(self):
```
erlendvollset: Can you not use the keys from `_get_non_empty_data_fields`?

cognite/client/data_classes/datapoints.py
```diff
@@ -271,19 +333,69 @@ def __str__(self):
...
+            not_supported = set(aggregates_used) - set(supported_aggregates + ["timestamp"])
+            if not_supported:
+                raise ValueError(
+                    "The aggregate(s) {} is not supported for dataframe completion with dropna, only {} are".format(
+                        [utils._auxiliary.to_camel_case(a) for a in not_supported],
+                        [utils._auxiliary.to_camel_case(a) for a in supported_aggregates],
+                    )
+                )
+            df.dropna(inplace=True)
+        return df
```
erlendvollset: Maybe try splitting this into a few methods for readability aswell?

cognite/client/data_classes/datapoints.py
```diff
@@ -271,19 +333,69 @@ def __str__(self):
...
+            complete_start = min(starts) if starts else None
+            complete_end = max(ends) if starts else None
+            if interp_ids:  # get isStep flag and pass it to completion to handle interpolation correctly
+                ts_meta = self._cognite_client.time_series.retrieve_multiple(ids=interp_ids)
+                is_step_dict.update({ts.id: bool(ts.is_step) for ts in ts_meta})
+        dfs = [
+            df.to_pandas(
+                column_names=column_names,
+                include_aggregate_name=include_aggregate_name,
+                complete=fill_flag,
```
erlendvollset: won't this just always be True?

cognite/client/data_classes/datapoints.py
sanderland: the dropna depends on all rows. This is a bit of an issue when retrieve can happily return a datapoints or datapointslist -- fixing now

sanderland: actually really annoying - any thoughts on how to 'always get a list' ?

cognite/client/data_classes/datapoints.py
sanderland: I was worried about the .copy, but then copied that as well -- and it's shallow anyway, right?

cognite/client/data_classes/datapoints.py
sanderland: ok

cognite/client/data_classes/datapoints.py
sanderland: done

cognite/client/data_classes/datapoints.py
sanderland: fill_flag = "fill" in complete
(maybe clearer in the refactored version)

cognite/client/data_classes/datapoints.py
sanderland: -> just making this a string as suggested, since the user may want to to_pandas it anyway

cognite/client/_api/datapoints.py

cognite/client/_api/datapoints.py
sanderland: added some more newlines after checking readability

cognite/client/_api/datapoints.py

cognite/client/data_classes/datapoints.py
sanderland: done

cognite/client/data_classes/datapoints.py
```diff
@@ -189,15 +208,74 @@ def to_pandas(self, column_names="externalId") -> "pandas.DataFrame":
...
+        complete = Datapoints._get_complete_options(complete)
+        if "fill" in complete and df.shape[0] > 1:
+
+            if complete_is_step is None and "interpolation" in aggregates:
+                client = getattr(self, "_cognite_client", None)
+                if client:
+                    complete_is_step = Datapoints._is_step(client, [self.id])
+            complete_granularity = getattr(self, "_granularity", None)
+            if not complete_granularity:
+                raise AttributeError("Can only complete Datapoints objects returned from DatapointsAPI.retrieve")
```
erlendvollset: Raising an AttributeError here is a little off. I would create a custom exception for this.

cognite/client/_api/datapoints.py
```diff
@@ -96,7 +96,9 @@ def retrieve(
             limit=limit,
         )
         dps_list = fetcher.fetch(query)
-
+        for dp in dps_list:
+            dp._granularity = granularity
```
erlendvollset: This should be set in the DatapointsFetcher, so that you can use the complete functionality on Datapoints objects returned from .query() aswell. And should have unit tests which make sure they have it set.

Also, can't you do the same thing for ._complete_start, ._complete_end, and ._complete_is_step. Why do those need to be passed to .to_pandas()?

cognite/client/_api/datapoints.py
```diff
@@ -96,7 +96,9 @@ def retrieve(
...
+            dp._cognite_client = self._cognite_client
```
erlendvollset: This is already being set in the DatapointsFetcher

cognite/client/_api/datapoints.py
```diff
@@ -419,9 +440,64 @@ def retrieve_dataframe(
                 aggregates=aggregates,
                 granularity=granularity,
                 limit=limit,
-            ).to_pandas()
+            ).to_pandas(include_aggregate_name=include_aggregate_name, complete=complete)
         return pd.concat([id_df, external_id_df], axis="columns")
 
+    def retrieve_dataframe_dict(
```
erlendvollset: I am leaning towards dropping this method. I really want to try to keep the sdk code base as light as possible, and we have not gotten any explicit feature requests for this. We have the "include_aggregate_name" parameter now, which solves the same problem no? 

I'm not saying that this one method will make the SDK bloated, but we should always be mindful when adding new methods as to what purpose they serve.

We can always create a package where we patch methods in to the SDK. cognite-sdk-plus or something where we can put in all kinds of fancy functionality.

cognite/client/_api/datapoints.py
erlendvollset: Maybe something like "missing values completed or dropped depending on the nature of the aggregate"

cognite/client/data_classes/datapoints.py
```diff
@@ -271,19 +372,66 @@ def __str__(self):
...
+            ends = [dp.timestamp[-1] for dp in self.data if dp.timestamp]
+            complete_start = min(starts) if starts else None
+            complete_end = max(ends) if starts else None
+            if interp_ids:  # get isStep flag and pass it to completion to handle interpolation correctly
+                is_step_dict.update(Datapoints._is_step(self._cognite_client, interp_ids))
+        return [
+            df.to_pandas(
+                column_names=column_names,
+                include_aggregate_name=include_aggregate_name,
+                complete="fill" if fill else None,  # don't pass dropna
```
erlendvollset: Wont `fill` here still always evaluate to `True` since you are doing `if fill` on line 416?

cognite/client/_api/datapoints.py
sanderland: - support for query is noted

- complete_is_step we don't want to request if not needed, and then we'd like to request all at once. This is one of the most awkward things of all.
- start/end have to do with Datapoints having different ranges, where count and such should fill it in. what should we set these to in DatapointsFetcher?

cognite/client/_api/datapoints.py
sanderland: - it's set on the DatapointsList, but not on the Datapoints objects, and since retrieve can return one of those as well for single ids, the user can easily call to_pandas on one.

cognite/client/data_classes/datapoints.py
sanderland: this is not in the block starting at 416

cognite/client/_api/datapoints.py
sanderland: the idea would be that you can do things like
quality = df['count'] > 10 & df['totalVariance'] > 0.1
but yes, without feedback it's hard to see who aside from me has use for this

sanderland:
any idea why tox is failing?

cognite/client/_api/datapoints.py
erlendvollset: Let's keep it, I think it sounds very useful.

cognite/client/data_classes/datapoints.py
```diff
@@ -171,11 +172,12 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
             dumped = {utils._auxiliary.to_camel_case(key): value for key, value in dumped.items()}
         return {key: value for key, value in dumped.items() if value is not None}
 
-    def to_pandas(self, column_names="externalId") -> "pandas.DataFrame":
+    def to_pandas(self, column_names: str = "externalId", include_aggregate_name: bool = True) -> "pandas.DataFrame":
```
erlendvollset: `include_aggregate_name` is not used anywhere so it won't have any effect.

erlendvollset: Should have a unittest for this.

cognite/client/data_classes/datapoints.py
```diff
@@ -271,16 +278,21 @@ def __str__(self):
             i["datapoints"] = utils._time.convert_time_attributes_to_datetime(i["datapoints"])
         return json.dumps(item, default=lambda x: x.__dict__, indent=4)
 
-    def to_pandas(self, column_names="externalId") -> "pandas.DataFrame":
+    def to_pandas(self, column_names: str = "externalId", include_aggregate_name: bool = True) -> "pandas.DataFrame":
```
erlendvollset: Should have a unit test for `include_aggregate_name` here aswell.

cognite/client/_api/datapoints.py
```diff
@@ -402,25 +411,154 @@ def retrieve_dataframe(
...
+            )
+            external_id_df = external_id_dpl.to_pandas()
+        else:
+            external_id_df = pd.DataFrame()
+            external_id_dpl = DatapointsList([])
+
+        df = pd.concat([id_df, external_id_df], axis="columns")
+
+        complete = [s.strip() for s in (complete or "").split(",")]
+        if set(complete) - set(["fill", "dropna", ""]):
```
erlendvollset: This "set" function call can be replaced with a set literal, so `{"fill", "dropna", ""}` instead.

cognite/client/data_classes/datapoints.py
sanderland: nice catch. fixed and added tests all around

cognite/client/_api/datapoints.py
```diff
@@ -402,25 +411,154 @@ def retrieve_dataframe(
...
+            )
+            is_step_dict = {
+                str(field): bool(ts.is_step) for ts in ts_meta for field in [ts.id, ts.external_id] if field
+            }
+            df = self._dataframe_fill(df, granularity, is_step_dict)
+
+            if "dropna" in complete:
+                self._dataframe_safe_dropna(df, set([ag for id, ags in ag_used_by_id.items() for ag in ags]))
+
+        if not include_aggregate_name:
```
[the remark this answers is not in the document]
sanderland: I was going to let the user shoot themselves in the foot if they wanted to, but very well. this is now checked in the strip names function and tested

cognite/client/data_classes/datapoints.py
sanderland: done

cognite/client/_api/datapoints.py
sanderland: fixed

PR #490, opened 2019-08-12
Increase create limit of RawRowsAPI

HMEiding:


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Changes are grouped as follows
 ## [Unreleased]
 ### Changed
 - Fixed naming for sequencesdata iteritems() -> items()
+- The insert method of the raw.rows API now has a create limit of 10 000 (increased from 1000).
```
erlendvollset: Rather, "will chunk raw rows into batches of 10,000 instead of 1,000". There is no limit.

HMEiding: Yes, I agree.

Issue #492, opened 2019-08-13
Making copying data types between tenants easier by allowing to insert retrieved objects directly


senickel:
**Is your feature request related to a problem? Please describe.**

As a data scientist working with customers, I want to copy different data types from the production to the test tenant, so that I can develop without fear.

**Describe the solution you'd like**
```
ts1 = client_source.time_series.retrieve(id=123).to_dict()
written_ts = client_dev.time_series.create(time_series=ts1)
dps1 = client_source.datapoints.retrieve(start="1h-ago", end="now", external_id=ts1.external_id)to_dict()
written_dps = client_dev.datapoints.insert(dps1, external_id=ts1.external_id)
```

**Describe alternatives you've considered**
The approach for this is now:

```
ts1 = client_source.time_series.retrieve(id=123)
ts2 = {
                "externalId": ts1.external_id,
                "name": ts1.name,
                "isString": ts1.is_string,
                "metadata": ts1.metadata,
                "unit": ts1.unit,
                "isStep": ts1.is_step,
                "description": ts1.description,
                "legacyName": ts1.external_id
            }
dps1 = client_source.datapoints.retrieve(start="1h-ago", end="now", external_id=ts2['externalId'])
dps2 =  list(zip(dps.timestamp, dps.value))
written_ts = client_dev.time_series.create(time_series=ts2)
written_dps = client_dev.datapoints.insert(dps2, external_id=ts1.external_id)
```

sanderland:
two issues here:
1. datapoint insertion from retrieval
this works client.datapoints.insert(dps1.dump()['datapoints'],id=ts.id)
but client.datapoints.insert(dps1,id=ts.id) could be supported

2. timeseries (and presumably other objects like asset)
- id / updated+created time could be stripped from the object via some helper methods

erlendvollset:
Enabling this requires generating some notion of read/write attributes on the data classes. Then we can use those constraints to remove readOnly fields when creating resources. This is not trivial and will require some work. I have noted that it is requested, but we have other tools (like spark data source) which can handle this today, so it will not be prioritized right now.

sanderland:
I think the good first issue was on (1) only, sorry.

sanderland:
PR #500 implements the datapoints copying for this.

haakonvt:
Solved by the new write classes implemented in #1567 (and #1582)

PR #493, opened 2019-08-13
Time series dumb

senickel:
I added a dump method to TimeSeries to address parts of this issue: https://github.com/cognitedata/cognite-sdk-python/issues/492


erlendvollset:
This breaks the current .dump() implementation on the TimeSeries object, and is not how we want to solve this issue. If you want to contribute to enabling copying data between tenants, please come discuss the topic with the data tooling team. But as I have commented on the related issue this is not a trivial thing to do in a consistent and maintainable way. Closing this PR, and we can continue the discussion on the related issue #492

PR #494, opened 2019-08-13
Add new method create_hierarchy() to AssetsAPI

HMEiding:


cognite/client/_api/assets.py
```diff
@@ -213,9 +213,29 @@ def create(self, asset: Union[Asset, List[Asset]]) -> Union[Asset, AssetList]:
                 >>> res = c.assets.create(assets)
         """
         utils._auxiliary.assert_type(asset, "asset", [Asset, list])
-        if isinstance(asset, Asset) or len(asset) <= self._CREATE_LIMIT:
-            return self._create_multiple(asset)
-        return _AssetPoster(asset, client=self).post()
+        return self._create_multiple(asset)
+
+    def create_hierarchy(self, assets: List[Asset]) -> AssetList:
+        """Create asset hierarchy.
```
erlendvollset: You can add a little more information here about how this differs from .create(), i.e. which validation steps it will perform and how it will resolve the hierarchy locally and post assets in the correct order.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ Changes are grouped as follows
 
 ### Added
 - The files API can now create a file without uploading content.
+- New method create_hierarchy() added to assets API.
```
erlendvollset: Under ### Changed, add that assets.create() no longer will validate and resolve asset hierarchy locally and post assets in correct order, and that that functionality has been replaced by assets.create_hierarchy(). 

We should probably wait with merging this PR until we have merged in the other PRs currently waiting and made a patch release. Then we bump the minor version for this change since it is a breaking change. 

Breaking changes would really mean that we should bump the major version. But since it is a very small change and bumping major would mean maintaining 2 major versions and porting bugfixes to 1.x, let's just bump the minor version and hope that no one gets too upset about it.

cognite/client/_api/assets.py
```diff
@@ -409,10 +427,11 @@ def __init__(self, assets: List[Asset], client: AssetsAPI):
     def _validate_asset_hierarchy(assets) -> None:
         external_ids_seen = set()
         for asset in assets:
-            if asset.external_id:
-                if asset.external_id in external_ids_seen:
-                    raise AssertionError("Duplicate external_id '{}' found".format(asset.external_id))
-                external_ids_seen.add(asset.external_id)
+            if asset.external_id is None:
+                raise AssertionError("An asset does not have external_id.")
```
eventh: Isn't external_id optional?

eventh: Nevermind, I see the requirements further down, resolving it :)

PR #504, opened 2019-08-19
Deduplicate during fetching when retrieving related resources on AssetList

HMEiding:


tests/tests_unit/test_data_classes/test_assets.py
```diff
@@ -79,19 +79,22 @@ def test_get_files(self):
...
         r3 = resource_class(id=3)
         resources_a1 = resource_list_class([r1])
         resources_a2 = resource_list_class([r2, r3])
         resources_a3 = resource_list_class([r2, r3])
 
-        mock_concurrency.execute_tasks_concurrently.return_value.results = [resources_a1, resources_a2, resources_a3]
+        mock_cognite_client = mock.MagicMock()
+        mock_method = getattr(mock_cognite_client, method)
+        mock_method.list.side_effect = [resources_a1, resources_a2, resources_a3]
+        mock_method._config = mock.Mock(max_workers=3)
```
erlendvollset: Is this necessary?

HMEiding: It's necessary in order for the comparison on line 108 of _concurrency.py to work. The exact value of 3 is not important, though.

cognite/client/data_classes/assets.py
```diff
@@ -173,6 +174,11 @@ class AssetList(CogniteResourceList):
     _RESOURCE = Asset
     _UPDATE = AssetUpdate
 
+    def __init__(self, resources: List[Any], cognite_client=None):
+        super().__init__(resources, cognite_client)
+        self._lock = threading.Lock()
```
erlendvollset: This lock should be instantiated in the function scope.

cognite/client/data_classes/assets.py
```diff
@@ -204,21 +210,28 @@ def files(self) -> "FileMetadataList":
         return self._retrieve_related_resources(FileMetadataList, self._cognite_client.files)
 
     def _retrieve_related_resources(self, resource_list_class, resource_api):
+        seen = set()
+
+        def retrieve_and_deduplicate(asset_ids):
+            res = resource_api.list(asset_ids=asset_ids, limit=-1)
+            resources = resource_list_class([])
+            for resource in res:
+                with self._lock:
```
erlendvollset: Actually, move this lock outside the for loop. No point in releasing/acquiring the lock in every iteration 😅

HMEiding: Yes, I thought so too.

PR #511, opened 2019-08-26
notebook output

sanderland:
These functions make it so that resources in a notebook show up like their pandas equivalent instead of json dumps.
Convenient, or confusing?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,12 @@ Changes are grouped as follows
 
 ## [Unreleased]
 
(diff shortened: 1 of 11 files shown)


erlendvollset:
Should add a note about it in the changelog though :)

erlendvollset:
Maybe default string representation for CogniteResourceList should be .to_pandas() aswell?

sanderland:
As in to_pandas. __str? I think that might break some console scripts/core sdk.

erlendvollset:
Ah yes, of course.

PR #521, opened 2019-08-29
Fix bug causing CogniteResource.to_pandas() to fail if metadata not set

sanderland:


tests/tests_unit/test_api/test_time_series.py
```diff
@@ -258,3 +258,6 @@ def test_time_series_to_pandas(self, mock_ts_response):
         assert "metadata" not in df.columns
         assert [0] == df.loc["securityCategories"][0]
         assert "metadata-value" == df.loc["metadata-key"][0]
+
+    def test_pandas_no_metadata(self):
+        TimeSeries(id=123)._repr_html_()
```
erlendvollset: This test should be in test_base.TestCogniteResource. And it shouldn't test _repr_html(), but to_pandas().

sanderland: saw where the test was insufficient before and added it there.

PR #526, opened 2019-09-05
Fix invalid external link in README

f1cognite:


sanderland:
api link doc->docs?

sanderland:
and the jenkins fail probably requires you to
python generate_from_spec.py

hhenriques:
> and the jenkins fail probably requires you to
> python generate_from_spec.py

The test is also failing in master though.

PR #549, opened 2019-09-20
Add root_asset_external_ids to events

sanderland:
was mentioned to me as being inconsistent with assets, which was a fair point.


cognite/client/_api/events.py
```diff
@@ -36,8 +37,8 @@ def __call__(
             type (str): Type of the event, e.g 'failure'.
             subtype (str): Subtype of the event, e.g 'electrical'.
             metadata (Dict[str, Any]): Customizable extra data about the event. String key -> String value.
-            asset_ids (List[int]): Asset IDs of related equipments that this event relates to.
-            root_asset_ids (List[Dict[str, Any]]): The IDs of the root assets that the related assets should be children of.
+            root_asset_ids (List[int]): The IDs of the root assets that the related assets should be children of.
```
erlendvollset: You removed the asset_ids param from the docstring here

PR #558, opened 2019-10-01
Add support for using bearer token and release 1.2.0

erlendvollset:
Adds a parameter on CogniteClient named `token`.

`token` is either a jwt (str) or a function which takes no args and returns a jwt (str) to authenticate against CDF.
If `token` has been supplied, every time a request is made the token factory is called and the request will authenticate using that jwt. The jwt will override any API key which as been set.

This way the developer can handle refreshing of that token in the token factory. This is a lot cleaner than passing a token explicitly to every request, and we won't have to add an extra parameter on every single method in the SDK.

### DIFF
--- a/cognite/client/_cognite_client.py
+++ b/cognite/client/_cognite_client.py
@@ -1,5 +1,5 @@
 import warnings
-from typing import Any, Dict
+from typing import Any, Callable, Dict
 
 from cognite.client import utils
 from cognite.client._api.assets import AssetsAPI
@@ -30,6 +30,8 @@ class CogniteClient:
         max_workers (int): Max number of workers to spawn when parallelizing data fetching. Defaults to 10.
         headers (Dict): Additional headers to add to all requests.
         timeout (int): Timeout on requests sent to the api. Defaults to 30 seconds.
+        token_factory(Callable[[], str]): A method which takes no arguments and returns a token to use for authentication.
+            This will override any api-key set.
         debug (bool): Configures logger to log extra request details to stderr.
     """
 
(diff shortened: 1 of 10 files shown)


sanderland:
Could the user pass something like a:

`class TokenProvider:
    def __init__(self,token):
        self.token = token
    def __call__(self):
        return self.token`

(possibly we could add this class for them and show an example, if we think this is a common use case)

cognite/client/_cognite_client.py
```diff
@@ -30,6 +30,8 @@ class CogniteClient:
         max_workers (int): Max number of workers to spawn when parallelizing data fetching. Defaults to 10.
         headers (Dict): Additional headers to add to all requests.
         timeout (int): Timeout on requests sent to the api. Defaults to 30 seconds.
+        token_factory(Callable[[], str]): A method which takes no arguments and returns a token to use for authentication.
+            This will override any api-key set.
```
NilsBarlaug-cognite: Would we reduce confusion and potential for bugs if we enforce mutual exclusivity instead (throwing an exception if both are set)?

sanderland: I think a single parameter is more pythonesque, and fine if well documented.

erlendvollset: I prefer it overriding. That way I won't need to unset the COGNITE_API_KEY environment variable each time I want to use tokens.

erlendvollset: @sanderland I think he was referring to api-key/token not token_factory/token

sanderland: ah, ignore the above then :)

sanderland: test for string option?

erlendvollset:
fyi @dlcognite

PR #560, opened 2019-10-01
sequences no defaults for insert / no longer assumes order

sanderland:
Different endpoints give columns in different ordering, even a create call does not preserve ordering. This is weird, but apparently as intended. Changed defaults/examples to protect users a bit from our API.


erlendvollset:
Change the PR title so it explains the change you made, not the problem that caused the bug.

erlendvollset:
Do not need to explain why or the how. Just the what.

PR #568, opened 2019-10-08
refactor datapoints and fix bugs in include_outside_points

sanderland:
Based on 
- the bugs in include outside points, 
- aggregate/ts_items being checked sometimes, sometimes not, 
- general query id spaghetti dicts that were growing out of control

datapoints refactored to be hopefully cleaner


sanderland:
- user limit needs a test
- exception raising in one task needs a test

sanderland:
- found bugs through / now giving the same results as cognite-async
- limit + include_outside_points is strange to define, since the API will not respect limit (and return outside points on top of limit). Current behaviour here is to throw away the point 'after' and respect user limit exactly, unlike the API.

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -88,6 +101,43 @@ def test_retrieve_include_outside_points(self, test_time_series):
...
+        dps_non_outside = COGNITE_CLIENT.datapoints.retrieve(id=ts.id, start=start, end=end, limit=1234)
+        count_first = COGNITE_CLIENT.datapoints._post.call_count
+
+        tmp = COGNITE_CLIENT.datapoints._DPS_LIMIT
+        COGNITE_CLIENT.datapoints._DPS_LIMIT = test_lim
+        dps = COGNITE_CLIENT.datapoints.retrieve(
+            id=ts.id, start=start, end=end, include_outside_points=True, limit=1234
+        )
+        COGNITE_CLIENT.datapoints._DPS_LIMIT = tmp
+        assert len(dps) >= len(dps_non_outside)
```
runevalle-cognite: In this case it should be equal?

sanderland: correct

PR #569, opened 2019-10-09
Add devx-team as code owners

fredrik955:


.github/CODEOWNERS
```diff
@@ -1,2 +1,5 @@
 * @erlendvollset
 * @sanderland
+* @fredrik955
+* @polomani
+* hhenriques
```
polomani:
```suggestion
* @hhenriques
```
maybe?

erlendvollset:
I think you can do @cognitedata/devx-team instead of listing individually

polomani:
still don't think it's a good idea..
@cognitedata/devx-team is any of us able to actually review it properly?

erlendvollset:
@polomani This is not a handover. I will still be involved and can answer any questions you may have. For any PR you are not confident in approving you can leave the review to me. I just think it would be good for devx to be more involved so that you can get more comfortable with the code base and doing reviews on this project.

polomani:
@erlendvollset ok, let's start doing that

PR #573, opened 2019-10-14
Change CogniteUpdate to merge instead of overwrite for ._add() and ._remove()

### LINKED ISSUE #524: later calls to timeseriesupdate.metadata.add overwrite the former
**Describe the bug**
the .add method overwrites the .add fields to be posted instead of merging, this is counter-intuitive and restrictive

```
            update = TimeSeriesUpdate(id=ts_old.id)
            update = update.metadata.add( {k: v05_meta[k] for k in missing} )
            update = update.metadata.add( {k: v1_meta[k] for k in added} )
```
**Expected behavior**
not silently throw away the first call, but have later .add calls to overwrite former, or to raise an exception if duplicate keys are given

sanderland:
Should fix #524


cognite/client/data_classes/_base.py
```diff
@@ -260,10 +260,21 @@ def _set_null(self, name):
         self._update_object[name] = {"setNull": True}
 
     def _add(self, name, value):
-        self._update_object[name] = {"add": value}
+        update_obj = self._update_object.get(name, {})
+        if "set" in update_obj:
+            del update_obj["set"]
+        if isinstance(value, list):
+            update_obj["add"] = update_obj.get("add", []) + value
```
erlendvollset: This will fail if the user already has done `update_obj.<name>.add(<dict>)`.

cognite/client/data_classes/_base.py
```diff
@@ -260,10 +260,21 @@ def _set_null(self, name):
...
+        else:
+            update_obj["add"] = {**update_obj.get("add", {}), **value}
```
erlendvollset: This will fail if the user already has done `update_obj.<name>.add(<list>)`.

cognite/client/data_classes/_base.py
```diff
@@ -260,10 +260,21 @@ def _set_null(self, name):
...
+        self._update_object[name] = update_obj
 
     def _remove(self, name, value):
-        self._update_object[name] = {"remove": value}
+        update_obj = self._update_object.get(name, {})
+        if "set" in update_obj:
+            del update_obj["set"]
```
erlendvollset: I think this behaviour is very implicit and not necessarily very intuitive. I would rather this raised an error saying that you cannot remove from something you have already set.

sanderland: good idea!

cognite/client/data_classes/_base.py
sanderland: In what case will a user reasonably do this?

sanderland:
The API allows for you to combine remove and add, so not supporting this and silently dropping is a bit strange.
We could raise an error on double add, but I think this behaviour is fairly safe and intuitive.

cognite/client/data_classes/_base.py
erlendvollset: A reasonable user 😮 ❤️

I just think that a double add is in many cases a user error. Maybe they've mixed up two names. I would prefer raising an error on double anything. Combining add and remove makes sense as you say though.

cognite/client/data_classes/_base.py
```diff
@@ -260,10 +260,19 @@ def _set_null(self, name):
...
+        if isinstance(value, list):
+            update_obj["add"] = update_obj.get("add", []) + value
+        else:
+            update_obj["add"] = {**update_obj.get("add", {}), **value}
+        self._update_object[name] = update_obj
 
     def _remove(self, name, value):
-        self._update_object[name] = {"remove": value}
+        update_obj = self._update_object.get(name, {})
+        assert not ("set" in update_obj), "Can not call remove or add fields after calling set on an update object."
```
erlendvollset: should disallow `set` if `add` or `remove` already exists aswell then.

cognite/client/data_classes/_base.py
```diff
@@ -260,10 +260,19 @@ def _set_null(self, name):
...
+        update_obj["remove"] = sorted(list(set(update_obj.get("remove", [])) | set(value)))
```
erlendvollset: how does this work? not familiar with `|` syntax and for some reason google won't help me 🤔

sanderland: {1,2} | {1,3} == {1,2,3}
set union

cognite/client/data_classes/_base.py
sanderland: I think there's a reasonable case for add in a loop or something, but as long as we give an error it should be easy enough for a user to fix. changes as suggested

PR #576, opened 2019-10-14
Fix chunking bug in list_generator

sanderland:


cognite/client/_api_client.py
```diff
@@ -306,21 +306,22 @@ def _list_generator(
             else:
                 raise ValueError("_list_generator parameter `method` must be GET or POST, not %s", method)
```
haakonvt: typo?

PR #579, opened 2019-10-16
Snippets generator and documentation links

sanderland:
Whenever a link is present in the first line of the docstring, it turns the examples for that function into a code snippet in documentation.

Open questions:
- Multiple endpoints being used/relevant (JS does this with another link)
- Empty line or not between multiple examples.


sanderland:
@erlendvollset it wasn't immediately clear how to cleanly link to two pages without making the sdk docs messy. suggestions?

PR #580, opened 2019-10-16
fix typo, unify str fmt, flake8, whitespace

haakonvt:
Initially; wanted to fix a typo, then the ball started rolling..^^ If any of the "stylistic" changes are unwanted, I'll remove them.

Regarding comparison to `False` or `True`. Do we need this or is it sufficient to use `assert some_value / assert not some_value`?

### DIFF
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -636,7 +636,7 @@ class APIClient:
             error_details["duplicated"] = duplicated
         error_details["headers"] = res.request.headers.copy()
         APIClient._sanitize_headers(error_details["headers"])
-        log.debug("HTTP Error {} {} {}: {}".format(code, res.request.method, res.request.url, msg, extra=error_details))
+        log.debug("HTTP Error {} {} {}: {}".format(code, res.request.method, res.request.url, msg), extra=error_details)
         raise CogniteAPIError(msg, code, x_request_id, missing=missing, duplicated=duplicated, extra=extra)
 
     @staticmethod
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -304,7 +304,7 @@ class APIClient:
                 body = {"filter": filter, "limit": current_limit, "cursor": next_cursor, **(other_params or {})}
                 res = self._post(url_path=resource_path + "/list", json=body, headers=headers)
             else:
-                raise ValueError("_list_generator parameter `method` must be GET or POST, not %s", method)
+                raise ValueError("_list_generator parameter `method` must be GET or POST, not {}".format(method))
             last_received_items = res.json()["items"]
             total_items_retrieved += len(last_received_items)
 
@@ -624,7 +624,7 @@ class APIClient:
                         extra[k] = v
             else:
                 msg = res.content
-        except:
+        except Exception:
             msg = res.content
 
         error_details = {"X-Request-ID": x_request_id}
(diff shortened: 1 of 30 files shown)


cognite/client/_api_client.py
```diff
@@ -636,7 +636,7 @@ def _raise_API_error(res: Response, payload: Dict):
             error_details["duplicated"] = duplicated
         error_details["headers"] = res.request.headers.copy()
         APIClient._sanitize_headers(error_details["headers"])
-        log.debug("HTTP Error %s %s %s: %s", code, res.request.method, res.request.url, msg, extra=error_details)
+        log.debug("HTTP Error {} {} {}: {}".format(code, res.request.method, res.request.url, msg, extra=error_details))
```
sanderland: I think log.debug supports sprintf statements, so maybe keep the original here

tests/tests_integration/test_api/test_sequences_data.py
```diff
@@ -171,7 +167,7 @@ def test_delete_range(self, new_seq_long):
             column_external_ids=new_seq_long.column_external_ids, rows=data, id=new_seq_long.id
         )
         COGNITE_CLIENT.sequences.data.delete_range(start=4, end=15, id=new_seq_long.id)
-        dps = COGNITE_CLIENT.sequences.data.retrieve(start=0, end=None, id=new_seq_long.id)
+        dps = COGNITE_CLIENT.sequences.data.retrieve(start=0, end=None, id=new_seq_long.id)  # noqa F841
```
sanderland: no noqa comments please, can just fix dps = if you want

tests/tests_unit/test_api/test_datapoints.py
```diff
@@ -1284,7 +1276,7 @@ def mock_get_dps_count(rsps):
     def request_callback(request):
         payload = jsgz_load(request.body)
         granularity = payload["granularity"]
-        aggregates = payload["aggregates"]
+        # aggregates = payload["aggregates"]  # TODO: Unused
```
sanderland: just remove

tests/tests_unit/test_api/test_relationships.py
```diff
@@ -131,7 +128,7 @@ def test_create_single_types(self, mock_rel_response):
 
     def test_create_wrong_type(self, mock_rel_response):
         with pytest.raises(ValueError):
-            res = REL_API.create(
+            _ = REL_API.create(
```
sanderland: remove _=

haakonvt: Removed all occurences 👍

tests/tests_integration/test_api/test_sequences_data.py
haakonvt: I added it since `dps` is used in the out-commented lines below. Removing `noqa`, but keeping assignment to `dps`.

cognite/client/_api_client.py
haakonvt: Some lines below we have the following:
`log.debug("HTTP/{} {} {} {}".format(http_protocol_version, method, url, status_code), extra=extra)`

erlendvollset:
Sweet! Did flake8 pick up the whitespace/missing import stuff? In that case maybe we could add a check in jenks and a pre-commit hook for flake8?

Regarding comparison to true/false, I prefer `assert x is False` since `assert not x` will evaluate to False if x is empty string, `None`, or 0.

haakonvt:
> Sweet! Did flake8 pick up the whitespace/missing import stuff? In that case maybe we could add a check in jenks and a pre-commit hook for flake8?
> 
> Regarding comparison to true/false, I prefer `assert x is False`/`assert x is True` since `assert not x` will evaluate to False if x is empty string, `None`, or 0.

Yeah! I'm a fan of `flake8`, but this repo would need some changes before we could add it as an enforced merge requirement. I'll make another draft PR 😊

PR #590, opened 2019-10-24
Sub type generation

runevalle-cognite:
Code generation recognizes objects having schemas at "resource level" and assigns schema names to these. These were formerly Dict[str, Any]. These are still dict, with just a layer of __get/setattr__ to make them behave more like the schema. Hence no code should break.

Also make typehint for metadata field Dict[str, str] instead of Dict[str, Any]

Even if all of this should be non-breaking, we might want to consider bumping minor version due to these changes.

### DIFF
--- a/cognite/client/_api/assets.py
+++ b/cognite/client/_api/assets.py
@@ -41,8 +41,8 @@ class AssetsAPI(APIClient):
             root_external_ids (List[str], optional): List of root external ids to filter on.
             metadata (Dict[str, str]): Custom, application specific metadata. String key -> String value
             source (str): The source of this asset
-            created_time EpochTimestampRange: Range between two timestamps
-            last_updated_time EpochTimestampRange: Range between two timestamps
+            created_time (EpochTimestampRange): Range between two timestamps
+            last_updated_time (EpochTimestampRange): Range between two timestamps
             root (bool): filtered assets are root assets or not
             external_id_prefix (str): Filter by this (case-sensitive) prefix for the external ID.
             aggregated_properties (List[str]): Set of aggregated properties to include.
@@ -166,8 +166,8 @@ class AssetsAPI(APIClient):
             root_external_ids (List[str], optional): List of root external ids to filter on.
             metadata (Dict[str, str]): Custom, application specific metadata. String key -> String value.
             source (str): The source of this asset.
-            created_time EpochTimestampRange: Range between two timestamps.
-            last_updated_time EpochTimestampRange: Range between two timestamps.
+            created_time (EpochTimestampRange): Range between two timestamps.
+            last_updated_time (EpochTimestampRange): Range between two timestamps.
             root (bool): filtered assets are root assets or not.
             external_id_prefix (str): Filter by this (case-sensitive) prefix for the external ID.
             aggregated_properties (List[str]): Set of aggregated properties to include.
--- a/cognite/client/_api/assets.py
+++ b/cognite/client/_api/assets.py
@@ -44,7 +44,7 @@ class AssetsAPI(APIClient):
             created_time EpochTimestampRange: Range between two timestamps
             last_updated_time EpochTimestampRange: Range between two timestamps
             root (bool): filtered assets are root assets or not
-            external_id_prefix (str): External Id provided by client. Should be unique within the project
+            external_id_prefix (str): Filter by this (case-sensitive) prefix for the external ID.
             aggregated_properties (List[str]): Set of aggregated properties to include.
             limit (int, optional): Maximum number of assets to return. Defaults to return all items.
 
@@ -169,7 +169,7 @@ class AssetsAPI(APIClient):
             created_time EpochTimestampRange: Range between two timestamps.
             last_updated_time EpochTimestampRange: Range between two timestamps.
             root (bool): filtered assets are root assets or not.
-            external_id_prefix (str): External Id provided by client. Should be unique within the project.
+            external_id_prefix (str): Filter by this (case-sensitive) prefix for the external ID.
             aggregated_properties (List[str]): Set of aggregated properties to include.
             partitions (int): Retrieve assets in parallel using this number of workers. Also requires `limit=None` to be passed.
             limit (int, optional): Maximum number of assets to return. Defaults to 25. Set to -1, float("inf") or None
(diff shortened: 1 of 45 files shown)


cognite/client/_api/assets.py
```diff
@@ -39,10 +39,10 @@ def __call__(
             parent_ids (List[int]): No description.
             root_ids (List[int], optional): List of root ids ids to filter on.
             root_external_ids (List[str], optional): List of root external ids to filter on.
-            metadata (Dict[str, Any]): Custom, application specific metadata. String key -> String value
+            metadata (Dict[str, str]): Custom, application specific metadata. String key -> String value
             source (str): The source of this asset
-            created_time (Dict[str, Any]): Range between two timestamps
-            last_updated_time (Dict[str, Any]): Range between two timestamps
+            created_time EpochTimestampRange: Range between two timestamps
```
sanderland: ()

cognite/client/_api/assets.py
```diff
@@ -147,10 +147,10 @@ def list(
         parent_ids: List[int] = None,
         root_ids: List[int] = None,
         root_external_ids: List[str] = None,
-        metadata: Dict[str, Any] = None,
+        metadata: Dict[str, str] = None,
```
sanderland: we do allow other types in metadata values even if they're persisted as strings, no?

cognite/client/data_classes/shared.py
```diff
@@ -0,0 +1,25 @@
...
+    """Range between two timestamps.
+
+    Args:
+        max (int): The number of milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds.
+        min (int): The number of milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds.
+        cognite_client (CogniteClient): The client to associate with this object.
+    """
+
+    def __init__(self, max: int = None, min: int = None, cognite_client=None):
+        self.max = max
```
sanderland: since we have a centralized point here, could we support datetime objects?

cognite/client/data_classes/shared.py
```diff
@@ -0,0 +1,25 @@
...
+        self.min = min
+        self._cognite_client = cognite_client
+
+    # GenStop
+    def __getattr__(self, item):
```
sanderland: why not a  @ property or something?

tests/tests_unit/test_api/test_files.py
```diff
@@ -179,6 +179,16 @@ def test_list_root_ids_list(self, mock_files_response):
             "filter": {"rootAssetIds": [{"id": 1}, {"id": 2}, {"externalId": "a"}]},
         } == jsgz_load(calls[0].request.body)
 
+    def test_list_with_time_dict(self, mock_files_response):
+        FILES_API.list(created_time={"min": 20})
```
sanderland: if the user can do this, is the type hint in the function incorrect?

cognite/client/_api/assets.py
runevalle-cognite: Good catch

cognite/client/_api/assets.py
runevalle-cognite: While we do allow it I guess this shouldn't be encouraged? I don't know of anywhere we use any sort of implisit casting our self, and then I guess it is good manners to suggest users serializing before putting into the dictionary, rather than rely on us doing it the way they would expect? Type hints are not enforced, so if someone is adventurous this shouldn't break anything at least.

cognite/client/data_classes/shared.py
runevalle-cognite: The why is me being a python newb. Will fix.

tests/tests_unit/test_api/test_files.py
runevalle-cognite: I would say that the type hint is correct, but that our api is permissive of still using the old code. This test is here to confirm that user can still use the old way (not breaking code), while we encourage going over to the more type safe way.

cognite/client/_api/assets.py
sanderland: we do allow this in e.g. all filters and it's pretty standard, since people don't want to import 25 different objects for every simple call

cognite/client/data_classes/shared.py
sanderland: fair enough :)

cognite/client/data_classes/shared.py
sanderland: actually I think you can leave it out completely and it will work, the defaults on __init__ take care of None, no?

tests/tests_unit/test_api/test_files.py
sanderland: we do allow this in e.g. all filters and it's a pretty standard we support. i think typehints there have Union as well

runevalle-cognite: Ah, ok. Can allow union for now.

cognite/client/data_classes/shared.py
runevalle-cognite: Property si having other nice properties I didn't feel like I managed to duplicate with this method as well, so I was quite motivated to fix.

runevalle-cognite:
> some quick comments form a first read.
> One question is whether we want the type names to match the API classes (EpochTimestampRange is a bit strange compared to other SDK classes for example)

It might be nice to introduce a map from schema name to SDK name at some point. I just didn't think it was necessary with the current batch where the schema names seem to be sane. I am not sure what name you would think would be better match for the SDK than EpochTimestampRange?

The Relationship schema did have a schema without caps name, so that is the background for the is upper filter. With a mapping this could be included.

openapi/tests/input_output/input.py
```diff
@@ -22,6 +22,22 @@ class CogniteListUpdate:
     pass
 
 
+class CognitePropertyClassUtil:
```
sanderland: why does this need mocking instead of importing?

runevalle-cognite: I don't want to import anything from the cognite.client here due to the added dependency. Mocked instead of just declared as it is called at collection stage by generated code.

PR #594, opened 2019-10-30
chore: change code snippets label name [DEVX-157]

polomani:


polomani:
yeah, first contribution!! 🎊

PR #596, opened 2019-11-01
[DT-513] Breaking changes Model Hosting v1

HMEiding:
This PR updates the model hosting client according to the incoming breaking changes of Nostromo v1. I think it would be a good idea to trigger an alpha-release from this branch for testing purposes before merging to master.


erlendvollset:
> This PR updates the model hosting client according to the incoming breaking changes of Nostromo v1. I think it would be a good idea to trigger an alpha-release from this branch for testing purposes before merging to master.

Why do you need to make an alpha-release? That will block our ability to release other stuff in the mean time. Integration tests in the sdk run against greenfield, so that should be all the testing you need.

cognite/client/data_classes/model_hosting/schedules.py
```diff
@@ -7,9 +7,8 @@ class Schedule(CogniteResource):
     """A representation of a Schedule in the model hosting environment.
 
     Args:
-        id (int): Id of the schedule.
-        model_id (int): The id of the model associated with this schedule.
         name (str): Name of the schedule.
```
wjoel: For consistency we should maybe use `schedule_name` here as well?

cognite/client/data_classes/model_hosting/models.py
```diff
@@ -57,13 +54,12 @@ class ModelVersion(CogniteResource):
     """A representation of a Model version in the model hosting environment.
 
     Args:
-        id (int): Id of the model version.
+        name (str): Name of the model version.
```
wjoel: `version_name` maybe?

cognite/client/_api/model_hosting/schedules.py
```diff
@@ -17,8 +17,8 @@ def create_schedule(
         """Create a new schedule on a given model.
 
         Args:
-            model_id (int): Id of model to create schedule on
-            name (str): Name of schedule
+            model_name (str): Name of model to create schedule on
+            schedule_name (str): Name of schedule
```
wjoel: If we decide to leave the others as `name`, this one should be `name` instead of `schedule_name`.

tests/tests_integration/test_api/test_model_hosting/test_source_packages.py
```diff
@@ -12,6 +12,7 @@
 SOURCE_PACKAGES_API = CogniteClient().model_hosting.source_packages
 
 
+@pytest.mark.skip(reason="wait for model hosting v1")
```
wjoel: Shouldn't/couldn't this run in playground and be ok?

cognite/client/_api/model_hosting/schedules.py
HMEiding: I agree.

tests/tests_integration/test_api/test_model_hosting/test_source_packages.py
HMEiding: All methods now expect that the "data"-wrapper is removed from the json-bodies and will therefore fail until we deploy v1.

erlendvollset: All the integration tests run against Greenfield. So this should pass.

HMEiding:
> > This PR updates the model hosting client according to the incoming breaking changes of Nostromo v1. I think it would be a good idea to trigger an alpha-release from this branch for testing purposes before merging to master.
> 
> Why do you need to make an alpha-release? That will block our ability to release other stuff in the mean time. Integration tests in the sdk run against greenfield, so that should be all the testing you need.

@erlendvollset It is perhaps not strictly necessary, but @wjoel suggested that we make an alpha-release directly from this branch by editing the Jenkinsfile. This way other people could pip install and test Model Hosting on Greenfield. But people could of course just git clone instead.

cognite/client/data_classes/model_hosting/models.py
HMEiding: Agreed. So I think we should have `name` only when there is a single name in the method.

When there are more than one name i a method, I think it makes sence to keep a consistent order throughout. I suggest `model_name` before `version_name` and the same for schedules.

cognite/client/data_classes/model_hosting/schedules.py
HMEiding: Here as well, since this is a resource object, it needs to match the json returned from Nostromo.

HMEiding:
@erlendvollset , @wjoel : I've made a couple of changes to the API, and I'm currently updating the examples in cognite-python-docs. Can you have a look to see that we agree on the latest changes?

cognite/client/_api/model_hosting/schedules.py
wjoel: Also it should maybe say "the model" and "the schedule", not sure.

cognite/client/data_classes/model_hosting/models.py
wjoel: Hmm, ok.

cognite/client/_api/model_hosting/schedules.py
HMEiding: Yes, agree.

docs/source/cognite.rst
```diff
@@ -897,63 +897,63 @@ Delete model
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-.. automethod:: cognite.client._api.model_hosting.models.ModelsAPI.deploy_model_version
+.. automethod:: cognite.client._api.model_hosting.versions.ModelVersionsAPI.deploy_model_version
 
 Create model version without deploying
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-.. automethod:: cognite.client._api.model_hosting.models.ModelsAPI.create_model_version
+.. automethod:: cognite.client._api.model_hosting.versions.ModelVersionsAPI.create_model_version
 
 Deploy awaiting model version version
```
erlendvollset:
```suggestion
Deploy awaiting model version
```

tests/tests_integration/test_api/test_model_hosting/test_schedules.py
```diff
@@ -10,47 +10,47 @@
 @pytest.mark.skip(reason="temporarily disable while model hosting v1 is being tested in greenfield")
```
erlendvollset: Should be removed?

HMEiding: Indeed

PR #597, opened 2019-11-01
sequences create from dataframe function

sanderland:
inspired by @cognite-fholm 's boat case.
Particularly I'd like someone to check the function signature/options on how many things we do by default / have flags for.


cognite/client/_api/sequences.py
```diff
@@ -185,6 +186,74 @@ def list(
...
+            if allow_integers and np.issubdtype(t, np.integer):
+                return "LONG"
+            elif np.issubdtype(t, np.number):
+                return "DOUBLE"
+            else:
+                return "STRING"
+
+        return [{"externalId": col, "name": col, "valueType": map_type(t)} for col, t in dataframe.dtypes.iteritems()]
+
+    def create_from_dataframe(
```
erlendvollset: Would it be better to piggyback on insert_dataframe()?

i.e. adding a `create_sequence_if_missing` parameter to SequencesDataApi.insert_dataframe().

cognite/client/_api/sequences.py
```diff
@@ -185,6 +186,74 @@ def list(
...
+        force_create: bool = False,
+        clear_existing_data: bool = False,
+        *args
+    ):
+        """`Create and populate a sequence from a pandas dataframe . <https://docs.cognite.com/api/v1/#operation/createSequence>`_
+
+                Args:
+                    dataframe (pandas.DataFrame): the data to insert. column names will be used for column name and externalId, column dtypes will be used to determine valueType.
+                    external_id (str): external id of the sequence.
+                    allow_integers (bool):: allow the valueType LONG if the dtype of the column indicates this. if False, reverts to DOUBLE.
```
erlendvollset:
```suggestion
                    allow_integers (bool): allow the valueType LONG if the dtype of the column indicates this. if False, reverts to DOUBLE.
```

cognite/client/_api/sequences.py
```diff
@@ -185,6 +186,74 @@ def list(
...
+                    force_create (bool):: if sequence exists and column types do not match, re-create the sequence.
```
erlendvollset:
```suggestion
                    force_create (bool): if sequence exists and column types do not match, re-create the sequence.
```

erlendvollset: Also, what does "column types do not match" mean? This explanation should be improved.

cognite/client/_api/sequences.py
```diff
@@ -185,6 +186,74 @@ def list(
...
+                        >>> seq = c.sequences.create_from_dataframe(df,force_create=True,clear_existing_data=True)
+                """
+        cols = self._dataframe_to_columns_definition(dataframe=dataframe, allow_integers=allow_integers)
+        existing_seq = self.retrieve(external_id=external_id)
+        if existing_seq:
+            existing_cols_map = {col["externalId"]: col["valueType"] for col in existing_seq.columns}
+            df_cols_map = {col["externalId"]: col["valueType"] for col in cols}
+            if existing_cols_map != df_cols_map:
+                if not force_create:
+                    raise CogniteCompatibilityError(
```
erlendvollset: This exception should have a more specific name. `IncompatibleColumnTypes` for example

cognite/client/_api/sequences.py
```diff
@@ -185,6 +186,74 @@ def list(
...
+        *args
+    ):
+        """`Create and populate a sequence from a pandas dataframe . <https://docs.cognite.com/api/v1/#operation/createSequence>`_
+
+                Args:
+                    dataframe (pandas.DataFrame): the data to insert. column names will be used for column name and externalId, column dtypes will be used to determine valueType.
+                    external_id (str): external id of the sequence.
+                    allow_integers (bool):: allow the valueType LONG if the dtype of the column indicates this. if False, reverts to DOUBLE.
+                    force_create (bool):: if sequence exists and column types do not match, re-create the sequence.
+                    clear_existing_data (bool): if sequence exists, clear all existing data before inserting.
```
erlendvollset: This would be natural to include for `insert_dataframe` aswell.

cognite/client/_api/sequences.py
erlendvollset: And maybe `force_recreate` is better?

cognite/client/_api/sequences.py
sanderland: interesting idea, I'm not sure.

cognite/client/_api/sequences.py
sanderland: For example, someone added a column to their CSV, or their column that formerly looked like all integers is now shown to have doubles.

sanderland:
removed allow_integers as it's probably too specific and confusing, and force_recreate kind of solves the issue.
made clear_existing_data default as I think the common use case probably has this true.
after thinking about it, I think we shouldn't delete / create objects in insert_dataframe.

cognite/client/_api/sequences.py
```diff
@@ -185,6 +186,72 @@ def list(
...
+                    )
+                else:
+                    self.delete(external_id=external_id)
+                    existing_seq = None
+            else:
+                if clear_existing_data:
+                    self.data.delete_range(external_id=external_id, start=0, end=None)
+
+        if not existing_seq:
+            created_seq = self.create(Sequence(external_id=external_id, columns=cols, *args))
```
erlendvollset: You cannot specify args after kwargs in Python, so this will fail.

tests/tests_integration/test_api/test_sequences.py
```diff
@@ -70,3 +72,27 @@ def test_get_new(self, new_seq):
...
+        seq = COGNITE_CLIENT.sequences.create_from_dataframe(dataframe=df, external_id=extid, force_recreate=True)
+        assert 3 == len(seq.columns)
+        assert "LONG" == seq.columns[0]["valueType"]
+        assert "DOUBLE" == seq.columns[1]["valueType"]
+        assert "STRING" == seq.columns[2]["valueType"]
+
+        with pytest.raises(IncompatibleColumnTypesError):
+            COGNITE_CLIENT.sequences.create_from_dataframe(dataframe=df2, external_id=extid)
+
+        COGNITE_CLIENT.sequences.create_from_dataframe(dataframe=df2, external_id=extid, force_recreate=True)
```
erlendvollset: This only tests that this doesn't raise an exception. Bare minimum would be to retrieve the dataframe and ensure that it is correct.

We also in essence have multiple tests within the same test here, which is not good.

tests/tests_integration/test_api/test_sequences.py
```diff
@@ -70,3 +72,27 @@ def test_get_new(self, new_seq):
...
+        COGNITE_CLIENT.sequences.delete(external_id=extid)
```
erlendvollset: Setup and teardown should happen in a fixture, so that the cleanup is executed even if the test fails at some point.

cognite/client/_api/sequences.py
```diff
@@ -185,6 +186,72 @@ def list(
...
+        return [{"externalId": col, "name": col, "valueType": map_type(t)} for col, t in dataframe.dtypes.iteritems()]
+
+    def create_from_dataframe(
+        self,
+        dataframe: "pandas.DataFrame",
+        external_id: str,
+        force_recreate: bool = False,
+        clear_existing_data: bool = True,
+        *args
+    ):
```
erlendvollset: Missing return type

tests/tests_integration/test_api/test_sequences.py
sanderland: we basically never test retrieve right after insert, due to the delay.
the test sets up a state that's necessary for the other tests which makes it hard to disentangle.

cognite/client/_api/sequences.py
sanderland: you're right, should be kwargs, fixed

tests/tests_integration/test_api/test_sequences.py
sanderland: but setup is kind of what's being tested, so then the fixture will just fail?

cognite/client/_api/sequences.py
sanderland: fixed

sanderland:
happy birthday! :cake:

PR #598, opened 2019-11-05
Fix issue with trailing slashes in base url for login endpoint

mathialo:
When a client is configured with a trailing slash in the base url, the login endpoint was breaking due to the `urljoin` in `_get_base_url_with_base_path` joining with an empty string, leaving the trailing slash behind.


mathialo:
> should the rstrip just be in _get_base_url_with_base_path then? or even the ClientConfig constructor?

Sure, I'll move it. If it's in the ClientConfig, we fix the problem once, and not every API call, which is better IMO.

PR #599, opened 2019-11-05
chore: service-contracts auto-pr with updated code snippets [DEVX-168]

polomani:


Jenkinsfile
```diff
@@ -106,10 +106,18 @@ podTemplate(
             println("This version: " + currentVersion)
             def versionExists = sh(returnStdout: true, script: 'pipenv run python3 cognite/client/utils/_version_checker.py -p cognite-sdk -v ' + currentVersion).trim()
             println("Version Exists: " + versionExists)
+
             if (env.BRANCH_NAME == 'master' && versionExists == 'no') {
                 stage('Release') {
                     sh("pipenv run twine upload --config-file /pypi/.pypirc dist/*")
                 }
+                stage('After release') {
```
sanderland: This is the name that shows up in jenkins right, can we name it something sensible like 'Update code snippets'

polomani: done

generate_code_snippets.py
```diff
@@ -51,4 +51,7 @@ def collect_apis(obj, done):
             if openapi_ident in duplicate_operations:
                 snippets["operations"][duplicate_operations[openapi_ident]] += code
 
-print(json.dumps(snippets, indent=4))
+filename = "python-sdk-examples.json"
```
fredrik955: I would recommend to run the script with `python3 generate_code_snippets.py > python-sdk-examples.json` instead. More flexible and less code

polomani: fixed

Jenkinsfile
```diff
@@ -106,10 +106,18 @@ podTemplate(
             println("This version: " + currentVersion)
             def versionExists = sh(returnStdout: true, script: 'pipenv run python3 cognite/client/utils/_version_checker.py -p cognite-sdk -v ' + currentVersion).trim()
             println("Version Exists: " + versionExists)
+
             if (env.BRANCH_NAME == 'master' && versionExists == 'no') {
                 stage('Release') {
                     sh("pipenv run twine upload --config-file /pypi/.pypirc dist/*")
                 }
+                stage('Update code snippets on service-contracts') {
+                    sh("pip3 install requests")
```
erlendvollset: Requests should already be installed.

Jenkinsfile
```diff
@@ -106,10 +106,18 @@ podTemplate(
...
+                    sh("python3 generate_code_snippets.py > python-sdk-examples.json")
```
erlendvollset:
```suggestion
                    sh("python3 scripts/generate_code_snippets.py > python-sdk-examples.json")
```
?

Jenkinsfile
polomani: you sure? it did fail for me last time though

polomani: ok, we realised that I need `pipenv`

PR #604, opened 2019-11-12
Typing API

sanderland:
TODO: 
* TypeUpdate after behaviour is fixed
* AssetUpdate to types discussion on how to handle this


polomani:
@sanderland do you still need this?

sanderland:
@polomani this is in https://github.com/cognitedata/cognite-sdk-python-experimental now, so should be fine. was just leaving the branch in case people were using it.

polomani:
ok, we'll keep the branch then

PR #607, opened 2019-11-13
Add concept of writable fields to openapi spec and generated classes

sanderland:


polomani:
I don't have much context, but what's the point of having spec with 57k lines of code?
can you read it now? can you read it if changes?
Instead, can you extract only needed parts for test purposes?

sanderland:
> I don't have much context, but what's the point of having spec with 57k lines of code?
> can you read it now? can you read it if changes?
> Instead, can you extract only needed parts for test purposes?

the current tests fail if the spec changes in service contracts since it grabs it online and tests that regenerating doesn't cause changes. this is annoying. adding a fixed spec to the repo stops this.

runevalle-cognite:
> > I don't have much context, but what's the point of having spec with 57k lines of code?
> > can you read it now? can you read it if changes?
> > Instead, can you extract only needed parts for test purposes?
> 
> the current tests fail if the spec changes in service contracts since it grabs it online and tests that regenerating doesn't cause changes. this is annoying. adding a fixed spec to the repo stops this.

Is it any reason to want to have the spec synced with actual spec? Else might it make more sense to just use a lot shorter dummy spec for the code generation testing? I feel like adding the full spec gives the worst of both worlds: Missing sync with anything "real", and still having a huge unmaintainable blob for new developers to wonder if is safe to touch..

PR #612, opened 2019-11-18
Make disable_pypi_version_check configurable from client creation

mathialo:
Add an option to turn off the version check from an argument to init

### DIFF
--- a/cognite/client/_cognite_client.py
+++ b/cognite/client/_cognite_client.py
@@ -1,5 +1,5 @@
 import warnings
-from typing import Any, Callable, Dict, Union
+from typing import Any, Callable, Dict, Optional, Union
 
 from cognite.client import utils
 from cognite.client._api.assets import AssetsAPI
@@ -49,7 +49,7 @@ class CogniteClient:
         headers: Dict[str, str] = None,
         timeout: int = None,
         token: Union[str, Callable[[], str], None] = None,
-        disable_pypi_version_check: bool = None,
+        disable_pypi_version_check: Optional[bool] = None,
         debug: bool = False,
     ):
         self._config = ClientConfig(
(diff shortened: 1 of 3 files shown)


cognite/client/_cognite_client.py
```diff
@@ -48,6 +49,7 @@ def __init__(
         headers: Dict[str, str] = None,
         timeout: int = None,
         token: Union[str, Callable[[], str], None] = None,
+        disable_pypi_version_check: bool = None,
```
erlendvollset:
```suggestion
        disable_pypi_version_check: bool = False,
```

mathialo: I agree thats a bit clunky, but then how we detect no argument so we can fall back to env vars?

erlendvollset: Ah right. Set type hint to Optional[bool] in that case.

cognite/client/utils/_client_config.py
```diff
@@ -75,6 +76,7 @@ def __init__(
         self.headers = headers or self.headers
         self.timeout = timeout or self.timeout
         self.token = token
+        self.disable_pypi_version_check = disable_pypi_version_check or self.disable_pypi_version_check
```
erlendvollset:
```suggestion
        self.disable_pypi_version_check = disable_pypi_version_check if disable_pypi_version_check is not None else self.disable_pypi_version_check
```

cognite/client/utils/_client_config.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         headers: Dict[str, str] = None,
         timeout: int = None,
         token: Union[Callable[[], str], str] = None,
+        disable_pypi_version_check: bool = None,
```
erlendvollset: Optional[bool]

cognite/client/utils/_client_config.py
erlendvollset: Will use envvar when this is false.

sanderland:
my inclination is to call the parameter `check_version` with default True, rather than this double negative.

mathialo:
> my inclination is to call the parameter `check_version` with default True, rather than this double negative.

I don't disagree (while on the topic of double negatives...), my reasoning was to keep it equal to the existing envvar-way of disabling it.

PR #614, opened 2019-11-19
Calling Model Hosting endpoints in playground + release v1.4.5

HMEiding:


tests/tests_unit/test_api_client.py
```diff
@@ -903,12 +903,12 @@ class TestHelpers:
         "input, emulator_url, expected",
         [
             (
-                "http://localtest.com/api/1.0/projects/test-project/analytics/models",
+                "http://localtest.com/api/1.0/projects/test-project/modelhosting/models",
```
wjoel: We're mixing 1.0 (which I guess is Nostromo 1.0, not CDF v1) and 0.1 here, looks a bit odd?

cognite/client/_api_client.py
```diff
@@ -676,13 +674,3 @@ def _sanitize_headers(headers: Optional[Dict]):
             headers["api-key"] = "***"
         if "Authorization" in headers:
             headers["Authorization"] = "***"
-
-    def _apply_model_hosting_emulator_url_filter(self, full_url):
```
HMEiding: There's no reason to keep this any longer, right @erlendvollset ?

CHANGELOG.md
```diff
@@ -19,12 +19,13 @@ Changes are grouped as follows
 - Upserts for all resource types
 - Separate read/write fields on data classes
 
-## Unreleased
+## [1.4.4] - 2019-11-29
```
sanderland: note 1.4.4 is out now, this should be bumped when you decide to merge

PR #618, opened 2019-11-27
dont camel case notebook output / asset subtree filtering / 1.4.2

sanderland:
https://cognitedata.slack.com/archives/CG10VQPFX/p1574830580236400

then openapi generator tests and signature tests fail, needed to implement filter options to pass

### DIFF
--- a/tests/tests_unit/test_api/test_assets.py
+++ b/tests/tests_unit/test_api/test_assets.py
@@ -122,6 +122,16 @@ class TestAssets:
             calls[0].request.body
         )
 
+    def test_list_subtree(self, mock_assets_response):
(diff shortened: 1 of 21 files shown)


tests/tests_unit/test_api/test_assets.py
```diff
@@ -122,6 +122,16 @@ def test_list_root(self, mock_assets_response):
             calls[0].request.body
         )
 
+    def test_list_subtree(self, mock_assets_response):
+        ASSETS_API.list(asset_subtree_ids=1, asset_subtree_external_ids=["a"], limit=10)
```
erlendvollset: We should use these subtree filter params in Asset.subtree() and AssetsApi.retrieve_subtree().

PR #625, opened 2019-12-02
gremlin queries

sanderland:


sanderland:
> I notice that the numeric fields in the filter take dicts, which is in line with the sdk asset filter, but seems strange, considering the dicts should always contain just min and/or max. Would min_confidence and max_confidence as separate fields make more sense? Seems easier to understand from a user perspective.

Yes, these filters are auto generated from the spec, any fields which are not top level turn into dicts. not super great to document, but consistent throughout the api

cognite/client/_api/relationships.py
```diff
@@ -244,3 +244,20 @@ def delete(self, external_id: Union[str, List[str]]) -> None:
...
+
+        Returns:
+            Dict: Raw JSON result of the query.
+        """
+        utils._auxiliary.assert_type(filter, "filter", [dict, RelationshipFilter], allow_none=True)
+        if isinstance(filter, RelationshipFilter):
+            filter = filter.dump(camel_case=True)
+        elif isinstance(filter, dict):
+            filter = utils._auxiliary.convert_all_keys_to_camel_case(filter)
+        return self._post(url_path="/restrictedGraphQuery", json={"items": [{"query": query, "filter": filter}]}).json()
```
erlendvollset: Could we have a data class for this response instead?

sanderland: Such as? The output is rather a mess and not worth investing much time to model until that changes.

PR #626, opened 2019-12-02
Synthetic time series query in playground/experimental + 1.4.7

sanderland:


cognite/client/_api/datapoints.py
```diff
@@ -8,9 +8,11 @@
 
 import cognite.client.utils._time
 from cognite.client import utils
+from cognite.client._api.synthetic_time_series import SyntheticDatapointsAPI
 from cognite.client._api_client import APIClient
 from cognite.client.data_classes import Datapoints, DatapointsList, DatapointsQuery
 from cognite.client.exceptions import CogniteAPIError
+from cognite.client.utils._experimental_warning import experimental_api
```
erlendvollset: not used?

cognite/client/_api/datapoints.py
```diff
@@ -22,6 +24,9 @@ def __init__(self, *args, **kwargs):
         self._DPS_LIMIT = 100000
         self._POST_DPS_OBJECTS_LIMIT = 10000
         self._RETRIEVE_LATEST_LIMIT = 100
+        self.synthetic = SyntheticDatapointsAPI(
```
erlendvollset: Doesn't this mean it will be exposed on the CogniteClient from the non-experimental namespace?

cognite/client/_api/synthetic_time_series.py
```diff
@@ -0,0 +1,55 @@
...
+            end (Union[int, str, datetime]): Exclusive end.
+
+        Returns:
+            Datapoints: A Datapoints object containing the calculated data.
+
+        Examples:
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> dps = c.datapoints.synthetic.retrieve(expression="TS{id:123} + TS{externalId:'abc''}", start="2w-ago", end="now")
```
erlendvollset:
```suggestion
                >>> dps = c.datapoints.synthetic.retrieve(expression="TS{id:123} + TS{externalId:'abc'}", start="2w-ago", end="now")
```

cognite/client/_api/synthetic_time_series.py
```diff
@@ -0,0 +1,55 @@
...
+            expression (str): Function to be calculated.
+            start (Union[int, str, datetime]): Inclusive start.
+            end (Union[int, str, datetime]): Exclusive end.
+
+        Returns:
+            Datapoints: A Datapoints object containing the calculated data.
+
+        Examples:
+
+                >>> from cognite.client import CogniteClient
```
erlendvollset:
```suggestion
                >>> from cognite.client.experimental import CogniteClient
```

sanderland: fixed

sanderland:
Since when do we add integration tests on experimental apis?

erlendvollset:
Since when do we not?

sanderland:
> 
> 
> Since when do we not?

relationships does not. sequences did not until some level of stability was reached. 
this is a good thing imho, there are already enough tests that force you to (e.g.) implement all changes to assets before you can merge anything else. 
blocking PRs until all unstable playground apis are patched, particularly when reviews can take weeks, seems rather uneccesary.

erlendvollset:
You can always mark tests as skipped. This shows up when running the test suite, so developers will be aware of the fact that something is not working and can fix it ASAP. Bug fixes should always have priority over new features. Not knowing that something is broken, and letting the users find out for you is the worst case scenario. In my opinion everything should be integration tested. If something is breaking so often that we don't want to maintain it, we shouldn't add it to the SDK.

If it is taking weeks for PRs to be reviewed, we should address it. @f1cognite @maciejmrowiec

hhenriques:
I'm of the opinion that unstable APIs shouldn't be in the production SDK. However, for now, it's what we have. We can not have extensive test coverage for playground features on the SDK while the endpoint is still unstable, once we have an indication that the API won't suffer any changes then we can focus on integration testing.
This isn't an ideal situation and we have plans for how to support playground in SDK. But that's still something we're designing. 

@erlendvollset, do you have any objection (other than the test) not to merge this PR?

PR #627, opened 2019-12-03
more notebook output fixes / 1.4.6

sanderland:


cognite/client/data_classes/raw.py
```diff
@@ -24,7 +24,7 @@ def __init__(
         self._cognite_client = cognite_client
 
     # GenStop
-    def to_pandas(self):
+    def to_pandas(self, **kwargs):
```
erlendvollset: Should really try to avoid using kwargs here. It makes for some unexpected behaviour on the users part. They will pass in arguments that are never used and not understand why they don't have an effect, I've seen it many times before.

Why do you need it?

sanderland: These derived objects implement to_pandas in a way which  already accepted fewer arguments than the base class before the camel case change (ignore, expand).  Adding some parameters explicitly with docstring 'this is ignored' didn't seem right either.

erlendvollset: Is it because these methods are called from the superclass' _repr_html_ method with the camel_case argument?

sanderland: yep, that's why they all need to accept camel_case.

erlendvollset:
I suggest adding some tests for _repr_html_.. Maybe there are more corner cases we haven't thought of that we might catch.

amorken:
Checking in from onboarding, just stumbled across this bug when going through https://cognite.talentlms.com/learner/courseinfo/id:133
Any idea when the fixed release will be visible in PyPi? Or alternatively, what's the latest release that isn't impacted by this issue? =)

Thanks,
-@M

sanderland:
@amorken when build succeeds it should be out in pypi, jenkins is a bit troublesome recently

amorken:
Alright, thanks @sanderland, and I appreciate the pointer to .to_pandas() =)

amorken:
1.4.6 from PyPi tested and found to be in good working order. Thanks!

PR #633, opened 2019-12-11
support for ignore unknown ids in time series

sanderland:


cognite/client/_api/datapoints.py
```diff
@@ -84,7 +86,7 @@ def retrieve(
                 ...                         external_id={"externalId": "1", "aggregates": ["max"]},
                 ...                         start="1d-ago", end="now", granularity="1h")
         """
-        fetcher = DatapointsFetcher(client=self)
+        fetcher = DatapointsFetcher(client=self, ignore_unknown_ids=ignore_unknown_ids)
```
erlendvollset: To stay aligned with how we handle all other parameters, it should rather be an argument to `DatapointsQuery`.

cognite/client/_api/datapoints.py
```diff
@@ -99,7 +101,7 @@ def retrieve(
             limit=limit,
         )
         dps_list = fetcher.fetch(query)
-        if is_single_id:
+        if is_single_id and not ignore_unknown_ids:
             return dps_list[0]
```
erlendvollset:
```
if is_single_id:
    if ignore_unknown_ids is True:
        return None
    return dps_list[0]
```
And change the return type hint to `Union[Datapoints, DatapointsList, None]`

cognite/client/_api/datapoints.py
```diff
@@ -165,7 +172,7 @@ def retrieve_latest(
         return DatapointsList._load(res, cognite_client=self._cognite_client)
 
     def query(
-        self, query: Union[DatapointsQuery, List[DatapointsQuery]]
+        self, query: Union[DatapointsQuery, List[DatapointsQuery]], ignore_unknown_ids=False
```
erlendvollset: Should be on the `DatapointsQuery` instead.

cognite/client/_api/datapoints.py
sanderland: having two return types for retrieve is already a pain, do you really think three different ones will help?

cognite/client/_api/datapoints.py
sanderland: I think it is prefered by users as is, but we can ask them

cognite/client/_api/datapoints.py
sanderland: to follow up:
2 requested, ignoreunknown False -> DatapointsList of length 2
single requested, ignoreunknown False -> Datapoints
2 requested, 1 found, ignoreunknown True -> DatapointsList of length 1
2 requested, 0 found, ignoreunknown True -> DatapointsList of length 0
single requested, 0 found ignoreunknown True -> new behaviour: None ?
agreed?

cognite/client/_api/datapoints.py

cognite/client/_api/datapoints.py
erlendvollset: The nested if there should of course be `if ignore_unknown_ids and len(dps_list) == 0`. Sorry..

IMO it is better to have behaviour consistent with other methods for single retrieval than returning an empty list upon single retrieval not found.

sanderland: I fixed it in the last commit. It's a bit painful in retrieve_dataframe to handle 3 return types, but I think it's ok now

cognite/client/_api/datapoints.py
```diff
@@ -97,9 +99,12 @@ def retrieve(
             granularity=granularity,
             include_outside_points=include_outside_points,
             limit=limit,
+            ignore_unknown_ids=ignore_unknown_ids,
         )
         dps_list = fetcher.fetch(query)
         if is_single_id:
+            if len(dps_list) == 0 and ignore_unknown_ids is True:
```
erlendvollset: Return type hint and and return value in docstring are still wrong.

cognite/client/_api/datapoints.py
```diff
@@ -389,6 +399,7 @@ def retrieve_dataframe(
         limit: int = None,
         include_aggregate_name=True,
         complete: str = None,
+        ignore_unknown_ids=False,
```
erlendvollset: Type hint

cognite/client/_api/datapoints.py
```diff
@@ -528,6 +551,7 @@ def retrieve_dataframe_dict(
             str, List[str], Dict[str, Union[int, List[str]]], List[Dict[str, Union[int, List[str]]]]
         ] = None,
         limit: int = None,
+        ignore_unknown_ids=False,
```
erlendvollset: Type hint

PR #640, opened 2020-01-17
sequence multi retrieve

sanderland:
Based on statnett needs

### DIFF
--- a/cognite/client/data_classes/sequences.py
+++ b/cognite/client/data_classes/sequences.py
@@ -276,21 +276,31 @@ class SequenceData:
             dumped = {utils._auxiliary.to_camel_case(key): value for key, value in dumped.items()}
         return {key: value for key, value in dumped.items() if value is not None}
 
-    def to_pandas(self) -> "pandas.DataFrame":
+    def to_pandas(self, column_names: str = "columnExternalId") -> "pandas.DataFrame":
         """Convert the sequence data into a pandas DataFrame.
 
         Args:
-            column_names (str): Which field to use as column header. Either "externalId", "id" or "externalIdIfExists" for externalId if it exists for all columns and id otherwise.
+            column_names (str):  Which field to use as column header. Can use any combination of "externalId", "columnExternalId", "id" and other characters as a template.
 
         Returns:
             pandas.DataFrame: The dataframe.
         """
         pd = utils._auxiliary.local_import("pandas")
 
+        column_names = (
+            column_names.replace("columnExternalId", "{columnExternalId}")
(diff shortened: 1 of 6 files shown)


sanderland:
> LGTM. But maybe the `column_names` argument should include the curly brackets? I think it would be easier to understand how the argument works. Also I think they should be snake_cased since that is how they pass the arguments elsewhere. So instead of passing for example "externalId|columnExternalId" you would pass "{external_id}|{column_external_id}".

I was trying to keep it more similar to what we do for datapoints, where we do snake case the parameters. But the templating has no precedent really, so I'm not sure there, will ask around a bit.

erlendvollset:
With my example it would work like the standard f-string, which seems natural to me.

sanderland:
> With my example it would work like the standard f-string, which seems natural to me.

Yes, I got that, and I think it's a good idea if we didn't have anything yet, but it would also lead to:

`seq_data.to_pandas(column_names='{external_id}')` being different and not a natural extension from the already existing `dpt.to_pandas(column_names='externalId')`

erlendvollset:
Right. Maybe just stick to the way it is done there and not support templating here then. Do we have requests for multiple identifiers in the same column header?

sanderland:
f

> Right. Maybe just stick to the way it is done there and not support templating here then. Do we have requests for multiple identifiers in the same column header?

It's hard to make them unique otherwise. We need to make them columnExernalId for backward compatibility if 1 sequence is used, but multi sequence single column retrieval is the use case that led to this PR, where (sequence) externalId would be natural. Without assumptions and supporting both you're led to this kind of combination template.

cognite/client/data_classes/sequences.py
```diff
@@ -276,21 +276,42 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
...
+        if column_names not in [
+            "externalId",
+            "id",
+            "columnExternalId",
+            "id|columnExternalId",
+            "externalId|columnExternalId",
+        ]:
+            raise ValueError(
+                'Invalid column_names value, should be one of "externalId", "id", "columnExternalId", "id|columnExternalId", "externalId|columnExternalId"'
+            )
```
erlendvollset:
```
options = [
            "externalId",
            "id",
            "columnExternalId",
            "id|columnExternalId",
            "externalId|columnExternalId",
        ]
if column_names not in options:
    raise ValueError(f"Invalid column_names value, should be one of {options}")
```

sanderland: fixed in a py3.5 compatible way (although using f-strings would be awesome)

PR #646, opened 2020-01-30
datapoint insertion memory reduction

sanderland:
based on SAs running into issues on datapoint insertion. This is a simple fix that delays generating the json long form to just before posting, with around a 50% reduction in memory usage.

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -269,7 +269,7 @@ class DatapointsAPI(APIClient):
         utils._auxiliary.assert_exactly_one_of_id_or_external_id(id, external_id)
         post_dps_object = self._process_ids(id, external_id, wrap_ids=True)[0]
         if isinstance(datapoints, Datapoints):
-            datapoints = datapoints.dump()["datapoints"]
+            datapoints = [(t, v) for t, v in zip(datapoints.timestamp, datapoints.value)]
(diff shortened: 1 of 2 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -756,7 +752,12 @@ def _insert_datapoints_concurrently(self, dps_object_lists: List[List[Dict[str,
         )
 
     def _insert_datapoints(self, post_dps_objects: List[Dict[str, Any]]):
+        # convert to memory intensive format as late as possible and clean up after
+        for it in post_dps_objects:
+            it["datapoints"] = [{"timestamp": t, "value": v} for t, v in it["datapoints"]]
         self.client._post(url_path=self.client._RESOURCE_PATH, json={"items": post_dps_objects})
+        for it in post_dps_objects:
+            del it["datapoints"]
```
erlendvollset: Does this make any difference? Won't the garbage collector take care of these as soon as the method returns and references to post_dbs_objects are dropped?

sanderland: I thought there is still a reference to it via the `tasks` object whose elements are being modified, and these won't be cleared up until _insert_datapoints_concurrently returns (and all of these converted objects pile up). No?

erlendvollset: Right. That's probably correct

PR #647, opened 2020-02-04
Adding patch as allowed method to support noma

sanderland:


polomani:
can you add tests?

hhenriques:
@sanderland what's the use-case for Patch on Python?

sanderland:
@hhenriques I ran into this when building on top of the sdk for noma, which used patch. there's no reason to error out on this request type anyway

hhenriques:
So al this does is avoid raising an error if Patch is used?

sanderland:
> So all this does is avoid raising an error if Patch is used?

yes if you use the base request method with patch

PR #648, opened 2020-02-06
output fix for nullable int columns when camel case is off / release 1.4.11

sanderland:


polomani:
can we test it?

sanderland:
this is for jupyter notebook output, tests are not so relevant

polomani:
how do I know that the code works then?

sanderland:
> how do I know that the code works then?
By looking at the output?
Anyway since you insisted I added a test, I think it's more complicated than reading the code, but hope it helps you understand.

PR #660, opened 2020-03-02
Send in write filemetadata when uploading files from path / release 1.4.12

lynochka:


lynochka:
Tests failing on event api  
```
_____________________ TestEventsAPI.test_partitioned_list ______________________
self = <tests.tests_integration.test_api.test_events.TestEventsAPI object at 0x7faa4706ccf8>
    def test_partitioned_list(self):
        # stop race conditions by cutting off max created time
        maxtime = utils.timestamp_to_ms(datetime(2019, 6, 20, 17, 30))
        res_flat = COGNITE_CLIENT.events.list(limit=None, type="test-data-populator", start_time={"max": maxtime})
        res_part = COGNITE_CLIENT.events.list(
            partitions=8, type="test-data-populator", start_time={"max": maxtime}, limit=None
        )
>       assert len(res_flat) > 0
E       assert 0 > 0
E        +  where 0 = len([])
```

lynochka:
The test looked fine, asserting for the entire FileMetaData 
https://github.com/cognitedata/cognite-sdk-python/blob/3327cb01d0e6760175a9cb04ffae2ca01db4b9c1/tests/tests_unit/test_api/test_files.py#L280

cognite/client/_api/files.py
```diff
@@ -437,12 +437,35 @@ def upload(
                     for file in files:
                         file_path = os.path.join(root, file)
                         basename = os.path.basename(file_path)
-                        tasks.append((FileMetadata(name=basename), file_path, overwrite))
+                        file_metadata = FileMetadata(
+                            name=basename,
+                            source=source,
```
polomani: seems like these 6 lines appear trice in this function 
can we copy initial `file_metadata` and change `name` only?
```
source=source,
mime_type=mime_type,
metadata=metadata,
asset_ids=asset_ids,
source_created_time=source_created_time,
source_modified_time=source_modified_time,
```

sanderland: sure, I suppose that would lower the risk for such bugs in the future if fields are added

PR #665, opened 2020-03-23
Update documentation of limit on __call__ functions.

eventh:
The documentation stated the default limit was 25, but on
__call__ the limit is set to None, which means you will
get all resources.

This change only changes the documentation to correctly
say that the default is to return all results.


sanderland:
I see some fine copy paste there as usual, thanks for the fix!

PR #670, opened 2020-04-01
Add security_categories for files and active_at_time for relationships

erlendvollset:


vegarsti:
Should `security_categories` be added as a field to the `upload_bytes` method as well?

sanderland:
We should probably mark the relationships API as potentially out of date and redirect people to the experimental extensions.

PR #671, opened 2020-04-02
Add security categories to file methods

vegarsti:
File methods weren't updated with security categories


vegarsti:
oops, dismissed your review @erlendvollset that wasn't my intention. Is the changelog entry OK?

PR #672, opened 2020-04-06
Add jitter to the delay between retries

alexbrasetvik:


hhenriques:
@alexbrasetvik do you want to remove this from Draft? Maybe also mention the change in the changelog.

PR #673, opened 2020-04-17
feat: aggregation endpoint

mrtraser:
Add aggregation endpoint for:
- assets
- data_sets
- events
- files
- sequences
- time_series

### DIFF
--- a/cognite/client/_api/assets.py
+++ b/cognite/client/_api/assets.py
@@ -5,7 +5,7 @@ from typing import *
 
 from cognite.client import utils
 from cognite.client._api_client import APIClient
-from cognite.client.data_classes import Asset, AssetFilter, AssetList, AssetUpdate, TimestampRange
+from cognite.client.data_classes import Asset, AssetAggregate, AssetFilter, AssetList, AssetUpdate, TimestampRange
 from cognite.client.exceptions import CogniteAPIError
 
 
@@ -267,6 +267,25 @@ class AssetsAPI(APIClient):
             other_params={"aggregatedProperties": aggregated_properties} if aggregated_properties else {},
             partitions=partitions,
         )
+
+    def aggregate(self, filter: Union[AssetFilter, Dict] = None) -> List[AssetAggregate]:
+        """`Aggregate assets <https://docs.cognite.com/api/v1/#operation/aggregateAssets>`_
+
+        Args:
+            filter (Union[AssetFilter, Dict]): Filter on assets filter with exact match
+
+        Returns:
+            List[AssetAggregate]: List of asset aggregates
+
+        Examples:
+
+            Aggregate assets:
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> aggregate_by_prefix = c.assets.aggregate(filter={"external_id_prefix": "prefix"})
+        """
+        return self._aggregate(filter=filter, cls=AssetAggregate)
 
     def create(self, asset: Union[Asset, List[Asset]]) -> Union[Asset, AssetList]:
         """`Create one or more assets. <https://docs.cognite.com/api/v1/#operation/createAssets>`_
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -412,6 +412,19 @@ class APIClient:
             raise tasks_summary.exceptions[0]
         return cls._load(tasks_summary.joined_results(), cognite_client=self._cognite_client)
 
+    def _aggregate(
+        self, resource_path: str = None, filter: Union[CogniteFilter, Dict] = None, headers: Dict = None, cls=None
+    ):
+        utils._auxiliary.assert_type(filter, "filter", [dict, CogniteFilter], allow_none=True)
+        if isinstance(filter, CogniteFilter):
+            filter = filter.dump(camel_case=True)
+        elif isinstance(filter, Dict):
+            filter = utils._auxiliary.convert_all_keys_to_camel_case(filter)
+        resource_path = resource_path or self._RESOURCE_PATH
+        body = {"filter": filter or {}}
+        res = self._post(url_path=resource_path + "/aggregate", json=body, headers=headers)
+        return [cls(**agg) for agg in res.json()["items"]]
+
     def _create_multiple(
         self,
         items: Union[List[Any], Any],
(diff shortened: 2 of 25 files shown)


cognite/client/data_classes/aggregates.py
```diff
@@ -0,0 +1,9 @@
+from cognite.client.data_classes._base import CognitePropertyClassUtil
+
+# GenClass: DataSet
```
erlendvollset, line 3: This means the DataSet class will be generated from the openapi spec here and overwrite this implementation next time `generate_from_spec.py` is run, which I guess you don't want.

cognite/client/_api_client.py
```diff
@@ -412,6 +412,20 @@ def get_partition(partition):
             raise tasks_summary.exceptions[0]
         return cls._load(tasks_summary.joined_results(), cognite_client=self._cognite_client)
 
+    def _aggregate(
```
erlendvollset, line 415: This should be unit tested in test_api_client.py

cognite/client/_api_client.py
```diff
@@ -412,6 +412,20 @@ def get_partition(partition):
...
+        self,
+        cls=None,
```
erlendvollset, line 417: Remove this.

cognite/client/_api_client.py
```diff
@@ -412,6 +412,20 @@ def get_partition(partition):
...
+        resource_path: str = None,
+        filter: Dict = None,
+        headers: Dict = None,
+    ): 
+        resource_path = resource_path or self._RESOURCE_PATH
+        body = {
+            "filter": filter or {},
+        }
+        res = self._post(url_path=resource_path + "/aggregate", json=body, headers=headers)
+        return [cls(agg["count"]) for agg in res.json()["items"]]
```
erlendvollset, line 427:
```suggestion
        return [Aggregate(agg["count"]) for agg in res.json()["items"]]
```

cognite/client/data_classes/aggregates.py
mrtraser: oops, thx for the catch!

cognite/client/_api_client.py
mrtraser: added

cognite/client/_api_client.py
mrtraser: added

cognite/client/_api_client.py
mrtraser: fixed

mrtraser:
@erlendvollset, could you please take a look at Jenkins errors? I don't have access to check what's going wrong

cognite/client/data_classes/aggregates.py
```diff
@@ -0,0 +1,10 @@
+from typing import *
+
+from cognite.client.data_classes._base import CognitePropertyClassUtil
+
+
+class Aggregate(dict):
+    def __init__(self, count: int = None):
+        self.count = count
```
polomani, line 8: there a few things I don't like about this file:

First of all it assumes at aggregated responses will always have consistent format across all resource types. Which is not always the case if we look retrospectively

Secondly, this class is not a resource type, so it probably shouldn't have its own file. Should be more like be a part of something else, like `shared.py`? (this might not make sense after you look at the next point)

And lastly we need to generate as much as we can, so I suggest to use the generator, (look how `# GenPropertyClass: AggregateResultItem` is generated)
Then we'll have it's own aggregation response in every resource type

You can have a look for schema names here https://storage.googleapis.com/cognitedata-api-docs/dist/v1.json
for assets it's called "AssetAggregate" for instance.
They have inconsistent names across resource type so you might need to rename some (generator can do that)

cognite/client/_api/assets.py
```diff
@@ -268,6 +268,70 @@ def list(
...
+        parent_external_ids: List[str] = None,
+        asset_subtree_ids: List[int] = None,
+        asset_subtree_external_ids: List[str] = None,
+        data_set_ids: List[int] = None,
+        data_set_external_ids: List[str] = None,
+        metadata: Dict[str, str] = None,
+        source: str = None,
+        created_time: Union[Dict[str, Any], TimestampRange] = None,
+        last_updated_time: Union[Dict[str, Any], TimestampRange] = None,
+        root: bool = None,
```
polomani, line 284: I was trying to look into future here, and as far as I know there will be other props in this query, like `groupBy` for instance. Because of this and because of the fact that it's hard to maintain this I was thinking to take filter as an object parameter (as it is in api).

@erlendvollset I am not sure if it's convenient enough for pydevs to pass nested params into function. could you apply a little bit of your judgement here?

mrtraser: I've found a few places in code where it is done with the same approach, so I didn't think too much about it. For instance, list methods for timeseries, sequences, events, etc. are done in the same way.

cognite/client/data_classes/aggregates.py
mrtraser: Fixed

cognite/client/_api/assets.py
polomani: yep, you're right. list is basically just a filter endpoint, so it doesn't do anything else. but I am not sure if we should do the same here

polomani: @sanderland do you have opinion on this one?

sanderland: you mean make it more like search rather than list? Does the API use the same filter spec as search? If so I think as long as you allow a dict for the parameter (again, like search) it's fine. it's a bit of a developer vs user convenience tradeoff. 
If not, you end up implementing another possibly autogenerated filter object which is incompatible with the other and will be confusing I think

cognite/client/data_classes/files.py
```diff
@@ -214,6 +217,23 @@ def remove(self, value: List) -> FileMetadataUpdate:
     # GenStop
 
 
+# GenPropertyClass: FilesAggregate
+class FilesAggregate(dict):
```
polomani, line 221:
```suggestion
class FileAggregate(dict):
```
need too rename for consistency

mrtraser: the name comes from the [schema](https://storage.googleapis.com/cognitedata-api-docs/dist/v1.json), I just left it without the changes

sanderland: schemas are not user facing and can be changed, the name we choose here is more permanent

mrtraser: got it, will change

PR #676, opened 2020-04-28
feat: aggregation endpoint, publish package

mrtraser:


mrtraser:
was accidentally opened

PR #679, opened 2020-05-12
use github actions

sanderland:
* Tests vs 3.5.9 and not 3.5.0, but I think the chance this will affect anything is minute.


.github/workflows/test_and_build.yml
```diff
@@ -0,0 +1,81 @@
...
+    - name: Check codestyle
+      if: matrix.python-version != 3.5
+      run: pipenv run black -l 120 --check .
+
+    - name: Test OpenAPI Generator
+      if: matrix.python-version != 3.5
+      run: pipenv run pytest openapi/tests
+
+    - name: Run tests
+      if: matrix.python-version != 3.5
```
erlendvollset, line 54: So we don't test 3.5?

Could we have an extra job only running a compatible Python version which installs dev dependencies where we check the code and all that. And then only install prod-dependencies + what is necessary for running tests in this python-version matrix?

sanderland: we test 3.5 core only

sanderland: I will see what I can do

hhenriques:
@sanderland shouldn't we remove the Jenkins file?

sanderland:
> @sanderland shouldn't we remove the Jenkins file?

I thought I'd keep it until this is merged & maybe until @polomani adds the snippets action, then we can change branch protection and remove the jenkinsfile.

polomani:
@sanderland, we can add snippet action to this branch, so we merge everything at once. 
didn't get about branch protection?

sanderland:
> @sanderland, we can add snippet action to this branch, so we merge everything at once.
> didn't get about branch protection?

Can you add the snippet to the release script in this branch?

I mean we have branch protection set to not merge unless jenkins passes, so if you delete the jenkinsfile you can't merge. of course we can change branch protection to actions before merge as well.

polomani:
> Can you add the snippet to the release script in this branch?
yep

PR #681, opened 2020-05-14
feat: add assets&events dataSetId.isNull filters

dmivankov:


cognite/client/data_classes/events.py
```diff
@@ -96,6 +96,7 @@ class EventFilter(CogniteFilter):
         asset_external_ids (List[str]): Asset External IDs of equipment that this event relates to.
         root_asset_ids (List[Dict[str, Any]]): This parameter is deprecated. Use assetSubtreeIds instead. Only include events that have a related asset in a tree rooted at any of these root assetIds.
         asset_subtree_ids (List[Dict[str, Any]]): Only include events that have a related asset in a subtree rooted at any of these assetIds (including the roots given). If the total size of the given subtrees exceeds 100,000 assets, an error will be returned.
+        data_set_id (Union[Dict[str, Any], EndTimeIsNull]): No description.
```
sanderland, line 99: can we rename EndTimeIsNull ?

dmivankov: poor generator, not quite sure how it ended up with that choice of types, spec is more sane. Will try to fix

dmivankov:
some 3d tests failed :thinking:

sanderland:
@msusag files and 3d still unstable - what's the status on this?

msusag:
> @msusag files and 3d still unstable - what's the status on this?

FILES
------
For the files issue, I wonder if the problem is due to a file being deleted just prior in one test before the test that fails. As the list endpoint will return a non-strongly-consistent view of files in the db, the retrieve /byids call on the specific set returned from elasticsearch may lead to 400 with missing IDs, if the entry was actually deleted in the db.
For example this delete request:
https://github.com/cognitedata/cognite-sdk-python/blob/master/tests/tests_integration/test_api/test_files.py#L37

It looks like some of the tests for other resource use the ignoreUnknownIds=true.
This should also be used for files, if the tests are not supposed to handle these inherent inconsistencies between DB and ES endpoints.
The files API does support ignoreUnknownIds for the /byids endpoint,
https://docs.cognite.com/api/v1/#operation/byIdsFiles

3D
---
For the test failure in test_three_d.py in https://github.com/cognitedata/cognite-sdk-python/pull/681/checks?check_run_id=705935399, I did not see a request Id in the logs, so it is hard to know when and how things go wrong.
But found 12:56 in the test report header.
The closest entry in the log might be this:
https://console.cloud.google.com/logs/viewer?project=cognitedata-greenfield&authuser=1&organizationId=486095736426&minLogLevel=0&expandAll=false&timestamp=2020-05-25T12:09:14.085000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeEnd=2020-05-25T12:08:55.750Z&interval=PT1H&resource=k8s_cluster%2Flocation%2Feurope-west1%2Fcluster_name%2Fcdp&advancedFilter=resource.type%3D%22k8s_container%22%0Aresource.labels.cluster_name%3D%22cdp%22%0Aresource.labels.namespace_name%3D%22default%22%0Alabels.k8s-pod%2Fapp%3D%22api%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fmanaged-by%3D%22spinnaker%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fname%3D%22api%22%0AjsonPayload.api_path:%22%2F3d%2F%22&scrollTimestamp=2020-05-25T10:56:18.119000000Z&dateRangeUnbound=backwardInTime&pinnedLogId=c9x4vyluegambwmkr&pinnedLogTimestamp=2020-05-25T10:56:16.513Z

It would be really good if the test could report the request ID that fails, so we can try to figure out what goes wrong.

dmivankov:
noticed following instabilities
- api_keys list
- 3d revisions list
- retries on raw tables/delete exceeded

sanderland:
@msusag that seems like an explanation, but also means your inconsistency is worse than assets and worse than it's been in the past. or the ordering is different in that the top results are newer/more inconsistent. we should look into not relying on list to test retrieve.

msusag:
> @msusag that seems like an explanation, but also means your inconsistency is worse than assets and worse than it's been in the past. or the ordering is different in that the top results are newer/more inconsistent. we should look into not relying on list to test retrieve.

The search loader based endpoints such as /list will have varying degree of inconsistency with the db. That applies to all endpoints.
It is not a error as such in the files API, and there is to my knowledge no specific reason for degradation in the /list endpoint consistency for files data.

For strong consistency checking you should use the byids endpoint.

sanderland:
> For strong consistency checking you should use the byids endpoint.

I think the test literally just wants an example file to see if retrieve (byids) works and gives a correct value. It does this by taking the first entry from list.

dmivankov:
3d test does something similar, assumes there is some model and it has revisions, though there are also correct tests that create a model, do checks, delete the model

dmivankov:
for now, hitting a rerun few times helps))

cognite/client/data_classes/assets.py
```diff
@@ -2,7 +2,7 @@
 from typing import *
 
 from cognite.client.data_classes._base import *
-from cognite.client.data_classes.shared import TimestampRange
+from cognite.client.data_classes.shared import *
```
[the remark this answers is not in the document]
dmivankov: based on the line above, but can do TimestampRange, DataSetIdIsNull too

polomani:
@dmivankov is it final design? can be merged?

cognite/client/_api/assets.py
```diff
@@ -177,6 +188,7 @@ def list(
         asset_subtree_external_ids: List[str] = None,
         data_set_ids: List[int] = None,
         data_set_external_ids: List[str] = None,
+        data_set_id: Union[Dict[str, Any], DataSetIdIsNull] = None,
```
polomani, line 191: add this to a python doc as well? line 212-ish

tests/tests_unit/test_api/test_assets.py
```diff
@@ -114,6 +114,10 @@ def test_list_with_aggregated_properties_param_when_snake_cased(self, mock_asset
         ASSETS_API.list(name="bla", aggregated_properties=["child_count"])
         assert ["childCount"] == jsgz_load(mock_assets_response.calls[0].request.body)["aggregatedProperties"]
 
+    def test_list_with_dataset_id_null(self, mock_assets_response):
+        ASSETS_API.list(name="bla", data_set_id={"isNull": "true"})
```
polomani, line 118:
```suggestion
        ASSETS_API.list(name="bla", data_set_id={"isNull": True})
```

tests/tests_unit/test_api/test_assets.py
```diff
@@ -114,6 +114,10 @@ def test_list_with_aggregated_properties_param_when_snake_cased(self, mock_asset
...
+        assert {"isNull": "true"} == jsgz_load(mock_assets_response.calls[0].request.body)["filter"]["dataSetId"]
```
polomani, line 119:
```suggestion
        assert {"isNull": True} == jsgz_load(mock_assets_response.calls[0].request.body)["filter"]["dataSetId"]
```

tests/tests_unit/test_api/test_events.py
```diff
@@ -62,6 +62,10 @@ def test_list_with_dataset_ids(self, mock_events_response):
             "dataSetIds"
         ]
 
+    def test_list_with_dataset_id_null(self, mock_events_response):
+        EVENTS_API.list(source="bla", data_set_id={"isNull": "true"})
```
polomani, line 66:
```suggestion
        EVENTS_API.list(source="bla", data_set_id={"isNull": True})
```

tests/tests_unit/test_api/test_events.py
```diff
@@ -62,6 +62,10 @@ def test_list_with_dataset_ids(self, mock_events_response):
...
+        assert {"isNull": "true"} == jsgz_load(mock_events_response.calls[0].request.body)["filter"]["dataSetId"]
```
polomani, line 67:
```suggestion
        assert {"isNull": True} == jsgz_load(mock_events_response.calls[0].request.body)["filter"]["dataSetId"]
```

polomani:
@dmivankov do you have time to address issues, or should we help you a bit here?

dmivankov:
I'll draft it for now, need to confirm the api fields naming...

erlendvollset:
Closing to due to inactivity. Feel free to reopen if this is wrong or you wish to resume the work (I'll leave the branch).

PR #682, opened 2020-05-18
chore(gh-action): push code snippets

polomani:


polomani:
@sanderland gh token supposed to be native and set automatically

PR #689, opened 2020-06-03
retryable list to static var

sanderland:


polomani:
do we need to add/remove them on a fly?
or just once per init?
adding it to `ClientConfig` maybe is a bit better idea? but in that case there will be no option to delete an item from the list (only extend the list).

sanderland:
> do we need to add/remove them on a fly?
> or just once per init?
> adding it to `ClientConfig` maybe is a bit better idea? but in that case there will be no option to delete an item from the list (only extend the list).

I expect only the experimental sdk to use this to add a bunch of entries in it's initializer. clientconfig would signal a user might need to edit this, and I hope they don't.

polomani:
ah, I see, I though it was a feature request

PR #691, opened 2020-06-12
experimental warnings update

sanderland:
more of the start of a discussion on how we should handle diverging implementations - synthetic ts and model hosting seems somewhat stable, but relationships is a bit outdated and going for some big changes.


hhenriques:
We're looking into bringing experimental back into the main repo (we're already doing that with the JS SDK https://github.com/cognitedata/cognite-sdk-js/pull/373 - still wip). The idea is to have a core package that includes all infrastructural logic and then have a folder for the stable, beta, and alpha SDK with their own packages.

That should solve the diverging problem and make it easier to bring things into stable once it's ready

PR #694, opened 2020-06-19
feat: move synthetic TS to main client

mrtraser:


cognite/client/_api/synthetic_time_series.py
```diff
@@ -79,7 +78,7 @@ def retrieve(
         datapoints.external_id = short_expression  # for dataframe readability
         while True:
             query["limit"] = min(limit, self._DPS_LIMIT)
-            resp = self._post(url_path=self._SYNTHETIC_RESOURCE_PATH + "/query", json={"items": [query]})
+            resp = self._post(url_path=self._RESOURCE_PATH + "/query", json={"items": [query]})
```
polomani, line 81: api supports multiple, but we support only single expression per request?

cognite/client/_api/synthetic_time_series.py
```diff
@@ -46,28 +45,28 @@ def retrieve(
 
             Request a synthetic time series query with direct syntax
 
-                >>> from cognite.client.experimental import CogniteClient
+                >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
-                >>> dps = c.datapoints.synthetic.retrieve(expression="TS{id:123} + TS{externalId:'abc'}", start="2w-ago", end="now")
+                >>> dps = c.synthetic_time_series.retrieve(expression="TS{id:123} + TS{externalId:'abc'}", start="2w-ago", end="now")
```
polomani, line 50: I am not sure we can call it retrieve, since there might be "persistent synthetic time series", when you can save them and retrieve later.
maybe we should go for `query()` instead?
@sanderland thoughs?

sanderland: why was this moved at all? wtf?

sanderland: normal datapoints are client.datapoints.retrieve
persistent synthetic ts will be compatible and under the same endpoint
synthetic query was right where it belonged and there was no need to move it away from c.datapoints.synthetic

mrtraser: @sanderland, got it, will move to `datapoints.synthetic`

sanderland: but...why was it moved in the first place? who did you discuss this with?

cognite/client/_api/synthetic_time_series.py
mrtraser: redone

cognite/client/_api/synthetic_time_series.py
mrtraser: revert to previous endpoint

sanderland: I disagree with ivan on query, see client.datapoints using retrieve, so this should match that endpoint as much as possible

cognite/client/_api/synthetic_time_series.py
```diff
@@ -46,46 +45,68 @@ def retrieve(
...
+            datapoints.append(query_datapoints)
+
         while True:
-            query["limit"] = min(limit, self._DPS_LIMIT)
-            resp = self._post(url_path=self._SYNTHETIC_RESOURCE_PATH + "/query", json={"items": [query]})
-            data = resp.json()["items"][0]
-            datapoints._extend(Datapoints._load(data, expected_fields=["value", "error"]))
-            limit -= len(data["datapoints"])
-            if len(data["datapoints"]) < self._DPS_LIMIT or limit <= 0:
+            resp = self._post(url_path=self._RESOURCE_PATH + "/query", json={"items": queries})
```
sanderland, line 92: if you insist on implementing multi query support, I'd at least expect it to be parallelized like other endpoints.

mrtraser: @sanderland , right, but how is it possible to parallelize it if we're passing multiple expressions to single request and decide if we should fetch additional datapoints based on response?

sanderland: By not passing multiple expressions to a single request. See e.g. the normal data points api.

mrtraser: so, we can divide it into two endpoints – `retrieve` and `retrive_multiple`?

sanderland: No, that would be weird. Try to mirror how the normal datapoints API works.

mrtraser: got it, redone with the suggested solution

cognite/client/_api/synthetic_time_series.py
```diff
@@ -46,40 +45,57 @@ def retrieve(
...
-        }
-        datapoints = Datapoints(value=[], error=[])
-        datapoints.external_id = short_expression  # for dataframe readability
+
+        if datapoints_summary.exceptions:
+            raise datapoints_summary.exceptions[0]
+
+        return datapoints_summary.results
+
+    def func(self, query, datapoints, limit):
```
sanderland, line 95: func? please make this a sensibly named method starting with _

cognite/client/_api/synthetic_time_series.py
```diff
@@ -10,17 +10,16 @@
...
+    _RESOURCE_PATH = "/timeseries/synthetic"
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._DPS_LIMIT = 10000
 
     def retrieve(
         self,
-        expression: Union[str, "sympy.Expr"],
+        expressions: List[Union[str, "sympy.Expr"]],
```
sanderland, line 22: I'd like this to accept a single expression/string as well for convenience/backward compatibility. @hhenriques thoughts?

cognite/client/_api/synthetic_time_series.py
polomani: I don't agree with `retrieve` since retrieve is used for endpoints that can reference something by id.
query expressions is quite far from being just an id

polomani: and I would reserve retrieve for persistent sts

cognite/client/_api/synthetic_time_series.py
mrtraser: ah, missed to change it, thx

cognite/client/_api/synthetic_time_series.py
sanderland: @polomani please see the datapoints api

cognite/client/_api/synthetic_time_series.py
```diff
@@ -46,27 +45,28 @@ def retrieve(
 
             Request a synthetic time series query with direct syntax
 
-                >>> from cognite.client.experimental import CogniteClient
+                >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
-                >>> dps = c.datapoints.synthetic.retrieve(expression="TS{id:123} + TS{externalId:'abc'}", start="2w-ago", end="now")
+                >>> dps = c.datapoints.synthetic.query(expression="TS{id:123} + TS{externalId:'abc'}", start="2w-ago", end="now")
```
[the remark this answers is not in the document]
mrtraser: done

cognite/client/_api/synthetic_time_series.py
polomani: @hhenriques someone needs to break a tie

cognite/client/_api/synthetic_time_series.py
hhenriques: agreed

mrtraser: @sanderland, done

cognite/client/_api/synthetic_time_series.py
hhenriques: @mrtraser Let's go with `datapoints.synthetic.query`. Also aligned with Time Lords and their opinion is also leaning to `retrieve` should be used for persistent sts

cognite/client/_api/synthetic_time_series.py
```diff
@@ -46,40 +45,57 @@ def retrieve(
...
-            "start": cognite.client.utils._time.timestamp_to_ms(start),
-            "end": cognite.client.utils._time.timestamp_to_ms(end),
-        }
-        datapoints = Datapoints(value=[], error=[])
-        datapoints.external_id = short_expression  # for dataframe readability
+
+        if datapoints_summary.exceptions:
+            raise datapoints_summary.exceptions[0]
+
+        return datapoints_summary.results
```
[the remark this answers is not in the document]
mrtraser: I see now, fixed

cognite/client/_api/synthetic_time_series.py
mrtraser: done

cognite/client/_api/synthetic_time_series.py
```diff
@@ -5,81 +5,102 @@
...
 
         Examples:
 
             Request a synthetic time series query with direct syntax
 
-                >>> from cognite.client.experimental import CogniteClient
+                >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
-                >>> dps = c.datapoints.synthetic.retrieve(expression="TS{id:123} + TS{externalId:'abc'}", start="2w-ago", end="now")
+                >>> dps = c.datapoints.synthetic.query(expressions=["TS{id:123} + TS{externalId:'abc'}"], start="2w-ago", end="now")
```
sanderland, line 50: why list

cognite/client/_api/synthetic_time_series.py
```diff
@@ -5,81 +5,102 @@
...
 
             Use variables to re-use an expression:
 
-                >>> from cognite.client.experimental import CogniteClient
+                >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
                 >>> vars = {"A": "my_ts_external_id", "B": client.time_series.retrieve(id=1)}
-                >>> dps = c.datapoints.synthetic.retrieve(expression="A+B", start="2w-ago", end="now", variables=vars)
+                >>> dps = c.datapoints.synthetic.query(expressions=["A+B"], start="2w-ago", end="now", variables=vars)
```
sanderland, line 57: why list

cognite/client/_api/synthetic_time_series.py
```diff
@@ -5,81 +5,102 @@
...
 
             Use sympy to build complex expressions:
 
-                >>> from cognite.client.experimental import CogniteClient
+                >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
                 >>> from sympy import symbols, cos, pi
                 >>> a = sympy.symbols('a')
-                >>> dps = c.datapoints.synthetic.retrieve(pi * cos(a), start="2w-ago", end="now", variables={"a": "my_ts_external_id"},aggregate='interpolation',granularity='1m')
+                >>> dps = c.datapoints.synthetic.query([pi * cos(a)], start="2w-ago", end="now", variables={"a": "my_ts_external_id"},aggregate='interpolation',granularity='1m')
```
sanderland, line 65: maybe change this to a list with sin(a) , cos(a) to show list?

tests/tests_integration/test_api/test_synthetic_time_series.py
```diff
@@ -55,20 +64,20 @@ def test_retrieve_with_errors(self, test_time_series, post_spy):
     def test_expression_builder_time_series_vs_string(self, test_time_series):
         from sympy import symbols
 
-        dps1 = COGNITE_CLIENT.datapoints.synthetic.retrieve(
-            expression=symbols("a"),
+        dps1 = COGNITE_CLIENT.datapoints.synthetic.query(
```
sanderland, line 67: please change one of them to single item, one to single item list, one to multi item list and check return types

cognite/client/_api/synthetic_time_series.py
mrtraser: done

cognite/client/_api/synthetic_time_series.py
mrtraser: done

cognite/client/_api/synthetic_time_series.py
mrtraser: yeah, good point, thx

tests/tests_integration/test_api/test_synthetic_time_series.py
mrtraser: ah, you're right, changed regarding your suggestion

cognite/client/_api/synthetic_time_series.py
```diff
@@ -5,81 +5,102 @@
...
 
             Use sympy to build complex expressions:
 
-                >>> from cognite.client.experimental import CogniteClient
+                >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
                 >>> from sympy import symbols, cos, pi
                 >>> a = sympy.symbols('a')
-                >>> dps = c.datapoints.synthetic.retrieve(pi * cos(a), start="2w-ago", end="now", variables={"a": "my_ts_external_id"},aggregate='interpolation',granularity='1m')
+                >>> dps = c.datapoints.synthetic.query([sin(a), cos(a)], start="2w-ago", end="now", variables={"a": "my_ts_external_id"},aggregate='interpolation',granularity='1m')
```
polomani, line 65:
```suggestion
                >>> dps = c.datapoints.synthetic.query(["sin(a), cos(a)"], start="2w-ago", end="now", variables={"a": "my_ts_external_id"},aggregate='interpolation',granularity='1m')
```
doesn't this need quotes?

mrtraser: no, I think, cause `sin(a)` and `cos(a)` is two different expressions with `sympy.Expr` type

sanderland: ["sin(a)", "cos(a)"]

mrtraser: @sanderland , `[sin(a), cos(a)]` is also valid, isn't it?

sanderland: yes since this is sympy

PR #695, opened 2020-06-23
feat(events): unique value aggregation

fredrik955:


cognite/client/_api/events.py
```diff
@@ -291,7 +292,30 @@ def aggregate(self, filter: Union[EventFilter, Dict] = None) -> List[EventAggreg
                 >>> aggregate_type = c.events.aggregate(filter={"type": "failure"})
         """
 
-        return self._aggregate(filter=filter, cls=EventAggregate)
+        return self._aggregate(filter=filter, cls=AggregateResult)
+
+    def unique_values_aggregate(
+        self, filter: Union[EventFilter, Dict] = None, fields: List[str] = None
+    ) -> List[UniqueValuesAggregateResult]:
+        """`Aggregate unique values for events <https://docs.cognite.com/api/v1/#operation/aggregateEvents>`_
```
sanderland, line 300: is having 2 functions for one endpoint not confusing?

cognite/client/data_classes/shared.py
```diff
@@ -21,3 +21,32 @@ def __init__(self, max: int = None, min: int = None, **kwargs):
...
+    """
+
+    def __init__(self, count: int = None, **kwargs):
+        self.count = count
+        self.update(kwargs)
+
+    count = CognitePropertyClassUtil.declare_property("count")
+
+
+class UniqueValuesAggregateResult(AggregateResult):
```
sanderland, line 40: I would just make this one type with value being None if not relevant

cognite/client/_api/events.py
hhenriques: we decided to go down this route to hide the complexity of the API a bit. If it's a single endpoint, then the user won't know what fields are valid for what aggregation types.

cognite/client/_api/events.py
```diff
@@ -291,7 +292,30 @@ def aggregate(self, filter: Union[EventFilter, Dict] = None) -> List[EventAggreg
                 >>> aggregate_type = c.events.aggregate(filter={"type": "failure"})
         """
 
-        return self._aggregate(filter=filter, cls=EventAggregate)
+        return self._aggregate(filter=filter, cls=AggregateResult)
+
+    def unique_values_aggregate(
```
hhenriques, line 297: I'd prefer something like `aggregates.unique_values` so it's less messy once we add the rest of the aggregations. Otherwise, I propose `aggregate_unique_values`. Then all aggregations start with `aggregate` which makes it easier to use with autocomplete

fredrik955: Ah, we didn't do that for the javascript sdk... https://github.com/cognitedata/cognite-sdk-js/blob/v1/src/resources/events/eventsApi.ts#L74

polomani: @fredrik955 yeah I know..

cognite/client/__init__.py
```diff
@@ -1,3 +1,3 @@
 from cognite.client._cognite_client import CogniteClient
 
-__version__ = "1.8.0"
```
hhenriques: think we can have the same version for unique values and STS?

fredrik955: Yes, I undid the version change to bulk it

PR #696, opened 2020-06-25
feat: labels with asset support

fredrik955:


vincent-cognite:
seems there are a lot of changes that are not labels related.
maybe a separate PR would have been nice

fredrik955:
@vincent-cognite @hhenriques I tried to add an integration test for asset filter with labels but it fails due to eventual consistency :( What do you suggest?

PR #697, opened 2020-06-26
Removed outdated planned version from change log.

maciejmrowiec:


polomani:
added that here https://github.com/cognitedata/cognite-sdk-python/pull/698

PR #700, opened 2020-06-30
Use only nodeId and assetId in 3d mappings delete to avoid 400 response

msusag:
3d model asset mappings delete will return http status 400 if the asset mappings objects contain all fields returned from 3d model mappings list request.

The fix makes sure only the nodeId and assetId fields are used in the delete request.

### DIFF
--- a/cognite/client/_api/three_d.py
+++ b/cognite/client/_api/three_d.py
@@ -547,11 +547,11 @@ class ThreeDAssetMappingAPI(APIClient):
         path = utils._auxiliary.interpolate_and_url_encode(self._RESOURCE_PATH, model_id, revision_id)
         utils._auxiliary.assert_type(asset_mapping, "asset_mapping", [list, ThreeDAssetMapping])
         if isinstance(asset_mapping, ThreeDAssetMapping):
-            asset_mapping = [asset_mapping]
+            asset_mapping = [ThreeDAssetMapping(asset_mapping.node_id, asset_mapping.asset_id)]
         chunks = utils._auxiliary.split_into_chunks(
             [a.dump(camel_case=True) for a in asset_mapping], self._DELETE_LIMIT
         )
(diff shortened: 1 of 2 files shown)


msusag:
> This does not appear to be valid code, a chunk is a list and does not have these properties.

Of course, thanks.
Will fix.

cognite/client/_api/three_d.py
```diff
@@ -547,11 +547,11 @@ def delete(
         path = utils._auxiliary.interpolate_and_url_encode(self._RESOURCE_PATH, model_id, revision_id)
         utils._auxiliary.assert_type(asset_mapping, "asset_mapping", [list, ThreeDAssetMapping])
         if isinstance(asset_mapping, ThreeDAssetMapping):
-            asset_mapping = [asset_mapping]
+            asset_mapping = [ThreeDAssetMapping(asset_mapping.node_id, asset_mapping.asset_id)]
         chunks = utils._auxiliary.split_into_chunks(
             [a.dump(camel_case=True) for a in asset_mapping], self._DELETE_LIMIT
```
sanderland, line 552: [ThreeDAssetMapping(a.node_id, a.asset_id).dump(camel_case=True) for a in asset_mapping]

msusag: See c6eb97d

sanderland: lgtm, please update changelog to add a fixed entry

polomani:
maybe add a unit test at least?

cognite/client/_api/three_d.py
msusag: Fixed in d6732b2

msusag:
> maybe add a unit test at least?

I think it is covered by existing tests to some extent.

polomani:
@msusag if it was we probably wouldn't have this issue at all, right? :)
anyway I merged

msusag:
> @msusag if it was we probably wouldn't have this issue at all, right? :)
> anyway I merged

Agree a more specific test case would have been good.

PR #707, opened 2020-07-06
fix: attach/detach multiple labels in a single method

fredrik955:
This PR will do the following:
- Makes it easy to attach/detach multiple labels to an asset:
```
my_update = AssetUpdate(id=1).labels.add(["PUMP, "another_label"]).labels.remove(["VALVE"])
```
- Don't fail `asset.update` requests when doing a full replacement if the resource contains labels (the API doesn't support labels.set yet). We will not support full label replacement in the SDK for now
- Adds more test to verify update of labels on assets

### DIFF
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -372,12 +372,15 @@ class CogniteLabelUpdate:
         self._update_object = update_object
         self._name = name
 
+    def _wrap_labels(self, value: List):
+        return [{"externalId": label} for label in value]
+
     def _add(self, value: List):
-        self._update_object._add(self._name, value)
+        self._update_object._add(self._name, self._wrap_labels(value))
         return self._update_object
 
     def _remove(self, value: List):
-        self._update_object._remove(self._name, value)
+        self._update_object._remove(self._name, self._wrap_labels(value))
         return self._update_object
 
 
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -315,7 +315,7 @@ class CogniteUpdate:
 
     @classmethod
     def _get_update_properties(cls):
-        return [key for key in cls.__dict__.keys() if not key.startswith("_")]
+        return [key for key in cls.__dict__.keys() if (not key.startswith("_")) and (not key == "labels")]
 
 
 class CognitePrimitiveUpdate:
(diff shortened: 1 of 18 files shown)


cognite/client/data_classes/_base.py
```diff
@@ -315,7 +315,7 @@ def dump(self):
 
     @classmethod
     def _get_update_properties(cls):
-        return [key for key in cls.__dict__.keys() if not key.startswith("_")]
+        return [key for key in cls.__dict__.keys() if (not key.startswith("_")) and (not key == "labels")]
```
fredrik955, line 318: We don't support the `set` operation on assets.labels yet

knuto: Thanks - this should serve as a workaround for now!
How about this commit string instead: "We currently ignore labels (e.g. leave any labels unchanged) in full updates"
The commit message + the one for the test in   ' test: add test to discard labels on full resource update' were a bit confusing.

Would probably also be good for other users running into this to have a sentence about labels being ignored in full updates in the doc.

vincent-cognite:
I see you add support for label updates to timeseries, datasets, events...
Is that intentional?

fredrik955:
> I see you add support for label updates to timeseries, datasets, events...
> Is that intentional?

@vincent-cognite It isn't added. The generator now adds a label update class on all resources but they are not exposed.

cognite/client/data_classes/_base.py
fredrik955: Great feedback!
I added a new commit to add a sentence to the documentation about the lack of support for labels when performing a full asset replacement. I will also squash-merge this PR and provide a better overall commit message

openapi/generator.py
```diff
@@ -265,6 +268,7 @@ def generate_attr_update_classes(self, class_name, identation):
             "Primitive": [("set", "Any")],
             "Object": [("set", "Dict"), ("add", "Dict"), ("remove", "List")],
             "List": [("set", "List"), ("add", "List"), ("remove", "List")],
+            "Label": [("add", "List"), ("remove", "List")],
```
hhenriques, line 271: @fredrik955 this might be a bit too general. For example, will we ever want Labels on `ThreeDModelRevision`

fredrik955: Even though we add the class there doesn't mean it will be exposed/used. We add the ObjectPatchClass in many resources without them using it. The generator doesn't care if it is being used as the property definition is defined another place in the generator.

hhenriques: Fair enough! 👍

PR #710, opened 2020-07-13
fix: properly support labels

BugGambit:
This PR is a breaking change to our current labels interface. The reason for that is that the current interface is partially broken and we have gotten several support tickets that it is hard to use.

For reference see:
- https://cognitedata.atlassian.net/browse/CDF-7896
- https://cognitedata.atlassian.net/browse/CDF-7951
- https://cognitedata.atlassian.net/browse/CDF-7889
- https://cognitedata.atlassian.net/browse/CSS-1154
- https://cognitedata.atlassian.net/browse/AC-1641
- https://cognitedata.atlassian.net/browse/CDF-7763
- https://cognitedata.atlassian.net/browse/CSS-1152
- https://cognitedata.slack.com/archives/CG10VQPFX/p1594817270359100
- https://cognitedata.slack.com/archives/CEXTYVBMG/p1593672603068800

With this PR the user experience will be the following:

#### Create label(s)
```python
client.labels.create(LabelDefinition(external_id="PUMP", name="Pump", description="Pump equipment"))
```
and multiple:
```python
client.labels.create([LabelDefinition(external_id="PUMP"), LabelDefinition(external_id="VALVE")])
```

#### List labels:
```python
labels = client.labels.list(name="Pump")
```

#### Delete labels:
```python
client.labels.delete("PUMP")
```
or multiple:
```python
client.labels.delete(["PUMP", "VALVE"])
```

#### Create asset with labels:
```python
asset = Asset(name="my_pump", labels=[Label(external_id="PUMP")])
client.assets.create(assets)
```

#### Filter assets based on labels:
```python
my_label_filter = LabelFilter(contains_all=["PUMP", "VERIFIED"])
asset_list = client.assets.list(labels=my_label_filter)
```

#### Attach/detach labels to/from assets:
```python
my_update = AssetUpdate(id=1).labels.add(["PUMP"]).labels.remove(["VALVE"])
res = c.assets.update(my_update)
```


cognite/client/data_classes/assets.py
```diff
@@ -354,24 +354,40 @@ def retrieve_and_deduplicate(asset_ids):
...
     """
 
-    def __init__(
-        self, contains_any: List[Dict[str, Any]] = None, contains_all: List[Dict[str, Any]] = None, cognite_client=None
-    ):
-        self.contains_any = contains_any
-        self.contains_all = contains_all
+    def __init__(self, contains_any: List[str] = None, contains_all: List[str] = None, cognite_client=None):
+
+        self.contains_any = self._wrap_external_ids(contains_any)
```
BugGambit, line 380: @sanderland Any suggestions here? I don't like to transform it before we actually do the API call as the properties had a setter

sanderland: implement a dump method of the filter that makes the dict?

BugGambit: Done

tests/tests_integration/test_api/test_labels.py
```diff
@@ -27,29 +27,28 @@ def post_spy():
...
-        res = COGNITE_CLIENT.labels.list()
-        assert 0 < len(res) < 100
+        res = COGNITE_CLIENT.labels.list(limit=100)
+        assert 0 < len(res) <= 100
         assert 1 == COGNITE_CLIENT.labels._post.call_count
 
     def test_create_asset_with_label(self, new_label):
-        labels = [{"externalId": new_label.external_id}]
-        ac = COGNITE_CLIENT.assets.create(Asset(name="any", labels=labels))
+        ac = COGNITE_CLIENT.assets.create(
```
sanderland, line 35: do we support a dict for a label here?

tests/tests_unit/test_api/test_assets.py
```diff
@@ -239,31 +239,40 @@ def test_update_multiple(self, mock_assets_response):
...
     def test_ignore_labels_resource_class(self, mock_assets_response):
-        ASSETS_API.update(Asset(id=1, labels=[{"external_id": "PUMP"}], name="Abc"))
-        request_body = json.loads(mock_assets_response.calls[0].request.body)["items"][0]["update"]
-        expected = {"name": {"set": "Abc"}}
-        assert request_body == expected
+        ASSETS_API.update(Asset(id=1, labels=[Label(external_id="Pump")], name="Abc"))
+        assert {"name": {"set": "Abc"}} == jsgz_load(mock_assets_response.calls[0].request.body)["items"][0]["update"]
+
+    def test_labels_filter_contains_all(self, mock_assets_response):
+        my_label_filter = LabelFilter(contains_all=["PUMP", "VERIFIED"])
```
sanderland, line 258: likewise here, does it support dict?
also what happens when you give both contains_all and contains_any?

tests/tests_unit/test_api/test_labels.py
```diff
@@ -29,28 +29,43 @@ def test_list(self, mock_labels_response):
...
     def test_call(self, mock_labels_response):
         list(LABELS_API(limit=10))
         calls = mock_labels_response.calls
         assert 1 == len(calls)
         assert {"cursor": None, "limit": 10, "filter": {}} == jsgz_load(calls[0].request.body)
 
     def test_create_single(self, mock_labels_response):
-        res = LABELS_API.create(Label(external_id="1"))
-        assert isinstance(res, Label)
+        res = LABELS_API.create(LabelDefinition(external_id="1", name="my_label", description="mandatory"))
```
sanderland, line 46: I think this is going to give persistent support questions with people using Label.
Should we use Label for this and sth like AttachedLabel for the other?

CHANGELOG.md
```diff
@@ -16,10 +16,14 @@ Changes are grouped as follows
 
 ## Unreleased	
 
-## [1.8.2] - 2020-07-08
```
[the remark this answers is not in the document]
hhenriques: 😱

cognite/client/__init__.py
```diff
@@ -1,3 +1,3 @@
 from cognite.client._cognite_client import CogniteClient
 
-__version__ = "1.8.1"
+__version__ = "1.9.0"
```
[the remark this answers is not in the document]
hhenriques: Not sure about this. Based on [metrics from the last two weeks](https://docs.google.com/spreadsheets/d/1O_w-4T_csqp20on8UxVg1o687LQRbtATYFyBVygtrf4/edit?usp=sharing) we have quite a lot of usage on Labels. Unfortunately, we don't know if the Asset usage is using Labels.

I think we should bump major. Ultimately, this is a breaking change and we need users to trust that we won't break things on minors.

hhenriques:
I'll leave the actual code review to @sanderland since I don't feel comfortable approving with my current Python knowledge. Just wanted to mention that the syntax is much nicer and I loved how detailed your PR description was :D

tests/tests_unit/test_api/test_labels.py
hhenriques: > I think this is going to give persistent support questions with people using Label.

What do you mean here?

cognite/client/__init__.py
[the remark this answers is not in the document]
sanderland: I think a new major will make people hesitant to upgrade, and may lead to confusion that there is an api v2.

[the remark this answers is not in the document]
hhenriques: > I think a new major will make people hesitant to upgrade, and may lead to confusion that there is an api v2.

We want to separate the SDK version from the API version (especially since the API will move into data versions while the SDK will remain on semantic versioning). For example, we're releasing JS SDK v3 this month.

tests/tests_unit/test_api/test_labels.py
sanderland: people will .create(Label()) and then ask support

cognite/client/__init__.py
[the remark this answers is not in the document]
sanderland: I think that break is better done when you actually move to date versions, rather than now.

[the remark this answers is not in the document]
hhenriques: I'm a firm believer that if we break a running app/service with a minor release then it'll be hard to ask the developers to trust us in the future. This is a breaking change and we should provide that visibility along with an accurate description of what was changes so people feel safe upgrading.

CHANGELOG.md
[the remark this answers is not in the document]
BugGambit: Forgot to update this file: https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/__init__.py

cognite/client/__init__.py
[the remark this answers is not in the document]
BugGambit: I am fine with both. Just a bid sad that we "waste" a major bump on such a silly mistake

tests/tests_integration/test_api/test_labels.py
BugGambit: Nope

tests/tests_unit/test_api/test_assets.py
BugGambit: Doesn't support dict... I am not sure if the user will ever use the dict

BugGambit: The api should return 400 in that case

tests/tests_unit/test_api/test_labels.py
BugGambit: Hm, this is at least consistent with the API. I can rename labels API to label definition api

cognite/client/__init__.py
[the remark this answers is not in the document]
BugGambit: I bumped the version to 2.0.0 and updated the changelog

tests/tests_unit/test_api/test_labels.py
BugGambit: I added a type check for LabelDefinition and raise an exception if they use(d) Label

tests/tests_unit/test_api/test_assets.py
sanderland: but filter does support dict, so isn't that inconsistent here? the main use is not having to import :)

openapi/tests/input_output/input.py
```diff
@@ -46,20 +46,7 @@ class AggregateResultItem(dict):
     # GenStop
 
 
-# GenClass: Asset
-class Asset(CogniteResource):
-    # GenStop
-    def to_pandas(self):
-        pass
```
erlendvollset, line 53: Should we change this to test for example Event instead? So that we still test the `# GenClass` functionality?

tests/tests_integration/test_api/test_iam.py
```diff
@@ -1,82 +0,0 @@
-import pytest
```
erlendvollset, line 1: Why are we deleting these?

BugGambit: See this thread: https://cognitedata.slack.com/archives/CAYSVGPUN/p1594724150121500

openapi/tests/input_output/input.py
BugGambit: Yes, I can change that

cognite/client/data_classes/labels.py
```diff
@@ -51,3 +51,62 @@ class LabelList(CogniteResourceList):
     _RESOURCE = Label
     _UPDATE = None
     _ASSERT_CLASSES = False  # because no Update
+
+
+class AttachedLabel(dict):
```
polomani, line 56: is it possible to have some kind of name with a postfix instead?

PR #715, opened 2020-07-30
Docs fix for events.aggregate_unique_values

haakonvt:
Example in docs specifies wrong function so it throws:
`TypeError: aggregate() got an unexpected keyword argument 'field'`


haakonvt:
@haavardhaugen fyi

PR #716, opened 2020-08-04
Fix for new pandas v1.1.0 behaviour of DataFrame.interpolate to throw…

haakonvt:
… TypeError if called on empty DataFrame


haakonvt:
> checked and ffill is fine - bump patch version and add changelog?

On it 🚀

haakonvt:
Oooh, my editor really hates unnecessary whitespace it seems 🙈

PR #718, opened 2020-08-10
Make HTTP requests *not* follow redirects by default

alexbrasetvik:


alexbrasetvik:
@sanderland In #692 you added tracking of redirects. Are you aware of use cases where handling redirects is necessary?

This PR disables following redirects by default. (If the redirect target can be influenced externally, then it's pretty easy to get access to the token the client is configured with)

sanderland:
Did you read the slack thread in #692? the redirect is http to https, and by some obscure line in the standard this can change POST to GET, leading to very confusing errors

alexbrasetvik:
@sanderland That seems to be about trying to hit our API over plain-text http, which we most definitely should not allow at all. 

So we should be able to apply this.

PR #720, opened 2020-08-11
Update CONTRIBUTING.md

me-ydv-5:


CONTRIBUTING.md
```diff
@@ -1,47 +1,51 @@
...
 If you have an appropriate API key, you can run the integratino tests like this:
 
-`$ pytest tests/tests_integration`
+`python3 -m pytest tests/tests_integration`
 
 If you want to generate code coverage reports run:
 
 ```
-pytest tests/tests_unit --cov-report html \
+python3 -m pytest tests/tests_unit --cov-report html \
```
polomani, line 48: do you don't need `python3 -m` if you run it inside of `pipenv`?

me-ydv-5: Oh, I was switching terminals because of another project and I think I executed tests in a terminal without pipenv. So you're correct.

me-ydv-5:
I removed the dollar signs so we can just double click and copy the command.

PR #722, opened 2020-08-13
feat: CDF-8344 uptake label support in files

maaland:


CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Changes are grouped as follows
 ### Fixed
 - Calls to datapoints `retrieve_dataframe` with `complete="fill"` would break using Pandas version 1.1.0 because it raises TypeError when calling `.interpolate(...)` on a dataframe with no columns.
 
+## [2.2.0] - 2020-08-13
```
BugGambit, line 29: You also need to bump https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/__init__.py

tests/tests_unit/test_api/test_files.py
```diff
@@ -152,6 +155,20 @@ def test_create(self, mock_file_create_response):
         assert FileMetadata._load(response_body) == returned_file_metadata
         assert response_body["uploadUrl"] == upload_url
 
+    def test_create_with_label(self, mock_file_create_response):
```
BugGambit, line 158: Could you also verify the request body and not only the response?

cognite/client/data_classes/files.py
```diff
@@ -46,13 +49,16 @@ def __init__(
         last_updated_time: int = None,
         cognite_client=None,
     ):
+        if labels is not None and len(labels) > 0 and not isinstance(labels[0], Label):
```
vegarsti, line 52: Should the last check there go through the entire list? i.e. `and not all(isinstance(l, Label) for l in labels)`

maaland: hmm good point, this would pass if the first element *is* a `Label` while the rest are not...

vegarsti: I'm not sure how defensive it should be, or how defensive other similar parts of the code are.

maaland: I copied this from assets, so I will need to change it there as well

cognite/client/data_classes/assets.py
```diff
@@ -79,7 +79,7 @@ def __init__(
         aggregates: Union[Dict[str, Any], AggregateResultItem] = None,
         cognite_client=None,
     ):
-        if labels is not None and len(labels) > 0 and not isinstance(labels[0], Label):
+        if labels is not None and len(labels) > 0 and not all(isinstance(l, Label) for l in labels):
```
hhenriques, line 82: Nice improvement here!

Issue #714, opened 2020-07-28
CDF throws internal & OS exceptions


alexchandel:
**Describe the bug**

All `CogniteClient` API calls are prone to throwing implementation-specific exceptions when communication with CDF fails.  Handling these requires catching `requests.exceptions.ConnectionError` (or similar) and retrying the API call.

Users' code should not be coupled to `CogniteClient`'s internal dependencies/implementation when handling network & server-side errors.

Please wrap this in `CogniteMultiException` instead.

**To Reproduce**
This can happen with every single API call.  The log below came from `TimeSeries.update`.  However reproducing it requires a server-side error, or possibly a network failure.

**Expected behavior**
A `CogniteMultiException`, which claims to encompass API failures, should be thrown.

**Log**
```log
…
2020-07-28 11:02:54.161 [ERROR   ] Agent main.run_config_main:    HTTPSConnectionPool(host='api.cognitedata.com', port=443): Max retries exceeded with url: /api/v1/projects/noc-test/timeseries/update (Caused by ProtocolError('Connection aborted.', ConnectionResetError(10054, 'An existing connection was forcibly closed by the remote host', None, 10054, None)))
Traceback (most recent call last):
  File "lib\site-packages\urllib3\connectionpool.py", line 677, in urlopen
  File "lib\site-packages\urllib3\connectionpool.py", line 426, in _make_request
  File "<string>", line 3, in raise_from
  File "lib\site-packages\urllib3\connectionpool.py", line 421, in _make_request
  File "http\client.py", line 1336, in getresponse
  File "http\client.py", line 306, in begin
  File "http\client.py", line 267, in _read_status
  File "socket.py", line 589, in readinto
  File "ssl.py", line 1071, in recv_into
  File "ssl.py", line 929, in read
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "lib\site-packages\requests\adapters.py", line 449, in send
  File "lib\site-packages\urllib3\connectionpool.py", line 725, in urlopen
  File "lib\site-packages\urllib3\util\retry.py", line 439, in increment
[...]

PR #726, opened 2020-08-17
Rewrite http transport logic and improve error handling

erlendvollset:
A few improvements to transport logic. Hoperfully this fixes the connection errors we have been seeing, need someone who can reproduce to verify though (I was told you could help with that @pavelzubarev ?).

Changes:
- Retry logic no longer relies on urllib3's builtin Retry class. This should ensure that a new connection is used when retrying, instead of retrying on the same broken connection. This also means we no longer need two separate connection pools, as retry logic is decoupled from HTTPAdapter. ConnectionRefused (i.e. ECONNREFUSED) is treated as non-retryable, all other connection errors are treated as retryable.
- Moves retry and transport configuration into a _http_client module.
- Adds jitter to exponential backoff so no need for #672 
- Addresses #714 

requests is very inconsistent as to how it communicates connection errors, so I had to do some hacking and search the exception context chain for different requests/urrlib3/built-in exceptions, and classify the failure as either CogniteReadTimeout, CogniteConnectionError, or CogniteConnectionRefused.

### DIFF
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -5,7 +5,7 @@ import numbers
 import os
 import re
 from collections import UserList
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 from urllib.parse import urljoin
 
 import requests.utils
@@ -55,7 +55,7 @@ class APIClient:
         "/relationships/byids",
     }
 
-    def __init__(self, config: utils._client_config.ClientConfig, api_version: str = None, cognite_client=None):
+    def __init__(self, config: utils._client_config.ClientConfig, api_version: str = None, cognite_client=None) -> None:
         self._config = config
         self._api_version = api_version
         self._cognite_client = cognite_client
@@ -90,23 +90,33 @@ class APIClient:
         self._DELETE_LIMIT = 1000
         self._UPDATE_LIMIT = 1000
 
-    def _delete(self, url_path: str, params: Dict[str, Any] = None, headers: Dict[str, Any] = None):
+    def _delete(
+        self, url_path: str, params: Optional[Dict[str, Any]] = None, headers: Optional[Dict[str, Any]] = None
+    ) -> requests.Response:
         return self._do_request("DELETE", url_path, params=params, headers=headers, timeout=self._config.timeout)
 
-    def _get(self, url_path: str, params: Dict[str, Any] = None, headers: Dict[str, Any] = None):
+    def _get(
+        self, url_path: str, params: Optional[Dict[str, Any]] = None, headers: Optional[Dict[str, Any]] = None
+    ) -> requests.Response:
         return self._do_request("GET", url_path, params=params, headers=headers, timeout=self._config.timeout)
 
     def _post(
-        self, url_path: str, json: Dict[str, Any] = None, params: Dict[str, Any] = None, headers: Dict[str, Any] = None
-    ):
+        self,
+        url_path: str,
+        json: Optional[Dict[str, Any]] = None,
+        params: Optional[Dict[str, Any]] = None,
+        headers: Optional[Dict[str, Any]] = None,
+    ) -> requests.Response:
         return self._do_request(
             "POST", url_path, json=json, headers=headers, params=params, timeout=self._config.timeout
         )
 
-    def _put(self, url_path: str, json: Dict[str, Any] = None, headers: Dict[str, Any] = None):
+    def _put(
+        self, url_path: str, json: Optional[Dict[str, Any]] = None, headers: Optional[Dict[str, Any]] = None
+    ) -> requests.Response:
         return self._do_request("PUT", url_path, json=json, headers=headers, timeout=self._config.timeout)
 
-    def _do_request(self, method: str, url_path: str, **kwargs):
+    def _do_request(self, method: str, url_path: str, **kwargs) -> requests.Response:
         is_retryable, full_url = self._resolve_url(method, url_path)
 
         json_payload = kwargs.get("json")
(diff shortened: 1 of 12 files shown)


cognite/client/utils/_client_config.py
```diff
@@ -41,7 +41,7 @@ def __init__(self):
         self.status_forcelist = self._get_status_forcelist()
         self.max_retries = int(os.getenv("COGNITE_MAX_RETRIES", 10))
         self.max_retry_backoff = int(os.getenv("COGNITE_MAX_RETRY_BACKOFF", 30))
-        self.max_connection_pool_size = int(os.getenv("COGNITE_MAX_CONNECTION_POOL_SIZE", 50))
+        self.max_connection_pool_size = int(os.getenv("COGNITE_MAX_CONNECTION_POOL_SIZE", self.max_workers))
```
sanderland, line 44: is this expected, given that someone may thread-parallelize multiple sdk calls?

erlendvollset: Good point. That was probably the reasoning for setting it this high to begin with.

cognite/client/_api_client.py
```diff
@@ -117,13 +56,36 @@ class APIClient:
...
-        self._request_session_with_retry = _REQUESTS_SESSION_WITH_RETRY
-
         self._config = config
         self._api_version = api_version
         self._cognite_client = cognite_client
 
+        self._http_client = HTTPClient(
+            session=GLOBAL_REQUEST_SESSION,
+            config=HTTPClientConfig(
+                status_codes_to_retry={429},
```
pavelzubarev, line 66: I'm not sure I undrestood the change. Are you retry only HTTP/429?

pavelzubarev:
> A few improvements to transport logic. Hoperfully this fixes the connection errors we have been seeing, need someone who can reproduce to verify though (I was told you could help with that @pavelzubarev ?). Although this does not change any public interface, I believe a lot of people have created hotfixes which rely on the SDK's leaky abstractions. IMO this means that we should bump the major version when introducing this change.

The idea is that connection becomes stale or dies (due to the pod being replaced on CDF end f.ex). Current implementation throws errors that connection was forcefully closed and that may lead to a data loss as well if client doesn't handle such case. Unfortunately, I can't reproduce it easily. we can create any long-running application which simply reads the same thing every 15 minutes and just wait until either ambassador or the service node is rolled out by k8s. with the fix there should be no error messages in client's logs

tests/tests_unit/test_http_client.py
```diff
@@ -0,0 +1,155 @@
...
+        c = HTTPClient(
+            config=cnf,
+            session=MagicMock(
+                request=MagicMock(
+                    side_effect=lambda *args, **kwargs: raise_exception_wrapped_as_in_requests_lib(socket.timeout())
+                )
+            ),
+        )
+
+        with pytest.raises(CogniteReadTimeout):
```
rbtcollins, line 93: this strikes me as really odd; either use unittest tests and support libraries, or pytest testing idioms, mixing both is liable to confuse well, everyone :). Not least test suite tooling authors like me :P.

sanderland: I thought we were mostly using pytest with unittest mainly for mock, the entire test suite is like this.

cognite/client/_api_client.py
erlendvollset: There is noe change here. This client is used for create/update endpoints. It will retry 429's and connection errors. Other endpoints use a http client which is configured to retry 429/502/503, connection errors, and read timeouts.

tests/tests_unit/test_http_client.py
erlendvollset: Yeah, not sure what is odd here. Only using MagicMock from unittest, otherwise this is all pytest.

erlendvollset: (This class is not a unittest.TestCase)

erlendvollset:
> > A few improvements to transport logic. Hoperfully this fixes the connection errors we have been seeing, need someone who can reproduce to verify though (I was told you could help with that @pavelzubarev ?). Although this does not change any public interface, I believe a lot of people have created hotfixes which rely on the SDK's leaky abstractions. IMO this means that we should bump the major version when introducing this change.
> 
> The idea is that connection becomes stale or dies (due to the pod being replaced on CDF end f.ex). Current implementation throws errors that connection was forcefully closed and that may lead to a data loss as well if client doesn't handle such case. Unfortunately, I can't reproduce it easily. we can create any long-running application which simply reads the same thing every 15 minutes and just wait until either ambassador or the service node is rolled out by k8s. with the fix there should be no error messages in client's logs

Yeah I understand. The built-in retry logic in urllib3 would perform retries on the same connection over and over again, even if it was stale/dead. Now we retry outside, so if there is a connection error, the connection will be dropped, and we will perform the next attempt on a new connection.

cognite/client/_api_client.py
erlendvollset: see https://github.com/cognitedata/cognite-sdk-python/pull/726/files/a505e0f8d7eca1c9b74fe7070edd5e9f001b0411#diff-e559e86e19971731b9d9a837e8158c5dR76-R88 and https://github.com/cognitedata/cognite-sdk-python/pull/726/files/a505e0f8d7eca1c9b74fe7070edd5e9f001b0411#diff-e559e86e19971731b9d9a837e8158c5dR128-R130

erlendvollset:
@pavelzubarev Been running such a script for 1.5 hours. If there are still no errors tomorrow, this should be ready to ship.

erlendvollset:
Script ran all night without any errors. Can I get a sign-off on this? @cognitedata/core-api @cognitedata/devx-team

pavelzubarev (approved):
I don't have full understanding of SDK flow, but LGTM

sanderland (approved):

BugGambit (approved):

hhenriques (approved):
👍 Thanks for taking the time @erlendvollset !

PR #727, opened 2020-08-25
[SBS-2094] Add py.typed file to inform that package is typed

vegarsti:
We (unstructured search) painfully discovered that mypy doesn't know types of an installed package, even when it has been typed.

This change is apparently needed, and is per [this accepted Python Enhancement Proposal](https://www.python.org/dev/peps/pep-0561/). From [mypy docs](https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages):

> If you would like to publish a library package to a package repository (e.g. PyPI) for either internal or external use in type checking, packages that supply type information via type comments or annotations in the code should put a py.typed file in their package directory.

See also [this PR for the cognite-auth package](https://github.com/cognitedata/python-auth/pull/58).

I updated the minor version. The changelog has to be updated as well, right?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ Changes are grouped as follows
 
 ## Unreleased
 
+## [2.3.0] - 2020-08-25
+
+### Changed
+- Provide type information to users of this package
+
 ## [2.2.2] - 2020-08-18
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Changes are grouped as follows
 ## [2.3.0] - 2020-08-25
 
 ### Changed
-- Provide type information to users of this package
+- Add support for mypy and other type checking tools by adding packaging type information
 
 ## [2.2.2] - 2020-08-18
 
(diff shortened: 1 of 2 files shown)


hhenriques (commented):
> The changelog has to be updated as well, right?

Yup, it needs to be done manually. Otherwise looks good

vegarsti:
Thanks! Will do that then

vegarsti:
@hhenriques something like this?

CHANGELOG.md
```diff
@@ -16,6 +16,11 @@ Changes are grouped as follows
 
 ## Unreleased
 
+## [2.3.0] - 2020-08-25
+
+### Changed
+- Provide type information to users of this package
```
hhenriques, line 22: Maybe something a bit more specific so users know about the new functionality.

```suggestion
- Add Mypy support by adding packaging type information
```

Something like that? Keep in mind I'm not a python developer

vegarsti: Great suggestion!

hhenriques (approved):
Looks great! Thanks for contributing!

vegarsti:
Thank you!

PR #728, opened 2020-08-27
SBS-2037 add directory with prefix filter to files

maaland:


cognite/client/_api/files.py
```diff
@@ -525,6 +534,8 @@ def upload(
         if os.path.isfile(path):
             if not name:
                 file_metadata.name = os.path.basename(path)
+            if not directory:
+                file_metadata.directory = os.path.dirname(path)
```
vegarsti, line 538: Are we sure that we want to fill this in if the user does not provide it?

vegarsti: I see now that this is in the `upload` method, which uploads from a file on disk.

Still, I'm not sure we want to do this. The file name is nice to provide, but the user might not expect that the SDK reads and fills in the directory.

maaland: Yeah I wasn't really sure about this. On one hand it's really convenient *when* the user wants it but as you say, they might not always want that to happen. And since this is the absolute path it might also contain some personal-_ish_ data, like their computer name. 

So perhaps it's better to just not do this at all?

vegarsti: I agree! I vote for no! But I'd be interested to hear others chime in as well

hhenriques: I'm also going to vote for no. Not only could we hit personal data problem but the original file directory might not match the upload directory

hhenriques (changes requested):
Please remove the automatic dirname. 

Also, any chance you could please add a directory to an existing file upload integration test and check if it's there?

cognite/client/_api/files.py
erlendvollset: +1 for not doing this

hhenriques (commented):
Looks good! To release you need to bump the version on `cognite/client/__init__.py` and also add a message to the changelog

hhenriques (approved):
lgtm!

PR #729, opened 2020-08-31
Add partitioning for generator-based listing

mathialo:


cognite/client/_api_client.py
```diff
@@ -263,57 +263,115 @@ def _list_generator(
...
+
+        while len(next_cursors) > 0:
+            tasks_summary = utils._concurrency.execute_tasks_concurrently(
+                get_partition, [(partition,) for partition in next_cursors], max_workers=partitions
+            )
+            if tasks_summary.exceptions:
+                raise tasks_summary.exceptions[0]
+
+            for item in tasks_summary.joined_results():
+                yield cls._RESOURCE._load(item, cognite_client=self._cognite_client)
```
sanderland, line 371: I don't see the point of this, it's not returning as it's retrieving, so has no benefit over for a in c.bla.list()

mathialo: It yields in-between the partitions so only a single page pr partition is read before starting to return, reducing memory usage quite substantially for very large sets of data

cognite/client/_api_client.py
```diff
@@ -263,57 +263,115 @@ def _list_generator(
...
-                    yield cls._RESOURCE._load(item, cognite_client=self._cognite_client)
-            else:
-                current_items.extend(last_received_items)
-                if len(current_items) >= chunk_size:
-                    items_to_yield = current_items[:chunk_size]
-                    current_items = current_items[chunk_size:]
-                    yield cls._load(items_to_yield, cognite_client=self._cognite_client)
+        if partitions:
+            if limit is not None:
+                raise ValueError("When using partitions, limit should be `None`, `-1` or `inf`.")
```
sebbert, line 276: This check won't catch `-1` or `inf`, unless it's preprocessed elsewhere?

```
>>> -1 is None
False
>>> import math
>>> math.inf is None
False
```

mathialo: I assume it's called from `_list_generator`, which has 
``` python
if limit == -1 or limit == float("inf"):
    limit = None
```

cognite/client/_api_client.py
sanderland: aha I get it now, that's a nice solution!

sanderland (changes requested):
looks good, but please add an integration test at least, e.g. list and generator return the same.

cognite/client/_api_client.py
sebbert: Ah, didn't catch that, great 👍

sanderland (approved):

PR #733, opened 2020-09-14
Add Relationships support on Python SDK

me-ydv-5:
Corresponding service contract: https://pr-1019.specs.preview.cogniteapp.com/v1.json.html#tag/Relationships

### DIFF
--- a/cognite/client/__init__.py
+++ b/cognite/client/__init__.py
@@ -1,3 +1,3 @@
 from cognite.client._cognite_client import CogniteClient
 
-__version__ = "2.4.4"
+__version__ = "2.5.0"
--- a/cognite/client/__init__.py
+++ b/cognite/client/__init__.py
@@ -1,3 +1,3 @@
 from cognite.client._cognite_client import CogniteClient
 
-__version__ = "2.4.3"
+__version__ = "2.4.4"
--- a/cognite/client/_api/relationships.py
+++ b/cognite/client/_api/relationships.py
@@ -3,6 +3,7 @@ from typing import *
 from cognite.client import utils
 from cognite.client._api_client import APIClient
 from cognite.client.data_classes import Relationship, RelationshipFilter, RelationshipList
+from cognite.client.data_classes.labels import LabelFilter
 
 
 class RelationshipsAPI(APIClient):
@@ -15,66 +16,48 @@ class RelationshipsAPI(APIClient):
 
     def _create_filter(
         self,
-        source_resource: str = None,
-        source_resource_id: str = None,
-        sources: List[Dict[str, Any]] = None,
-        target_resource: str = None,
-        target_resource_id: str = None,
-        targets: List[Dict[str, Any]] = None,
-        start_time: Dict[str, Any] = None,
-        end_time: Dict[str, Any] = None,
-        confidence: Dict[str, Any] = None,
-        last_updated_time: Dict[str, Any] = None,
-        created_time: Dict[str, Any] = None,
-        data_set: Optional[Union[str, List[str]]] = None,
-        relationship_type: Optional[Union[str, List[str]]] = None,
+        source_external_ids: List[str] = None,
+        source_types: List[str] = None,
+        target_external_ids: List[str] = None,
+        target_types: List[str] = None,
+        data_set_ids: List[Dict[str, Any]] = None,
+        start_time: Dict[str, int] = None,
+        end_time: Dict[str, int] = None,
+        confidence: Dict[str, int] = None,
+        last_updated_time: Dict[str, int] = None,
+        created_time: Dict[str, int] = None,
+        active_at_time: int = None,
+        labels: LabelFilter = None,
     ):
-        if sources and (source_resource or source_resource_id):
-            raise ValueError("Can not set both sources and source_resource/source_resource_id.")
-        if targets and (target_resource_id or target_resource):
-            raise ValueError("Can not set both targets and target_resource/target_resource_id.")
-        if isinstance(relationship_type, str):
-            relationship_type = [relationship_type]
-        if isinstance(data_set, str):
-            data_set = [data_set]
-
-        if source_resource or source_resource_id:
-            sources = [{"resource": source_resource, "resourceId": source_resource_id}]
-        if target_resource or target_resource_id:
-            targets = [{"resource": target_resource, "resourceId": target_resource_id}]
-        if sources:  # remove keys with null values
-            sources = [{k: v for k, v in source.items() if v is not None} for source in sources]
-        if targets:
-            targets = [{k: v for k, v in target.items() if v is not None} for target in targets]
-
         return RelationshipFilter(
-            sources=sources,
-            targets=targets,
+            source_external_ids=source_external_ids,
+            source_types=source_types,
+            target_external_ids=target_external_ids,
+            target_types=target_types,
+            data_set_ids=data_set_ids,
             start_time=start_time,
             end_time=end_time,
             confidence=confidence,
             last_updated_time=last_updated_time,
             created_time=created_time,
-            data_sets=data_set,
-            relationship_types=relationship_type,
+            active_at_time=active_at_time,
+            labels=labels,
         ).dump(camel_case=True)
 
     def __call__(
         self,
-        chunk_size: int = None,
-        source_resource: str = None,
-        source_resource_id: str = None,
-        sources: List[Dict[str, Any]] = None,
-        target_resource: str = None,
-        target_resource_id: str = None,
-        targets: List[Dict[str, Any]] = None,
-        start_time: Dict[str, Any] = None,
-        end_time: Dict[str, Any] = None,
-        confidence: Dict[str, Any] = None,
-        last_updated_time: Dict[str, Any] = None,
-        created_time: Dict[str, Any] = None,
-        data_set: Optional[Union[str, List[str]]] = None,
-        relationship_type: Optional[Union[str, List[str]]] = None,
+        source_external_ids: List[str] = None,
+        source_types: List[str] = None,
+        target_external_ids: List[str] = None,
+        target_types: List[str] = None,
+        data_set_ids: List[Dict[str, Any]] = None,
+        start_time: Dict[str, int] = None,
+        end_time: Dict[str, int] = None,
+        confidence: Dict[str, int] = None,
+        last_updated_time: Dict[str, int] = None,
+        created_time: Dict[str, int] = None,
+        active_at_time: int = None,
+        labels: LabelFilter = None,
         limit: int = None,
     ) -> Generator[Union[Relationship, RelationshipList], None, None]:
         """Iterate over relationships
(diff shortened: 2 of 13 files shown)


cognite/client/beta.py
```diff
@@ -0,0 +1,8 @@
+from cognite.client._api.relationships import RelationshipsAPI
+from cognite.client._cognite_client import CogniteClient as Client
+
+
+class CogniteClient(Client):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.relationships = RelationshipsAPI(self._config, api_version="beta", cognite_client=self)
```
hhenriques, line 8: The problem with doing this is that `api_version` replaced the version on the URL (see `_get_base_url_with_base_path` in `_api_client.py`). 

A possible solution is to add an if on `_get_base_url_with_base_path` to check if the version is `beta` and builds the url with `v1`. You'll also need a if on `_do_request` to add a header with the `headers["version"] = "beta"`.

cognite/client/data_classes/relationships.py
```diff
@@ -1,139 +1,129 @@
...
 
     def __init__(
         self,
-        source: Dict[str, Any] = None,
-        target: Dict[str, Any] = None,
+        external_id: str = None,
+        source_external_id: str = None,
+        source_type: str = None,
+        target_external_id: str = None,
+        target_type: str = None,
```
andreavs, line 32: Wondering if it using None as a default for mandatory fields just makes it more difficult for users - @sanderland is this standard practice in the sdk?

cognite/client/data_classes/relationships.py
```diff
@@ -1,139 +1,129 @@
...
-        rel.target = self._resolve_target(rel.target)
+        self._validate_target(rel.source_type)
+        self._validate_target(rel.target_type)
         return rel
 
     @staticmethod
-    def _resolve_target(target):
+    def _validate_target(target):
         if isinstance(target, dict) or target is None:
             return target
```
andreavs, line 65: source/target type should be a string, not a dict, in the v1 implementation. I would remove this if statement

cognite/client/utils/_auxiliary.py
```diff
@@ -19,6 +19,7 @@
 
 import cognite.client
 from cognite.client import utils
+from cognite.client.data_classes import LabelFilter
```
andreavs, line 22: is this used for anything?

andreavs (changes requested):

cognite/client/data_classes/relationships.py
me-ydv-5: I left them as optional because the API will anyway give back an error

cognite/client/utils/_auxiliary.py
me-ydv-5: Not sure how it got added here.

cognite/client/data_classes/relationships.py
andreavs: its bad design imo if the default arguments cause errors to be raised

cognite/client/utils/_auxiliary.py
me-ydv-5: Ah, I was I think doing some experimentation.

cognite/client/data_classes/relationships.py
sanderland: typically we don't =None for mandatory arguments in function calls. Data classes used to be autogenerated so they have this =None to make ._load stable and such

cognite/client/data_classes/relationships.py
```diff
@@ -1,139 +1,129 @@
...
         rel = copy.copy(self)
-        rel.source = self._resolve_target(rel.source)
-        rel.target = self._resolve_target(rel.target)
+        self._validate_target(rel.source_type)
+        self._validate_target(rel.target_type)
         return rel
 
     @staticmethod
-    def _resolve_target(target):
+    def _validate_target(target):
```
sanderland, line 63:
```suggestion
    def _validate_resource_type(resource_type):
```

cognite/client/data_classes/relationships.py
```diff
@@ -1,139 +1,129 @@
...
-            Event: "Event",
-            Sequence: "Sequence",
-        }
-        typestr = _TARGET_TYPES.get(target.__class__)
-        if typestr:
-            return {"resource": typestr, "resourceId": target.external_id}
-        raise ValueError("Invalid source or target '{}' of type {} in relationship".format(target, target.__class__))
+        _TARGET_TYPES = {"asset", "timeseries", "file", "event", "sequence"}
+        if target.lower() not in _TARGET_TYPES:
+            raise ValueError("Invalid source or target '{}' in relationship".format(target))
```
sanderland, line 69: type

cognite/client/data_classes/relationships.py
```diff
@@ -1,139 +1,129 @@
...
 
 
 class RelationshipFilter(CogniteFilter):
-    """Filter on relationships with exact match. Multiple filter elments in one property, e.g. `dataSets: [ "a", "b" ]`, will return all relationships where the dataSet field is either `a` or `b`. Filters in multiple properties will return the relationships that match all criteria. Filters on a `resourceId` without a `resource` (type) in sources and targets will return relationships that match the resourceId and match any resource type. If the filter is not specified it default to an empty filter.
+    """Filter on relationships with exact match. Multiple filter elments in one property, e.g. sourceExternalIds: [ "a", "b" ], will return all relationships where the sourceExternalId field is either a or b. Filters in multiple properties will return the relationships that match all criteria. If the filter is not specified it default to an empty filter.
```
sanderland, line 73: weird documentation change - the previous one was autogenerated so changing it to a different non-sdk related one doesn't make sense.

sanderland (changes requested):

cognite/client/data_classes/relationships.py
me-ydv-5: I just copied the service contract here. Should this be reverted?

hhenriques:
@me-ydv-5 don't forget to add relationships to the changelog and to bump the version on `cognite\client\__init__.py`

cognite/client/_api/relationships.py
```diff
@@ -15,109 +16,86 @@ def __init__(self, *args, **kwargs):
...
+        source_types: List[str] = None,
+        target_external_ids: List[str] = None,
+        target_types: List[str] = None,
+        data_set_ids: List[Dict[str, Any]] = None,
+        start_time: Dict[str, int] = None,
+        end_time: Dict[str, int] = None,
+        confidence: Dict[str, int] = None,
+        last_updated_time: Dict[str, int] = None,
+        created_time: Dict[str, int] = None,
+        active_at_time: int = None,
```
caelyn-cognite, line 29: `active_at_time` is a range filter like the other time based filters.

caelyn-cognite (changes requested):

cognite/client/data_classes/relationships.py
sanderland: I would make the docstring relevant python code `e.g. RelationshipFilter( .... )`

sanderland:
> @sanderland any recommendations on the version bump since it's a beta? The JS is a separate package so version management is a bit easier

I would just bump minor, experimental already has warnings for such. Major version bumps are quite costly throughout all the million dependencies that need to be adjusted.

cognite/client/experimental.py
```diff
@@ -13,4 +12,3 @@ class CogniteClient(Client):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.model_hosting = ModelHostingAPI(self._config, api_version="playground", cognite_client=self)
```
sanderland, line 14: discussed with @hhenriques and we decided to remove MH here as well, and just have experimental inherit from beta - to be removed on the next version bump. Probably would be good to give a warning in init as well to use beta.

caelyn-cognite (commented):
The issue I saw has been fixed 👍

cognite/client/data_classes/relationships.py
```diff
@@ -1,139 +1,126 @@
...
+        source_type: str = None,
+        target_external_id: str = None,
+        target_type: str = None,
         start_time: int = None,
         end_time: int = None,
         confidence: float = None,
-        data_set: str = None,
-        external_id: str = None,
-        relationship_type: str = None,
+        data_set_id: Dict[str, Any] = None,
```
Tapped, line 36: This doesn't seem to be correct. The API expects this to be just a number, not an object. The documentation and I tried myself as well. See https://pr-1019.specs.preview.cogniteapp.com/v1.json.html#tag/Relationships/paths/~1api~1v1~1projects~1{project}~1relationships~1list/post. 

You should add a test where you use data_set_id when creating a relationship.

Tapped: This is also consistent with our other APIs, though I think it's weird that we don't support setting it directly from an externalId when creating, while being able to filter just by the externalId.

me-ydv-5: Good catch! I didn't add a test for datasetId because there's no way to delete them and so every time a new datasetId would get created but yeah there should be one.
And we can filter both by external and internal id, if I understood your statement correctly.

cognite/client/__init__.py
```diff
@@ -1,3 +1,3 @@
 from cognite.client._cognite_client import CogniteClient
 
-__version__ = "2.4.3"
+__version__ = "2.4.4"
```
sanderland, line 3: this seems more than a patch

cognite/client/_api/relationships.py
```diff
@@ -15,109 +16,86 @@ def __init__(self, *args, **kwargs):
...
-            created_time (Dict[str, Any]): Range to filter the field for. (inclusive)
-            data_set (Union[str,List[str]): Filter on any of a given list of dataSets.
-            relationship_type (Union[str,List[str]):  Filter on any of a given list o relationship types.
-            limit (int, optional): Maximum number of relationships to return. Defaults to 100. Set to -1, float("inf") or None
-                to return all items.
+            created_time (Dict[str, int]): Range to filter the field for. (inclusive)
+            active_at_time (int): Limits results to those active at any point within the given time range, i.e. if there is any overlap in the intervals [activeAtTime.min, activeAtTime.max] and [startTime, endTime], where both intervals are inclusive. If a relationship does not have a startTime, it is regarded as active from the begining of time by this filter. If it does not have an endTime is will be regarded as active until the end of time. Similarly, if a min is not supplied to the filter, the min will be implicitly set to the beginning of time, and if a max is not supplied, the max will be implicitly set to the end of time.
```
sanderland, line 78: this still has int type all over but mentions 'the given range'.

cognite/client/utils/_experimental_warning.py
```diff
@@ -14,7 +14,9 @@ def decorator(*args, **kwargs):
             warnings.warn(
                 "\nThe {} API is currently experimental, so this functionality does not adhere to semantic versionining."
                 "\nThis means that this API may be subject to breaking changes even between patch versions."
-                "\nYou should NOT use the {} API in any production code.".format(api_name, api_name),
+                "\nYou should NOT use the {} API in any production code. Consider using the 'beta' client instead.".format(
```
sanderland, line 17:
```suggestion
              "\nThe experimental client is deprecated and will be removed soon. Consider using the 'beta' client instead.".format(
```

tests/tests_unit/test_api/test_relationships.py
```diff
@@ -84,74 +79,96 @@ def test_create_single(self, mock_rel_response):
...
+                )
             )
             assert isinstance(res, Relationship)
 
         for call in mock_rel_response.calls:
-            it = json.loads(gzip.decompress(call.request.body).decode("utf-8"))["items"][0]
-            assert isinstance(it["source"], dict)
-            assert isinstance(it["target"], dict)
+            x = json.loads(gzip.decompress(call.request.body).decode("utf-8"))["items"]
+            print(x)
```
sanderland, line 135: no print

tests/tests_unit/test_api/test_relationships.py
```diff
@@ -176,42 +189,63 @@ def test_delete_multiple(self, mock_rel_response):
...
         assert mock_rel_response.calls[0].response.json()["items"] == res.dump(camel_case=True)
 
     def test_source_target_packing(self, mock_rel_response):
         res = REL_API.list(
-            source_resource="asset",
-            source_resource_id="bla",
-            target_resource="timeseries",
-            target_resource_id="foo",
-            relationship_type="belongs_to",
+            source_types="asset",
```
sanderland, line 202: you specify this as List[str] but then call it with str  - it's nice to accept both, but make sure it can in terms of api, and update docs

cognite/client/__init__.py
me-ydv-5: `2.5.0` ?

tests/tests_unit/test_api/test_relationships.py
me-ydv-5, line 135:
```suggestion
```

tests/tests_unit/test_api/test_relationships.py
me-ydv-5: Great catch!

sanderland (approved):
LGTM

andreavs (approved):
LGTM

PR #735, opened 2020-09-16
move directory field and update description to differentiate it from path

maaland:
I realized that the new `directory` field could be a bit confused when placed right next to `path` in the documentation, so I wanted to make the difference a bit more clear.


vegarsti (approved):
Great!

erlendvollset (approved):

PR #740, opened 2020-10-02
Add more contraints to relationships test

me-ydv-5:


tests/tests_integration/test_api/test_relationships.py
```diff
@@ -100,7 +103,7 @@ class TestRelationshipsAPI:
     def test_get_single(self, new_relationship):
         res = API_REL.retrieve(external_id=new_relationship.external_id)
         assert isinstance(res, Relationship)
-        assert res.dump()["confidence"] == 0.9
+        assert res.dump()["confidence"] == 1
```
me-ydv-5, line 106: This was a failing edge case a few days ago in data viz team.

sanderland: what was wrong? floating point roundoff ?

andreavs: this edge case never failed for the retrieve endpoint - it would fail for the filter endpoint (for the curious, the confidence filter would run on all filter requests, and the default values were set to the extreme allowed values for confidence, and since the upper limit was also inclusive, every single filter request would run 0 <= confidence < 1 - big oopsie)

tests/tests_integration/test_api/test_relationships.py
```diff
@@ -17,16 +17,19 @@
 
 
 @pytest.fixture
-def new_relationship():
+def new_relationship(new_label):
     external_id = uuid.uuid4().hex[0:20]
+    pre_existing_data_set = 3603480159997264
```
andreavs, line 22: i guess this means the tests will fail if run on any other tenant than the sdk integration test tenant?

me-ydv-5: yes. I couldn't think of any other way to by pass the mess of data-sets 😅

andreavs: its good to test that data sets are working, but i would argue for either creating the data set with an external id, and then getting the internal id by fetching the data set, or setting the data set id by an env variable, although neither solution is very clean - how are data set tests handled elsewhere in the sdk?

me-ydv-5: There is no test that includes data sets other than data set integration tests. Nevertheless I agree it should be fetched. So changed that part.

sanderland (approved):

tests/tests_integration/test_api/test_relationships.py
me-ydv-5: I've added the test for that in filter

andreavs (approved):
lgtm :+1:

PR #743, opened 2020-10-22
implement geoLocation in files

maaland:


cognite/client/data_classes/shared.py
```diff
@@ -47,3 +47,94 @@ def __init__(self, count: int = None, value: Union[int, str] = None, **kwargs):
         self.value = value
 
     value = CognitePropertyClassUtil.declare_property("value")
+
+
+class Geometry(dict):
```
me-ydv-5, line 52: Shouldn't it be put inside `data_classes/files.py`?

tests/tests_integration/test_api/test_files.py
```diff
@@ -100,3 +135,12 @@ def test_retrieve_file_with_labels(self, new_file_with_label):
         res = COGNITE_CLIENT.files.retrieve(id=file.id)
         assert len(res.labels) == 1
         assert res.labels[0].external_id == label_external_id
+
+    def test_filter_file_on_geoLocation(self, new_file_with_geoLocation, mock_geo_location):
+        file = new_file_with_geoLocation
```
vegarsti, line 140: This line can be removed, I think? I don't see it being used elsewhere

cognite/client/data_classes/shared.py
maaland: Eh, I thought this was cleaner because otherwise the geolocation related classes would have to be defined at the top of `data_classes/files.py` and I thought that made the `data_classes/files.py` file less readable. 

But no, they're technically not shared (yet, at least). No strong opinions on my side so I can move them back if you like.

tests/tests_integration/test_api/test_files.py
maaland: It is used in the assertion at the bottom. But it doesn't have to be its own line, so I moved it down to be inline :)

cognite/client/_api/files.py
```diff
@@ -36,6 +40,7 @@ def __call__(
         data_set_ids: List[int] = None,
         data_set_external_ids: List[str] = None,
         labels: LabelFilter = None,
+        geo_location: GeoLocationFilter = None,
```
me-ydv-5, line 43: nit: Mabe adding a new argument at the end of the argument list maybe more suitable for better backwards compatibility? Similarly for next two changes.

cognite/client/_api/files.py
```diff
@@ -393,7 +411,6 @@ def update(
                 >>> from cognite.client.data_classes import FileMetadataUpdate
                 >>> c = CogniteClient()
                 >>> my_update = FileMetadataUpdate(id=1).source.set("new source").metadata.add({"key": "value"})
-                >>> res = c.files.update(my_update)
```
me-ydv-5, line 396: Is this intentional?

cognite/client/_api/files.py
```diff
@@ -527,6 +553,7 @@ def upload(
             asset_ids=asset_ids,
             data_set_id=data_set_id,
             labels=labels,
+            geo_location=geo_location,
```
me-ydv-5, line 556: Similar suggestion as first one. If there's any specific reason then I guess it's fine here as well.

cognite/client/_api/files.py
maaland: Oops. no. No idea how that happened 😅

cognite/client/data_classes/shared.py
me-ydv-5: Yeah IMHO it would be better to put it in files.py for as long as they are not being shared anywhere else? :)

cognite/client/_api/files.py
maaland: No particular reason. I think I added `labels` right below where `data_set_ids` were added, not sure why they were added in the middle and not on the bottom. And then now I added `geo_location` right below where I added `labels`.

Shouldn't matter when named arguments are used though?

cognite/client/data_classes/shared.py
```diff
@@ -47,3 +47,94 @@ def __init__(self, count: int = None, value: Union[int, str] = None, **kwargs):
...
+    """Represents the points, curves and surfaces in the coordinate space.
+
+    Args: type (str): The geometry type. One of 'Point', 'MultiPoint, 'LineString', 'MultiLineString', 'Polygon', or 'MultiPolygon'.
```
me-ydv-5, line 55: I think it's better to put a sanity check on the `type` arg supplied in this constructor? ValueError if there's some input that you don't want.

cognite/client/data_classes/shared.py
```diff
@@ -47,3 +47,94 @@ def __init__(self, count: int = None, value: Union[int, str] = None, **kwargs):
...
+          coordinates (List): An array of the coordinates of the geometry. The structure of the elements in this array is determined by the type of geometry.
```
me-ydv-5, line 56: Umm, that's bit unclear to me what structure to use for let's say `MultiLineString` or anything else.

cognite/client/data_classes/shared.py
```diff
@@ -47,3 +47,94 @@ def __init__(self, count: int = None, value: Union[int, str] = None, **kwargs):
...
+    type = CognitePropertyClassUtil.declare_property("type")
+    coordinates = CognitePropertyClassUtil.declare_property("coordinates")
+
+
+class GeoLocation(dict):
+    """A GeoLocation object conforming to the GeoJSON spec.
+
+    Args: type (str): The GeoJSON type. Currently only 'Feature' is supported.
+          geometry (object): The geometry type. One of 'Point', 'MultiPoint, 'LineString', 'MultiLineString', 'Polygon', or 'MultiPolygon'.
+          properties (object): Additional properties in a String key -> Object value format.
```
me-ydv-5, line 95: Is this a metadata associated to files?

tests/tests_integration/test_api/test_files.py
```diff
@@ -21,6 +37,18 @@ def new_file():
     assert COGNITE_CLIENT.files.retrieve(id=res.id) is None
 
 
+@pytest.fixture(scope="class")
+def new_file_with_geoLocation(mock_geo_location):
+    res = COGNITE_CLIENT.files.upload_bytes(content="blabla", name="geoLocationFile", geo_location=mock_geo_location)
+    while True:
+        if COGNITE_CLIENT.files.retrieve(id=res.id).uploaded:
+            break
+        time.sleep(0.5)
```
me-ydv-5, line 46: What is this for?

tests/tests_unit/test_api/test_files.py
```diff
@@ -170,12 +188,20 @@ def test_create_with_label_request(self, mock_file_create_response):
...
         assert all(body["labels"][0]["externalId"] == "WELL LOG" for body in [request_body, response_body])
 
-    def test_create(self, mock_file_create_response):
-        file_metadata = FileMetadata(name="bla")
+    def test_create_with_geoLocation(self, mock_file_create_response, mock_geo_location):
+        file_metadata = FileMetadata(name="bla", geo_location=mock_geo_location)
         returned_file_metadata, upload_url = FILES_API.create(file_metadata)
         response_body = mock_file_create_response.calls[0].response.json()
         assert FileMetadata._load(response_body) == returned_file_metadata
-        assert response_body["uploadUrl"] == upload_url
```
me-ydv-5, line 178: Might as well keep this assertion.

cognite/client/_api/files.py
me-ydv-5: Yup.

me-ydv-5 (commented):
Looks good! please see some comments left.

cognite/client/data_classes/shared.py
maaland: Good point.

cognite/client/data_classes/shared.py
maaland: Fair. I'll flesh out the documentation a bit, based on what I've written in `service-contracts`. 
@maciejmrowiec do you think it's worth it to go for the approach you suggested of building up separate projects from point -> array[point] etc. here as well?

cognite/client/data_classes/shared.py
maaland: Similar to metadata, but it's related to the `geoLocation`. Optional though, I'll clarify that.

tests/tests_unit/test_api/test_files.py
maaland: Copied it from the `new_file_with_label()` fixture below, but I believe it's because the file is not always available immediately after upload.

tests/tests_integration/test_api/test_files.py
maaland: Copied it from the new_file_with_label() fixture below, but I believe it's because the file is not always available immediately after upload.

cognite/client/data_classes/shared.py
me-ydv-5: @maaland please look into this.

cognite/client/data_classes/shared.py
me-ydv-5: Did we reach on any consensus here?

cognite/client/data_classes/shared.py
maaland: Ah I implemented it on Friday but forgot to `git push` 😆

cognite/client/data_classes/shared.py
maaland: I've been trying various ways of doing this. in the `service-contracts` PR with hopes of drawing inspiration from that. Haven't found a solution I'm happy with there yet. 

While we wait for @maciejmrowiec 's input, what are your thoughts, @me-ydv-5? The options are to either just explain the different shapes a bit more in text (with a link to [this page](https://github.com/cognitedata/doc.cognitedata.com/pull/543) , or to construct `Point`, `LineString` etc. as objects of each other. The latter is naturally more work but I think it will look cleaner.

me-ydv-5: Hmm, I think it's better to have python objects for each of the geojson objects (GO) given we derive one GO from other. Overall I agree it might cost some of your developer time but I guess that gets compensated with better user experience and early input validation (even though partially). :)

maaland: Just got the time to look at this again. I like the cleanliness of deriving the GO from each other. However I worry that the usability will suffer, since developers can't just paste in a big array anymore (unless we add some sort of parsing and transformation).

Examples
With the current approach, a developer can construct a `GeoLocation` object with a `Polygon` like this:
```
    geometry = Geometry(type="Polygon", coordinates=[
        [[30, 10], [40, 40], [20, 40], [10, 20], [30, 10]]
    ])
    gl = GeoLocation(type="Feature", geometry=geometry, properties=dict())
```

With the suggested approach, the developer would have to construct each object individually, as they rely on each other:
```
p1 = Point(coordinates=[30, 10])
p2 = Point(coordinates=[40, 40])
... etc

ls = LineString(coordinates=[p1, p2, ...])
polygon = Polygon(coordinates=[ls])
```

So more steps. Depending on how this feature will actually be used, this might be fine. And we can add constructors to just take in the big array and then parse it accordingly. But that's even more work.

me-ydv-5: Letting it go for the time being. Can be changed in the future based on reviews from users, etc.

me-ydv-5 (dismissed):

maciejmrowiec:
Where can I find the render of sdk ref docs?

maaland:
> Where can I find the render of sdk ref docs?

@maciejmrowiec the only way to see the sdk docs of a non-master branch is to generate them locally. See https://github.com/cognitedata/cognite-sdk-python/blob/master/CONTRIBUTING.md#documentation

maaland:
@maciejmrowiec here's a snippet of the documentation, I used the same descriptions as in the `service-contracts`
image

me-ydv-5 (approved):

PR #745, opened 2020-10-23
Fix docs example in relationships

me-ydv-5:


CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.5.5] - 2020-10-23
+
+### Fixed
+- Incorrect example for creating relationships
```
hhenriques, line 20: Given that this is a docs change, I don't think it should be it in the change log.

cognite/client/__init__.py
```diff
@@ -1,3 +1,3 @@
 from cognite.client._cognite_client import CogniteClient
 
-__version__ = "2.5.4"
+__version__ = "2.5.5"
```
hhenriques, line 3: We're on making a change to docs right? If so, we shouldn't bump

hhenriques (changes requested):

cognite/client/__init__.py
me-ydv-5: That makes my work easier. Thanks!

hhenriques (approved):
lgtm

PR #753, opened 2020-11-03
Add support for Azure native token generation - DRAFT

caelyn-cognite:
Draft PR for adding Azure native token support to python sdk.

Known issues:

- There is no token lifecycle management. When this token expires, the client will become invalid. The client returns a `401 - Unauthorized` error when the token expires. Token creation returns time til expiry. If we wanted to detect token expiration, we would have to track time ourselves. We don't.
- The msal library is being pulled in as a local import and has not been added to `install_requires` in `setup.py` - there are several other points in the sdk where we use local import on packages missing from that list. So, either those are also mistakes, or we can be comfortable with this import generating an error if someone does not have the package locally (and then only if they have all three azure related environment variables set!)
- There is no integration test for this. I have run it against my own tenant; but there would be some work to set one up and would mostly be replicating an auth smoke test.
- The use of the default scope was copied from [here](https://github.com/cognitedata/auth-smoke/blob/0494008358050e746592e15d340e9cbf3d5183e0/auth-smoke/login_tests.py#L51).

### DIFF
--- a/cognite/client/utils/_azure_token_generation.py
+++ b/cognite/client/utils/_azure_token_generation.py
@@ -1,0 +1,40 @@
+import os
+
+from cognite.client.exceptions import CogniteAPIKeyError
+from cognite.client.utils import _auxiliary
+
+
+def _azure_environment_vars_set():
+    azure_api_client_id = os.getenv("AZURE_API_CLIENT_ID")
+    azure_api_client_secret = os.getenv("AZURE_API_CLIENT_SECRET")
+    oidc_authority = os.getenv("OIDC_AUTHORITY")
+
+    return azure_api_client_id is not None and azure_api_client_secret is not None and oidc_authority is not None
+
+
+def _generate_access_token():
+    # Microsoft Authentication Library (MSAL) for Python
+    # https://github.com/AzureAD/microsoft-authentication-library-for-python
+    msal = _auxiliary.local_import("msal")
+    azure_api_client_id = os.getenv("AZURE_API_CLIENT_ID")
+    azure_api_client_secret = os.getenv("AZURE_API_CLIENT_SECRET")
+    oidc_authority = os.getenv("OIDC_AUTHORITY")
+    cognite_base_url = os.getenv("COGNITE_BASE_URL", "https://api.cognitedata.com")
+
+    app = msal.ConfidentialClientApplication(
+        azure_api_client_id, client_credential=azure_api_client_secret, authority=oidc_authority
+    )
+    # When using the client-credentials flow with msal the only valid scope is .default
+    scopes = [f"{cognite_base_url}/.default"]
+    print(app)
+    response = app.acquire_token_for_client(scopes=scopes)
+    print(response)
+    access_token = response.get("access_token")
+    print(access_token)
+    if access_token is None:
+        print("should be here")
+        raise CogniteAPIKeyError(
+            "Could not generate azure access token from provided azure related environment variables"
+        )
+
+    return access_token
--- a/tests/tests_unit/test_utils/test_azure_token_generation.py
+++ b/tests/tests_unit/test_utils/test_azure_token_generation.py
@@ -1,0 +1,57 @@
+import os
+from unittest.mock import patch
+
+import pytest
+
+from cognite.client import utils
+from cognite.client.exceptions import CogniteAPIKeyError
+from tests.utils import unset_env_var
+
+
+@pytest.fixture
+def azure_environment_vars():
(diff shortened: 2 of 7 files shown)


ngerstle-cognite:
I think we should have a discussion about where token lifecycle management should live, but it's indeed simplest to start without it.

In general it would be great to not focus too much on AzureAD (or ADFS), but rather generic oauth2 client credentials flow use.

cognite/client/utils/_azure_token_generation.py
```diff
@@ -0,0 +1,40 @@
...
+    azure_api_client_secret = os.getenv("AZURE_API_CLIENT_SECRET")
+    oidc_authority = os.getenv("OIDC_AUTHORITY")
+
+    return azure_api_client_id is not None and azure_api_client_secret is not None and oidc_authority is not None
+
+
+def _generate_access_token():
+    # Microsoft Authentication Library (MSAL) for Python
+    # https://github.com/AzureAD/microsoft-authentication-library-for-python
+    msal = _auxiliary.local_import("msal")
```
ngerstle-cognite, line 18: We'll likely end up using the ADAL python library (https://github.com/AzureAD/azure-activedirectory-library-for-python) for Aramco (they use ADFS 2016, MSAL only supports ADFS2019), but we don't yet have an ADFS to test with.

cognite/client/utils/_azure_token_generation.py
```diff
@@ -0,0 +1,40 @@
+import os
+
+from cognite.client.exceptions import CogniteAPIKeyError
+from cognite.client.utils import _auxiliary
+
+
+def _azure_environment_vars_set():
+    azure_api_client_id = os.getenv("AZURE_API_CLIENT_ID")
+    azure_api_client_secret = os.getenv("AZURE_API_CLIENT_SECRET")
+    oidc_authority = os.getenv("OIDC_AUTHORITY")
```
ngerstle-cognite, line 10: This is very Azure AD specific (not Azure, but Azure AD). We'll want to make this more general 2021Q1..

cognite/client/utils/_azure_token_generation.py
ngerstle-cognite: (if we want to use such an IdP specific library at all. The client credentials flow is pretty simple, IMHO).

cognite/client/utils/_azure_token_generation.py
```diff
@@ -0,0 +1,40 @@
...
+    cognite_base_url = os.getenv("COGNITE_BASE_URL", "https://api.cognitedata.com")
+
+    app = msal.ConfidentialClientApplication(
+        azure_api_client_id, client_credential=azure_api_client_secret, authority=oidc_authority
+    )
+    # When using the client-credentials flow with msal the only valid scope is .default
+    scopes = [f"{cognite_base_url}/.default"]
```
timothy-mehay-cognite, line 28: Does this result in a usable token? I recall needing to set the scope to something along the lines of `$AAD_APP_REGISTRATION_CLIENT_ID/.default` in order for the resulting token to be useable.

tests/tests_unit/test_utils/test_azure_token_generation.py
```diff
@@ -0,0 +1,57 @@
...
+    @pytest.mark.parametrize("missing", ["AZURE_API_CLIENT_ID", "AZURE_API_CLIENT_SECRET", "OIDC_AUTHORITY"])
+    def test_missing_azure_environment_vars(self, missing, azure_environment_vars):
+        with unset_env_var(missing):
+            assert False == utils._azure_token_generation._azure_environment_vars_set()
+
+
+class TestGenerateAccessToken:
+    @patch("msal.ConfidentialClientApplication")
+    def test_access_token_generated(self, msal_client_mock):
+        msal_client_mock().acquire_token_for_client.return_value = {"access_token": "azure_token"}
```
timothy-mehay-cognite, line 45: It might be worth adding a dedicated AD tenant for testing purposes and using that to create tokens. While they wouldn't work for auth purposes you'd be able to test that the corresponding claims have the expected values.

cognite/client/utils/_azure_token_generation.py
ngerstle-cognite: The `${cognite_base_ul}` isn't guaranteed to work- it's fully possible to use some other scope if so configured in the IDP. 
In general, this should be something like `{resource_id}/.default` or `{api_identifier}/.default` or just `{scope}`

caelyn-cognite:
Closed in favour of a more generic follow up PR

PR #756, opened 2020-11-05
[CDF-9834] Return empty AssetList if asset does not exist on retrieve_subtree

me-ydv-5:
https://cognitedata.atlassian.net/browse/CDF-9834

### DIFF
--- a/cognite/client/_api/assets.py
+++ b/cognite/client/_api/assets.py
@@ -421,7 +421,10 @@ class AssetsAPI(APIClient):
                 >>> from cognite.client.data_classes import AssetUpdate
                 >>> c = CogniteClient()
                 >>> my_update = AssetUpdate(id=1).description.set("New description").metadata.add({"key": "value"})
-                >>> res = c.assets.update(my_update)
+                >>> res1 = c.assets.update(my_update)
+                >>> # Remove an already set field like so
+                >>> another_update = AssetUpdate(id=1).description.set(None)
+                >>> res2 = c.assets.update(another_update)
 
             Attach labels to an asset::
 
@@ -499,7 +502,7 @@ class AssetsAPI(APIClient):
             search={"name": name, "description": description, "query": query}, filter=filter, limit=limit
         )
 
-    def retrieve_subtree(self, id: int = None, external_id: str = None, depth: int = None) -> AssetList:
+    def retrieve_subtree(self, id: int = None, external_id: str = None, depth: int = None) -> Optional[AssetList]:
         """Retrieve the subtree for this asset up to a specified depth.
 
         Args:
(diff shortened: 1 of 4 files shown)


sanderland:
Why should this return None and not raise an exception?

me-ydv-5:
> Why should this return None and not raise an exception? Not sure myself, but let's get consensus.

Because retrieving a non-existing asset does not raise an error. I had first raised an error but found that to be a little non uniform with other endpoints.

erlendvollset:
The other endpoints returning an list would not return None. It should be either an empty AssetList or an exception imo.

erlendvollset:
> > Why should this return None and not raise an exception? Not sure myself, but let's get consensus.
> 
> Because retrieving a non-existing asset does not raise an error. I had first raised an error but found that to be a little non uniform with other methods.

retrieve does not raise an exception, but retrieve_multiple does.

me-ydv-5:
crap! How do I undelete this?

pymishi:
I think that makes more sense to return None as this will reflect that the data does not exists based on your criteria and not an exception that the returned type was not the one expected, I would expect that kind of behavior in a native code but in a SDK it should return information based on what you are working with. In this case I am looking for data and if get nothing I expect that the SDK would give me that answer.

me-ydv-5:
> I think that makes more sense to return None as this will reflect that the data does not exists based on your criteria and not an exception that the returned type was not the one expected, I would expect that kind of behavior in a native code but in a SDK it should return information based on what you are working with. In this case I am looking for data and if get nothing I expect that the SDK would give me that answer.

@pymishi yeah that is what's confusing me, for example [retrieving an asset](https://cognite-docs.readthedocs-hosted.com/projects/cognite-sdk-python/en/latest/cognite.html#retrieve-an-asset-by-id) returns None while it should tell the user that the asset does not exist instead, correct?

pymishi:
> > I think that makes more sense to return None as this will reflect that the data does not exists based on your criteria and not an exception that the returned type was not the one expected, I would expect that kind of behavior in a native code but in a SDK it should return information based on what you are working with. In this case I am looking for data and if get nothing I expect that the SDK would give me that answer.
> 
> @pymishi yeah that is what's confusing me, for example [retrieving an asset](https://cognite-docs.readthedocs-hosted.com/projects/cognite-sdk-python/en/latest/cognite.html#retrieve-an-asset-by-id) returns None while it should tell the user that the asset does not exist instead, correct?

Yes, my expectation from a SDK that even that is a development kit, it still a product, and I expect that the error handling to be better refined than working directly with the API calls where I would need to have that kind of error handling. I may be wrong but that is what I would like for how a SDK as I expect that if facilitate my work and reduce the need that I need to code as much I would need if working without it.

me-ydv-5:
@sanderland @erlendvollset seems like returning `None` is better than throwing error after reading Patrick's comments.

sanderland:
I think an empty assetlist may be a good option then @me-ydv-5 - this would be like a list/filter call with missing root ids, which I think also returns an empty list. the result is also Falsy so people can if res: blah

cognite/client/_api/assets.py
```diff
@@ -509,10 +512,12 @@ def retrieve_subtree(self, id: int = None, external_id: str = None, depth: int =
                 subtree.
 
         Returns:
-            AssetList: The requested assets.
+            AssetList: The requested assets or None if asset does not exist.
         """
         utils._auxiliary.assert_exactly_one_of_id_or_external_id(id, external_id)
         asset = self.retrieve(id=id, external_id=external_id)
+        if asset is None:
+            return AssetList([])
```
sanderland, line 520: it's good to set the client on this regardless I think

sanderland (approved):

PR #757, opened 2020-11-07
[CDF-8371] Allow resource metadata update using .metadata.set(None)

me-ydv-5:
Non breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.6.3] - 2020-11-06
+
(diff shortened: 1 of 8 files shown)


sanderland:
hmm, isn't there a special method/api syntax for set none?

me-ydv-5:
> hmm, isn't there a special method/api syntax for set none?

@sanderland yes there is one, all other fields use `.set(None)` but metadata is different as it is `AssetUpdate._ObjectAssetUpdate` while others are `AssetUpdate._PrimitiveAssetUpdate` and metadata can only be set as empty if we update it to `{}`. I didn't put much effort into this as we're anyway going to refactor Python SDK in near term.

sanderland (approved):

PR #759, opened 2020-11-09
Add fetch_resources to relationships read methods

andreavs:
This PR allows users to set a parameter `fetch_resources` on the relationships list and byIds endpoint, which will, on the api side, attempt to fetch the resources referenced in the relationship and add those to the response. In the sdk, it will convert the raw jsons to the right class instances based on the `source_type` and `target_type?` fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,15 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [2.6.4] - 2020-11-10
+## [2.6.5] - 2020-11-10
 
 ### Added
 - `fetch_resources` parameter to the relationships `list` and `retrieve_multiple` methods, which attempts to fetch the resource referenced in the relationship.
+
+## [2.6.4] - 2020-11-10
+
+### Fixed
+- Fixed a bug where 429 was not retried on all endpoints
 
 ## [2.6.3] - 2020-11-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.6.2] - 2020-11-05
+
+### Fixed
+- Asset retrieve subtree should return empty AssetList if asset does not exist.
+
 ## [2.6.1] - 2020-10-30
 
 ### Added
@@ -22,7 +27,7 @@ Changes are grouped as follows
 ## [2.6.0] - 2020-10-26
 
 ### Changed
-- Relationships list should take dataset internal and external id as different parameters
+- Relationships list should take dataset internal and external id as different parameters.
 
 ## [2.5.4] - 2020-10-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.6.3] - 2020-11-06
+
+### Fixed
+- Resource metadata should be able to set empty using `.metadata.set(None)` or `.metadata.set({})`.
+
 ## [2.6.2] - 2020-11-05
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Changes are grouped as follows
 ## [2.6.4] - 2020-11-10
 
 ### Added
-- added `fetch_resources` parameter to the relationships list and retrieve_multiple methods, which attempts to fetch the resource referenced in the relationship.
+- `fetch_resources` parameter to the relationships `list` and `retrieve_multiple` methods, which attempts to fetch the resource referenced in the relationship.
 
 ## [2.6.3] - 2020-11-06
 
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.6.4] - 2020-11-10
+
+### Added
+- added `fetch_resources` parameter to the relationships list and retrieve_multiple methods, which attempts to fetch the resource referenced in the relationship.
```
me-ydv-5, line 20:
```suggestion
- `fetch_resources` parameter to the relationships `list` and `retrieve_multiple` methods, which attempts to fetch the resource referenced in the relationship.
```

cognite/client/_api/relationships.py
```diff
@@ -186,6 +229,8 @@ def list(
             labels (LabelFilter): Return only the resource matching the specified label constraints.
             limit (int): Maximum number of relationships to return. Defaults to 100. Set to -1, float("inf") or None
                 to return all items.
+            fetch_resources (bool): if true, will try to return the full resources referenced by the relationship in the
+                source and target fields.
```
me-ydv-5, line 233: Maybe
```suggestion
                source_type and target_type fields.
```

cognite/client/_api/relationships.py
```diff
@@ -149,7 +165,33 @@ def retrieve_multiple(self, external_ids: List[str]) -> RelationshipList:
...
+        if relationship is not None:
+            if relationship.source is not None:
+                relationship.source = RelationshipsAPI._convert_resource(relationship.source, relationship.source_type)
+            if relationship.target is not None:
+                relationship.target = RelationshipsAPI._convert_resource(relationship.target, relationship.target_type)
+
+    @staticmethod
+    def _convert_resource(resource: Dict, resource_type: str):
+        resource = {utils._auxiliary.to_snake_case(key): value for key, value in resource.items()}
+        if resource_type == "timeSeries":
```
sanderland, line 185: better to use `TimeSeries._load` etc, so the SDK doesn't crash when fields are added before it updates. That also snake cases for you.
A dict from string to class could prevent some repetitive code here, but that's more personal style.

cognite/client/_api/relationships.py
andreavs: it is also based on the source_external_id and target_external_id

cognite/client/_api/relationships.py
```diff
@@ -129,13 +139,19 @@ def retrieve(self, external_id: str) -> Optional[Relationship]:
                 >>> c = CogniteClient()
                 >>> res = c.relationships.retrieve(external_id="1")
         """
-        return self._retrieve_multiple(external_ids=external_id, wrap_ids=True)
+        relationship = self._retrieve_multiple(
+            external_ids=external_id, wrap_ids=True, other_params={"fetchResources": fetch_resources}
+        )
+        RelationshipsAPI._convert_resources(relationship)
```
sanderland, line 145: Could `Relationship._load` handle this in one place?

cognite/client/_api/relationships.py
me-ydv-5, line 185: This if-else ladder could be changed to something cleaner but I'm not exactly sure what. maybe a dict?

tests/tests_integration/test_api/test_relationships.py
```diff
@@ -51,6 +52,30 @@ def new_label():
     CogniteClient().labels.delete(external_id=tp.external_id)
 
 
+@pytest.fixture
+def new_asset():
+    # create a asset to use in relationships
+    from cognite.client import CogniteClient
+
+    external_id = uuid.uuid4().hex[0:20]
+    tp = CogniteClient().assets.create(Asset(external_id=external_id, name="mandatory"))
```
sanderland, line 61: just make a global instead of CogniteClient()

tests/tests_integration/test_api/test_relationships.py
```diff
@@ -51,6 +52,30 @@ def new_label():
...
+    assert isinstance(tp, Asset)
+    yield tp
+    CogniteClient().assets.delete(external_id=tp.external_id)
```
sanderland, line 64: maybe clean up all these CogniteClient() calls

cognite/client/_api/relationships.py
me-ydv-5: Oh no, this seems to be correct, sorry!

sanderland (changes requested):
suggesting to use _load consistently

cognite/client/_api/relationships.py
andreavs: Figured this was implemented somewhere already, but could not find it - thanks for the tip :+1:

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.6.4] - 2020-11-10
```
me-ydv-5, line 17: I think we should bump it to minor version as it brings back `target` and `source` fields. Also, the change, even though non breaking, still has some meat in it.
```suggestion
## [2.7.0] - 2020-11-10
```

sanderland (approved):
LGTM

PR #761, opened 2020-11-11
Support for generic native token generation and lifecycle management - DRAFT

caelyn-cognite:
Using the `requests_oauthlib` to handle fetching access tokens. This also includes basic token lifecycle management in the form of re-fetching an access token when the token expires.

This has been tested against the existing azure credentials; but not against any other credentials providers. It can fetch an access token, use it to make requests, re-use it up until its expiry time, and then automatically re-fetch an access token upon the first request after expiry.

NOTES FOR REVIEWERS:
- Tests are mostly not written, waiting on feedback on this approach before writing them.
- Please suggest better environment variable names.
- Scope can be set; but has a default. Please let me know if this is a sensible default or not.

### DIFF
--- a/cognite/client/utils/_token_generator.py
+++ b/cognite/client/utils/_token_generator.py
@@ -1,0 +1,57 @@
+import datetime
+import os
+
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+
+from cognite.client.exceptions import CogniteAPIKeyError
+
+
+class TokenGenerator:
+    def __init__(self):
+        if not self.environment_vars_set():
+            raise CogniteAPIKeyError("Missing required environment vars for token creation")
+
+        self._idp_client_id = os.getenv("COGNITE_IDP_CLIENT_ID")
+        self._idp_client_secret = os.getenv("COGNITE_IDP_CLIENT_SECRET")
+        self._idp_token_endpoint = os.getenv("COGNITE_IDP_TOKEN_ENDPOINT")
+
+        cognite_base_url = os.getenv("COGNITE_BASE_URL", "https://api.cognitedata.com")
+        self._idp_scope = [os.getenv("COGNITE_IDP_SCOPE", f"{cognite_base_url}/.default")]
+        self._generate_access_token()
+
+    def return_access_token(self):
+        if self._expires_at < datetime.datetime.now().timestamp():
+            self._generate_access_token()
+
+        if self._access_token is None:
+            raise CogniteAPIKeyError(
+                "Could not generate access token from provided token related environment variables"
+            )
+
+        return self._access_token
+
+    def _generate_access_token(self):
+        client = BackendApplicationClient(client_id=self._idp_client_id)
+        oauth = OAuth2Session(client=client)
+        token_result = oauth.fetch_token(
+            token_url=self._idp_token_endpoint,
+            client_id=self._idp_client_id,
+            client_secret=self._idp_client_secret,
+            scope=self._idp_scope,
+        )
+
+        self._access_token = token_result["access_token"]
+        self._expires_at = token_result["expires_at"]
+
+    @staticmethod
+    def environment_vars_set():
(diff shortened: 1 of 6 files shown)


cognite/client/utils/_token_generator.py
```diff
@@ -0,0 +1,57 @@
...
+        )
+
+        self._access_token = token_result["access_token"]
+        self._expires_at = token_result["expires_at"]
+
+    @staticmethod
+    def environment_vars_set():
+        local_idp_client_id = os.getenv("COGNITE_IDP_CLIENT_ID")
+        local_idp_client_secret = os.getenv("COGNITE_IDP_CLIENT_SECRET")
+        local_idp_token_endpoint = os.getenv("COGNITE_IDP_TOKEN_ENDPOINT")
```
ngerstle-cognite, line 51:
```suggestion
        local_idp_client_id = os.getenv("CLIENT_ID")
        local_idp_client_secret = os.getenv("CLIENT_SECRET")
        local_idp_token_endpoint = os.getenv("TOKEN_URL")
```
To line up with suggestion in https://docs.google.com/document/d/1H2N5Cjtua5awJq24xsU4YLsOELNwoZhwMD8TUzyyRwU/edit# ?
Thoughts?

cognite/client/utils/_token_generator.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self):
+        if not self.environment_vars_set():
+            raise CogniteAPIKeyError("Missing required environment vars for token creation")
+
+        self._idp_client_id = os.getenv("COGNITE_IDP_CLIENT_ID")
+        self._idp_client_secret = os.getenv("COGNITE_IDP_CLIENT_SECRET")
+        self._idp_token_endpoint = os.getenv("COGNITE_IDP_TOKEN_ENDPOINT")
+
+        cognite_base_url = os.getenv("COGNITE_BASE_URL", "https://api.cognitedata.com")
+        self._idp_scope = [os.getenv("COGNITE_IDP_SCOPE", f"{cognite_base_url}/.default")]
```
ngerstle-cognite, line 20:
```suggestion
```suggestion
        self._idp_client_id = os.getenv("CLIENT_ID")
        self._idp_client_secret = os.getenv("CLIENT_SECRET")
        self._idp_token_endpoint = os.getenv("TOKEN_URL")

        cognite_base_url = os.getenv("COGNITE_BASE_URL", "https://api.cognitedata.com")
        self._idp_scope = [os.getenv("SCOPE", f"{cognite_base_url}/.default")]
```

The default scope works for Azure AD, but is likely to be different for other IdPs..

caelyn-cognite: If we think scope is likely to just not have a sensible default, I can make it a required parameter.

cognite/client/utils/_token_generator.py
caelyn-cognite: My main concern is that `CLIENT_ID`, `CLIENT_SECRET` and `TOKEN_URL` are very generic environment variables. We could end up with a name conflict. How do you feel about `COGNITE_CLIENT_ID`?

ngerstle-cognite: Are all variables prefixed with `COGNITE_`? 
1) that seems excessive. If not, then I don't see why these should be? was it CDF_API_KEY to distinguish from other api keys? Ideally, the client_id & secret is specific to the extractor, but not to CDF- they could (if support) be associated with access to other systems as well (to safely get different tokens for other APIs).
2) why `COGNITE` and not `CDF`?
3) these are generic values- especially the token_url- that will likely be the same for all extractors in the customer environment, as well as applications, etc.

caelyn-cognite: I can confirm that all environment variables references by this SDK are prefixed with `COGNITE_` - including `COGNITE_API_KEY`

I can only guess as to why `COGNITE` and not `CDF`; but my guess is that these environment variables predated the term `CDF`.

caelyn-cognite:
Closed in favour of https://github.com/cognitedata/cognite-sdk-python/pull/764

PR #764, opened 2020-11-23
Support for token creation + lifecycle management

caelyn-cognite:
Adding support for native token generation and lifecycle management to the SDK. The variables to generate tokens can be supplied either as environment variables or as optional arguments passed into the `CogniteClient`.

This change adds the `requests_oauthlib` as a required package to support a variety of different oauth providers. For oauth flows that do not fit into the structure of the `requests_oauthlib`, there is still the approach of passing in a token generating method.


cognite/client/utils/_client_config.py
```diff
@@ -76,12 +84,25 @@ def __init__(
...
+            # If no api_key or token is present; try setting up a token generator
+            token_generator = utils._token_generator.TokenGenerator(
+                self.token_url, self.token_client_id, self.token_client_secret, self.token_scopes
+            )
+
+            if token_generator.token_params_set():
+                self.token = lambda: token_generator.return_access_token()
+
+            if self.token is None:
+                raise CogniteAPIKeyError("No API key or token has been specified")
```
sanderland, line 105: somethingsomething or token generator parameters?

cognite/client/utils/_token_generator.py
```diff
@@ -0,0 +1,48 @@
...
+
+        if self.token_params_set():
+            self._generate_access_token()
+        else:
+            self._access_token = None
+            self._access_token_expires_at = None
+
+    def return_access_token(self):
+        if self._access_token is None or not self.token_params_set():
+            raise CogniteAPIKeyError("Could not generate access token from provided token related variables")
```
sanderland, line 25: different errors for failed vs not set?

cognite/client/utils/_token_generator.py
```diff
@@ -0,0 +1,48 @@
...
+
+class TokenGenerator:
+    def __init__(self, token_url: str, client_id: str, client_secret: str, scopes: List[str]):
+        self.token_url = token_url
+        self.client_id = client_id
+        self.client_secret = client_secret
+        self.scopes = scopes
+
+        if self.token_params_set():
+            self._generate_access_token()
```
sanderland, line 18: this will potentially cause a fairly obscure exception to go up to user code, right? Can we convert that to something more sensible?

cognite/client/utils/_token_generator.py
```diff
@@ -0,0 +1,48 @@
...
+        oauth = OAuth2Session(client=client)
+        token_result = oauth.fetch_token(
+            token_url=self.token_url, client_id=self.client_id, client_secret=self.client_secret, scope=self.scopes
+        )
+
+        self._access_token = token_result["access_token"]
+        self._access_token_expires_at = token_result["expires_at"]
+
+    def token_params_set(self):
+        return (
```
sanderland, line 43: are empty strings sensible for any of these? if not, should we just check without 'is None' ?

cognite/client/_cognite_client.py
```diff
@@ -35,6 +35,14 @@ class CogniteClient:
         timeout (int): Timeout on requests sent to the api. Defaults to 30 seconds.
         token (Union[str, Callable[[], str]]): A jwt or method which takes no arguments and returns a jwt to use for authentication.
             This will override any api-key set.
+        token_url (str): Optional url to use for token generation.
+            This will override the COGNITE_TOKEN_URL environment variable and only be used if both api-key and token are not set.
+        token_client_id (str): Optional client id to use for token generation.
+            This will override the COGNITE_CLIENT_ID environment variable and only be used if both api-key and token are not set.
+        token_client_secret (str): Optional client secret to use for token generation.
+            This will override the COGNITE_CLIENT_SECRET environment variable and only be used if both api-key and token are not set.
+        token_scopes (list): Optional list fo scopes to use for token generation.
```
sanderland, line 44: does token_scopes have a sensible default, possible from base_url?

sanderland (changes requested):
As someone who logs into a lot of different projects, I am a bit concerned about the 4 (env) variables needed in client creation. Do any have sensible defaults, or can we consider an additional env var which 'packs' them together ?

cognite/client/_cognite_client.py
caelyn-cognite: I initially went with `base_url` + `/.default`; but was told that it was likely to be different for each of the IDPs.

cognite/client/utils/_token_generator.py
caelyn-cognite: Usually this is either a deliberately vague 401 error or something very clear like 'Tenant ABC not found'. For those very useful errors informing someone that they have put the arguments in wrong, I would rather throw that error up to user code for the user to handle.

We could do something like catching all errors, and pushing the error message inside an error of our own; but I am not sure if that is more sensible. Do we do that elsewhere in this sdk?

cognite/client/utils/_token_generator.py
caelyn-cognite: I looked into this and `client_secret` can be empty for certain flows.

caelyn-cognite: Also `token_url` for the implicit flow

caelyn-cognite:
The good news is that if the same OpenId Connect configuration is being set on all of the projects (which you might choose to do on a set of different development projects), then the same 4 env vars can be used across multiple projects. But if not, then, apart from scope, they will likely have to differ.

cognite/client/utils/_token_generator.py
sanderland: HTTP errors and such get converted to CogniteAPIError
`raise CogniteAPIKeyError("somethingsomething token error") from actualtokenexception` ?

caelyn-cognite: Better than catching all errors: how about catching `Oauth2Error`s, e.g. all of the expected errors that might come from calling `fetch_token` on `OAuth2Session`?

cognite/client/_cognite_client.py
```diff
@@ -52,6 +60,10 @@ def __init__(
         headers: Dict[str, str] = None,
         timeout: int = None,
         token: Union[str, Callable[[], str], None] = None,
+        token_url: str = None,
+        token_client_id: str = None,
+        token_client_secret: str = None,
+        token_scopes: List[str] = None,
```
erlendvollset, line 66:
```suggestion
        token_url: Optional[str] = None,
        token_client_id: Optional[str] = None,
        token_client_secret: Optional[str] = None,
        token_scopes: Optional[List[str]] = None,
```

cognite/client/utils/_client_config.py
```diff
@@ -63,6 +67,10 @@ def __init__(
         headers: Dict[str, str] = None,
         timeout: int = None,
         token: Union[Callable[[], str], str] = None,
+        token_url: str = None,
+        token_client_id: str = None,
+        token_client_secret: str = None,
+        token_scopes: List[str] = None,
```
erlendvollset, line 73:
```suggestion
        token_url: Optional[str] = None,
        token_client_id: Optional[str] = None,
        token_client_secret: Optional[str] = None,
        token_scopes: Optional[List[str]] = None,
```

setup.py
```diff
@@ -17,7 +17,7 @@
     download_url="https://github.com/cognitedata/cognite-sdk-python/archive/{}.tar.gz".format(version),
     author="Erlend Vollset",
     author_email="erlend.vollset@cognite.com",
-    install_requires=["requests>=2.21.0,<3.0.0", "pandas"],
+    install_requires=["requests>=2.21.0,<3.0.0", "pandas", "requests_oauthlib"],
```
erlendvollset, line 20: We should lock this to `=1.*`

erlendvollset (approved):
Nice work!

erlendvollset (changes requested):
Meant to request changes

cognite/client/_cognite_client.py
caelyn-cognite: Any reason not do this for all of these Optional arguments while I am at it?

sanderland (dismissed):
lgtm!

erlendvollset (approved):

cognite/client/_cognite_client.py
```diff
@@ -36,6 +36,16 @@ class CogniteClient:
...
             This will override any api-key set.
+        token_url (str): Optional url to use for token generation.
+            This will override the COGNITE_TOKEN_URL environment variable and only be used if both api-key and token are not set.
+        token_client_id (str): Optional client id to use for token generation.
+            This will override the COGNITE_CLIENT_ID environment variable and only be used if both api-key and token are not set.
+        token_client_secret (str): Optional client secret to use for token generation.
+            This will override the COGNITE_CLIENT_SECRET environment variable and only be used if both api-key and token are not set.
+        token_scopes (list): Optional list of scopes to use for token generation.
+            This will override the COGNITE_TOKEN_SCOPES environment variable and only be used if both api-key and token are not set.
+        token_custom_args (Dict): Additional arguments to use for token generation.
```
sanderland, line 47: I would note that these are passed to the OAuth library fetch_token function or something

sanderland (approved):

PR #771, opened 2020-12-02
Chain requests for big relationships filters

andreavs:
This PR makes it so that if a user asks to filter on a set of e.g. source_external_ids which is larger than the what the list endpoint allows, the sdk will split it up into allowed chunks, in a similar manner to how we treat the limit on the list methods. The pattern was already in place in the experimental sdk (thanks @sanderland !) and is needed to port the power-sdk to v1 relationships

### DIFF
--- a/cognite/client/_api/relationships.py
+++ b/cognite/client/_api/relationships.py
@@ -264,9 +264,9 @@ class RelationshipsAPI(APIClient):
                 for si in range(0, max(1, len(source_external_ids)), self._LIST_SUBQUERY_LIMIT):
                     task_filter = copy.copy(filter)
                     if target_external_ids:  # keep null if it was
-                        task_filter["target_external_ids"] = target_external_ids[ti : ti + self._LIST_SUBQUERY_LIMIT]
+                        task_filter["targetExternalIds"] = target_external_ids[ti : ti + self._LIST_SUBQUERY_LIMIT]
                     if source_external_ids:  # keep null if it was
-                        task_filter["source_external_ids"] = source_external_ids[si : si + self._LIST_SUBQUERY_LIMIT]
+                        task_filter["sourceExternalIds"] = source_external_ids[si : si + self._LIST_SUBQUERY_LIMIT]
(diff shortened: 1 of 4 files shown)


cognite/client/_api/relationships.py
```diff
@@ -242,6 +249,39 @@ def list(
...
+                lambda filter: self._list(
+                    method="POST", limit=limit, filter=filter, other_params={"fetchResources": fetch_resources}
+                ),
+                tasks,
+                max_workers=self._config.max_workers,
+            )
+            tasks_summary.raise_compound_exception_if_failed_tasks()
+            res_list = tasks_summary.results
+            rels = RelationshipList([], cognite_client=self._cognite_client)
+            for res in res_list:
```
sanderland, line 282: I think there is a joined_results thingy that's more normal here

tests/tests_unit/test_api/test_relationships.py
```diff
@@ -258,3 +258,31 @@ def test_multi_source_target_list(self, mock_rel_response):
...
+        res = REL_API.list(source_external_ids=source_external_ids, limit=-1)
+        assert 3 == len(mock_rel_response.calls)
+        assert isinstance(res, RelationshipList)
+        assert 3 == len(res)
+        requested_sources = []
+        for call in mock_rel_response.calls:
+            json = jsgz_load(call.request.body)
+            assert "targetExternalIds" not in json["filter"]
+            requested_sources.extend([s for s in json["filter"]["sourceExternalIds"]])
+        assert set([s for s in source_external_ids]) == set(requested_sources)
```
sanderland, line 288: i like this test

sanderland (changes requested):
one comment
+changelog
+version

caelyn-cognite:
I doubt anyone is going to be attempting to filter on a large number of data sets or labels anytime soon; but we should keep this approach in mind if/when that becomes relevant.

sanderland (approved):

PR #773, opened 2020-12-08
[SBS-3069] Set geoLocation.type to "Feature" if missing

vegarsti:
It's not supposed to be allowed to do this, but it was. [This PR](https://github.com/cognitedata/ark/pull/1646) fixed it, so now the API should never allow it. But there may be some files where `geolocation.type` is not set.

### DIFF
--- a/cognite/client/__init__.py
+++ b/cognite/client/__init__.py
@@ -1,3 +1,3 @@
 from cognite.client._cognite_client import CogniteClient
 
-__version__ = "2.10.1"
+__version__ = "2.10.2"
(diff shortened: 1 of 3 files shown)


maaland (approved):
LGTM! Ideally we wouldn't do this and just clean up the files, but this will at least fix the errors showing up.

maaland (approved):
LGTM! Ideally we wouldn't do this and just clean up the files, but this will at least fix the errors showing up.

vegarsti:
The test `TestAssetsAPI.test_get_subtree` is failing only on 3.5 with `AttributeError: 'NoneType' object has no attribute 'id'`. Strange. I restarted it, but same thing happened.

vegarsti:
Now the same test is failing on 3.8. I'm wondering if there's a race condition here, since these two are executed in parallel.

maaland (approved):

me-ydv-5 (approved):

PR #775, opened 2020-12-08
fix race condition in backup time/error tracking

sanderland:


cognite/client/_http_client.py
```diff
@@ -93,29 +93,27 @@ class HTTPClient:
     def __init__(self, config: HTTPClientConfig, session: requests.Session = GLOBAL_REQUEST_SESSION):
         self.session = session
         self.config = config
-        self.retry_tracker = None
+        self.last_retry_tracker = None  # for tests
```
caelyn-cognite, line 96: I hate having something like this in production code that is only used by tests; but I want this issue fixed sooner rather than later.

caelyn-cognite (dismissed):

cognite/client/_http_client.py
erlendvollset: You use a factory with default to deal with this.
e.g.
```
class HttpClient:
    def __init__(self, retry_tracker_factory: Callable[[HttpClientConfig], _RetryTracker] = _RetryTracker, ...):
        self._retry_tracker_factory = retry_tracker_factory

    def request(self, method: str, url: str, **kwargs) -> requests.Response:
        retry_tracker = self._retry_factory(self.config)
        ....
```
That way you can get a retry tracker per request, and access the retry tracker instance by replacing the `retry_tracker_factory` with an actual function returning the retry tracker instance.

erlendvollset (changes requested):
Nice catch!

sanderland:
A test failed in sequences, fixed delete_ranges

sanderland:
workaround for the race condition in assets added, @erlendvollset could you approve?

erlendvollset (approved):

PR #779, opened 2020-12-21
[CDF-10343] Ignore subtree_size and tree_index on ThreeDAssetMapping create

me-ydv-5:


hhenriques:
This seems like a pretty hacky solution, `ThreeDAssetMapping` should be used as an input for create.

me-ydv-5:
> This seems like a pretty hacky solution, `ThreeDAssetMapping` should be used as an input for create.

Umm, isn't that happening here? We're still using `ThreeDAssetMapping`

hhenriques:
> > This seems like a pretty hacky solution, `ThreeDAssetMapping` should be used as an input for create.
> 
> Umm, isn't that happening here? We're still using `ThreeDAssetMapping`

That was a typo 🤦‍♂️ I mean shouldn't. `ThreeDAssetMapping` contains required fields that aren't required for creation.

me-ydv-5:
That would be a breaking change if I create a new data class called `ThreeDAssetMappingCreate` with only two input arguments. Not sure if the effort is worth the value.

sanderland (commented):
This seems convenient, but I am not sure why we are doing this specifically here and not in other similar cases (e.g. retrieved asset objects can't be passed to create)

me-ydv-5:
This was asked in #3d channel but yeah your argument makes sense. @sanderland 
Closing this one for good!

PR #783, opened 2021-02-02
Adding update for relationships

caelyn-cognite:
Adding update for relationships. Also updating example code in doc strings for relationships to not use the beta client and adding entity matching to the doc string test.

### DIFF
--- a/cognite/client/_api/relationships.py
+++ b/cognite/client/_api/relationships.py
@@ -3,7 +3,7 @@ from typing import *
 
 from cognite.client import utils
 from cognite.client._api_client import APIClient
-from cognite.client.data_classes import Relationship, RelationshipFilter, RelationshipList
+from cognite.client.data_classes import Relationship, RelationshipFilter, RelationshipList, RelationshipUpdate
 from cognite.client.data_classes.labels import LabelFilter
 
 
@@ -142,7 +142,7 @@ class RelationshipsAPI(APIClient):
 
             Get relationship by external id:
 
-                >>> from cognite.client.beta import CogniteClient
+                >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
                 >>> res = c.relationships.retrieve(external_id="1")
         """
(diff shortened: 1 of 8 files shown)


tests/tests_integration/test_api/test_relationships.py
```diff
@@ -156,6 +157,13 @@ def test_retrieve_unknown(self, new_relationship):
             API_REL.retrieve_multiple(external_ids=["this does not exist"])
         assert API_REL.retrieve(external_id="this does not exist") is None
 
+    def test_update_single(self, new_relationship):
+        new_rel, ext_id = new_relationship
+        updated_rel = API_REL.update(RelationshipUpdate(ext_id).target_type.set("timeseries").confidence.set(None))
+        assert isinstance(updated_rel, Relationship)
+        assert updated_rel.dump()["target_type"] == "timeSeries"
```
sanderland, line 164: why dump?

sanderland (changes requested):
one little change, other than that LGTM!

cognite/client/data_classes/relationships.py
```diff
@@ -156,7 +156,66 @@ def dump(self, camel_case: bool = False):
 
 
 class RelationshipUpdate(CogniteUpdate):
-    pass
+    """Update applied to a single relationship
+
+    Args:
+        id (int): A server-generated ID for the object.
```
me-ydv-5, line 162: maybe?
```suggestion
```

cognite/client/_api/relationships.py
```diff
@@ -311,6 +310,48 @@ def create(self, relationship: Union[Relationship, List[Relationship]]) -> Union
 
         return self._create_multiple(items=relationship)
 
+    def update(
+        self, item: Union[Relationship, RelationshipUpdate, List[Union[Relationship, RelationshipUpdate]]]
```
me-ydv-5, line 314: might as well add an example that takes in `Relationship` instead of `RelationshipUpdate`.

tests/tests_integration/test_api/test_relationships.py
caelyn-cognite: Copypasta from an above test. Changed here and above.

cognite/client/data_classes/relationships.py
caelyn-cognite: Good catch. I specifically added a custom init for RelationshipUpdate so that it could not be instantiated with `id`

sanderland (approved):

PR #786, opened 2021-02-17
More lenient label input

andreavs:


cognite/client/data_classes/assets.py
```diff
@@ -103,8 +100,7 @@ def _load(cls, resource: Union[Dict, str], cognite_client=None):
         if isinstance(resource, Dict):
             if instance.aggregates is not None:
                 instance.aggregates = AggregateResultItem(**instance.aggregates)
-        if instance.labels is not None:
-            instance.labels = [Label._load(label) for label in instance.labels]
+        instance.labels = Label._load_list(instance.labels)
```
sanderland, line 103: this seems a bit meh, either Label._load or LabelList._load?

sanderland: if LabelList is already a thing that is

andreavs: agree it does not really belong perfectly in this place - initially tried putting it in utils._auxillary, but it caused a nasty circular dependency issue that I wasn't able to sort out :cry:

sanderland (approved):
:1st_place_medal:

PR #787, opened 2021-02-18
CONTEXT-1425 flatten metadata in em fit etc

sanderland:
This supports matchfields like "metadata.key" which is a convention used in pipelines.
Also noticed that we never released relationships update

### DIFF
--- a/cognite/client/_api/entity_matching.py
+++ b/cognite/client/_api/entity_matching.py
@@ -141,7 +141,7 @@ class EntityMatchingAPI(APIClient):
         capabilities in the project, are able to access the data sent to this endpoint.
 
         Args:
-            sources: entities to match from, should have an 'id' field. Tolerant to passing more than is needed or used (e.g. json dump of time series list). Metadata fields are flattened to "metadata.key" entries for use in match_fields.
+            sources: entities to match from, should have an 'id' field. Tolerant to passing more than is needed or used (e.g. json dump of time series list). Metadata fields are automatically flattened to "metadata.key" entries, such that they can be used in match_fields.
             targets: entities to match to, should have an 'id' field.  Tolerant to passing more than is needed or used.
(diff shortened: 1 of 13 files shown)


cognite/client/_api/entity_matching.py
```diff
@@ -141,7 +141,7 @@ def fit(
         capabilities in the project, are able to access the data sent to this endpoint.
 
         Args:
-            sources: entities to match from, should have an 'id' field. Tolerant to passing more than is needed or used (e.g. json dump of time series list)
+            sources: entities to match from, should have an 'id' field. Tolerant to passing more than is needed or used (e.g. json dump of time series list). Metadata fields are flattened to "metadata.key" entries for use in match_fields.
```
andreavs, line 144: This kind of makes it sound like the user is supposed to do that, but the code looks like it does it for the user.

cognite/client/data_classes/contextualization.py
```diff
@@ -214,15 +214,19 @@ def refit(self, true_matches: List[Union[Dict, Tuple[Union[int, str], Union[int,
         )
         return self._load(response.json(), cognite_client=self._cognite_client)
 
+    @staticmethod
+    def _flatten_entity(entity: Dict) -> Dict:
+        if isinstance(entity, CogniteResource):
+            entity = entity.dump(camel_case=True)
+        if "metadata" in entity:
+            for k, v in entity["metadata"].items():
+                entity["metadata.{}".format(k)] = v
```
andreavs, line 223: should there be a `del entity["metadata"]` afterwards here?

sanderland: the dict comprehension takes care of this

andreavs (dismissed):
lgtm

cognite/client/_api/entity_matching.py
sanderland: rephrased

PR #793, opened 2021-03-04
CDF-11202 Move entity matching from beta to v1.

rebwib:


me-ydv-5 (dismissed):
LGTM!

sanderland (changes requested):
can we change the tests to import the normal and not the beta client?

sanderland (approved):

PR #795, opened 2021-03-10
Add List 3D Nodes query parameters

haakonflatval-cognite:


haakonflatval-cognite:
The reason for the changes to `cognite/client/_api_client.py`, is that the function `get_partition()` previously only supported getting partitions through the HTTP `POST` method, while `List 3D Nodes` expects `GET`.

cognite/client/_api_client.py
```diff
@@ -446,14 +448,22 @@ def get_partition(partition):
...
+                    body = {
+                        "filter": filter or {},
+                        "limit": self._LIST_LIMIT,
+                        "cursor": next_cursor,
+                        "partition": partition,
+                        **(other_params or {}),
+                    }
+                    res = self._post(url_path=resource_path + "/list", json=body, headers=headers)
+                elif method == "GET":
+                    params = filter.copy()
```
sanderland, line 461: params = {**filter.copy(),
              "limit": ...
              "cursor": ...
                **other_params}

sanderland (changes requested):
one refactoring suggestion, and would like a test

me-ydv-5 (commented):
Please also add changelog and update version.

haakonflatval-cognite:
The test is currently skipped, just as the rest of the revision tests. This is because we don't have a setup for creating a revision with a 3d file. Adding such a setup is out of scope of this PR.

cognite/client/_api/three_d.py
```diff
@@ -378,7 +378,14 @@ def update_thumbnail(self, model_id: int, revision_id: int, file_id: int) -> Non
         self._post(resource_path, json=body)
 
     def list_nodes(
-        self, model_id: int, revision_id: int, node_id: int = None, depth: int = None, limit: int = 25
+        self,
+        model_id: int,
+        revision_id: int,
+        node_id: int = None,
+        depth: int = None,
+        limit: int = 25,
```
sanderland, line 386: I think we tend to keep limit as the last parameter, as it makes the docs a little more readable

sanderland (changes requested):
tiny change i missed before

sanderland (approved):

PR #796, opened 2021-03-11
Improve error message when missing raw key

mathialo:
If a user tries to insert a raw row without specifying key, the current error message is somewhat cryptic since there is an internal KeyError during the error handling, so
```
client.raw.rows.insert("tb", "table", Row(key=None, columns={"Key": "Hello"}), ensure_parent=True)
```
fails with `KeyError: 'key'` and not a `CogniteAPIError`. With this change it would give 
```
CogniteAPIError: The request body was invalid. | code: 400 | X-Request-ID: None
The API Failed to process some items.
Successful (2xx): []
Unknown (5xx): []
Failed (4xx): [None]
```
instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ Changes are grouped as follows
 - `Security` in case of vulnerabilities.
 
 
+## [2.12.2] - 2021-03-11
+
+### Fixed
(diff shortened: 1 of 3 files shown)


erlendvollset (commented):
Thanks! Please update the changelog aswell, and bump the version if you want to release this.

erlendvollset (approved):
🥇

PR #801, opened 2021-03-24
Add templates, add support for date-based versioning

andreavs:
Two things going on in this pr: 
- We want to add templates to the main sdk, and rely on the new API versioning system: https://cognitedata.atlassian.net/wiki/spaces/CDF/pages/1395228698/Core+CDF+Releases+Versioning - by version locking the sdk we can ensure that the users application does not suddenly break. 
- I found some inconsistencies/bad choices in the current versioning implementation: 
  1) It only implemented the beta part of the versioning, not the date-based part 
  2) It mixed the old api versioning concept (0.5/v1/playground) with the new, representing both in the same variable, which is hard to understand 

This PR completes/fixes the version header implementation, and adds templates to the main sdk.

### DIFF
--- a/cognite/client/data_classes/templates.py
+++ b/cognite/client/data_classes/templates.py
@@ -1,0 +1,240 @@
+from typing import List, Optional
+
+from cognite.client.data_classes._base import *
+
+
+class TemplateGroup(CogniteResource):
+    """A template group is a high level concept encapsulating a schema and a set of template instances. It also has query capability support.
+
+    Template groups are versioned, so there can be multiple template groups with the same external ID.
+    The versioning is happening automatically whenever a template groups is changed.
+
+    GraphQL schema definition language is used as the language to describe the structure of the templates and data types.
+
+    Args:
+        external_id (str): The external ID provided by the client. Must be unique for the resource type.
+        description (str): The description of the template groups.
+        owners (List[str]): The list of owners for the template groups.
+    """
+
+    def __init__(
+        self, external_id: str = None, description: str = None, owners: Optional[List[str]] = None, cognite_client=None,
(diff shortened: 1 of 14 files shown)


cognite/client/utils/_client_config.py
```diff
@@ -8,6 +8,8 @@
 from cognite.client import utils
 from cognite.client.exceptions import CogniteAPIKeyError
 
+_DEFAULT_API_SUBVERSION = "V20210323"
```
andreavs, line 11: this is a bad place for this to live long term, but i wasn't sure where to put it

sanderland: does this mean using the sdk will lock in api bugs fixed after 2 days ago?

andreavs: we can change the default api version when creating new sdk releases. And generally bugfixes will not trigger a new API version. It's more: the sdk will lock into a version of the API so it does not suddenly break if we make a breaking change to the API

sanderland: maybe we should make it live next to the version variable so you notice it more?

cognite/client/data_classes/templates.py
```diff
@@ -0,0 +1,240 @@
+from typing import List, Optional
+
+from cognite.client.data_classes._base import *
```
erlendvollset, line 3: import * is evil, please be explicit

erlendvollset (approved):
LGTM just fix `import *`. And agree with sander that default api version would be better to have alongside sdk verison - but we can fix that later.

EDIT: NVM, I see we do import * many places so we can fix that in all places in another PR.

PR #812, opened 2021-04-13
Update template data models

andreavs:
This PR updated the template data models to include created_time, last_updated_time and data_set_id (where applicable)

### DIFF
--- a/cognite/client/data_classes/templates.py
+++ b/cognite/client/data_classes/templates.py
@@ -18,11 +18,19 @@ class TemplateGroup(CogniteResource):
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


cognite/client/data_classes/templates.py
```diff
@@ -154,11 +172,17 @@ def __init__(
         external_id: str = None,
         template_name: str = None,
         field_resolvers: Dict[str, FieldResolvers] = None,
+        data_set_id: Optional[int] = None,
+        created_time: int = None,
```
dmivankov, line 176: not sure how int = None is different from having Optional[int], but otherwise looks ok :+1:

dmivankov (approved):

PR #813, opened 2021-04-19
Add COGNITE_DISABLE_SSL to docs

erlendvollset:


docs/source/cognite.rst
```diff
@@ -185,6 +185,7 @@ You can set default configurations with these environment variables:
     # Global Configuration
     $ export COGNITE_DISABLE_PYPI_VERSION_CHECK = "1"
     $ export COGNITE_DISABLE_GZIP = "1"
+    $ export COGNITE_DISABLE_SSL = "1"
```
ponnapz, line 188: are these public docs? don't think default should be to disable ssl..

erlendvollset: default is not to disable. Example just shows how they can be set. I'll update those to reflect the defaults though 👍

ponnapz: yeah, I meant the example

ponnapz (approved):

PR #818, opened 2021-04-23
Fix template instance docs

andreavs:
There was an issue with the layout of the code that caused the docs for template instances to not build (not having the docstring at the top). This PR fixes that, and also updates the docs, as they turned out to be a bit out of date 

I do not think we need to trigger a release for this, so i skipped that part

### DIFF
--- a/cognite/client/data_classes/templates.py
+++ b/cognite/client/data_classes/templates.py
@@ -154,17 +154,15 @@ FieldResolvers = Union[ConstantResolver, RawResolver, SyntheticTimeSeriesResolve
 
 
 class TemplateInstance(CogniteResource):
-    field_resolver_mapper = {
-        "constant": ConstantResolver,
-        "syntheticTimeSeries": SyntheticTimeSeriesResolver,
-        "raw": RawResolver,
-    }
-    """A template instance that implements a template by specificing a resolver per field.
+    """A template instance that implements a template by specifying a resolver per field.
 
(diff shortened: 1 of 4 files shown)


sanderland (approved):

cognite/client/data_classes/templates.py
```diff
@@ -154,17 +154,15 @@ def __init__(
...
-    """A template instance that implements a template by specificing a resolver per field.
+    """A template instance that implements a template by specifying a resolver per field.
 
     Args:
         external_id (str): The id of the template instance.
         template_name (str): The template name to implement.
         field_resolvers (Dict[str, FieldResolvers]): A set of field resolvers where the dictionary key correspond to the field name.
+        data_set_id (int): The id of the dataset this instance belongs to.
+        created_time (int): The number of milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds.
+        last_updated_time (int): The number of milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds.
```
me-ydv-5, line 165: shouldnt it be something else, like num of milli since it was last updated? lol

andreavs: maybe, i just copied this from assets. Better to be consistent on how we refer to the same concept imo :)

PR #820, opened 2021-04-27
'list' is missing in the section of 'list'

tgib23:
raw() and raw.list() is similar in that both show the list of raws, but raw() does not support some parameters, which list() does.

L.445 needs list() because raw() does not support `columns` param.
L.452 works as it is, but we haven't explained about raw(), so it should be somewhere else, and we should focus on list() here.

### DIFF
--- a/cognite/client/_api/raw.py
+++ b/cognite/client/_api/raw.py
@@ -442,14 +442,14 @@ class RawRowsAPI(APIClient):
 
                 >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
-                >>> for row in c.raw.rows(db_name="db1", table_name="t1", columns=["col1","col2"]):
+                >>> for row in c.raw.rows.list(db_name="db1", table_name="t1", columns=["col1","col2"]):
                 ...     row # do something with the row
 
             Iterate over chunks of rows to reduce memory load::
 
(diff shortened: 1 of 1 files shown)


sanderland:
I think the correct thing would be to add columns to the generator. list does not have chunk_size etc.
The generator doesn't generate good docs, this is why it's explained in list.

tgib23:
@sanderland I'm not sure about the design of raw(), but should it have `minLastUpdatedTime` and `maxLastUpdatedTime` as well as parameters or only `columns` is fine?

>  list does not have chunk_size

true.

erlendvollset:
> @sanderland I'm not sure about the design of raw(), but should it have `minLastUpdatedTime` and `maxLastUpdatedTime` as well as parameters or only `columns` is fine?
> 
> > list does not have chunk_size
> 
> true.
Agree with Sander here. Docs should stay as is.

The iterator impl should accept any parameters which `.list()` accepts. Feel free to make a PR for that, or I can file a ticket and have it triaged.

erlendvollset:
I'm gonna go ahead and close this. Feel free to open a a new PR with my suggested change.

tgib23:
@erlendvollset I've been away from this because of support tasks. I'll create another PR for that.

PR #821, opened 2021-05-06
Add support for /token/inspect endpoint

erlendvollset:


cognite/client/data_classes/iam.py
```diff
@@ -126,3 +124,52 @@ def __init__(self, name: str = None, id: int = None, cognite_client=None):
...
+
+    @classmethod
+    def _load(cls, api_response):
+        return cls(
+            subject=api_response["subject"],
+            projects=[ProjectSpec._load(p) for p in api_response["projects"]],
+            capabilities=api_response["capabilities"],
+        )
+
+    def dump(self, camel_case: bool = False) -> Dict[str, Any]:
```
dmivankov, line 167: is this method used/tested? Also would camel_case=false change data?

erlendvollset: Added test now. Does not mutate data.

dmivankov (approved):

PR #828, opened 2021-06-04
templates: Fix missing owner after change in backend

Tapped:
The backend did before add the subject creating a template group to the owner list. However, now the user has to be manually added as owner.

### DIFF
--- a/tests/tests_integration/test_api/test_templates.py
+++ b/tests/tests_integration/test_api/test_templates.py
@@ -23,9 +23,11 @@ API_INSTANCES = API.templates.instances
 @pytest.fixture
 def new_template_group():
     external_id = uuid.uuid4().hex[0:20]
-
+    username = API.login.status().user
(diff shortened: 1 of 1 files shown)


erlendvollset (approved):

Tapped:
Hmm, I need to retrieve the subject name of the current token. This approach doesn't seem to work, do you have another suggestion @erlendvollset?

I think the status endpoint should work but didn't find the endpoint exposed, but I might have been looking in the wrong place.

erlendvollset:
> Hmm, I need to retrieve the subject name of the current token. This approach doesn't seem to work, do you have another suggestion @erlendvollset?
> 
> I think the status endpoint should work but didn't find the endpoint exposed, but I might have been looking in the wrong place.

Can you use the `user` field from `client.login.status().user`?

Tapped:
Ah perfect, I will use

PR #830, opened 2021-06-21
context job statuses

chingyulin:
in a recent update in context-api, we introduced three more statues (distributed, distributing and collecting), which breaks the current setup and test in experimental sdk, this pr introduces these statuses and refactor them as constants

### DIFF
--- a/cognite/client/data_classes/contextualization.py
+++ b/cognite/client/data_classes/contextualization.py
@@ -64,7 +64,7 @@ class ContextualizationJob(CogniteResource):
         """Data class for the result of a contextualization job."""
         self.job_id = job_id
         self.model_id = model_id
-        self.status = status
+        self.status = JobStatus(status)
         self.created_time = created_time
         self.start_time = start_time
         self.status_time = status_time
@@ -76,7 +76,7 @@ class ContextualizationJob(CogniteResource):
     def update_status(self) -> str:
         """Updates the model status and returns it"""
         data = self._cognite_client.__getattribute__(self._JOB_TYPE)._get(f"{self._status_path}{self.job_id}").json()
-        self.status = data["status"]
+        self.status = JobStatus(data["status"])
         self.status_time = data.get("statusTime")
         self.start_time = data.get("startTime")
         self.created_time = self.created_time or data.get("createdTime")
--- a/cognite/client/data_classes/contextualization.py
+++ b/cognite/client/data_classes/contextualization.py
@@ -11,6 +11,21 @@ from cognite.client.exceptions import ModelFailedException
 from cognite.client.utils._auxiliary import convert_true_match
 
 
+class JobStatus:
+    QUEUED = "Queued"
+    RUNNING = "Running"
+    COMPLETED = "Completed"
+    FAILED = "Failed"
+    DISTRIBUTING = "Distributing"
+    DISTRIBUTED = "Distributed"
+    COLLECTING = "Collecting"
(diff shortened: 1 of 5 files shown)


me-ydv-5 (commented):
Would suggest adding this change to integration and unit tests as well.

cognite/client/data_classes/contextualization.py
```diff
@@ -162,10 +178,10 @@ def wait_for_completion(self, timeout: int = None, interval: int = 1) -> None:
         start = time.time()
         while timeout is None or time.time() < start + timeout:
             self.update_status()
-            if self.status not in ["Queued", "Running"]:
+            if self.status not in [JobStatus.QUEUED, JobStatus.RUNNING]:
```
HMEiding, line 181: I see that above you're using `JobStatus.NOT_FINISHED_STATUSES`. Might not be appropriate here, but letting you know since I'm not sure if it's intended.

HMEiding (approved):
Added one question, but otherwise LGTM.

cognite/client/data_classes/contextualization.py
chingyulin: Currently, Entitymatching does not have other non-finished statuses than queued, running, so just to be explicit here.

cognite/client/data_classes/contextualization.py
```diff
@@ -11,6 +11,21 @@
...
 
+class JobStatus:
+    QUEUED = "Queued"
+    RUNNING = "Running"
+    COMPLETED = "Completed"
+    FAILED = "Failed"
+    DISTRIBUTING = "Distributing"
+    DISTRIBUTED = "Distributed"
+    COLLECTING = "Collecting"
+    NOT_FINISHED_STATUSES = [QUEUED, RUNNING, DISTRIBUTED, DISTRIBUTING, COLLECTING]
```
erlendvollset, line 22: Use an enum instead?
```suggestion
class JobStatus(Enum):
    QUEUED = "Queued"
    RUNNING = "Running"
    COMPLETED = "Completed"
    FAILED = "Failed"
    DISTRIBUTING = "Distributing"
    DISTRIBUTED = "Distributed"
    COLLECTING = "Collecting"
    
    def is_finished() -> bool:
        return self in [JobStatus.QUEUED, JobStatus.RUNNING, JobStatus.DISTRIBUTED, JobStatus.DISTRIBUTING, JobStatus.COLLECTING]
```

cognite/client/data_classes/contextualization.py
```diff
@@ -68,10 +84,10 @@ def wait_for_completion(self, timeout=None, interval=1):
         start = time.time()
         while timeout is None or time.time() < start + timeout:
             self.update_status()
-            if self.status not in ["Queued", "Running"]:
+            if self.status not in JobStatus.NOT_FINISHED_STATUSES:
```
erlendvollset, line 87: With above suggestion:
```suggestion
            if not self.status.is_finished():
```

cognite/client/data_classes/contextualization.py
```diff
@@ -11,6 +11,21 @@
...
+    RUNNING = "Running"
+    COMPLETED = "Completed"
+    FAILED = "Failed"
+    DISTRIBUTING = "Distributing"
+    DISTRIBUTED = "Distributed"
+    COLLECTING = "Collecting"
+    NOT_FINISHED_STATUSES = [QUEUED, RUNNING, DISTRIBUTED, DISTRIBUTING, COLLECTING]
+
+
+class ContextualizationJobType:
```
erlendvollset, line 25: Would use enum here aswell.

erlendvollset (approved):
lgtm

PR #833, opened 2021-07-02
Add filter3dNodes endpoint

haakonflatval-cognite:


haakonflatval-cognite:
TODO: Add tests

cognite/client/_api/three_d.py
```diff
@@ -426,6 +426,47 @@ def list_nodes(
...
+        Example:
+
+            Filter nodes from the hierarchy in the 3d model that have one of the values "AB76", "AB77" or "AB78" for property PDMS/Area AND that also have one of the values "PIPE", "BEND" or "PIPESUP" for the property PDMS/Type.
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.three_d.revisions.filter_nodes(model_id=1, revision_id=1, properties={ "PDMS": { "Area": ["AB76", "AB77", "AB78"], "Type": ["PIPE", "BEND", "PIPESUP"] } }, limit=10)
+        """
+        resource_path = utils._auxiliary.interpolate_and_url_encode(
+            self._RESOURCE_PATH + "/{}/nodes", model_id, revision_id
```
msusag, line 459: URL should end in /nodes/list ?

haakonflatval-cognite: I tried that - turns out the SDK adds the `/list` automatically when the method is `POST`. I don't even want to ask why :sweat_smile:

haakonflatval-cognite: Here, to be exact: https://github.com/cognitedata/cognite-sdk-python/blob/404425dc20a560bc52baccdc691991fa707b925d/cognite/client/_api_client.py#L331

msusag: 😮

msusag: strange behavior!

msusag:
I see there are some non-related timeseries tests that fail.

msusag (approved):
I am ok with this change.
But would be good to add a test case.

me-ydv-5 (commented):
Please also add changelog and bump version :)

cognite/client/_version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = "2.24.0"
-__api_subversion__ = "V20210423"
+__version__ = "2.25.0"
+__api_subversion__ = "V20210701"
```
haakonflatval-cognite, line 2: The way I understand it, this variable holds the date of the earliest API change that is needed to support the SDKs functions. Is this correct? The `filter3dNodes` endpoint was added to the V1 API at this date.

me-ydv-5: ah no, it's a different thing that is used for alpha/beta versioning of features. You just need to change `__version__`.

me-ydv-5:
```suggestion
__api_subversion__ = "V20210423"
```

haakonflatval-cognite: Turns out, in fact, that it is not correct :P

erlendvollset (commented):
Lgtm, but in general we don't want to have both .filter and .list methods in the clients, only .list to keep a simple interface. For other resources we use the 'POST <resource>/list' endpoints to power the .list methods.

Any particular reason for having both here?

haakonflatval-cognite:
> Lgtm, but in general we don't want to have both .filter and .list methods in the clients, only .list to keep a simple interface. For other resources we use the `POST <resource>/list` endpoints to power the .list methods.
> 
> Any particular reason for having both here?

I see. The list and filter endpoints in the 3D APIs have slightly different APIs and offer different functionality:
List nodes: https://docs.cognite.com/api/v1/#operation/get3DNodes
Filter nodes: https://docs.cognite.com/api/v1/#operation/filter3DNodes

In particular, filter3DNodes allows for a filter on multiple node property values simultaneously, which will be used internally in the upcoming 2.0 release of Reveal. Strictly speaking, we have no external pressure on us to add this to the Python SDK, I primarily add it here for completeness.

Provided it's so different to the listNodes endpoint, can it justify merging it either way?

erlendvollset:
> > Lgtm, but in general we don't want to have both .filter and .list methods in the clients, only .list to keep a simple interface. For other resources we use the `POST <resource>/list` endpoints to power the .list methods.
> 
> > 
> 
> > Any particular reason for having both here?
> 
> 
> 
> @erlendvollset I see. The list and filter endpoints in the 3D APIs have slightly different APIs and offer different functionality:
> 
> List nodes: https://docs.cognite.com/api/v1/#operation/get3DNodes
> 
> Filter nodes: https://docs.cognite.com/api/v1/#operation/filter3DNodes
> 
> 
> 
> In particular, filter3DNodes allows for a filter on multiple node property values simultaneously, which will be used internally in the upcoming 2.0 release of Reveal. Strictly speaking, we have no external pressure on us to add this to the Python SDK, I primarily add it here for completeness.
> 
> 
> 
> Provided it's so different to the listNodes endpoint, can it justify merging it either way?

I see. For other resources the functionality is the same, we have the two endpoints for a few reasons:
- GET endpoint enables caching the browser
- very large filters may trigger `414 URI too long` on GET endpoints, so POST is needed.
- maybe other reasons 😄

If the functionality actually differs, I agree that we should have both.

erlendvollset:
Should add the method to docs/source/cognite.rst to have it included in the documentation.

erlendvollset (approved):

PR #842, opened 2021-07-20
Make client pickleable

haakonvt:
## Description
When using CogniteClient with the client-secret auth flow, the object would not be pickle-able (e.g. when using multiprocessing) because of an anonymous function:
```py
AttributeError: Can't pickle local object 'ClientConfig.__init__.<locals>.<lambda>'
```

## Checklist:
- [ ] Tests added/updated? Do I need this, and if so, can I get a pointer or two here? 🙏 
- [ ] Documentation updated.

### DIFF
--- a/cognite/client/utils/_client_config.py
+++ b/cognite/client/utils/_client_config.py
@@ -116,7 +116,7 @@ class ClientConfig(_DefaultConfig):
             )
 
             if token_generator.token_params_set():
-                self.token = lambda: token_generator.return_access_token()
+                self.token = token_generator.return_access_token
 
             if self.token is None:
                 raise CogniteAPIKeyError("No API key or token or token generation arguments have been specified")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.27.0] - 2021-07-20
+
+### Fixed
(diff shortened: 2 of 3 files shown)


cognite/client/utils/_client_config.py
```diff
@@ -116,7 +116,7 @@ def __init__(
             )
 
             if token_generator.token_params_set():
-                self.token = lambda: token_generator.return_access_token()
+                self.token = token_generator.return_access_token
```
me-ydv-5, line 119: this won't work because the issue is not just about lambda. you [cannot pickle any nested function](https://stackoverflow.com/a/12022055/15604332), which is what we have in return_access_token.

me-ydv-5: it'd be better to make changes to multiprocessing code in such a way that:
1. it first gets the token
2. then spawns up new processes that use that token.
@eighty20results

haakonvt: I seem to be a able to pickle it after the change:
```py
$ python aaaaaaaa.py
hey hoo from 'if token_generator.token_params_set()' WITH LAMBDA
Trying to pickle...
  File "aaaaaaaa.py", line 25, in <module>
    pickle.dumps(client)
AttributeError: Can't pickle local object 'ClientConfig.__init__.<locals>.<lambda>'
$ 
$ python aaaaaaaa.py
hey hoo from 'if token_generator.token_params_set()' WITHOUT LAMBDA
Trying to pickle...
SUCCESS!!!
```

me-ydv-5: okay, let's try it then. hope it works fine.

me-ydv-5 (approved):

PR #847, opened 2021-08-16
Cache .json() return value on Response to avoid redecoding if called multiple times

erlendvollset:


cognite/client/_api/files.py
```diff
@@ -153,9 +153,8 @@ def create(self, file_metadata: FileMetadata, overwrite: bool = False) -> Tuple[
             url_path=self._RESOURCE_PATH, json=file_metadata.dump(camel_case=True), params={"overwrite": overwrite}
         )
         returned_file_metadata = res.json()
-        upload_url = returned_file_metadata.pop("uploadUrl")
+        upload_url = returned_file_metadata["uploadUrl"]
```
erlendvollset, line 156: Tests for these methods started failing due to caching .json() response. uploadUrl was removed from the cached dict, so couldn't use it to compare with expected object. No need for pop here since when loading this dict into FileMetadata object this field is ignored.

cognite/client/_http_client.py
```diff
@@ -109,6 +110,8 @@ def request(self, method: str, url: str, **kwargs) -> requests.Response:
                 last_status = res.status_code
                 retry_tracker.status += 1
                 if not retry_tracker.should_retry(status_code=last_status):
+                    # Cache .json() return value in order to avoid redecoding JSON if called multiple times
+                    res.json = functools.lru_cache(maxsize=1)(res.json)
```
me-ydv-5, line 114: why cache only one call though?

me-ydv-5: TIL we can use decorators like that too :)

me-ydv-5 (approved):

cognite/client/_http_client.py
erlendvollset: Note that we create one lru_cache per object here, so there will only ever be one return value to store in the cache.

PR #848, opened 2021-08-16
Retry JSONDecodeError on datapoints retrieve

erlendvollset:
We sometimes receive a truncated response from this endpoint. Root cause is still unidentified. See this thread for details: https://cognitedata.slack.com/archives/CG10VQPFX/p1628357875042900

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -1,9 +1,8 @@
 import copy
 import math
 import re as regexp
-import threading
-from collections import defaultdict
 from datetime import datetime
+from json import JSONDecodeError
 from typing import *
 
 import cognite.client.utils._time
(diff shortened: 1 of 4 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -1041,7 +1041,13 @@ def _get_datapoints(
             "ignoreUnknownIds": task.ignore_unknown_ids,
             "limit": min(window.limit, task.request_limit),
         }
-        res = self.client._post(self.client._RESOURCE_PATH + "/list", json=payload).json()["items"]
+
+        # Retry JSONDecodeError because we sometimes receive a truncated body from this endpoint.
+        # TODO: follow up with timelords and remove this retrying once the root cause has been found and fixed.
```
ponnapz, line 1046: maybe add a Jira task also to make sure this isn't forgotten

ponnapz (approved):

cognite/client/_api/datapoints.py
erlendvollset: Will do 👍

PR #850, opened 2021-08-18
Add template views

Tapped:
## Description

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.29.1] - 2021-08-18
+### Added
+- Added support for Template Views
+
 ## [2.29.0] - 2021-08-16
 ### Added
 - Raw rows are retrieved using parallel cursors when no limit is set.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,9 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [2.29.1] - 2021-08-18
+## [2.30.0] - 2021-08-18
 ### Added
-- Added support for Template Views
+- Support for Template Views
 
 ## [2.29.0] - 2021-08-16
 ### Added
--- a/cognite/client/_api/templates.py
+++ b/cognite/client/_api/templates.py
@@ -74,7 +74,8 @@ class TemplateGroupsAPI(APIClient):
             Union[TemplateGroup, TemplateGroupList]: Created template group(s)
 
         Examples:
-            create a new template group:
+            Create a new template group:
+
                 >>> from cognite.client import CogniteClient
                 >>> from cognite.client.data_classes import TemplateGroup
                 >>> c = CogniteClient()
@@ -97,7 +98,7 @@ class TemplateGroupsAPI(APIClient):
             Union[TemplateGroup, TemplateGroupList]: Created template group(s)
 
         Examples:
-            create a new template group:
+            Upsert a template group:
 
                 >>> from cognite.client import CogniteClient
                 >>> from cognite.client.data_classes import TemplateGroup
@@ -309,8 +310,7 @@ class TemplateInstancesAPI(APIClient):
                 >>> from cognite.client import CogniteClient
                 >>> from cognite.client.data_classes import TemplateInstance
                 >>> c = CogniteClient()
-                >>> template_instance_1 = TemplateInstance(
-                >>>                               external_id="norway",
+                >>> template_instance_1 = TemplateInstance(external_id="norway",
                 >>>                               template_name="Country",
                 >>>                               field_resolvers={
                 >>>                                   "name": ConstantResolver("Norway"),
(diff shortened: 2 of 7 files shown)


cognite/client/_api/templates.py
```diff
@@ -494,3 +495,174 @@ def delete(self, external_id: str, version: int, external_ids: List[str], ignore
...
+
+        Examples:
+            Create new views:
+
+                >>> from cognite.client import CogniteClient
+                >>> from cognite.client.data_classes import View
+                >>> c = CogniteClient()
+                >>> view = View(
+                >>>             external_id="view",
+                >>>             source=Source(
```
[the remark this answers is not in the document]
Tapped: Good point, seems like a lot of the examples in this file has invalid python, thanks to identation.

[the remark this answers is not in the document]
Tapped: Hmm, since we prefix `>>>` you can't really just copy examples from the docs into CLI and such. Is that with purpose?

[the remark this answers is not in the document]
Tapped: Anyways, I fixed the docstring syntax now.

Tapped:
@erlendvollset Do you think you will have time to review this? Thanks!

cognite/client/_version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = "2.29.0"
-__api_subversion__ = "V20210423"
+__version__ = "2.29.1"
+__api_subversion__ = "V20210818"
```
me-ydv-5, line 2:
```suggestion
__version__ = "2.30.0"
__api_subversion__ = "V20210423"
```

This looks like a rather big change than patch. Also are we following alpha/beta versioning on this? if not we can leave subversion be. :)

tests/tests_integration/test_api/test_templates.py
```diff
@@ -77,6 +79,33 @@ def new_template_instance(new_template_group_version):
     API_INSTANCES.delete(ext_id, new_version.version, instance.external_id)
 
 
+@pytest.fixture
+def new_view(new_template_group_version):
+    events = []
+    for i in range(0, 1001):
+        events.append(Event(external_id="test_evt_templates_" + str(i), type="test_templates", start_time=i * 1000))
```
me-ydv-5, line 86: this will have potential race conditions when testing simultaneously with diff clients, better to gen a UUID or something.

Tapped: It doesn't fail if multiple clients are sending at the same time (or if it already exists). So it kind of works like insert or ignore at the moment, since I'm not cleaning them up afterwards. However, I can change it to generate it with UUIDs and then delete them afterwards instead, if that's current practice here.

cognite/client/_version.py
Tapped: Right, I've become biased by the NPM way of doing things, where patch is used for any non breaking things, while minor is used for minor breaking changes, and major for big breaking changes. I'll fix it.

me-ydv-5 (commented):
looks good mostly, please see the comments. :)

tests/tests_integration/test_api/test_templates.py
me-ydv-5: yes, that would be nice. we do like that generally.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.29.1] - 2021-08-18
+### Added
+- Added support for Template Views
```
me-ydv-5, line 19:
```suggestion
- Support for Template Views
```
Nit

Tapped:
Thank you for the review, I've fixed what you have mentioned here.

me-ydv-5 (approved):
A couple of tests in templates are failing. you can ignore the timeseries tests. those are a bit flaky, will pass on rerun, we've asked timelords but there has been no resolution as such.

Tapped:
I see, I think it's because of eventual consistency. How are we normally dealing with it in this repo?

Tapped:
The old approach worked without issues in this regard, as it created the events only once, and kept them forever.

Tapped:
I guess the test-data-polutator is basically doing something similar to what I did in the test. Populating the project once.

Tapped:
I think I'll just go back to the old way of dealing with this, instead of adding a sleep that most likely will break because of eventual consistency latency.

tests/tests_integration/test_api/test_templates.py
```diff
@@ -83,23 +83,26 @@ def new_template_instance(new_template_group_version):
 def new_view(new_template_group_version):
     events = []
     for i in range(0, 1001):
-        events.append(Event(external_id="test_evt_templates_" + uuid.uuid4().hex[0:20], type="test_templates", start_time=i * 1000))
-    API.events.create(events)
+        events.append(Event(external_id="test_evt_templates_1_" + str(i), type="test_templates_1", start_time=i * 1000))
+    try:
+        API.events.create(events)
+    except:
+        # We only generate this data once for a given project, to prevent issues with eventual consistency etc.
```
me-ydv-5, line 90: think I missed this comment earlier 😅

me-ydv-5:
yeah so if doesn't create events more than once then I guess that's good as well

tests/tests_integration/test_api/test_templates.py
Tapped: Hehe, I didn't have the comment before :)

PR #855, opened 2021-09-27
Make _get_response_content_safe safer

tommythorsen:
This function fails on most binary responses, such as the ones returned
from the document-preview endpoints (PDF/PNG).

This commit falls back to returning the string `"<binary>"` if both json
decoding and utf-8 string decoding fails.


erlendvollset (approved):
Good catch!

PR #867, opened 2021-11-19
Change default JSON behaviour to strict compliance

haakonvt:
- Changes default JSON behaviour to be in strict compliance with the standard.
- Adds improved error message.

Reason:
https://cognitedata.slack.com/archives/CG10VQPFX/p1637321494071900 

This PR is a oneliner if we don't want to make the error message nicer:
```py
if json_payload:
    data = _json.dumps(json_payload, default=utils._auxiliary.json_dump_default, allow_nan=False)
```

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -136,7 +136,12 @@ class APIClient:
         headers.update(kwargs.get("headers") or {})
 
         if json_payload:
-            data = _json.dumps(json_payload, default=utils._auxiliary.json_dump_default)
+            try:
+                data = _json.dumps(json_payload, default=utils._auxiliary.json_dump_default, allow_nan=False)
+            except ValueError as e:
+                raise ValueError(f"{e}. Make sure your data does not contain NaN(s) or +/- Inf!").with_traceback(
+                    e.__traceback__
+                ) from None
             kwargs["data"] = data
             if method in ["PUT", "POST"] and not os.getenv("COGNITE_DISABLE_GZIP", False):
                 kwargs["data"] = gzip.compress(data.encode())
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -1,5 +1,4 @@
 import gzip
-import json
 import json as _json
 import logging
 import numbers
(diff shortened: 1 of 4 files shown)


erlendvollset (commented):
Please add a test for this in test_api_client.py

cognite/client/_api_client.py
```diff
@@ -136,7 +136,12 @@ def _do_request(self, method: str, url_path: str, **kwargs) -> requests.Response
         headers.update(kwargs.get("headers") or {})
 
         if json_payload:
-            data = _json.dumps(json_payload, default=utils._auxiliary.json_dump_default)
+            try:
+                data = _json.dumps(json_payload, default=utils._auxiliary.json_dump_default, allow_nan=False)
+            except ValueError as e:
```
erlendvollset, line 141: Just catching a bare ValueError here seems unsafe. Let's verify the error message is `Out of range float values are not JSON compliant`

cognite/client/_api_client.py
```diff
@@ -136,7 +136,12 @@ def _do_request(self, method: str, url_path: str, **kwargs) -> requests.Response
...
+                raise ValueError(f"{e}. Make sure your data does not contain NaN(s) or +/- Inf!").with_traceback(
+                    e.__traceback__
+                ) from None
```
erlendvollset, line 144: Any particular reason for not just doing this?
```suggestion
                raise ValueError(f"{e}. Make sure your data does not contain NaN(s) or +/- Inf!") from e
```

erlendvollset:
Let's also bump version and update changelog so we get a new release.

cognite/client/_api_client.py
haakonvt: Just to avoid the huge "double traceback" when all I want to do is to make the JSON error more understandable.

cognite/client/_api_client.py
haakonvt: Good call, adding guard 👍 It seems that the error message is slightly longer when `_json.make_encoder` is not available:
https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Modules/_json.c#L1352 
vs
https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/json/encoder.py#L241

haakonvt:
> Let's also bump version and update changelog so we get a new release.

Bumped to `2.36.0` expecting #871 to be merged first 👍 

> Please add a test for this in test_api_client.py

I was quite unsure where to put the tests in the file (and how to interact with the client), so please advise 😅

PR #868, opened 2021-11-23
Add support for transformations API

silvavelosa:
## Description
Migrated the code for transformations API from experimental python SDK.

## Checklist:
- [✔️] Tests added/updated.
- [✔️] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
- [✔️] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [✔️] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) per [semantic versioning](https://semver.org/).


cognite/client/data_classes/transformations_common.py
```diff
@@ -0,0 +1,145 @@
...
+            "audience": self.audience,
+            "cdf_project_name": self.cdf_project_name,
+        }
+        if camel_case:
+            return {utils._auxiliary.to_camel_case(key): value for key, value in ret.items()}
+        return ret
+
+
+class TransformationBlockedInfo:
+    def __init__(self, reason: str = None, created_time: Optional[int] = None, time: Optional[int] = None):
```
emelvarol-cognite, line 143: excluding `time`  here gives a decoding error but we also don't want to support it as it is a deprecated field. Could we workaround it?

cognite/client/_api/transformations.py
```diff
@@ -0,0 +1,324 @@
...
+                >>> c = CogniteClient()
+                >>> transformations = [
+                >>>     Transformation(
+                >>>         name="transformation1",
+                >>>         destination=TransformationDestination.assets()
+                >>>     ),
+                >>>     Transformation(
+                >>>         name="transformation2",
+                >>>         destination=TransformationDestination.raw("myDatabase", "myTable"),
+                >>>     ),
```
emelvarol-cognite, line 54: can we make this more detailed, one with all required fields? we can do this after the documentation is added to `rst` file.

emelvarol-cognite:
I will approve but not sure if you can make changes when I approve, so waiting for your comments.

emelvarol-cognite (approved):
LGTM Please see my comments 🙏

emelvarol-cognite:
@silvavelosa Code owner review is required, my approval doesn't work. :/

CHANGELOG.md
```diff
@@ -14,6 +14,16 @@ Changes are grouped as follows
...
 
+
+## [2.34.1] - 2021-11-22
+### Added
+- `transformations` api client, which allows the creation, deletion, update and retrieval of transformations.
+- `transformations.schedules` api client, which allows the schedule, unschedule and retrieval of recurring runs of a transformation.
+- `transformations.notifications` api client, which allows the creation, deletion and retrieval of transformation email notifications.
+- `transformations.schema` api client, which allows the retrieval of the expected schema of sql transformations based on the destination data type.
+- `transformations.jobs` api client, which retrieves the  status of transformation runs.
+- `transformations.run` method, which runs transformations.
```
me-ydv-5, line 25: I think this can be included in the first line along with "deletion, update, run and retrieval"...

cognite/client/_api/transformation_jobs.py
```diff
@@ -0,0 +1,103 @@
...
+from cognite.client._api_client import APIClient
+from cognite.client.data_classes import TransformationJob, TransformationJobList, TransformationJobMetricList
+
+
+class TransformationJobsAPI(APIClient):
+    _RESOURCE_PATH = "/transformations/jobs"
+    _LIST_CLASS = TransformationJobList
+
+    def list(self, limit: Optional[int] = 25, transformation_id: Optional[int] = None) -> TransformationJobList:
+        """`List all running transformation jobs. <https://docs.cognite.com/api/playground/#operation/transformationJobs>`_
```
me-ydv-5, line 13: please refrain from using playground, we really want to move away from playground.

me-ydv-5: this hyperlink doesn't point anywhere. guess you need to change it to https://github.com/cognitedata/service-contracts/pull/1501/files#diff-8df5cf6e84a989d49d837ea04347ac6085957197f0ef01467745af40090de15dR1028

but the link above points to a v1 endpoint, so we should point this there too.

Also, the operation Id is a bit lengthy and cumbersome to read. Is there a way to shorten it? have a look at other op ids for example

cognite/client/_api/transformation_jobs.py
```diff
@@ -0,0 +1,103 @@
...
+        if transformation_id is not None:
+            resource_path = utils._auxiliary.interpolate_and_url_encode(
+                "/transformations/{}/jobs", str(transformation_id)
+            )
+            return self._list(method="GET", limit=limit, resource_path=resource_path)
+
+        return self._list(method="GET", limit=limit)
+
+    def retrieve(self, id: int) -> Optional[TransformationJob]:
+        """`Retrieve a single transformation job by id. <https://docs.cognite.com/api/playground/#operation/getTransformationJob>`_
```
me-ydv-5, line 45: this should be pointing to this https://github.com/cognitedata/service-contracts/pull/1501/files#diff-8df5cf6e84a989d49d837ea04347ac6085957197f0ef01467745af40090de15dR1322

https://docs.cognite.com/api/v1/#operation/getApiV1ProjectsProjectTransformationsJobsJobid

again, see if we can shorten the op id.

cognite/client/_api/transformation_jobs.py
silvavelosa: my bad, the API is already in V1, but forgot to update the links in the docstrings

cognite/client/_api/transformation_notifications.py
```diff
@@ -0,0 +1,117 @@
...
+                >>> notifications = [TransformationNotification(transformation_id = 1, destination="my@email.com"), TransformationNotification(transformation_external_id="transformation2", destination="other@email.com"))]
+                >>> res = c.transformations.notifications.create(notifications)
+        """
+        utils._auxiliary.assert_type(notification, "notification", [TransformationNotification, list])
+        return self._create_multiple(notification)
+
+    def list(
+        self,
+        transformation_id: Optional[int] = None,
+        transformation_external_id: str = None,
```
me-ydv-5, line 51: any particular reason for using internal ids here (and other places too)? we had a discussion last week to expose external ids only. some other core services like labels, relationships etc only expose external id for example.

cognite/client/_api/transformation_notifications.py
```diff
@@ -0,0 +1,117 @@
...
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> notifications_list = c.transformations.notifications.list(transformation_id = 1)
+
+            List all notifications by transformation external id::
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> notifications_list = c.transformations.notifications.list(transformation_external_id = "myExternalId")
```
me-ydv-5, line 84: can remove the first example, looks redundant.

cognite/client/_api/transformation_notifications.py
```diff
@@ -0,0 +1,117 @@
...
+            destination=destination,
+        ).dump(camel_case=True)
+
+        return self._list(method="GET", limit=limit, filter=filter)
+
+    def delete(self, id: Union[int, List[int]] = None) -> None:
+        """`Deletes the specified notification subscriptions on the transformation. Does nothing when the subscriptions already don't exist <https://doc.cognitedata.com/api/playground/#operation/deleteTransformationNotifications>`_
+
+        Args:
+            id (Union[int, List[int]): Id or list of ids
```
me-ydv-5, line 104:
```suggestion
            id (Union[int, List[int]): Id or list of transformation notification ids
```

me-ydv-5: why do we allow deletes only with internal Ids?

cognite/client/_api/transformation_notifications.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Examples:
+
+            Delete schedules by id or external id::
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> c.transformations.notifications.delete(id=[1,2,3])
+        """
+        self._delete_multiple(ids=id, wrap_ids=True)
```
me-ydv-5, line 117: might as well add 
`utils._auxiliary.assert_type(ids, "id", [List], allow_none=True)`

cognite/client/_api/transformation_schema.py
```diff
@@ -0,0 +1,31 @@
+from cognite.client import utils
+from cognite.client._api_client import APIClient
+from cognite.client.data_classes import TransformationDestination, TransformationSchemaColumnList
+
+
+class TransformationSchemaAPI(APIClient):
```
me-ydv-5, line 6: just one method? how is schema created?

cognite/client/_api/transformations.py
```diff
@@ -0,0 +1,324 @@
...
+
+        Examples:
+
+            Get multiple transformations:
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.transformations.retrieve_multiple(ids=[1,2,3], external_ids=['transform-1','transform-2'])
+        """
+        return self._retrieve_multiple(
```
me-ydv-5, line 160: `utils._auxiliary.assert_type(...)` for both internal and ext ids here?

cognite/client/_api/transformations.py
```diff
@@ -0,0 +1,324 @@
...
+        transformation_id: int = None,
+        transformation_external_id: str = None,
+        wait: bool = True,
+        timeout: Optional[float] = None,
+    ) -> TransformationJob:
+        """`Run a transformation. <https://docs.cognite.com/api/playground/#operation/runTransformation>`_
+
+        Args:
+            transformation_id (int): internal Transformation id
+            transformation_external_id (str): external Transformation id
```
me-ydv-5, line 206:
```suggestion
            transformation_external_id (str): Transformation external id
```

cognite/client/_api/transformations.py
```diff
@@ -0,0 +1,324 @@
...
+        self,
+        transformation_id: int = None,
+        transformation_external_id: str = None,
+        wait: bool = True,
+        timeout: Optional[float] = None,
+    ) -> TransformationJob:
+        """`Run a transformation. <https://docs.cognite.com/api/playground/#operation/runTransformation>`_
+
+        Args:
+            transformation_id (int): internal Transformation id
```
me-ydv-5, line 205:
```suggestion
            transformation_id (int): Transformation internal id
```

cognite/client/_api/transformations.py
```diff
@@ -0,0 +1,324 @@
...
+                >>> res = c.transformations.run(id = 1, wait = False)
+        """
+        utils._auxiliary.assert_exactly_one_of_id_or_external_id(transformation_id, transformation_external_id)
+
+        if transformation_external_id:
+            transformation_id = self.retrieve(external_id=transformation_external_id).id
+
+        response = self._post(
+            url_path=utils._auxiliary.interpolate_and_url_encode(
+                self._RESOURCE_PATH + "/{}/run", str(transformation_id)
```
me-ydv-5, line 236: this is interesting, running a transformation only takes internal id, so we need to make two calls, one to get internal id if an ext id is present and other to actually run the transformation. Is it possible to run transformation via external Id itself?

cognite/client/_api/transformations.py
```diff
@@ -0,0 +1,324 @@
...
+            transformation_id (int): internal Transformation id
+            transformation_external_id (str): external Transformation id
+            timeout (Optional[float]): maximum time (s) to wait, default is None (infinite time). Once the timeout is reached, it returns with the current status.
+
+        Returns:
+            Completed (if finished) or running (if timeout reached) transformation job.
+
+        Examples:
+
+            Run transformation asyncronously by id:
```
me-ydv-5, line 261: nice example!

cognite/client/_version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = "2.34.0"
-__api_subversion__ = "V20210423"
+__version__ = "2.34.1"
+__api_subversion__ = "V20211122"
```
me-ydv-5, line 2: api subversion doesn't need to be changed as you aren't using that afaik? it's related to header `cdf-version` as a consequence of not using playground endpoint.

cognite/client/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,249 @@
...
+    _RESOURCE = TransformationJobMetric
+    _ASSERT_CLASSES = False
+
+
+class TransformationJob(CogniteResource):
+    """The transformation job resource allows following the status of execution of a transformation run.
+
+    Args:
+        id (int): A server-generated ID for the object.
+        uuid (UUID): A server-generated UUID for the object.
```
me-ydv-5, line 51: whats the diff bw these two?

cognite/client/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,249 @@
...
+        status (TransformationJobStatus): Status of the job.
+        transformation_id (int): Server-generated ID of the transformation.
```
me-ydv-5, line 53: and this?

cognite/client/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,249 @@
...
+        self.error = error
+        self.ignore_null_fields = ignore_null_fields
+        self.created_time = created_time
+        self.started_time = started_time
+        self.finished_time = finished_time
+        self.last_seen_time = last_seen_time
+        self._cognite_client = cognite_client
+
+    def update(self):
+        """`Get updated job status. <https://docs.cognite.com/api/playground/#operation/runTransformation>`_"""
```
me-ydv-5, line 107: don't need this here I think. it's just a data class.

me-ydv-5: same for other methods.

me-ydv-5: instead should explain the args.

cognite/client/_api/transformation_schema.py
silvavelosa: it is a read-only endpoint, it tells the user the schema that a sql select should have to be able to write a given resource type

cognite/client/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,249 @@
...
+        self.error = updated.error
+        self.started_time = updated.started_time
+        self.finished_time = updated.finished_time
+        self.last_seen_time = updated.last_seen_time
+
+    def metrics(self):
+        """`Get job metrics. <https://docs.cognite.com/api/playground/#operation/runTransformation>`_"""
+        return self._cognite_client.transformations.jobs.list_metrics(self.id)
+
+    def wait(self, polling_interval: float = 1, timeout: Optional[float] = None) -> "TransformationJob":
```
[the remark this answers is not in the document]
me-ydv-5: these should belong to api client maybe?

cognite/client/_api/transformation_notifications.py
emelvarol-cognite: Because transformations are also referred to with ids in the UI and the user is aware that transformations both have id and external IDs. I think it is good to allow them to work with both because of this.

cognite/client/_api/transformation_notifications.py
emelvarol-cognite: transformation notification does not have external id

cognite/client/_api/transformation_schema.py
emelvarol-cognite: schema is not a resource type, it is a utility endpoint that returns the schema for a resource type.

cognite/client/data_classes/transformation_jobs.py
emelvarol-cognite: This is the legacy id for transformation jobs, user is exposed to uuid in the UI already so we kept it to allow them to work with it. But we can mask it from SDK? @silvavelosa

cognite/client/data_classes/transformation_jobs.py
emelvarol-cognite: this is introduced to standardize the API, aiming to replace uuid.

cognite/client/data_classes/transformations.py
```diff
@@ -0,0 +1,261 @@
...
+        destination_oidc_credentials (Optional[OidcCredentials]): Configures the transformation to authenticate with the given oidc credentials on the destination.
+        created_time (int): The number of milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds.
+        last_updated_time (int): The number of milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds.
+        owner (str): Owner of the transformation: requester's identity.
+        owner_is_current_user (bool): Indicates if the transformation belongs to the current user.
+        has_source_api_key (bool): Indicates if the transformation is configured with a source api key.
+        has_destination_api_key (bool): Indicates if the transformation is configured with a destination api key.
+        has_source_oidc_credentials (bool): Indicates if the transformation is configured with a source oidc credentials set.
+        has_destination_oidc_credentials (bool): Indicates if the transformation is configured with a destination oidc credentials set.
```
me-ydv-5, line 31: do we need these all?

tests/tests_integration/test_api/test_iam.py
```diff
@@ -26,18 +27,22 @@ def service_account_id(cognite_client):
 
 
 class TestAPIKeysAPI:
+    @pytest.mark.skip("most API keys methods failing, probably due to deprecation")
```
me-ydv-5, line 30: @erlendvollset has the sdk test tenant been moved to oidc?

tests/tests_integration/test_api/test_time_series.py
```diff
@@ -38,7 +38,7 @@ def test_retrieve_multiple(self, cognite_client):
     def test_retrieve_multiple_ignore_unknown(self, cognite_client):
         res = cognite_client.time_series.list(limit=2)
         retrieved_assets = cognite_client.time_series.retrieve_multiple(
-            [t.id for t in res] + [0, 1, 2, 3], ignore_unknown_ids=True
+            [t.id for t in res] + [1, 2, 3], ignore_unknown_ids=True
```
me-ydv-5, line 41: why this change?

cognite/client/data_classes/transformations.py
emelvarol-cognite: I think this is huge because we use the same type for TransformationCreate and TransformationRead

cognite/client/data_classes/transformations_common.py
erlendvollset: pop it from the dict here? https://github.com/cognitedata/cognite-sdk-python/pull/868/files#diff-4f82056777939cc794f3fd64dcbedce16cbcdd19ae67a7a4d1377957f5b524a8R123

erlendvollset (commented):
Looks pretty darn good to me. Will wait to approve until we finalize https://github.com/cognitedata/service-contracts/pull/1501#pullrequestreview-817913905.

tests/tests_integration/test_api/test_iam.py
erlendvollset: yup.

cognite/client/data_classes/transformations.py
emelvarol-cognite: `owner`, `owner_is_current_user`, `isPublic`, `source_api_key`, `destination_api_key`, `has_source_api_key` and `has_destination_api_key` are valid when working with API keys.  Hope that we can remove them after going v1 when API keys become deprecated fully.

me-ydv-5 (commented):
Some comments:

* As you might've seen in `_api` directory, we have one class file per resource. Goes the same for services that have different "sub service". For eg: going through the api clients, raw.py, sequences.py, iam.py etc. have their classes in the same file, so would be better to move these ones to a single one too. but since this would result in quite big file, maybe move it to a subdirectory like that of model_hosting? 

* Move all references from playground -> v1

* Make sure the operation ids are synced with what we have in service contract. this op id is important as it helps tag the examples that you write here to appear on the right side on documentation portal under "python sdk". So it matches that op id to the id on docs portal. 

* not sure about usage of internal ids, I wouldn't want those but you guys can decide. @hakontro  

* Kindly try splitting the PR into smaller reviewable PRs if possible, 2700+ lines of code is hard to go through in one sitting and might even miss some things.

silvavelosa:
@me-ydv-5 Thanks a lot for the effort, we know it's a HUGE pr, and that makes thinks harder.

> Some comments:
> 
> * As you might've seen in `_api` directory, we have one class file per resource. Goes the same for services that have different "sub service". For eg: going through the api clients, raw.py, sequences.py, iam.py etc. have their classes in the same file, so would be better to move these ones to a single one too. but since this would result in quite big file, maybe move it to a subdirectory like that of model_hosting?

The directory approach seems good to me, will give it a try 👍 

> * Move all references from playground -> v1

Yes, you're right, we forgot to update the  docstring urls, but as soon as we get the v1 documentation published and have the definitive urls we'll update them

> * not sure about usage of internal ids, I wouldn't want those but you guys can decide. @hakontro

Since this is more of a design decision, and existing transformations don't have an external ID (AFAIK) I'm guessing it won't change for v1. But, if there's an engineering decision to go with external ids over internal ones, you're right, we should look into it and start pushing the clients in that direction.

> * Kindly try splitting the PR into smaller reviewable PRs if possible, 2700+ lines of code is hard to go through in one sitting and might even miss some things.

We understand, sorry for the mess.... this is a merge of 23 prs from the experimentals python sdk repository (that's also the reason for the playground stuff). 

Will try to answer all your questions, and apply your suggestions.

tests/tests_integration/test_api/test_transformation_jobs.py
```diff
@@ -0,0 +1,214 @@
...
+
+from cognite.client import CogniteClient
+from cognite.client.data_classes import (
+    OidcCredentials,
+    Transformation,
+    TransformationDestination,
+    TransformationJobStatus,
+)
+
+COGNITE_CLIENT = CogniteClient()
```
erlendvollset, line 16: Should be using the `cognite_client` fixture

tests/tests_integration/test_api/test_transformation_notifications.py
```diff
@@ -0,0 +1,109 @@
+import random
+import string
+
+import pytest
+
+from cognite.client import CogniteClient
+from cognite.client.data_classes import Transformation, TransformationDestination, TransformationNotification
+
+COGNITE_CLIENT = CogniteClient()
```
erlendvollset, line 9: use the cognite_client fixture

tests/tests_integration/test_api/test_transformation_schedules.py
```diff
@@ -0,0 +1,129 @@
...
+from cognite.client import CogniteClient
+from cognite.client.data_classes import (
+    OidcCredentials,
+    Transformation,
+    TransformationDestination,
+    TransformationSchedule,
+    TransformationScheduleUpdate,
+)
+
+COGNITE_CLIENT = CogniteClient()
```
erlendvollset, line 12: use the fixture

tests/tests_integration/test_api/test_transformation_schema.py
```diff
@@ -0,0 +1,26 @@
+from cognite.client import CogniteClient
+from cognite.client.data_classes import TransformationDestination
+
+COGNITE_CLIENT = CogniteClient()
```
erlendvollset, line 4: use the fixture

cognite/client/_api/transformation_notifications.py
hakontro: We could change that though, and run some migration to make sure all Transformations have an externalId

emelvarol-cognite: What would we put into the externalId then? name is not unique, this would require us to decide on behalf of user and we exposed the ID in UI already. people see it and use it when reporting problems. So I don't think we should do this personally.

tests/tests_integration/test_api/test_time_series.py
silvavelosa: there was a change in the datapoints api, and now it checks that all received ids are positive numbers. That change broke this test.

cognite/client/data_classes/transformation_jobs.py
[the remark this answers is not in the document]
silvavelosa: `metrics()` is a shortcut to enable more readable use of the sdk, same as `run()` in `transformation` data class:

This two blocks do the same, start running a transformation, and retrieve the metrics when it finishes:

```
my_metrics = my_transformation.run().metrics()
```

```
my_job = client.transformations.jobs.run(id = my_transformation.id)
my_metrics = client.transformations.jobs.list_metrics(id = my_job.id)
```

we decided to use a callable method instead of a property, to make clear to the user that an expensive operation (an API call) is performed on that "navigation".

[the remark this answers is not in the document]
silvavelosa: `wait()`, `wait_async()` and `update()` were done as instance methods because they don't map directly to an API operation, but to operations involving local and server operations, to achieve something the user might want to actually do with a `TransformationJob` instance, e.g. the user might want to `update()` the status of its instance, instead of creating a new object (this calls the api and maps the received fields to the calling object), or to `wait()` blocking the current thread until the transformation finishes running on the server.

cognite/client/data_classes/transformation_jobs.py
silvavelosa: right, will remove it

cognite/client/_version.py
silvavelosa: reverted 👍

cognite/client/_api/transformations.py
silvavelosa: good catch! that's already implemented on the api (it wasn't when we first wrote this), so the double call isn't necessary

cognite/client/_api/transformation_notifications.py
silvavelosa: type assertion is done inside `base._delete_multiple()`, id can be a single integer or a list of integers, and it will handle the cases inside the base logic.

cognite/client/_api/transformation_schedules.py
```diff
@@ -0,0 +1,204 @@
...
+        self._delete_multiple(
+            ids=id, external_ids=external_id, wrap_ids=True, extra_body_fields={"ignoreUnknownIds": ignore_unknown_ids}
+        )
+
+    def update(
+        self,
+        item: Union[
+            TransformationSchedule,
+            TransformationScheduleUpdate,
+            List[Union[TransformationSchedule, TransformationScheduleUpdate]],
```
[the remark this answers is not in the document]
silvavelosa: short answer, yes.

For most resource types, the sdk allows full update, which receives the object and updates all its properties except id (or external_id, if no id is present); or partial update, which only updates specific properties via an update object.

cognite/client/data_classes/transformation_jobs.py
silvavelosa: you're right, since they don't correspond to a specific api call, no link is necessary.

cognite/client/data_classes/transformations/__init__.py
```diff
@@ -0,0 +1,262 @@
...
+    def dump(self, camel_case: bool = False) -> Dict[str, Any]:
+        """Dump the instance into a json serializable Python data type.
+
+        Args:
+            camel_case (bool): Use camelCase for attribute names. Defaults to False.
+
+        Returns:
+            Dict[str, Any]: A dictionary representation of the instance.
+        """
+        ret = CogniteResource.dump(self, camel_case=camel_case)
```
erlendvollset, line 139:
```suggestion
        ret = super().dump(self, camel_case=camel_case)
```

cognite/client/_api/transformations/jobs.py
```diff
@@ -0,0 +1,103 @@
...
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> transformation_jobs_list = c.transformations.jobs.list()
+
+            List transformation jobs of a single transformation::
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> transformation_jobs_list = c.transformations.jobs.list(transformation_id = 1)
```
erlendvollset, line 34:
```suggestion
                >>> from cognite.client import CogniteClient
                >>> c = CogniteClient()
                >>> transformation_jobs_list = c.transformations.jobs.list()

            List transformation jobs of a single transformation::

                >>> from cognite.client import CogniteClient
                >>> c = CogniteClient()
                >>> transformation_jobs_list = c.transformations.jobs.list(transformation_id = 1)
```

cognite/client/_api/transformations/jobs.py
```diff
@@ -0,0 +1,103 @@
...
+            id (int): Job internal Id
+
+        Returns:
+            Optional[TransformationJob]: Requested transformation job or None if it does not exist.
+
+        Examples:
+
+            Get transformation job by id:
+
+                >>> from cognite.experimental import CogniteClient
```
erlendvollset, line 57:
```suggestion
                >>> from cognite.client import CogniteClient
```

cognite/client/_api/transformations/jobs.py
```diff
@@ -0,0 +1,103 @@
...
+            id (int): Job internal Id
+
+        Returns:
+            TransformationJobMetricList: List of updated metrics of the given job.
+
+        Examples:
+
+            Get metrics by transformation job id:
+
+                >>> from cognite.experimental import CogniteClient
```
erlendvollset, line 76:
```suggestion
                >>> from cognite.client import CogniteClient
```

cognite/client/_api/transformations/jobs.py
```diff
@@ -0,0 +1,103 @@
...
+            ignore_unknown_ids (bool): Ignore IDs that are not found rather than throw an exception.
+
+        Returns:
+            TransformationJobList: Requested transformation jobs.
+
+        Examples:
+
+            Get jobs by id::
+
+                >>> from cognite.experimental import CogniteClient
```
erlendvollset, line 98:
```suggestion
                >>> from cognite.client import CogniteClient
```

cognite/client/_api/transformations/notifications.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Returns:
+            Created notification(s)
+
+        Examples:
+
+            Create new notifications:
+
+                >>> from cognite.experimental import CogniteClient
+                >>> from cognite.experimental.data_classes import TransformationNotification
```
erlendvollset, line 40:
```suggestion
                >>> from cognite.client import CogniteClient
                >>> from cognite.client.data_classes import TransformationNotification
```

cognite/client/_api/transformations/notifications.py
```diff
@@ -0,0 +1,105 @@
...
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> notifications_list = c.transformations.notifications.list()
+
+            List all notifications by transformation id::
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> notifications_list = c.transformations.notifications.list(transformation_id = 1)
```
erlendvollset, line 78:
```suggestion
            List all notifications::

                >>> from cognite.client import CogniteClient
                >>> c = CogniteClient()
                >>> notifications_list = c.transformations.notifications.list()

            List all notifications by transformation id::

                >>> from cognite.client import CogniteClient
                >>> c = CogniteClient()
                >>> notifications_list = c.transformations.notifications.list(transformation_id = 1)
```

cognite/client/_api/transformations/notifications.py
```diff
@@ -0,0 +1,105 @@
...
+            id (Union[int, List[int]): Id or list of transformation notification ids
+
+        Returns:
+            None
+
+        Examples:
+
+            Delete schedules by id or external id::
+
+                >>> from cognite.experimental import CogniteClient
```
erlendvollset, line 101:
```suggestion
                >>> from cognite.client import CogniteClient
```

cognite/client/_api/transformations/schedules.py
```diff
@@ -0,0 +1,197 @@
...
+
+        Returns:
+            Created schedule(s)
+
+        Examples:
+
+            Create new schedules:
+
+                >>> from cognite.experimental import CogniteClient
+                >>> from cognite.experimental.data_classes import TransformationSchedule
```
erlendvollset, line 41:
```suggestion
                >>> from cognite.client import CogniteClient
                >>> from cognite.client.data_classes import TransformationSchedule
```

cognite/client/_api/transformations/schedules.py
```diff
@@ -0,0 +1,197 @@
...
+
+            Get transformation schedule by transformation id:
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.transformations.schedules.retrieve(id=1)
+
+            Get transformation schedule by transformation external id:
+
+                >>> from cognite.experimental import CogniteClient
```
erlendvollset, line 69:
```suggestion
                >>> from cognite.client import CogniteClient
                >>> c = CogniteClient()
                >>> res = c.transformations.schedules.retrieve(id=1)

            Get transformation schedule by transformation external id:

                >>> from cognite.client import CogniteClient
```

cognite/client/_api/transformations/schedules.py
```diff
@@ -0,0 +1,197 @@
...
+
+            Get transformation schedules by transformation ids:
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.transformations.schedules.retrieve_multiple(ids=[1, 2, 3])
+
+            Get transformation schedules by transformation external ids:
+
+                >>> from cognite.experimental import CogniteClient
```
erlendvollset, line 102:
```suggestion
                >>> from cognite.client import CogniteClient
                >>> c = CogniteClient()
                >>> res = c.transformations.schedules.retrieve_multiple(ids=[1, 2, 3])

            Get transformation schedules by transformation external ids:

                >>> from cognite.client import CogniteClient
```

cognite/client/_api/transformations/schedules.py
```diff
@@ -0,0 +1,197 @@
...
+            limit (int): Limits the number of results to be returned. To retrieve all results use limit=-1, default limit is 25.
+
+        Returns:
+            TransformationScheduleList: List of schedules
+
+        Example:
+
+            List schedules::
+
+                >>> from cognite.experimental import CogniteClient
```
erlendvollset, line 127:
```suggestion
                >>> from cognite.client import CogniteClient
```

cognite/client/_api/transformations/schedules.py
```diff
@@ -0,0 +1,197 @@
...
+            ignore_unknown_ids (bool): Ignore IDs and external IDs that are not found rather than throw an exception.
+
+        Returns:
+            None
+
+        Examples:
+
+            Delete schedules by id or external id::
+
+                >>> from cognite.experimental import CogniteClient
```
erlendvollset, line 155:
```suggestion
                >>> from cognite.client import CogniteClient
```

cognite/client/data_classes/transformations/common.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Returns:
+            TransformationDestination pointing to the target table
+        """
+        return RawTable(type="raw", database=database, table=table)
+
+
+class RawTable(TransformationDestination):
+    def __init__(self, type: str = None, database: str = None, table: str = None):
+        super().__init__(type=type)
```
erlendvollset, line 91: I think this is safer. Or can should type ever be anything else for an instance of RawTable?
```suggestion
        return RawTable(database=database, table=table)


class RawTable(TransformationDestination):
    def __init__(self, database: str = None, table: str = None):
        super().__init__(type="raw")
```

erlendvollset (commented):
Didn't mark all occurrences of `cognite.experimental` in docstrings. But those should be fixed.

Also some other small comments.

erlendvollset (approved):
LGTM!

silvavelosa:
Thanks a lot for the reviews!
🚀

PR #869, opened 2021-11-25
Update transformations documentation

silvavelosa:
## Description

## Checklist:
- [ ❌ ] Tests added/updated.
    No code changes, nothing to test.
- [ ✔️ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change. 
- [ ✔️ ] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ ✔️ ] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) per [semantic versioning](https://semver.org/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.38.1] - 2021-12-07
+### Fixed
+- Bug where loading `transformations.jobs` from JSON fails for raw destinations.
 
 ## [2.38.0] - 2021-11-30
 ### Added
--- a/cognite/client/_version.py
+++ b/cognite/client/_version.py
@@ -1,2 +1,2 @@
-__version__ = "2.38.0"
+__version__ = "2.38.1"
 __api_subversion__ = "V20210423"
--- a/cognite/client/data_classes/__init__.py
+++ b/cognite/client/data_classes/__init__.py
@@ -101,10 +101,12 @@ from cognite.client.data_classes.time_series import (
 )
 from cognite.client.data_classes.transformations import (
     OidcCredentials,
+    RawTable,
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.35.0] - 2021-11-5
```
me-ydv-5, line 17: this should be a patch, so 2.34.1 instead

emelvarol-cognite:
Is this failing because the API itself is not present in the master?

CHANGELOG.md
silvavelosa: we were thinking of a 2.34.1 "ghost version" with all the functionality, but no documentation ( #868 ) to be released now, and a 2.35.0 (this) version with the documentation to make it official to the users as part of the December release.

What do you think would be the best aproach?

silvavelosa:
> Is this failing because the API itself is not present in the master?

It's failing because there are a couple API keys tests broken currently.... those are skipped in the other PR, so once that one is merged, this one should start passing

CHANGELOG.md
me-ydv-5: Hmm, I think it would be better to add this documentation as part of the functionality so that the feature change is complete in itself. So maybe close this PR and merge it to the other PR and version that one to 2.35.0?

emelvarol-cognite: This was because we need to make use of the official SDK before release time without announcing it. So that is the reason why we have 2 PRs. Is this a wrong approach?

erlendvollset: Don't need to bump version at all to update docs.

PR #877, opened 2021-12-09
add the possibility to pass extra body fields to APIClient._create_multiple

vincent-cognite:
Using the same principle as _delete_multiple(...)


erlendvollset (approved):
LGTM

PR #878, opened 2021-12-10
Not log content if streaming

tuanng-cognite:
Calling _log_request will make client consumes all streaming data into internal memory which could cause memory issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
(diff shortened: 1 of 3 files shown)


pawelzwolenskicognite (approved):

pawelzwolenskicognite:
should we log something else then?

tuanng-cognite:
> should we log something else then?

yeap, I made the change to only exclude the response body

erlendvollset (approved):
Excellent catch! 🚀

tuanng-cognite:
Build is not triggered. So I clone this PR to this one https://github.com/cognitedata/cognite-sdk-python/pull/879
@erlendvollset Can you have a look again?

PR #880, opened 2021-12-17
schedules creation, update and delete batch size

silvavelosa:
## Description
transformation schedules creation, update and delete should be made in batches of max 5 elements because of API limitations

## Checklist:
- [〰️] Tests added/updated.
- [〰️ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change. 
- [✔️] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [✔️] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) per [semantic versioning](https://semver.org/).

### DIFF
--- a/tests/tests_integration/test_api/test_transformations/test_transformations.py
+++ b/tests/tests_integration/test_api/test_transformations/test_transformations.py
@@ -117,7 +117,7 @@ class TestTransformationsAPI:
         )
 
     def test_list(self, cognite_client, new_transformation):
-        retrieved_transformations = cognite_client.transformations.list()
+        retrieved_transformations = cognite_client.transformations.list(limit=None)
         assert new_transformation.id in [transformation.id for transformation in retrieved_transformations]
 
     def test_preview(self, cognite_client):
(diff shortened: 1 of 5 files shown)


tests/tests_integration/test_api/test_transformations/test_transformations.py
```diff
@@ -117,7 +117,7 @@ def test_update_partial(self, cognite_client, new_transformation):
         )
 
     def test_list(self, cognite_client, new_transformation):
-        retrieved_transformations = cognite_client.transformations.list()
+        retrieved_transformations = cognite_client.transformations.list(limit=None)
```
emelvarol-cognite, line 120: why is this?

silvavelosa: the default limit is 25. So, if there are already more than 25 transformations in the environment, this test will fail (list will return the first 25, and the created one won't be there)

emelvarol-cognite (approved):

erlendvollset (approved):

PR #895, opened 2022-02-16
SBS-4754: implement delete properties and search specs for GeospatialAPI.update_feature_types()

vincent-cognite:
not sure why I suddenly get reformatting...

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,5 @@
 * @cognitedata/starbase
 geospatial.py @cognitedata/geospatial
 test_geospatial.py @cognitedata/geospatial
+CHANGELOG.md @cognitedata/starbase @cognitedata/geospatial
+_version.py @cognitedata/starbase @cognitedata/geospatial
--- a/cognite/client/_version.py
+++ b/cognite/client/_version.py
@@ -1,2 +1,2 @@
-__version__ = "2.40.0"
-__api_subversion__ = "V20220125"
+__version__ = "2.41.0"
+__api_subversion__ = "V20220216"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.41.0] - 2022-02-16
+### Added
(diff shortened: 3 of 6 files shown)


.github/CODEOWNERS
```diff
@@ -1,3 +1,5 @@
 * @cognitedata/starbase
 geospatial.py @cognitedata/geospatial
 test_geospatial.py @cognitedata/geospatial
+CHANGELOG.md @cognitedata/starbase @cognitedata/geospatial
```
vincent-cognite, line 4: do you mind if we do those changes to CODEOWNERS @erlendvollset (and @starbase) ?
should offload you a bit

cognite/client/_version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = "2.40.1"
-__api_subversion__ = "V20220125"
+__version__ = "2.41.0"
+__api_subversion__ = "V20220216"
```
dmivankov, line 2: not sure about this subversion thing, do you need newer on for some of services?
should be safe for ark services anyway as I think we don't use date version headers yet

dmivankov: I mean it is set per sdk, not per resource :thinking: but anyway, if it explodes tests should be added to prevent it in the future :grin:

dmivankov (approved):

cognite/client/_version.py
vincent-cognite: We don't need it in geospatial. But it felt like the right thing to do since it is in the version file. As you say, if it blows up somewhere, then we'll add a test.

pawelzwolenskicognite (approved):

cognite/client/_version.py
me-ydv-5: @vincent-cognite the __version__ shouldn't have been regressed 😅

me-ydv-5: also this api_subversion is used for beta release afaik so it could've been left unchanged

me-ydv-5: @dmivankov ^^

vincent-cognite: ok, will. set it back.
Any place that explains how this works?

dmivankov: hmm, I have no idea either...

PR #896, opened 2022-02-17
revert the api subversion to previous value V20220125

vincent-cognite:
revert the api subversion according to https://github.com/cognitedata/cognite-sdk-python/pull/895#discussion_r808799166

### DIFF
--- a/cognite/client/_version.py
+++ b/cognite/client/_version.py
@@ -1,2 +1,2 @@
 __version__ = "2.41.0"
-__api_subversion__ = "V20220216"
+__api_subversion__ = "V20220125"


erlendvollset:
Some context would be good.

erlendvollset:
Thx. We should definitely be setting this per resource.

erlendvollset (approved):

vincent-cognite:
> Thx. We should definitely be setting this per resource.

yes, that would make it clearer whether this has to change or not

vincent-cognite:
who is using api subversion today?

PR #904, opened 2022-03-23
treat http 204 as ok status

tuanng-cognite:


cognite/client/_version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = "2.42.0"
-__api_subversion__ = "V20220125"
+__version__ = "2.43.0"
+__api_subversion__ = "V20220323"
```
vincent-cognite, line 2: this one should remain untouched (I don't know when it should be changed, it is undocumented)

tuanng-cognite: Nice to know. We definitely need to document this somewhere

vincent-cognite: So please remove the change on `__api_subversion__` and fix the conflicts to get approval @tuanng-cognite :)

vincent-cognite: Documentation has been requested earlier. Hasn't happened yet.

vincent-cognite (approved):
👍

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.46.0] - 2022-03-29
+### Changed
+- Treat HTTP 204 as ok status
```
dmivankov, line 19: which services/endpoints return 204? That'd be more changelog-worthy :slightly_smiling_face: could be a bugfix or preparation for new endpoints, don't have context

vincent-cognite: raster support in geospatial api (experimental) is returning 204 when the corresponding raster is empty (NULL in db). We considered using 404 (or 400) too for the same request but it felt more logical to go with 204 (No content)

dmivankov: then suggest putting something like "fixed retrieval of empty raster in experimental geospatial api"

dmivankov (approved):
Looks safe code-wise, but consider expanding changelog entry

CHANGELOG.md
tuanng-cognite: https://pr-1632.specs.preview.cogniteapp.com/v1.json.html#operation/getRaster
We are returning 204 if there is no data for this endpoint.

This PR is more aligned with https://github.com/cognitedata/service-contracts/blob/master/design_guide/API_design_guide.md#status-codes where 204 should be supported.

PR #917, opened 2022-04-28
Reduce request multiple timeseries

doctrino:
## Description
**Problem** For the use case filtering + aggregation of multiple timeseries, the current implementation is prohibitively slow to use in a user application. For example 100 datapoints from 4000 timeseries which is only 4 00 000 datapoints in total are fetched through 400 calls, but could be fetched with 40 calls given the limitation of CDF.

**Solution** Chunk time series such that up to 100 timeseries are included in each requests. 

**Implementation Details**
 * Extended `_DPTask `to support timeseries items.
 * Refactored out `_DPResult `from `_DPTask `to handle results for individual timeseries.
 * Did some minor code clean ups, e.g., replacing for loops with list comprehension etc.
 * If any local aggregates are passed, chunks size 1 is used.

**Tested in Motivating use case**
The use case which motivated this PR was fetching 3 months of daily data for 3000-4000 timeseries. 
Result: 
```
No optimization
CALL TO client.datapoints.retrieve_dataframe: 18.77 seconds
Chunking approach
CALL TO client.datapoints.retrieve_dataframe: 9.71 seconds
```

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Changes are grouped as follows
 
 ## [2.47.0] - 2022-05-02
 ### Changed
-- Performance gain for datapoints.retrieve by utilizing multiple timeseries in on requests.
+- Performance gain for `datapoints.retrieve` by grouping together time series in single requests against the underlying API.
 
 
 ## [2.46.1] - 2022-04-22
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.47.0] - 2022-05-02
+### Changed
+- Performance gain for datapoints.retrieve by utilizing multiple timeseries in on requests.
+
+
 ## [2.46.1] - 2022-04-22
 ### Changed
 - POST requests to the `sessions/revoke`-endpoint are now automatically retried
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -1052,24 +1052,29 @@ class DatapointsFetcher:
 
         if isinstance(ids, List):
             is_list = True
-            for item in ids:
-                items.append(DatapointsFetcher._process_single_ts_item(item, False))
+            items.extend(
+                DatapointsFetcher._process_single_ts_item(item, False)
+                for item in ids
+            )
         elif ids is not None:
             items.append(DatapointsFetcher._process_single_ts_item(ids, False))
 
         if isinstance(external_ids, List):
             is_list = True
-            for item in external_ids:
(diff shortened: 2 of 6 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -1019,31 +1075,31 @@ def _align_window_end(start: int, end: int, granularity: str):
     def _get_datapoints_with_paging(self, task, window):
         ndp_retrieved_total = 0
         while window.end > window.start and ndp_retrieved_total < window.limit:
-            ndp_retrieved, last_time = self._get_datapoints(task, window)
+            _ = self._get_datapoints(task, window)
```
erlendvollset, line 1078:
```suggestion
            self._get_datapoints(task, window)
```

cognite/client/_api/datapoints.py
```diff
@@ -843,45 +819,115 @@ def custom_sort_key(x):
...
     def fetch(self, query: DatapointsQuery) -> DatapointsList:
-        return self.fetch_multiple([query])[0]
+        chunk_size = self.client._TS_LIMIT
+        if query.granularity:
+            start = utils._time.timestamp_to_ms(query.start)
+            end = utils._time.timestamp_to_ms(query.end)
+            freq = utils._time.granularity_to_ms(query.granularity)
+            max_datapoint_count = int((end - start) / freq) + 1 + (2 if query.include_outside_points else 0)
+            chunk_size = min(chunk_size, self.client._DPS_LIMIT_AGG // max_datapoint_count)
+        chunk_size = max(chunk_size, 1)
```
erlendvollset, line 898: This looks like a good candidate for splitting into its own method with unit tests.
```suggestion
        chunk_size = self.client._TS_LIMIT
        if query.granularity:
            start = utils._time.timestamp_to_ms(query.start)
            end = utils._time.timestamp_to_ms(query.end)
            freq = utils._time.granularity_to_ms(query.granularity)
            max_datapoint_count = int((end - start) / freq) + 1 + (2 if query.include_outside_points else 0)
            chunk_size = min(chunk_size, self.client._DPS_LIMIT_AGG // max_datapoint_count)
        chunk_size = max(chunk_size, 1)
```

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.47.0] - 2022-05-02
+### Changed
+- Performance gain for datapoints.retrieve by utilizing multiple timeseries in on requests.
```
erlendvollset, line 19:
```suggestion
- Performance gain for `datapoints.retrieve` by grouping together time series in single requests against the underlying API.
```

erlendvollset (approved):

PR #922, opened 2022-05-18
Remove obsolete typing.io import

srittau:
## Description
`typing.io` has been deprecated since Python 3.8 and will be removed in Python 3.12. `TextIO` and `BinaryIO` are available from the main `typing` namespace, though, so the existing star import should be sufficient.

## Checklist:
- [ ] Tests added/updated.

Does this require a changelog entry?


erlendvollset:
Thanks! This has been addressed in a different PR.

PR #926, opened 2022-05-24
improve geospatial feature type update

tuanng-cognite:


cognite/client/data_classes/geospatial.py
```diff
@@ -39,14 +40,18 @@ class FeatureTypeList(CogniteResourceList):
     _ASSERT_CLASSES = False
 
 
-class PropertyAndSearchSpec:
```
vincent-cognite, line 42: this will break backward compatibility...
I think we still need to support the old way.

tuanng-cognite: I'll revert the change. The only way I see that can keep backward compatibility is to introduce  new method.

cognite/client/_api/geospatial.py
```diff
@@ -184,6 +184,15 @@ def patch_feature_types(self, patch: Union[FeatureTypePatch, List[FeatureTypePat
                 ...         property_updates=Updates(add={"altitude": {"type": "DOUBLE"}}),
                 ...         search_spec_updates=Updates(add={"altitude_idx": {"properties": ["altitude"]}})
                 ... ))
+
+            Add an additional index to an existing property
+
+                >>> from cognite.client.data_classes.geospatial import Updates
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.geospatial.patch_feature_types(update=FeatureTypePatch(external_id="wells",
```
vincent-cognite, line 193: maybe redo indentation to make the example more readable

cognite/client/_api/geospatial.py
```diff
@@ -160,6 +163,51 @@ def mapper(it):
...
+        """`Patch feature types`
+        <https://docs.cognite.com/api/v1/#operation/updateFeatureTypes>
+
+        Args:
+            patch (Union[FeatureTypePatch, List[FeatureTypePatch]]): the patch to apply
+
+        Returns:
+            FeatureTypeList: The patched feature types.
+
+        Examples:
```
vincent-cognite, line 176: While you are at it, maybe add an example of multi property index, I think that is not obvious that we can do it.

cognite/client/_api/geospatial.py
```diff
@@ -160,6 +163,51 @@ def mapper(it):
...
+
+            Add one property to a feature type and add index on that property:
+
+                >>> from cognite.client.data_classes.geospatial import Updates
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.geospatial.patch_feature_types(update=FeatureTypePatch(external_id="wells",
```
vincent-cognite, line 183: should be patch

cognite/client/data_classes/geospatial.py
```diff
@@ -65,6 +66,19 @@ def __init__(
...
+@dataclasses.dataclass
+class Updates:
+    add: Dict[str, Any] = None
+    remove: List[str] = None
+
+
+@dataclasses.dataclass
+class FeatureTypePatch:
+    external_id: str = None
+    property_updates: Updates = None
```
vincent-cognite, line 78: should we be consistent and use the parameter name "property_patches" ?

vincent-cognite:
should mark the "old" method deprecated (@deprecation ?)

vincent-cognite (approved):

dmivankov (approved):

PR #931, opened 2022-06-07
[SBS-5101] geospatial: fix feature properies serialization

tuanng-cognite:
Define a feature type
```
FeatureType(external_id="test_snake_case",
                properties={"firstName": {"type": "DOUBLE"},
                            "temperature_Hot": {"type": "DOUBLE"},
                            "levelH20": {"type": "DOUBLE", "optional": True, "description": "level H2O"},
                            }
                )
```

Corresponding feature has following attributes: external_id, first_name, temperature__hot, level_h2_o.

The set of attributes should be: external_id, firstName, temperature_Hot, levelH20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.50.1] - 2022-06-07
+### Fixed
+- Geospatial: keep feature properties as is
+
 ## [2.50.0] - 2022-05-27
 ### Changed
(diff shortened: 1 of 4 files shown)


vincent-cognite (commented):
Any idea of how much client code this will break?

tuanng-cognite:
Yeah, it will break all codes related to Feature objects, as this change will keep the Feature's properties as is instead of converting to snake_case.

vincent-cognite:
> Yeah, it will break all codes related to Feature objects, as this change will keep the Feature's properties as is instead of converting to snake_case.

I'm ok to go for that breaking change (this is a syntax change really) but should warn our users in #geospatial

vincent-cognite (approved):

PR #932, opened 2022-06-08
geospatial: fix issue for FeatureList.from_geopandas

tuanng-cognite:
FeatureList.from_geopandas throws KeyError when feature type contains optional property which geopandas dataframe does not have the corresponding column.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Changes are grouped as follows
 
 ## [2.50.2] - 2022-06-09
 ### Fixed
-- Geospatial: fix FeatureList.from_geopandas issue with optional properties 
+- Geospatial: fix FeatureList.from_geopandas issue with optional properties
 
 ## [2.50.1] - 2022-06-07
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.50.1] - 2022-06-08
+### Fixed
+- Geospatial: fix FeatureList.from_geopandas issue with optional properties 
+
 ## [2.50.0] - 2022-05-27
 ### Changed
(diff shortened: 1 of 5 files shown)


vincent-cognite (commented):
I'm surprised we don't have to modify ` _is_reserved_property(...)` since the original problem relates to `dataSetId`...

tuanng-cognite:
> I'm surprised we don't have to modify ` _is_reserved_property(...)` since the original problem relates to `dataSetId`...
As dataSetId is a nullable property. The current change should catch it. However, it also relates to snake case conversion. Current change does not convert dataSetId into data_set_id. I'll change the code.

vincent-cognite (approved):

tuanng-cognite:
A feature type from list_feature_types
```
{
        "external_id": "test",
        "created_time": "2022-05-25 11:19:56",
        "last_updated_time": "2022-05-25 11:19:56",
        "properties": {
            "name": {
                "type": "STRING",
                "size": 256
            },
            "externalId": {
                "type": "STRING",
                "size": 255
            },
            "createdTime": {
                "type": "LONG"
            },
            "lastUpdatedTime": {
                "type": "LONG"
            },
            "dataSetId": {
                "type": "LONG",
                "optional": true
            }
        },
        "search_spec": {
            "externalIdIdx": {
                "properties": [
                    "externalId"
                ]
            },
            "createdTimeIdx": {
                "properties": [
                    "createdTime"
                ]
            },
            "lastUpdatedTimeIdx": {
                "properties": [
                    "lastUpdatedTime"
                ]
            },
            "dataSetIdIdx": {
                "properties": [
                    "dataSetId"
                ]
            }
        }
    }
```
The attribute names are original values returned from Rest API. In Feature, we allow to set dataSetId as Feature(data_set_id=x)

So it is not consistent between feature type definition and the call on Feature. 
FeatureList.from_geopandas gets property_column_mapping as input. There is high chance that user will misuse it and specify either dataSetId or data_set_id as a mapping. Therefore, use `data_set_id_column` explicitly.

tuanng-cognite:
@vincent-cognite

PR #935, opened 2022-06-14
improve geospatial tests

tuanng-cognite:


tests/tests_integration/test_api/test_geospatial.py
```diff
@@ -389,8 +389,8 @@ def test_update_feature_types(self, cognite_client, test_feature_type):
             )
         )
         assert len(res) == 1
-        assert len(res[0].properties) == 7
-        assert len(res[0].search_spec) == 5
+        assert len(res[0].properties) == len(test_feature_type.properties) + 1
```
tuanng-cognite, line 392: number of properties can change if we enable/disable data set. So compare it with value before the change.

emmabuoen (approved):
Nice:D

vincent-cognite (approved):

PR #943, opened 2022-06-17
SBS-5094: add geopandas support for the geospatial streaming search

vincent-cognite:


tuanng-cognite (commented):
I think it is better to provide a utility method which converts list of features into geopandas dataframe.
It is more flexible and can be applied in many cases.

cognite/client/_api/geospatial.py
```diff
@@ -603,6 +603,57 @@ def stream_features(
...
+                ...
+                >>> # do something with the GeoDataFrame
+        """
+        iterator = self.stream_features(
+            feature_type_external_id=feature_type_external_id,
+            filter=filter,
+            properties=properties,
+            allow_crs_transformation=allow_crs_transformation,
+        )
+        features = [f for f in iterator]
```
tuanng-cognite, line 654: there is a chance that memory overflow happens or server timeout response is returned.

tuanng-cognite: I think it is easier to let user prepare the features from iterator from build geospandas DF from it.

PR #944, opened 2022-06-18
Fix reduce multiple timeseries paging bug

doctrino:
## Description
Fixed bug introduced in v2.47 and active until 2.51

### Bug description
Triggered: When there are more data points than can be returned in a single request, and there is a need for paging.
Problem: Reuse of the task object responsible for storing results. With the introduction of multiple timeseries for a task object, when one timeseries needed more request, all timeseries in given task object were paged. This causes duplicated timestamp in returned datapoints.

### Fix
* Added a method `pop_ts_item` to `_DPTask`. This removes and creates a new task for a given timeseries item which is used for paging.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [2.55.0] - 2022-06-20
+### Changed
+- Reintroduced performance gain for `datapoints.retrieve` by grouping together time series in single requests against the underlying API.
+
+### Fixed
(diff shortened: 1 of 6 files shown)


larshagencognite:
While I am very happy that this is being fixed, I am not familiar with the code case or with python, so I cannot give any detailed feedback.

haakonvt:
Should we close this one @doctrino as we have moved the work to #966 ?

erlendvollset:
Doing a cleanup of PRs and issues, so closing this in favour of #966. Feel free to reopen if you feel this was not correct.

PR #952, opened 2022-07-05
[DOCS-134] Add missing 'to_pandas' to example

Tapped:


erlendvollset:
Hmm. Doesn't retrieve_dataframe return a pandas DataFrame already?

PR #954, opened 2022-07-07
Nonces support in transformations

silvavelosa:
## Description
- Transformations creation and updating now supports the use of nonces to bind sessions (rather than sending client credentials to the service).
- If client credentials are set on transformations creation and update, a session will be created and the nonce used instead.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/transformations/__init__.py
+++ b/cognite/client/_api/transformations/__init__.py
@@ -9,11 +9,11 @@ from cognite.client._api_client import APIClient
 from cognite.client.data_classes import Transformation, TransformationJob, TransformationList
 from cognite.client.data_classes.shared import TimestampRange
 from cognite.client.data_classes.transformations import (
+    NonceCredentials,
     TransformationFilter,
     TransformationPreviewResult,
     TransformationUpdate,
 )
-from cognite.client.data_classes.transformations.common import NonceCredentials
 
 
 class TransformationsAPI(APIClient):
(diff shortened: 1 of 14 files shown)


cognite/client/_api/transformations/__init__.py
```diff
@@ -57,6 +57,16 @@ def create(
...
         utils._auxiliary.assert_type(transformation, "transformation", [Transformation, list])
+
+        if isinstance(transformation, TransformationList):
+            sessions = {}
+            for t in transformation:
+                t._cognite_client = self._cognite_client
+                t._process_credentials(sessions)
+        elif isinstance(transformation, Transformation):
+            transformation._cognite_client = self._cognite_client
+            transformation._process_credentials({})
```
emelvarol-cognite, line 68: I couldn't understand why we procecess empty dictionary when the item is not a list. :'( could you explain?

cognite/client/data_classes/transformations/__init__.py
```diff
@@ -112,8 +116,45 @@ def __init__(
...
+            key = (
+                f"{oidc_credentials.client_id}:{oidc_credentials.client_secret.__hash__()}"
+                if oidc_credentials
+                else "DEFAULT"
+            )
+
+            ret = sessions_dict.get(key)
+            if not ret:
+                if oidc_credentials and oidc_credentials.client_id and oidc_credentials.client_secret:
+                    credentials = ClientCredentials(oidc_credentials.client_id, oidc_credentials.client_secret)
```
emelvarol-cognite, line 137: don't we have other parameters for client credentials, such as project, scopes etc..?

emelvarol-cognite: ok this is not transformations credentials type 👍

emelvarol-cognite: where do we use client credentials when we fail to generate the nonce?

cognite/client/_api/transformations/__init__.py
silvavelosa: the credentials that are processed are the ones inside the transformation, and the dictionary acts as a cache, to create only one session for repeated credentials.... maybe it would be clearer to just add a `_process_credentials()` method to TransformationList and change all this for a single line 😅 

```
transformation._process_credentials()
```

cognite/client/data_classes/transformations/__init__.py
silvavelosa: on line 150:
if the generation succeeds, OIDC client credentials are set to `None`

if the generation fails, OIDC client credentials are left intact and sent to the api

emelvarol-cognite: I seee thanks

emelvarol-cognite (approved):
awesome

PR #957, opened 2022-07-13
fix documentation issues introduced in 3.0.0

vincent-cognite:


pyproject.toml
```diff
@@ -1,9 +1,11 @@
 [tool.poetry]
 name = "cognite-sdk"
 
-version = "3.0.0"
+version = "3.0.1"
 
 description = "Cognite Python SDK"
+readme = "README.md"
```
vincent-cognite, line 7: @erlendvollset 
assuming this will fix the missing doc in pypy.org ?

erlendvollset: maybe? don't know if pypi will respect that. Normally it's the `long_description` field in setup.py which populates the doc on pypi.org, but when building the distribution on this branch the generated setup.py file still sets that field to `None`. The pyproject.toml file is included in the distribution too though, so perhaps pypi checks there too.

erlendvollset (approved):

PR #963, opened 2022-07-15
Cache version response from pypi in version checker

erlendvollset:
We only want to do this once per session, not per client instance.


rhuanbarreto (approved):
Great!

me-ydv-5 (approved):

PR #964, opened 2022-07-15
Increase parallelism in CI

erlendvollset:
split workflows into 3 jobs: lint, test-core, and test-all-and-build. These are run in parallel.

Also increase parallelism on tests from 4 to 8.

Now the entire workflow runs in ~3 minutes. (compared to 15 minutes a week ago).


me-ydv-5 (approved):
nice

PR #966, opened 2022-07-15
Refactor: Datapoints fetch logic

haakonvt:
#### New optional dependency: `numpy`
Adds `numpy` as an optional dependency to the SDK in order to use it with time series (datapoints), reading data from the API straight into familiar arrays and making `to_pandas` calls 100X++ times faster (faster the bigger). 
I would argue that numpy is "almost standard library" given how extremely prevalent it is in the Python ecosystem. It has none of the same install/compile issues as for example `fiona`, and installs without problems on all platforms with pre-built binaries.

#### New resource types `DatapointsArray` and `DatapointsArrayList`
Given the slow and dynamic nature of Python, storing millions of datapoints in basic containers is extremely inefficient; both in terms of performance and memory use. By using `numpy` arrays instead, we get the best of both worlds; amazing performance and Python's ease of use. The new _resource type_ and _resource list type_ are mere clones of the existing `Datapoints` and `DatapointsList`, except we have said goodbye to built-in `list`.

#### Fail-fast
The old SDK method awaits and collects all errors before raising (through the use of a "fire-and-forget" thread pool). This is great for e.g. updates/inserts to make sure you are aware of all partial changes. However, for reading datapoints, it should just fail fast (which it now does).

#### Using `limit=0`
The API endpoint allows specifying `limit=0` when fetching datapoints (returning no data except some time series info). The old SDK had faulty logic and interpreted this as "unlimited" - quite the opposite. This has been updated and now adheres to the API specs.

#### Optional `start` and `end`
The API endpoint specifies both these as optional with the assumed meaning when missing: `start=0` and `end="now"`. The old SDK enforced passing both of these. This has been updated and now adheres to the API specs.

#### Individual customization of queries
The API endpoint specifies top-level parameters which serve as defaults for the individual time series queries passed in `items`. The old SDK only allowed a very limited variant of this, namely `aggregates`. In the new update, all allowed parameters can be passed either as top-level or as individual settings (...and now adheres to the API specs).

#### Aggregates queries with limit
Due to how count aggregates are used and tallied up to parallelize queries, in most cases you'd only get the first few batches of datapoints. The reason is that the code exists the logic of creating "windows to fetch" once the count of the number of raw datapoints (floored to the max possible given by granularity and count aggregate granularity) hits the given limit by the user. This count is only meaningful for raw datapoints and is just a very bad estimate of an upper bound for aggregate queries. Fixing this in the current implementation is not feasible, as we cannot interact with the ThreadPoolExecutor before it finishes. The only way to make it work with the current code is to fetch pretending `limit=None` then truncate the result to the actual limit. This is of course bad for a lot of reasons.

#### Removing aggregate names from columns in `pd.DataFrame`
The current implementation, `Datapoints._strip_aggregate_name`, removes everything after the first pipe character (`|`) it finds, rendering it useless for all tags with such a character as part of their names.

#### Aggregate query time rounding bug
Due to the SDK rounding differently than the API, you could supply start and end (`start < end`) and still be given an error that `start is not before end`.

#### Datapoints.to_pandas dtype
When all aggregate datapoints were missing, an object array of `None` was returned instead of int/float array dtype.

#### Various bugfixes related to `start` ts minimum going from 1970 to 1900
- TimeSeries.first()
- TimeSeries.count()

#### Evaluation of `protobuf` performance
Garbage performance compared to built-in JSON with an additional dependency. If installed without binary, _complete and utter garbage_ performance.
#### Duplicates in query
Now allowed, making requests on the form "grab data from 2012 to 2014 and 2017 to 2021" possible. These were always allowed by the API, but raised an error in the SDK. Could get around it (if needing no more than two periods) by passing one with ID and one with external ID...
#### Fetching string datapoints
In the old SDK, this was done serially (one chunk after the other), both for limited- and unlimited queries. In the new logic, unlimited queries are split when needed. This also applies to limited queries, but they will be prioritized by order so as to avoid fetching more datapoints than necessary (we even try to cancel queued futures when a task is done).


CHANGELOG.md
```diff
@@ -14,6 +14,15 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [4.0.0] - xx-yy-zz
+### Added/Changed/Fixed
+- Increased speed of varied datapoints fetch queries, especially when fetching many time series. Now matches API endpoint more closely:
```
haakonvt, line 19: Better make it happen now
# 🤣

run_dps_refactor.py
```diff
@@ -0,0 +1,71 @@
+import math
+
+from local_cog_client import setup_local_cog_client
+
+# Specify the aggregates to return. Use default if null.
+# If the default is a set of aggregates,
+# specify an empty string to get raw data.
```
haakonvt, line 7: The old official API documentation

cognite/client/data_classes/datapoints.py
```diff
@@ -364,7 +386,7 @@ def plot(self, *args: Any, **kwargs: Any) -> None:
     int, List[int], Dict[str, Union[int, List[str]]], List[Dict[str, Union[int, List[str]]]]
 ]
 DatapointsExternalIdMaybeAggregate = Union[
-    str, List[str], Dict[str, Union[int, List[str]]], List[Dict[str, Union[int, List[str]]]]
+    str, List[str], Dict[str, Union[str, List[str]]], List[Dict[str, Union[str, List[str]]]]
```
haakonvt, line 389: Ask @  erlendvollset after 🍹 🌴  if this is copy/paste error

doctrino (commented):
I read through the code just to familiarize myself with it, and I hope you can forgive me for adding a few comments here and there.

haakonvt:
> I read through the code just to familiarize myself with it, and I hope you can forgive me for adding a few comments here and there.

Comments are always good! ...but I think it is a bit too early to start polishing the code when most of it might still change 😅

cognite/client/data_classes/datapoints.py
```diff
@@ -80,17 +128,93 @@ def to_pandas(self, camel_case: bool = True) -> "pandas.DataFrame":  # type: ign
         return df
 
 
+class DatapointsArray(CogniteResource):
```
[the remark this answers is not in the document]
haakonvt: I suggest when - and if this is done - all Cognite resource types be upgraded simultaneously.

cognite/client/data_classes/datapoints.py
```diff
@@ -318,6 +443,26 @@ def _repr_html_(self) -> str:
...
+    def to_pandas(self, column_names: str = "external_id", include_aggregate_name: bool = True) -> "pandas.DataFrame":
+        pd = cast(Any, utils._auxiliary.local_import("pandas"))
+        dfs = [dps_arr.to_pandas(column_names=column_names) for dps_arr in self.data]
+        if dfs:
+            df = pd.concat(dfs, axis="columns")
+            if not include_aggregate_name:  # do not pass in to_pandas above, so we check for duplicate columns
+                df = Datapoints._strip_aggregate_names(df)
+            return df
+
+        return pd.DataFrame()
```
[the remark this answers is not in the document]
haakonvt: Could you elaborate a bit more on this one? Some user has asked for the datapoints from N time series, but only want e.g. M (M < N) of them converted to a pandas dataframe, and thus passed the identifiers (id or xids) of the ones they want?

haakonvt:
Wait what, do some tests still work!? 😂 
image

haakonvt:
Will be split into several PRs - closing this!

PR #960, opened 2022-07-13
Add future warnings for datetime conversion functions

haakonvt:
## Description
Adds a `FutureWarning` when passing naive `datetime`s to `cognite.client.utils._time.datetime_to_ms` about a change in interpretation of these - and a suggestion recommendation to use aware datetimes instead.

I am unsure which warning to use though:

1. `UserWarning`
2. `DeprecationWarning`
3. `FutureWarning`
4. `PendingDeprecationWarning`

https://cognitedata.atlassian.net/browse/CDF-11744 

## Checklist:
- [ ] Tests added/updated.
- [ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change. 
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring. 
- [ ] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ ] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) per [semantic versioning](https://semver.org/).

PR #972, opened 2022-07-21
Remove all internal use of `utils._time.ms_to_datetime`

haakonvt:
## Description
In #960 a warning was added to two `datetime`-related functions both using a non-standard interpretation of what "naive" means in Python. 

In #965 these were changed to `FutureWarning` since `DeprecationWarning` is ignored by default. This change was not given a new version number and as such, has not been "released to the wild" yet... which is a good thing - because I missed two uses of one of these functions internally 😄

### DIFF
--- a/cognite/client/utils/_time.py
+++ b/cognite/client/utils/_time.py
@@ -126,7 +126,7 @@ def _convert_time_attributes_in_dict(item: Dict) -> Dict:
     for k, v in item.items():
         if k in TIME_ATTRIBUTES:
             try:
-                v = ms_to_datetime(v).strftime("%Y-%m-%d %H:%M:%S")
+                v = str(datetime.utcfromtimestamp(v / 1000))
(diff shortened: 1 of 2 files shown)


cognite/client/utils/_time.py
```diff
@@ -126,7 +126,7 @@ def _convert_time_attributes_in_dict(item: Dict) -> Dict:
     for k, v in item.items():
         if k in TIME_ATTRIBUTES:
             try:
-                v = ms_to_datetime(v).strftime("%Y-%m-%d %H:%M:%S")
+                v = str(datetime.utcfromtimestamp(v / 1000))
```
haakonvt, line 129: Comment: This now uses the better `.isoformat(" ")`, meaning that when `milliseconds` is not 0, they no longer get ignored (and when 0, we get exact same result as previously)

erlendvollset (approved):
LGTM

PR #975, opened 2022-07-21
Add thread lock when fetching new token and 10s time buffer

haakonvt:
Test to see if this solves long-running queries failing with 401

### DIFF
--- a/cognite/client/utils/_token_generator.py
+++ b/cognite/client/utils/_token_generator.py
@@ -1,10 +1,13 @@
 import datetime
+import threading
 from typing import Dict, List, Optional
 
 from oauthlib.oauth2 import BackendApplicationClient, OAuth2Error
 from requests_oauthlib import OAuth2Session
 
 from cognite.client.exceptions import CogniteAPIKeyError
+
+LOCK = threading.Lock()
 
 
 class TokenGenerator:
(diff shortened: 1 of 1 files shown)


cognite/client/utils/_token_generator.py
```diff
@@ -29,15 +32,16 @@ def __init__(
             self._access_token_expires_at = None
 
     def return_access_token(self) -> str:
-        if not self.token_params_set():
-            raise CogniteAPIKeyError("Could not generate access token - missing token generation arguments")
-        elif self._access_token is None:
-            raise CogniteAPIKeyError("Could not generate access token from provided token generation arguments")
+        with LOCK:
```
erlendvollset, line 35: Great catch! But we instantiate one TokenGenerator per CogniteClient, so I believe this lock should be scoped to the instance.

cognite/client/utils/_token_generator.py
```diff
@@ -29,15 +32,16 @@ def __init__(
...
+            if not self.token_params_set():
+                raise CogniteAPIKeyError("Could not generate access token - missing token generation arguments")
+            elif self._access_token is None:
+                raise CogniteAPIKeyError("Could not generate access token from provided token generation arguments")
 
-        if self._access_token_expires_at < datetime.datetime.now().timestamp():
-            self._generate_access_token()
+            if self._access_token_expires_at < datetime.datetime.now().timestamp() - 10:
```
erlendvollset, line 41: I'm probably missing something. Why is this buffer necessary?

haakonvt: In my mind (at least 😅 ) there is a race condition here; since the thread can pass this check a few milliseconds before expiry, then not get de-scheduled by the os (is that a word?) and not run before after expiry. Maybe these tokens already have this kind of buffer "built in" (I am absolutely not familiar with "auth" in general)

cognite/client/utils/_token_generator.py
```diff
@@ -1,11 +1,14 @@
 import datetime
+import threading
 from typing import Dict, List, Optional
 
 from oauthlib.oauth2 import BackendApplicationClient, OAuth2Error
 from requests_oauthlib import OAuth2Session
 
 from cognite.client.exceptions import CogniteAPIKeyError
 
+LOCK = threading.Lock()
```
doctrino, line 10: Shouldn't this be a reenter lock such that the same thread can acquire it form itself?
```suggestion
LOCK = threading.RLock()
```

haakonvt: Hmm, not sure understand how/why that would be needed? 🤔  Afaik, each thread is told to make one request, and while building the headers it makes one call to this non-recursive method to get a valid token.

erlendvollset:
I added these changes in the auth rework in #999 so I'm closing this one.

PR #984, opened 2022-07-26
Limit number of values per request insert rows

simenl:
## Description
The current implementation does not explicitly limit the number of values per insert request of sequence rows, only the number of rows. As a row can contain up to 400 values, a single request can contain up to 4 million values. 
With this change the number of values in an insert request can be at most 100k. The same limit that is documented in https://docs.cognite.com/api/v1/#tag/Sequences/operation/postSequenceData

## Checklist:
- [ ] Tests added/updated.


erlendvollset (approved):
LGTM

Issue #953, opened 2022-07-05
Allow other iterables than Lists to be passed into external_id(s) function variables


henrtk:
It is tedious to have to cast types like tuples (even tuples!?) or pd.Series to lists every time when passing them into API function calls. I see no reason why this should have to be the case for any data structure that eventually is to be iterated over. 
Definitely affected function variables are aggregates and external_id(s). 

A relatively easy solution to this would be to have the function cast the non-list class/type to a list inside the function instead. This moves workload from the user over to the SDK (which I'd say would be a primary reason to use this SDK in the first place).

A far more work intensive solution would be to not rely on list functionality when handling the function input, but after a quick check of the actual implementation the gain in flexibility seems negligible compared to the work needed in rewriting the existing functionality, seeing that Lists completely permeates the SDK.

This may perhaps a pet peave, but it feels unpythonic in a way that is hard to explain in other ways than "This SDK cannot do that on its own even though I'm sure it probably could".

erlendvollset:
Agree. We should accept `typing.Sequence` rather than `typing.List`.

PR #992, opened 2022-08-10
Accept Sequence instead of List where appropriate

erlendvollset:
Addresses #953


me-ydv-5 (approved):
LGTM! let's also change the rest of occurances with sequences as discussed. 😄

PR #993, opened 2022-08-11
Generate random external id for annotation integration test file to a void simultaneous tests conflicting with each other

olacognite:
## Description
Changed annotation integration test file external id to avoid tests conflicting with each other when run at the same time. 

## Checklist:
- [x] Tests added/updated.


erlendvollset (approved):
Nice!

PR #998, opened 2022-08-12
Introduce ClientConfig and remove support for envvar config

erlendvollset:
Client configuration no longer respects any environment variables. There are other libraries better 
suited for loading configuration from the environment (such as builtin `os` or `pydantic`). There have also
been several reports of ennvar name clash issues in tools built on top the SDK. We therefore 
consider this something that should be handled by the application consuming the SDK. All configuration of 
`cognite.client.CogniteClient` now happens using a `cognite.client.ClientConfig` object. Global configuration such as 
`max_connection_pool_size` and other options which apply to all client instances are now configured through 
the `cognite.client.global_config` object which is an instance of `cognite.client.GlobalConfig`. Examples 
have been added to the docs.


cognite/client/utils/__init__.py
```diff
@@ -1,6 +1,5 @@
-from cognite.client.utils import (
+from cognite.client.utils import (  # isort: skip
```
ponnapz, line 1: is this comment for linter?

ponnapz (commented):
this would be a breaking change and require a major version bump right? I don't think I can do an effective review of this PR. Would be a good idea to get some of the regular users of the SDK to take a look at the changes.

cognite/client/utils/__init__.py
erlendvollset: Yeah, the linter wants to remove the imports because it thinks they are unused.

erlendvollset:
> this would be a breaking change and require a major version bump right? I don't think I can do an effective review of this PR. Would be a good idea to get some of the regular users of the SDK to take a look at the changes.

Yes, we do the major version bump in #1000. Agree, I will solicit feedback from SDK users.

ponnapz (approved):
trusting the resident python guru :smile:

Issue #981, opened 2022-07-26
Rework oauth credential handling


erlendvollset:
Currently, the client credential flow is supported natively by the SDK. For other authentication flows, we just point to these guides here: https://github.com/cognitedata/python-oidc-authentication.

We should natively support all these, by supporting a `credentials` parameter to `CogniteClient`. Something along the lines of:
```
CogniteClient(credentials: AbstractCredentialProvider)

# Where we support something like these as subclasses of AbstractCredentialProvider
# ClientCredentials, InteractiveLogin, DeviceCode, InteractiveLoginTokenRefresh, APIKey
```

We should then also make the dependencies necessary for each of the CredentialProviders optional.

This should replace the current `token` and `api_key` parameters.

PR #999, opened 2022-08-13
Rework how credentials are handled by the SDK

erlendvollset:
Addresses #981 
Depends on #998 

The client configuration no longer accepts the `api_key` and `token_...` arguments. 
It accepts only a single `credentials` argument which must be a `CredentialProvider` object. A few 
implementations have been provided (`APIKey`, `Token`, `OAuthClientCredentials`). Example usage has 
been added to the docs. More credential provider implementations will be added in the future to accommodate 
other OAuth flows.


docs/source/cognite.rst
```diff
@@ -187,10 +200,11 @@ global configuration
 You can set global configuration options like this:
 
 .. code:: python
-    from cognite.client imporg global_config, ClientConfig
 
+    from cognite.client imporg global_config, ClientConfig
```
haakonvt, line 204:
```suggestion
    from cognite.client import global_config, ClientConfig
```

cognite/client/credentials.py
```diff
@@ -0,0 +1,103 @@
...
+from datetime import datetime
+from typing import Callable, List, Optional, Tuple, Union
+
+from oauthlib.oauth2 import BackendApplicationClient, OAuth2Error
+from requests_oauthlib import OAuth2Session
+
+from cognite.client.exceptions import CogniteAuthError
+
+
+class CredentialProvider(ABC):
```
haakonvt, line 12: Great cleanup! 🤩

emelvarol-cognite (commented):
LGTM, thanks for restucturing. this looks much better imo.

ponnapz (approved):

PR #1002, opened 2022-08-17
Update link to latest docs

qTipTip:
The docs-badge was pointing to old documentation containing references to model hosting, etc.


erlendvollset (approved):
Good catch!

PR #1003, opened 2022-08-17
Support token custom args in OAuthClientCredentials

emelvarol-cognite:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/credentials.py
+++ b/cognite/client/credentials.py
@@ -1,7 +1,7 @@
 import threading
 from abc import ABC, abstractmethod
 from datetime import datetime
-from typing import Callable, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 from oauthlib.oauth2 import BackendApplicationClient, OAuth2Error
 from requests_oauthlib import OAuth2Session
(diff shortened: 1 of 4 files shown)


cognite/client/credentials.py
```diff
@@ -120,6 +127,7 @@ def _refresh_access_token(self) -> None:
                 client_id=self.__client_id,
                 client_secret=self.__client_secret,
                 scope=self.__scopes,
+                token_custom_args=self.__token_custom_args,
```
erlendvollset, line 129: This should be
```suggestion
                token_custom_args=**self.__token_custom_args,
```
But I don't want to do this since it ties our interface to the underlying implementation. I'm interested in which specific arguments you need to pass, and then we can instead add them explicitly to the constructor of this class.

erlendvollset (changes requested):

cognite/client/credentials.py
emelvarol-cognite: I think aize needs `audience`, I haven't seen any other special arguments passed by people except for that. Do you think we should only support that one? I thought different IdPs may have specific headers, that's why kept the old way. 🤔

erlendvollset: Oh, I see. These are custom args for the request itself, not for `oauth.fetch_token`. But this key/value pairs in this dict can now also be used to pass args to `oauth.fetch_token`, which we don't want. I'll push a suggested fixup.

emelvarol-cognite: awesome, thank you so much for quick help

erlendvollset (approved):
LGTM with suggested fixups

emelvarol-cognite:
> LGTM with suggested fixups

LGTM as well  🤣 , should we merge

PR #1006, opened 2022-08-19
Added Pip as requirement for functions

alchemistake:
## Description
Changed how the import of pip works and then added an additional optional dependency to add support for pip-less pythons such as distroless.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ Changes are grouped as follows
 ## [4.0.2] - 2022-08-19
 ### Fixed
 - Fixed the issue on SDK when Python installation didn't include pip.
+
+### Added
+- Added Optional dependency called functions. Usage: `pip install cognite-sdk[functions]`
  
 ## [4.0.1] - 2022-08-17
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [4.0.2] - 2022-08-19
+### Fixed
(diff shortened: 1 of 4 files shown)


emelvarol-cognite (approved):
Lgtm 🚀😎

cognite/client/_api/functions.py
```diff
@@ -721,6 +719,8 @@ def _validate_and_parse_requirements(requirements: List[str]) -> List[str]:
     parsed_reqs: List[str] = []
     for req in requirements:
         try:
+            from pip._internal.req.constructors import install_req_from_line
```
erlendvollset, line 722: Don't import inside the try/except, or the import error gets transformed into a ValueError. You're also reimporting the module  for every iteration in the loop.

We also have a local_import function in utils which will raise an exception with a more friendly error message. I'd prefer if you used that.

The `except Exception` is also bad practice. But that can be addressed in a different PR.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [4.0.2] - 2022-08-19
```
erlendvollset, line 17: Also specify that you've added a new optional dependency and what the extra is called (functions).

erlendvollset (changes requested):

HMEiding (approved):
LGTM 🚀 Update the PR description?

erlendvollset (approved):
lgtm

emelvarol-cognite:
@tuanng-cognite Can you review on behalf of geospatial? 🙏

tuanng-cognite (approved):

PR #1012, opened 2022-08-23
Add some badges and update text in READMEs

erlendvollset:
Who doesn't love a good badge, huh?


amorken (approved):
Sorry for the late review, I distracted myself trying to find a funny gif of some cop with a badge.

PR #1016, opened 2022-08-30
Cleanup of rest of COGNITE_ prefixed env vars

haakonvt:
## Description
Removed the last rest of `COGNITE_` prefixed env.vars.

Also had to make some changes to shut `mypy` up...

### DIFF
--- a/cognite/client/data_classes/geospatial.py
+++ b/cognite/client/data_classes/geospatial.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union, cast
 
 from cognite.client import utils
 from cognite.client.data_classes._base import CogniteResource, CogniteResourceList
+from cognite.client.utils._auxiliary import to_snake_case
 
 if TYPE_CHECKING:
     import geopandas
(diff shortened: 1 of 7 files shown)


cognite/client/data_classes/geospatial.py
```diff
@@ -4,6 +4,7 @@
 
 from cognite.client import utils
 from cognite.client.data_classes._base import CogniteResource, CogniteResourceList
+from cognite.client.utils._auxiliary import to_snake_case
```
haakonvt, line 7: This entire file was just `mypy` complaining:
```
cognite/client/data_classes/geospatial.py:98: error: Cannot determine type of
"to_snake_case"  [has-type]
        PRE_DEFINED_SNAKE_CASE_NAMES: Set[str] = {utils._auxiliary.to_snak...
```

tests/conftest.py
```diff
@@ -81,9 +80,12 @@ def rsps():
 
 @pytest.fixture
 def disable_gzip():
-    os.environ["COGNITE_DISABLE_GZIP"] = "1"
-    yield
-    del os.environ["COGNITE_DISABLE_GZIP"]
+    try:
```
erlendvollset, line 83: try shouldn't be necessary, I believe. Pytest executes the teardown even when exceptions are raised

erlendvollset (approved):
Excellent! Thanks for cleaning that up 🐍💥

tests/conftest.py
haakonvt: Oh, thats cool, TIL 🤩

PR #1017, opened 2022-09-01
camelCase set objects on Transformations Update

silvavelosa:
## Description
Fixes the issue when updating nonce credentials on transformations

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [4.5.1] - 2022-09-09
+## [4.5.2] - 2022-09-09
 ### Fixed
 - Fixes the issue when updating transformations with new nonce credentials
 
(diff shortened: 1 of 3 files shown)


tests/tests_integration/test_api/test_transformations/test_transformations.py
```diff
@@ -184,6 +198,18 @@ def test_update_partial(self, cognite_client, new_transformation):
             and updated_transformation.query == retrieved_transformation.query == "SELECT * from _cdf.assets"
         )
 
+    def test_update_nonce(self, cognite_client, new_transformation):
```
emelvarol-cognite, line 201: could you also add a test to update the transformations using Transformation object (besides TransformationUpdate which is used in this test)?

tests/tests_integration/test_api/test_transformations/test_transformations.py
```diff
@@ -26,11 +27,27 @@ def new_datasets(cognite_client):
 @pytest.fixture
 def new_transformation(cognite_client, new_datasets):
```
emelvarol-cognite, line 28: can you also create a transformation without credentials and then update it with credentials and check if credentials exist?

emelvarol-cognite (approved):
Awesome, just requested one extra test

PR #1027, opened 2022-09-09
Vision Extract: Add support for beta parameters

marvrez:
Adds support for providing beta features to the extract v1 endpoint.

### DIFF
--- a/cognite/client/_api/vision.py
+++ b/cognite/client/_api/vision.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Any, Dict, List, Optional, Type, Union
 
 from cognite.client._api_client import APIClient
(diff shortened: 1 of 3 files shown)


cognite/client/_api/vision.py
```diff
@@ -88,13 +89,18 @@ def extract(
         if isinstance(features, VisionFeature):
             features = [features]
 
+        beta_features = [f for f in features if f in VisionFeature.beta_features()]
+        if len(beta_features) > 0:
+            warnings.warn(f"Features {beta_features} are in beta and are still in development")
```
marvrez, line 94: Didn't really know what to put in the warning message. Open for suggestions here

miladcognite (approved):
LGTM :+1:

cognite/client/_api/vision.py
miladcognite: Sounds good to me

PR #1028, opened 2022-09-09
Use typing module rather than typing_extensions

erlendvollset:
get_args is supported in the typing module. typing_extensions is not included as a dependency, so we fail on import errors at runtime now.

The reason we didn't catch this in CI is that our dev dependencies bring in typing_extensions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Changes are grouped as follows
 - `Security` in case of vulnerabilities.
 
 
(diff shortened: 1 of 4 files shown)


me-ydv-5 (approved):
lgtm but how did this work when someone did not install dev deps?

erlendvollset:
> lgtm but how did this work when someone did not install dev deps?

It will not have worked without dev deps

PR #1029, opened 2022-09-09
Fix: Move EM and Diagrams API to beta client

marvrez:
## Description
This work fixes the  tests for the entity matcher and the diagrams api. Many of the endpoints in the mentioned APIs are in beta and since the Context API now actually check for `cdf-version` in the header, an error will be returned if the corresponding header value is not `beta`. The fix is to therefore use the beta client in the EM and diagrams integration tests.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_cognite_client.py
+++ b/cognite/client/_cognite_client.py
@@ -62,7 +62,7 @@ class CogniteClient:
         self.three_d = ThreeDAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
         self.labels = LabelsAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
         self.relationships = RelationshipsAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
-        self.entity_matching = EntityMatchingAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
+        self.entity_matching = EntityMatchingAPI(self._config, api_version="beta", cognite_client=self)
         self.templates = TemplatesAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
         self.vision = VisionAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
         self.extraction_pipelines = ExtractionPipelinesAPI(self._config, api_version="playground", cognite_client=self)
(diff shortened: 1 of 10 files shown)


cognite/client/_cognite_client.py
```diff
@@ -62,15 +62,15 @@ def __init__(self, config: Optional[ClientConfig] = None) -> None:
...
+        self.entity_matching = EntityMatchingAPI(self._config, api_version="beta", cognite_client=self)
         self.templates = TemplatesAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
         self.vision = VisionAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
         self.extraction_pipelines = ExtractionPipelinesAPI(self._config, api_version="playground", cognite_client=self)
         self.extraction_pipeline_runs = ExtractionPipelineRunsAPI(
             self._config, api_version="playground", cognite_client=self
         )
         self.transformations = TransformationsAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
-        self.diagrams = DiagramsAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
+        self.diagrams = DiagramsAPI(self._config, api_version="beta", cognite_client=self)
```
erlendvollset, line 73: beta APIs should only be made available through the cognite.client.beta.CogniteClient class.

marvrez: https://github.com/cognitedata/cognite-sdk-python/pull/1029/commits/d6d62412e7e44915e848a30ed487157446347709

tests/tests_integration/conftest.py
```diff
@@ -4,21 +4,21 @@
...
 
 from cognite.client import ClientConfig, CogniteClient
-from cognite.client.credentials import OAuthClientCredentials, OAuthInteractive
+from cognite.client.beta import CogniteClient as CogniteBetaClient
+from cognite.client.credentials import CredentialProvider, OAuthClientCredentials, OAuthInteractive
 
 
-@pytest.fixture(scope="session")
-def cognite_client() -> CogniteClient:
+def _get_credentials() -> CredentialProvider:
```
erlendvollset, line 11: Let's make this a session-scoped fixture aswell so we don't instantiate the credential providers multiple times (the credential provider instantiation has side effects).

erlendvollset (approved):
You also need to move the APIs into the beta client, so they're not available through the main client.

erlendvollset (changes requested):
hit approve by mistake

tests/tests_integration/conftest.py
marvrez: Tried it in this commit https://github.com/cognitedata/cognite-sdk-python/pull/1029/commits/3c9f48809dad6ab1d527acb323ef6a4b6d703085, seems like it doesn't work

marvrez:
Moved the APIs to the beta client in https://github.com/cognitedata/cognite-sdk-python/pull/1029/commits/50dbda317c5d49c681986b37633acb49dc5b8480. Thanks for the suggestion.

marvrez:
Closing this PR as the fix is instead implemented in the Context API. The fix is to revert to the previous logic.

PR #1032, opened 2022-09-15
geospatial aggregation to support output

tuanng-cognite:


vincent-cognite (approved):
Looks good!
Can you show an example for 2 different properties since this was the core reason for the change in the api?
Also, can we annotate the parameters with some "deprecation" warning or something? (I'm not familiar with how this could be done)

tuanng-cognite:
@vincent-cognite https://github.com/cognitedata/cognite-sdk-python/pull/1032/files#diff-6e996bb467c7a2411a485aa9e3b1140f21379862ace9c80f0979d22fc44db42dR820 deprecation warning is shown when old properties are used.

PR #1033, opened 2022-09-15
[MLOPS-1227] Avoid opening a file for writing twice

qTipTip:
## Description

The FunctionsAPI had an issue whereby we opened a file twice. This works on UNIX, not on Windows. 
The change proposed here is to simply open it once, and pass the open file handler around instead.

From the `NamedTemporaryFile`-docs: 
> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows).

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/functions.py
+++ b/cognite/client/_api/functions.py
@@ -7,7 +7,7 @@ from inspect import getdoc, getsource
 from numbers import Integral, Number
 from pathlib import Path
 from tempfile import NamedTemporaryFile, TemporaryDirectory
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Sequence, Union, cast, IO
+from typing import IO, TYPE_CHECKING, Any, Callable, Dict, List, Optional, Sequence, Union, cast
 from zipfile import ZipFile
 
 from cognite.client import utils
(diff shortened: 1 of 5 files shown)


cognite/client/_api/functions.py
```diff
@@ -729,12 +729,12 @@ def _validate_and_parse_requirements(requirements: List[str]) -> List[str]:
     return parsed_reqs
 
 
-def _write_requirements_to_file(file_path: str, requirements: List[str]) -> None:
-    with open(file_path, "w+") as f:
-        f.write("\n".join(requirements))
+def _write_requirements_to_file(file: IO, requirements: List[str]) -> None:
+    if not file.closed:
```
HMEiding, line 733: What about the scenario where the file is closed? Is that not possible?

HMEiding (approved):
LGTM

erlendvollset (approved):
LGTM

cognite/client/_api/functions.py
qTipTip: Okay. I'll remove the if :)

PR #1036, opened 2022-09-19
Ensure raw row insert endpoint is retried

erlendvollset:


tests/tests_unit/test_api_client.py
```diff
@@ -1147,6 +1147,7 @@ class TestHelpers:
             ("PUT", "https://localhost:8000.com/api/v1/projects/blabla/assets", True),
             ("PATCH", "https://localhost:8000.com/api/v1/projects/blabla/patchy", True),
             ("GET", "https://another-cluster.cognitedata.com/login/status", True),
+            ("POST", "https://another-cluster.cognitedata.com/api/v1/projects/hey/raw", True),
```
dmivankov, line 1150: should be longer?
`https://api.cognitedata.com/api/v1/projects/{project}/raw/dbs/{dbName}/tables/{tableName}/rows`
https://docs.cognite.com/api/v1/#tag/Raw/operation/postRows

erlendvollset: Oh yeah. Forgot the en and table was specified in the path. Good catch!

cognite/client/_api_client.py
```diff
@@ -230,21 +213,27 @@ def _get_base_url_with_base_path(self) -> str:
...
         return False
 
+    @classmethod
+    @functools.lru_cache(64)
+    def _endpoint_path_is_retryable(cls, path: str) -> bool:
+        valid_path_pattern = r"(?:http|https)://[a-z\d.:\-]+(?:/api/(?:v1|playground)/projects/[^/]+)?(/[^\?]+)?(\?.+)?"
+        if not re.match(valid_path_pattern, path):
+            raise ValueError("Path {} is not valid. Cannot resolve whether or not it is retryable".format(path))
+        for pattern in cls._RETRYABLE_POST_ENDPOINT_REGEX_PATTERNS:
+            if re.search(pattern, path):
```
dmivankov, line 233:
```suggestion
            if re.search("^" + pattern + "$", path):
```
or something like that to exclude extra stuff?

cognite/client/_api_client.py
```diff
@@ -54,34 +55,16 @@ class APIClient:
...
-        "/timeseries/byids",
-        "/timeseries/search",
-        "/timeseries/list",
         "/timeseries/data",
         "/timeseries/data/list",
         "/timeseries/data/latest",
         "/timeseries/data/delete",
-        "/sequences/byids",
-        "/sequences/search",
-        "/sequences/data",
```
dmivankov, line 77: this one disappeared

erlendvollset: They're included in the first regex pattern.

cognite/client/_api_client.py
erlendvollset: That wont work since urls can include query params after and different protocols/hosts/ports before.

cognite/client/_api_client.py
erlendvollset: thanks

cognite/client/_api_client.py
dmivankov: but here it's path, so prefix part should be ok. Suffix yes, probably then something like `[#?].*$
just the prefix part should be ok more or less as working urls aren't prefixes of each other (mostly)

dmivankov (approved):
Detector can be improved, but should already be ok enough.
False positives are when api-like sub-path is part of api-path. And usually user-defined path segments are clamped between constants, which makes matches not very likely + most endpoints are retriable

PR #1041, opened 2022-09-23
Dps Fetch Refactor #003: Add ThreadPoolExecutor with PriorityQueue

haakonvt:


cognite/client/utils/_priority_tpe.py
```diff
@@ -0,0 +1,131 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+"""
+
+import atexit
```
erlendvollset, line 33: This file is missing type annotations. Did you reenable mypy?

haakonvt: Check out line 0 😅 I was hoping for some leniency here - or perhaps a shared effort 😬 Pretty much the same code as in the CPython source: https://github.com/python/cpython/blob/3.8/Lib/concurrent/futures/thread.py (except for the different queue ++)

haakonvt: # 😇

erlendvollset: Leniency granted, but if ignoring entire files, please add them to the mypy.ini file instead of adding a catch all `# type: ignore`. Easier to get an overview over what is missing type annotations then, so we can address it later.

haakonvt: Thanks! 😄 Ah, yes, that sounds way better 👌

PR #1043, opened 2022-09-23
Dps Fetch Refactor #005: New dps fetch code

haakonvt:
Smol PR, no? 👼

### DIFF
--- a/cognite/client/_api/datapoint_constants.py
+++ b/cognite/client/_api/datapoint_constants.py
@@ -1,0 +1,87 @@
+from datetime import datetime
+from typing import Dict, Iterable, List, Optional, TypedDict, Union
+
+try:
+    import numpy as np
+    import numpy.typing as npt
+
+    NUMPY_IS_AVAILABLE = True
+except ImportError:  # pragma no cover
+    NUMPY_IS_AVAILABLE = False
+
+if NUMPY_IS_AVAILABLE:
+    NumpyDatetime64NSArray = npt.NDArray[np.datetime64]
+    NumpyInt64Array = npt.NDArray[np.int64]
+    NumpyFloat64Array = npt.NDArray[np.float64]
+    NumpyObjArray = npt.NDArray[np.object_]
+
+# Datapoints API-limits:
+DPS_LIMIT_AGG = 10_000
+DPS_LIMIT = 100_000
+POST_DPS_OBJECTS_LIMIT = 10_000
+FETCH_TS_LIMIT = 100
+RETRIEVE_LATEST_LIMIT = 100
+
+
+ALL_SORTED_DP_AGGS = sorted(
+    [
+        "average",
+        "max",
+        "min",
+        "count",
+        "sum",
+        "interpolation",
+        "step_interpolation",
+        "continuous_variance",
+        "discrete_variance",
+        "total_variation",
+    ]
+)
+
+
+class CustomDatapointsQuery(TypedDict, total=False):
+    # No field required
+    start: Union[int, str, datetime, None]
+    end: Union[int, str, datetime, None]
+    aggregates: Optional[List[str]]
+    granularity: Optional[str]
+    limit: Optional[int]
+    include_outside_points: Optional[bool]
+    ignore_unknown_ids: Optional[bool]
+
+
+class DatapointsQueryId(CustomDatapointsQuery):
+    id: int  # required field
+
+
+class DatapointsQueryExternalId(CustomDatapointsQuery):
+    external_id: str  # required field
+
+
+class CustomDatapoints(TypedDict, total=False):
+    # No field required
+    start: int
+    end: int
+    aggregates: Optional[List[str]]
+    granularity: Optional[str]
+    limit: int
+    include_outside_points: bool
+
+
+class DatapointsPayload(CustomDatapoints):
+    items: List[CustomDatapoints]
+
+
+DatapointsTypes = Union[int, float, str]
+
+
+class DatapointsFromAPI(TypedDict):
+    id: int
+    externalId: Optional[str]
+    isString: bool
+    isStep: bool
+    datapoints: List[Dict[str, DatapointsTypes]]
+
+
+DatapointsIdTypes = Union[int, DatapointsQueryId, Iterable[Union[int, DatapointsQueryId]]]
+DatapointsExternalIdTypes = Union[str, DatapointsQueryExternalId, Iterable[Union[str, DatapointsQueryExternalId]]]
--- a/cognite/client/_api/datapoint_tasks.py
+++ b/cognite/client/_api/datapoint_tasks.py
@@ -1,0 +1,1167 @@
+from __future__ import annotations
+
+import math
+import numbers
+import operator as op
+import warnings
+from abc import abstractmethod
+from datetime import datetime
+from functools import cached_property
+from itertools import chain
+from threading import Lock
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Hashable,
+    Iterable,
+    Iterator,
+    List,
+    Literal,
+    NoReturn,
+    Optional,
+    Sequence,
+    Tuple,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+)
+
+from sortedcontainers import SortedDict, SortedList  # type: ignore [import]
+
+from cognite.client._api.datapoint_constants import (
+    DPS_LIMIT,
+    DPS_LIMIT_AGG,
+    NUMPY_IS_AVAILABLE,
+    CustomDatapoints,
+    CustomDatapointsQuery,
+    DatapointsExternalIdTypes,
+    DatapointsFromAPI,
+    DatapointsIdTypes,
+    DatapointsQueryExternalId,
+    DatapointsQueryId,
+    DatapointsTypes,
+)
+from cognite.client.data_classes.datapoints import Datapoints, DatapointsArray, DatapointsQuery
+from cognite.client.utils._auxiliary import convert_all_keys_to_snake_case, to_camel_case
+from cognite.client.utils._identifier import Identifier
+from cognite.client.utils._time import (
+    align_start_and_end_for_granularity,
+    granularity_to_ms,
+    split_time_range,
+    timestamp_to_ms,
+)
+
+if NUMPY_IS_AVAILABLE:
+    import numpy as np
+
+
+if TYPE_CHECKING:
+    import numpy.typing as npt
+
+
+T = TypeVar("T")
+
+
+class _SingleTSQueryValidator:
+    def __init__(self, user_query: DatapointsQuery) -> None:
+        self.user_query = user_query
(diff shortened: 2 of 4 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -243,38 +1057,39 @@ def insert(
...
-        if isinstance(datapoints, Datapoints):
-            datapoints = [(t, v) for t, v in zip(datapoints.timestamp, datapoints.value)]
-        post_dps_object.update({"datapoints": datapoints})
+        if isinstance(datapoints, (Datapoints, DatapointsArray)):
+            if datapoints.value is None:
+                raise ValueError(
+                    "When inserting data using a `Datapoints` or `DatapointsArray` object, only raw datapoints are supported"
+                )
+            datapoints = list(zip(datapoints.timestamp, datapoints.value))  # type: ignore [arg-type]
+        post_dps_object["datapoints"] = datapoints
```
haakonvt, line 1092: Note to self: test with `DatapointsArray`

cognite/client/_api/datapoint_tasks.py
```diff
@@ -0,0 +1,1165 @@
...
+        )
+
+    @staticmethod
+    def _validate_ts_query_dict_keys(
+        dct: Dict[str, Any], arg_name: str, exp_type: type
+    ) -> Union[DatapointsQueryId, DatapointsQueryExternalId]:
+        if arg_name not in dct:
+            if (arg_name_cc := to_camel_case(arg_name)) not in dct:
+                raise KeyError(f"Missing required key `{arg_name}` in dict: {dct}.")
+            # For backwards compatability we accept identifier in camel case: (Make copy to avoid side effects
```
HaydenCognite, line 148:
```suggestion
            # For backwards compatibility we accept identifiers in camel case: (Make copy to avoid side effects
```

HaydenCognite (commented):
Just a small typo.

cognite/client/_api/datapoints.py
haakonvt: Fixed in e3b4e5a

cognite/client/_api/datapoint_constants.py
```diff
@@ -0,0 +1,87 @@
...
+class DatapointsFromAPI(TypedDict):
+    id: int
+    externalId: Optional[str]
+    isString: bool
+    isStep: bool
+    datapoints: List[Dict[str, DatapointsTypes]]
+
+
+DatapointsIdTypes = Union[int, DatapointsQueryId, Iterable[Union[int, DatapointsQueryId]]]
+DatapointsExternalIdTypes = Union[str, DatapointsQueryExternalId, Iterable[Union[str, DatapointsQueryExternalId]]]
```
erlendvollset, line 87: These classes and type aliases would be more appropriately placed in `cognite.client.data_classes.datapoints`. But since we access the datapoints api through `CogniteClient::time_series::data` now, perhaps they should live in `cognite.client.data_classes.time_series` instead. But not in this module at least.

cognite/client/_api/datapoint_constants.py
```diff
@@ -0,0 +1,87 @@
...
+    granularity: Optional[str]
+    limit: int
+    include_outside_points: bool
+
+
+class DatapointsPayload(CustomDatapoints):
+    items: List[CustomDatapoints]
+
+
+DatapointsTypes = Union[int, float, str]
```
erlendvollset, line 75:
```suggestion
DatapointValue = Union[int, float, str]
```

cognite/client/_api/datapoint_constants.py
```diff
@@ -0,0 +1,87 @@
...
+
+class DatapointsFromAPI(TypedDict):
+    id: int
+    externalId: Optional[str]
+    isString: bool
+    isStep: bool
+    datapoints: List[Dict[str, DatapointsTypes]]
+
+
+DatapointsIdTypes = Union[int, DatapointsQueryId, Iterable[Union[int, DatapointsQueryId]]]
```
erlendvollset, line 86:
```suggestion
DatapointsId = Union[int, DatapointsQueryId, Iterable[Union[int, DatapointsQueryId]]]
```

cognite/client/_api/datapoint_constants.py
erlendvollset, line 87:
```suggestion
DatapointsExternalId = Union[str, DatapointsQueryExternalId, Iterable[Union[str, DatapointsQueryExternalId]]]
```

cognite/client/_api/datapoints.py
```diff
@@ -1,117 +1,969 @@
...
+        id: Optional[DatapointsIdTypes] = None,
+        external_id: Optional[DatapointsExternalIdTypes] = None,
+        start: Union[int, str, datetime, None] = None,
+        end: Union[int, str, datetime, None] = None,
+        aggregates: Optional[List[str]] = None,
+        granularity: Optional[str] = None,
+        limit: Optional[int] = None,
+        include_outside_points: bool = False,
         ignore_unknown_ids: bool = False,
     ) -> Union[None, Datapoints, DatapointsList]:
```
erlendvollset, line 579: Let's add overloads to this method to show what the. Return type will depend on which of types included by `DatapointsIdTypes` and `DatapointsExternalIdTypes` are used, so you can't really use those type aliases then.

cognite/client/_api/datapoints.py
```diff
@@ -1,117 +1,969 @@
...
-        """`Get datapoints for one or more time series. <https://docs.cognite.com/api/v1/#operation/getMultiTimeSeriesDatapoints>`_
+        """`Retrieve datapoints for one or more time series. <https://docs.cognite.com/api/v1/#operation/getMultiTimeSeriesDatapoints>`_
+
+        **Note**: All arguments are optional, as long as at least one identifier is given. When passing aggregates, granularity must also be given.
+        When passing dict objects with specific parameters, these will take precedence. See examples below.
```
erlendvollset, line 583: This note becomes redundant by adding some overloads.

cognite/client/_api/datapoints.py
```diff
@@ -1,117 +1,969 @@
...
+            granularity (str): The granularity to fetch aggregates at. e.g. '1s', '2h', '10d'. Default: None.
+            limit (int): Maximum number of datapoints to return for each time series. Default: None (no limit)
+            include_outside_points (bool): Whether or not to include outside points. Not allowed when fetching aggregates. Default: False
+            ignore_unknown_ids (bool): Whether or not to ignore missing time series rather than raising an exception. Default: False
 
         Returns:
-            Union[None, Datapoints, DatapointsList]: A Datapoints object containing the requested data, or a list of such objects. If `ignore_unknown_id` is True, single id is requested and it is not found, the function will return `None`.
+            Union[None, Datapoints, DatapointsList]: A `Datapoints` object containing the requested data, or a `DatapointsList` if multiple
+                time series were asked for. If `ignore_unknown_ids` is `True`, a single time series is requested and it is not found, the function
+                will return `None`. The ordering is first ids, then external_ids.
```
erlendvollset, line 601: This note becomes redundant by adding some overloads.

cognite/client/_api/datapoints.py
```diff
@@ -1,117 +1,969 @@
...
+        id: Optional[DatapointsIdTypes] = None,
+        external_id: Optional[DatapointsExternalIdTypes] = None,
+        start: Union[int, str, datetime, None] = None,
+        end: Union[int, str, datetime, None] = None,
+        aggregates: Optional[List[str]] = None,
+        granularity: Optional[str] = None,
+        limit: Optional[int] = None,
+        include_outside_points: bool = False,
+        ignore_unknown_ids: bool = False,
+    ) -> Union[None, DatapointsArray, DatapointsArrayList]:
```
erlendvollset, line 713: Seems to me like DatapointsArray and DatapointsArrayList have the exact same interface as Datapoints and DatapointsList, respectively. I think we can just make a common protocols `Datapoints` and `DatapointsList`, and then have separate implementations for the numpy and non-numpy versions. Then, we won't need a separate `retrieve_arrays`, we just make the enable the optimizations behind the scenes if numpy is installed.

cognite/client/_api/datapoints.py
```diff
@@ -1,117 +1,969 @@
...
+        end = pd.Timestamp(max(q.end for q in fetcher.agg_queries), unit="ms")
+        (granularity,) = grans_given
+        # Pandas understand "Cognite granularities" except `m` (minutes) which we must translate:
+        return df.reindex(pd.date_range(start=start, end=end, freq=granularity.replace("m", "T"), inclusive="left"))
+
+    @overload
+    def query(
+        self,
+        query: Union[Sequence[DatapointsQuery], DatapointsQuery],
+        use_numpy: Literal[False],
```
erlendvollset, line 915: Same note as above, if we give DatapointsList and DatapointsArrayList a common Protocol, we won't need this argument. We just enable behind the scenes.

erlendvollset (commented):
Let's just merge all the PRs into the v5 branch now. Reviewing becomes very fragmented when code we depend on doesn't exist in the same PR.

cognite/client/_api/datapoint_constants.py
haakonvt: Makes sense 👍 Will move

cognite/client/_api/datapoints.py
haakonvt: There will be a looot of overloaded methods though, but I’m fine with that solution 👍 Do you see any way to avoid optional Datapoints for when “single identifier is missing and ignore=True”?

cognite/client/_api/datapoints.py
haakonvt: I’ve done some more thinking and have come to the realization that I’m just gonna remove the query method entirely and allow individual customization of the ignore option. It will not mirror the API endpoint anymore, but imo be an improvement ☺️

cognite/client/_api/datapoints.py
haakonvt: Discussed and resolved on slack: https://cognitedata.slack.com/archives/C03PBAB4KUJ/p1664797575423249

cognite/client/_api/datapoint_constants.py
haakonvt: Moved to `cognite.client.data_classes.datapoints`

cognite/client/_api/datapoint_constants.py
haakonvt: Fixed in #1043

cognite/client/_api/datapoints.py
haakonvt: Implemented #1051. Ended up with 3 x 8 overloaded methods 🚀

cognite/client/_api/datapoints.py
haakonvt: Not entirely, `ignore_unknown_ids` may be overridden by an individual dictionary, so I've left it in. I think most people rather read this than look at the 8 overloaded methods anyway 😅

PR #1048, opened 2022-09-24
Dps Fetch Refactor #010: Integration tests, added, fixed and improved

haakonvt:


scripts/create_ts_for_integration_tests.py
```diff
@@ -0,0 +1,207 @@
...
+        df_add(pd.DataFrame({name: arr + i}, index=weekly_idx))
+
+    for i, name in enumerate(NAMES[53:103], i + 1):
+        ts_add(
+            TimeSeries(name=name, external_id=name, is_string=True, metadata={"offset": i, "delta": UNIT_IN_MS["w"]})
+        )
+        df_add(pd.DataFrame({name: (arr + i).astype(str)}, index=weekly_idx))
+
+    i = 103
+    for is_step in [False, True]:
```
HaydenCognite, line 92: Question: Would you get a noticeable speedup if you stored all the "i + 1", "i + 2", "i + 3", "i + 4", and "i + 5" variables that get computed repeatedly in this loop? I.e., set variables like `t_1 = i + 1`, `t_2 = i + 2`, etc and use those instead of the repeated calculations. 🤔

HaydenCognite (commented):
Just one comment for a potential speedup. 👍

scripts/create_ts_for_integration_tests.py
haakonvt: This is a script that - if no one touches the Cognite project for the "Python-SDK-integration-tests" - never has to be run again 😉

erlendvollset: Assume that someone will delete all your test data tomorrow. Ideally, you have a session-scoped fixture which validates the data you need and reingests it if any is missing.

haakonvt: That would not work; it wouldn’t be possible to know if the underlying data was wrong or the code that fetched it. I should probably add a note about that reingesting should always be done by a previous/released version

PR #1068, opened 2022-10-14
Fix integration test

BenChand:
## Description
Fix an integration test requesting too much data

## Checklist:
- [x] Tests added/updated.


soleh23:
Is the issue about too many datapoints in that range? If so, perhaps adding `limit = 1` should suffice

BenChand:
> Is the issue about too many datapoints in that range? If so, perhaps adding `limit = 1` should suffice

I believe so yeah, will try adding limit=1

me-ydv-5 (approved):

PR #1072, opened 2022-10-17
V5 proto check

haakonvt:
- Adds step to verify proto files AND the generated python files, by using `diff`
- Change `exit()` which is not guaranteed to be available on systems

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -58,6 +58,23 @@ jobs:
           poetry config virtualenvs.create false
           poetry install -E all
 
+      - name: Verify proto files
+        uses: arduino/setup-protoc@v1
+        env:
+          TMPDIR: latest_proto
+          PROTO_URL: https://raw.githubusercontent.com/cognitedata/protobuf-files/master/v1/timeseries
+        run: |
+          mkdir $TMPDIR
+          curl $PROTO_URL/data_points.proto --output $TMPDIR/data_points.proto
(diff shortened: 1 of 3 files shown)


scripts/check_version_match.py
```diff
@@ -15,4 +15,4 @@
         f"Version in pyproject.toml ({version_in_pyproject_toml}) does not match version in "
         f"cognite/client/_version ({__version__})"
     )
-    exit(1)
+    raise SystemExit(1)
```
erlendvollset, line 18: Is this the latest n' greatest in exit methods? 😁

.github/workflows/release.yaml
```diff
@@ -58,6 +58,25 @@ jobs:
...
           poetry install -E all
 
+      - name: Install protoc compiler
+        uses: arduino/setup-protoc@v1
+
+      - name: Verify proto files
+        env:
+          TMPDIR: latest_proto
+          PROTO_URL: https://raw.githubusercontent.com/cognitedata/protobuf-files/master/v1/timeseries
+        run: |
```
erlendvollset, line 68: Very nice!

erlendvollset (approved):

scripts/check_version_match.py
haakonvt: Haha yes! I learned it from the Python-Youtube-God himself ( @asottile ) 😂

PR #1075, opened 2022-10-26
add filter class for function calls

maur1:
## Description
Add function calls filter class

### Why:
None is read as filter input and the functions.list_calls() endpoint will not return any function calls. 
functions.calls.list(function_id=xxx) now returns a list of calls for the specified function id

### Still a bug to be solved:
Looks like the endpoint returns an empty list if any filter parameters are present

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "cognite-sdk"
 
-version = "4.11.1"
+version = "4.11.2"
 
 description = "Cognite Python SDK"
 readme = "README.md"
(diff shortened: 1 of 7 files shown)


erlendvollset:
Please also add a test which reproduces the issue.

pavelzubarev:
@erlendvollset ;)

erlendvollset:
You will need to bump version and update changelog aswell.

erlendvollset (approved):
Lgtm!

erlendvollset (changes requested):
Little too quick

pyproject.toml
```diff
@@ -21,6 +21,7 @@ pandas = { version = "*", optional = true }
 geopandas = { version = ">=0.10.0", optional = true }
 shapely = { version = ">=1.7.0", optional = true }
 pip = { version = ">=20.0.0", optional = true}
+coverage = "^6.5.0"
```
erlendvollset, line 24: This should be added to dev dependencies.

erlendvollset: Or why do we need to add it at all? pytest-cov should cover our needs.

erlendvollset (changes requested):
Little too quick

PR #1079, opened 2022-11-08
Fix: Use named arg in Vision extract doc + improve error message in `VisionExtractJob.save_annotations`

miladcognite:
## Description
* Use named arg in Vision extract doc
* Improve error message in save_predictions
* Add tests for `VisionExtractJob.save_annotations`

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/contextualization.py
+++ b/cognite/client/data_classes/contextualization.py
@@ -699,10 +699,14 @@ class VisionExtractJob(VisionJob):
             Union[Annotation, AnnotationList]: (suggested) annotation(s) stored in CDF.
 
         """
-        annotations = self._predictions_to_annotations(
-            creating_user=creating_user, creating_app=creating_app, creating_app_version=creating_app_version
+        if self.status == JobStatus.COMPLETED.value:
+            annotations = self._predictions_to_annotations(
+                creating_user=creating_user, creating_app=creating_app, creating_app_version=creating_app_version
+            )
+            if annotations:
+                return self._cognite_client.annotations.suggest(annotations=annotations)
+            raise CogniteException("Extract job does not contain any predictions.")
+
+        raise CogniteException(
+            "Extract job is not completed. If the job is queued or running, wait for completion and try again"
         )
-        if annotations:
-            return self._cognite_client.annotations.suggest(annotations=annotations)
-        else:
-            raise CogniteException("Extract job is not completed. Wait for completion and try again")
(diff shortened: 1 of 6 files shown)


tests/tests_unit/test_api/test_vision_extract.py
```diff
@@ -46,12 +48,24 @@ def mock_get_response_body_ok() -> Dict[str, Any]:
...
+                            },
+                        }
+                    ]
+                },
+            }
         ],
-        "useCache": True,
-        "partialMatch": True,
-        "assetSubtreeIds": [39468345],
+        "parameters": {"assetTagDetectionParameters": {"assetSubtreeIds": [], "partialMatch": False, "threshold": 0.4}},
```
torbjornopheim, line 68: If predictions is correct, maybe it should be updated in the response samples here: https://docs.cognite.com/api/v1/#tag/Vision/operation/postVisionExtract

torbjornopheim (approved):

tests/tests_unit/test_api/test_vision_extract.py
miladcognite: This is a sample response from the _GET_ endpoint (not POST), which I believe is compatible with what we have in the API docs (does not include all optional fields)

marvrez (approved):

tests/tests_unit/test_api/test_vision_extract.py
torbjornopheim: Yep, looked at the wrong one

erlendvollset (approved):
LGTM

cognite/client/data_classes/contextualization.py
```diff
@@ -699,10 +699,14 @@ def save_predictions(
...
         """
-        annotations = self._predictions_to_annotations(
-            creating_user=creating_user, creating_app=creating_app, creating_app_version=creating_app_version
+        if self.status == JobStatus.COMPLETED.value:
+            annotations = self._predictions_to_annotations(
+                creating_user=creating_user, creating_app=creating_app, creating_app_version=creating_app_version
+            )
+            if annotations:
+                return self._cognite_client.annotations.suggest(annotations=annotations)
+            raise CogniteException("Extract job does not contain any predictions.")
```
erlendvollset, line 708: Hmm. Should this be an exception? The job completed succesfully - so this would mean every invocation of this method needs to be wrappe din a try-except?

miladcognite: Tried to keep the previous behavior, but yes I agree would be better to return empty list or something. Will change it in an upcoming PR :+1:

PR #1084, opened 2022-11-15
WIP: Add datapoints.delete method for improved dps delete support

erlendvollset:
Wanna make this just datapoints.delete() instead. And that can accept timestamps and ranges, as well as Datapoint and Datapoints objects. Then we'll deprecate the other ones in V5 and remove in V6.

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -2,7 +2,7 @@ import copy
 import math
 import re as regexp
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Set, Tuple, Union, cast, overload
 
 import cognite.client.utils._time
 from cognite.client import utils
@@ -322,6 +322,55 @@ class DatapointsAPI(APIClient):
         dps_poster = DatapointsPoster(self)
         dps_poster.insert(datapoints)
 
+    @overload
+    def delete_point(self, point: Dict[str, Union[int, str, datetime]]) -> None:
+        ...
+
+    @overload
+    def delete_point(self, point: Sequence[Dict[str, Union[int, str, datetime]]]) -> None:
+        ...
+
+    def delete_point(
+        self, point: Union[Dict[str, Union[int, str, datetime]], Sequence[Dict[str, Union[int, str, datetime]]]]
+    ) -> None:
+        """Delete a single datapoint from one or more time series.
+
+        Args:
+            point (Union[Dict[str, Union[int, str, datetime]], Sequence[Dict[str, Union[int, str, datetime]]]]): The point or points to delete
+
+        Returns:
+            None
+
+        Examples:
+
+            Deleting a single datapoint from a single time series::
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> c.datapoints.delete_point(point={"id": 1, "timestamp": 12345})
+
+            Deleting a single datapoint from multiple time series::
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> c.datapoints.delete_point(point=[{"id": 1, "timestamp": 123}, {"externalId": "abc", "timestamp": datetime(year=1992)}])
+        """
+        point_list = point if isinstance(point, Sequence) else [point]
+        valid_ranges = []
+        for point in point_list:
+            for key in point:
+                if key not in ("id", "externalId", "timestamp"):
+                    raise AssertionError(
+                        f"Invalid key '{key}' in point. Must contain 'timestamp', and 'id' or 'externalId"
+                    )
+            id = cast(Optional[int], point.get("id"))
+            external_id = cast(Optional[str], point.get("externalId"))
+            valid_range = Identifier.of_either(id, external_id).as_dict()
+            start = utils._time.timestamp_to_ms(point["timestamp"])
+            valid_range.update({"start": start, "end": start + 1})
+            valid_ranges.append(valid_range)
+        self.delete_ranges(ranges=valid_ranges)
+
     def delete_range(
         self, start: Union[int, str, datetime], end: Union[int, str, datetime], id: int = None, external_id: str = None
     ) -> None:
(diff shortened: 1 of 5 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -322,6 +322,29 @@ def insert_multiple(self, datapoints: List[Dict[str, Union[str, int, List]]]) ->
         dps_poster = DatapointsPoster(self)
         dps_poster.insert(datapoints)
 
+    def delete_point(self, timestamp: Union[int, str, datetime], id: int = None, external_id: str = None) -> None:
+        """Delete a range of datapoints from a time series.
```
amorken, line 326:
```suggestion
        """Delete a single datapoint from a time series.
```

cognite/client/_api/datapoints.py
```diff
@@ -322,6 +322,29 @@ def insert_multiple(self, datapoints: List[Dict[str, Union[str, int, List]]]) ->
...
+        Returns:
+            None
+
+        Examples:
+
+            Deleting the last week of data from a time series::
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> c.datapoints.delete_range(start="1w-ago", end="now", id=1)
```
amorken, line 342:
```suggestion
                >>> c.datapoints.delete_point(timestamp="2000-01-01T00:00:00Z", id=1)
```
...assuming that the timestamp format is accepted, I haven't tested it. =)

amorken (commented):
+1 in principle but maybe adjust docs before merging?

cognite/client/_api/datapoints.py
```diff
@@ -322,6 +322,55 @@ def insert_multiple(self, datapoints: List[Dict[str, Union[str, int, List]]]) ->
...
+                >>> c.datapoints.delete_point(point=[{"id": 1, "timestamp": 123}, {"externalId": "abc", "timestamp": datetime(year=1992)}])
+        """
+        point_list = point if isinstance(point, Sequence) else [point]
+        valid_ranges = []
+        for point in point_list:
+            for key in point:
+                if key not in ("id", "externalId", "timestamp"):
+                    raise AssertionError(
+                        f"Invalid key '{key}' in point. Must contain 'timestamp', and 'id' or 'externalId"
+                    )
```
haakonvt, line 365: I think I've seen this type of logic in multiple places in the SDK, could be split out to utils.

cognite/client/_api/datapoints.py
```diff
@@ -322,6 +322,55 @@ def insert_multiple(self, datapoints: List[Dict[str, Union[str, int, List]]]) ->
...
+                >>> c = CogniteClient()
+                >>> c.datapoints.delete_point(point=[{"id": 1, "timestamp": 123}, {"externalId": "abc", "timestamp": datetime(year=1992)}])
+        """
+        point_list = point if isinstance(point, Sequence) else [point]
+        valid_ranges = []
+        for point in point_list:
+            for key in point:
+                if key not in ("id", "externalId", "timestamp"):
+                    raise AssertionError(
+                        f"Invalid key '{key}' in point. Must contain 'timestamp', and 'id' or 'externalId"
```
haakonvt, line 364: `KeyError` or `ValueError`?

cognite/client/_api/datapoints.py
```diff
@@ -322,6 +322,55 @@ def insert_multiple(self, datapoints: List[Dict[str, Union[str, int, List]]]) ->
...
+    @overload
+    def delete_point(self, point: Dict[str, Union[int, str, datetime]]) -> None:
+        ...
+
+    @overload
+    def delete_point(self, point: Sequence[Dict[str, Union[int, str, datetime]]]) -> None:
+        ...
+
+    def delete_point(
+        self, point: Union[Dict[str, Union[int, str, datetime]], Sequence[Dict[str, Union[int, str, datetime]]]]
```
haakonvt, line 334: To me it doesn't make much sense to allow `str` here; as these are all the "unprecise" or relative time offsets like "an hour ago" or "now".

haakonvt, line 334: ...what I would like to see though was `Datapoint` and `Datapoints` supported!

cognite/client/_api/datapoints.py
```diff
@@ -322,6 +322,55 @@ def insert_multiple(self, datapoints: List[Dict[str, Union[str, int, List]]]) ->
...
+    ) -> None:
+        """Delete a single datapoint from one or more time series.
```
haakonvt, line 336: Just a suggestion as the current sound like "delete t=200 from ts=a,b and c":
```suggestion
        """Delete individual datapoints from one or more time series.
```

cognite/client/_api/datapoints.py
```diff
@@ -322,6 +322,55 @@ def insert_multiple(self, datapoints: List[Dict[str, Union[str, int, List]]]) ->
...
+        for point in point_list:
+            for key in point:
+                if key not in ("id", "externalId", "timestamp"):
+                    raise AssertionError(
+                        f"Invalid key '{key}' in point. Must contain 'timestamp', and 'id' or 'externalId"
+                    )
+            id = cast(Optional[int], point.get("id"))
+            external_id = cast(Optional[str], point.get("externalId"))
+            valid_range = Identifier.of_either(id, external_id).as_dict()
+            start = utils._time.timestamp_to_ms(point["timestamp"])
```
haakonvt, line 369: May raise `KeyError`. Add a must-be-present check in loop above?

haakonvt (commented):
Just a few thoughts from me 😊

cognite/client/_api/datapoints.py
erlendvollset: Agree. But this dict also contains the externslId, which has a string value.

cognite/client/_api/datapoints.py
erlendvollset: Will do 👍

cognite/client/_api/datapoints.py
erlendvollset: The check above already ensures the existence of this key. Or did you mean smth else?

erlendvollset: Ah no it doesn't, it just checks that no other keys are present. I'll make a utility for doing both of these types of checks for dicts.

erlendvollset: Great catch!

cognite/client/_api/datapoints.py
```diff
@@ -346,17 +395,17 @@ def delete_range(
         """
         start = utils._time.timestamp_to_ms(start)
         end = utils._time.timestamp_to_ms(end)
-        assert end > start, "end must be larger than start"
+        assert end >= start, "end must be larger than start"
```
erlendvollset, line 398: Note to self: revert this. Having them equal doesn't make sense.

PR #1095, opened 2023-01-05
Add flake8 plugin `pep8-naming`

haakonvt:
## Description
Adds `pep8-naming` which extends flake8 with quite some nice rules for better consistency across the SDK: https://pypi.org/project/pep8-naming/

Only one was added to the ignore-list: N818, as too many existing exceptions violated it (and are not going to change): `CogniteException`, `CogniteConnectionRefused`, `CogniteReadTimeout`, `CogniteMultiException`, `ModelFailedException`

Copy of rule list:

Code | Sample message
-- | --
N801 | class names should use CapWords convention (class names)
N802 | function name should be lowercase (function names)
N803 | argument name should be lowercase (function arguments)
N804 | first argument of a classmethod should be named ‘cls’ (function arguments)
N805 | first argument of a method should be named ‘self’ (function arguments)
N806 | variable in function should be lowercase
N807 | function name should not start and end with ‘__’
N811 | constant imported as non constant (constants)
N812 | lowercase imported as non-lowercase
N813 | camelcase imported as lowercase
N814 | camelcase imported as constant (distinct from N817 for selective enforcement)
N815 | mixedCase variable in class scope (constants, method names)
N816 | mixedCase variable in global scope (constants)
N817 | camelcase imported as acronym (distinct from N814 for selective enforcement)
N818 | error suffix in exception names (exceptions)

### DIFF
--- a/cognite/client/_api/extractionpipelines.py
+++ b/cognite/client/_api/extractionpipelines.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Dict, Optional, Sequence, Union, overload
+from typing import Any, Dict, Optional, Sequence, Union, cast, overload
 
 from cognite.client import utils
 from cognite.client._api_client import APIClient
@@ -153,7 +153,10 @@ class ExtractionPipelinesAPI(APIClient):
                 >>> res = c.extraction_pipelines.create(extpipes)
         """
         # TODO: Remove support for old argument name in major version 6
-        extraction_pipeline = handle_deprecated_camel_case_argument(extraction_pipeline, "extractionPipeline", "create", kwargs)
+        extraction_pipeline = cast(
+            Union[ExtractionPipeline, Sequence[ExtractionPipeline]],
+            handle_deprecated_camel_case_argument(extraction_pipeline, "extractionPipeline", "create", kwargs),
+        )
         utils._auxiliary.assert_type(extraction_pipeline, "extraction_pipeline", [ExtractionPipeline, Sequence])
         return self._create_multiple(
             list_cls=ExtractionPipelineList, resource_cls=ExtractionPipeline, items=extraction_pipeline
--- a/cognite/client/_api/extractionpipelines.py
+++ b/cognite/client/_api/extractionpipelines.py
@@ -16,6 +16,7 @@ from cognite.client.data_classes import (
     TimestampRange,
 )
 from cognite.client.data_classes.extractionpipelines import StringFilter
+from cognite.client.utils._auxiliary import handle_deprecated_camel_case_argument
 from cognite.client.utils._identifier import IdentifierSequence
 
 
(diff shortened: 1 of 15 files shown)


cognite/client/_api/extractionpipelines.py
```diff
@@ -115,22 +116,26 @@ def list(self, limit: int = 25) -> ExtractionPipelineList:
...
-    def create(self, extractionPipeline: Sequence[ExtractionPipeline]) -> ExtractionPipelineList:
+    def create(
+        self, extraction_pipeline: Sequence[ExtractionPipeline], **kwargs: Dict[str, Any]
+    ) -> ExtractionPipelineList:
         ...
 
     def create(
-        self, extractionPipeline: Union[ExtractionPipeline, Sequence[ExtractionPipeline]]
+        self,
+        extraction_pipeline: Union[ExtractionPipeline, Sequence[ExtractionPipeline]],
```
erlendvollset, line 130: won't this break existing clients since `extraction_pipeline` is now a required argument?

cognite/client/data_classes/_base.py
```diff
@@ -443,8 +443,8 @@ def __getattribute__(self, item: Any) -> Any:
     @classmethod
     def _load(cls: Type[T_CogniteFilter], resource: Union[Dict, str]) -> T_CogniteFilter:
         if isinstance(resource, str):
-            return cls._load(json.loads(resource))
-        elif isinstance(resource, Dict):
+            resource = json.loads(resource)
```
erlendvollset, line 446: This doesn't get returned now?

cognite/client/data_classes/relationships.py
```diff
@@ -85,8 +85,8 @@ def _validate_resource_types(self) -> "Relationship":
 
     @staticmethod
     def _validate_resource_type(resource_type: Optional[str]) -> None:
-        _RESOURCE_TYPES = {"asset", "timeseries", "file", "event", "sequence"}
-        if resource_type is None or resource_type.lower() not in _RESOURCE_TYPES:
+        supported_res_types = {"asset", "timeseries", "file", "event", "sequence"}
```
erlendvollset, line 88: Since we're making a change here anyway, let's make this a class variable so it's only instantiated once.

cognite/client/data_classes/shared.py
```diff
@@ -130,8 +128,12 @@ class GeometryFilter(dict):
...
     """
 
-    def __init__(self, type: str, coordinates: List):
-        valid_types = ["Point", "LineString", "MultiLineString", "Polygon", "MultiPolygon"]
+    def __init__(
+        self,
+        type: Literal["Point", "LineString", "MultiLineString", "Polygon", "MultiPolygon"],
+        coordinates: List,
+    ):
+        valid_types = {"Point", "LineString", "MultiLineString", "Polygon", "MultiPolygon"}
```
erlendvollset, line 136: Same here, let's only instantiate this once.

erlendvollset (changes requested):

cognite/client/data_classes/_base.py
haakonvt: It does, since I also changed the `elif` below to `if`. But - I thought I had reversed this change. It dawned on me that a subclass might do something with the dict, but not a string, and so having this be recursive is probably correct. Will revert 👍

cognite/client/data_classes/relationships.py
haakonvt: Agreed 👍

cognite/client/_api/extractionpipelines.py
haakonvt: Oh, that's true, good catch! Btw, are you okey with these "camelCase depreations"? Or would you rather we `noqa`'d them?

haakonvt: Made optional in #29f8078

cognite/client/_api/extractionpipelines.py
```diff
@@ -117,22 +118,26 @@ def list(self, limit: int = 25) -> ExtractionPipelineList:
...
+    def create(
+        self, extraction_pipeline: Sequence[ExtractionPipeline], **kwargs: Dict[str, Any]
+    ) -> ExtractionPipelineList:
         ...
 
     def create(
-        self, extractionPipeline: Union[ExtractionPipeline, Sequence[ExtractionPipeline]]
+        self,
+        extraction_pipeline: Union[ExtractionPipeline, Sequence[ExtractionPipeline]] = None,
+        **kwargs: Dict[str, Any],
```
erlendvollset, line 133: Not a big fan of using `**kwargs` for this as it will silently swallow any additional named arguments. Why not this?
```
extraction_pipeline: Union[ExtractionPipeline, Sequence[ExtractionPipeline]] = None,
extractionPipeline: Union[ExtractionPipeline, Sequence[ExtractionPipeline]] = None
```

haakonvt: Agreed! …which is why I added `handle_deprecated_camel_case_argument`

haakonvt: (It raises and warns depending on usage and is reused for in all functions with deprecated parameter names 😄 )

erlendvollset: Ah, I missed the fact that it raises if it receives unexpected kwargs. Great! Let's add some unit tests for that function though.

haakonvt: Ah, yes, on it! 🙈

erlendvollset (approved):
Good stuff! 🏅

PR #1101, opened 2023-01-05
Add `yesqa` to pre-commit and run

haakonvt:
## Description
- Adds pre-commit check/plugin `yesqa`. It tries to find `# noqa` statements that are not needed.


erlendvollset (approved):
Nice. Didn't know about this one 💯

Issue #1108, opened 2023-01-10
Unstable GH workflows slows down development


haakonvt:
**Describe the bug**
```
Error: API rate limit exceeded for 172.xxxx. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

haakonvt:
Solved by #1109

PR #1109, opened 2023-01-10
Remove protoc-install step from workflows

haakonvt:
## Description
When more than just a few people push updates to PRs (that trigger our build workflow) in the same hour, we often get checks failing with `Error: API rate limit exceeded for 172.xxxx (...)`. The root cause is the step that installs the `protoc` compiler. After a quick inspection, I don't think it is necessary to check BOTH if proto-files have changed AND if the auto-generated python files from these have changed (which requires the `protoc` compiler).

Addresses #1108 .

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,9 +75,6 @@ jobs:
           poetry config virtualenvs.create false
           poetry install -E all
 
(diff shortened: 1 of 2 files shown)


erlendvollset (approved):
Sounds reasonable.

PR #1110, opened 2023-01-10
QoL changes in `data_classes/contextualization.py`

haakonvt:
## Description

1. Remove duplicate-but-different definitions of `ExternalId` and `InternalId` that are not subclasses of `Identifier`.
2. Use singleton property of enums
3. Change two uses of `__getattribute__` to `getattr`
4. Remove code that has been abstracted out but still left in (forgotten about I guess)
5. Write `@dataclass`-like `__str__` methods
6. Improve / fix docstrings
7. Narrowed two raised 'Exception's to 'ValueError's
8. Move information about what the function might raise to a proper Spinhx parseable "block"
9. Fix bug in function `_process_vertices` (correct dict with different insertion order than `x`-then-`y` would raise ValueError)

## Controversial? 😅 
10. Future-proofing (looking at you #1107 ): Change the dataclass `VisionExtractPredictions` to be made dynamically from `VISION_FEATURE_MAP` and not the other way around. Used to depend on type annotations being on an exact correct format, which isn't compatible with preferred annotations syntax in newer versions (or older, with `from __future__ import annotations`).

One way to do this:
```py
VISION_FEATURE_MAP: Dict[str, Any] = {
    "text_predictions": TextRegion,
    "asset_tag_predictions": AssetLink,
    "industrial_object_predictions": ObjectDetection,
    "people_predictions": ObjectDetection,
    "personal_protective_equipment_predictions": ObjectDetection,
}


class VisionExtractPredictions(VisionResource):
    __annotations__ = {param: f"Optional[List[{cls.__name__}]]" for param, cls in VISION_FEATURE_MAP.items()}


# Auto-generate the dataclass from the mapping VISION_FEATURE_MAP:
for attr in VISION_FEATURE_MAP:
    setattr(VisionExtractPredictions, attr, None)
VisionExtractPredictions = dataclass(VisionExtractPredictions)  # type: ignore [misc]
```

...and another:
```py
VISION_FEATURE_MAP: Dict[str, Any] = {
    "text_predictions": TextRegion,
    "asset_tag_predictions": AssetLink,
    "industrial_object_predictions": ObjectDetection,
    "people_predictions": ObjectDetection,
    "personal_protective_equipment_predictions": ObjectDetection,
}

# Auto-generate the dataclass from the mapping VISION_FEATURE_MAP:
@dataclass
class VisionExtractPredictions(VisionResource):
    __annotations__ = {param: f"Optional[List[{cls.__name__}]]" for param, cls in VISION_FEATURE_MAP.items()}
    for attr in VISION_FEATURE_MAP:
        locals()[attr] = None
```

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/contextualization.py
+++ b/cognite/client/data_classes/contextualization.py
@@ -533,6 +533,15 @@ class VisionFeature(str, Enum):
         return {cls.INDUSTRIAL_OBJECT_DETECTION, cls.PERSONAL_PROTECTIVE_EQUIPMENT_DETECTION}
 
 
+@dataclass
+class VisionExtractPredictions(VisionResource):
+    text_predictions: Optional[List[TextRegion]] = None
+    asset_tag_predictions: Optional[List[AssetLink]] = None
+    industrial_object_predictions: Optional[List[ObjectDetection]] = None
+    people_predictions: Optional[List[ObjectDetection]] = None
+    personal_protective_equipment_predictions: Optional[List[ObjectDetection]] = None
+
+
 VISION_FEATURE_MAP: Dict[str, Any] = {
     "text_predictions": TextRegion,
     "asset_tag_predictions": AssetLink,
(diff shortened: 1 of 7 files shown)


cognite/client/data_classes/contextualization.py
```diff
@@ -501,33 +511,25 @@ class VisionFeature(str, Enum):
...
+VISION_FEATURE_MAP: Dict[str, Any] = {
+    "text_predictions": TextRegion,
+    "asset_tag_predictions": AssetLink,
+    "industrial_object_predictions": ObjectDetection,
+    "people_predictions": ObjectDetection,
+    "personal_protective_equipment_predictions": ObjectDetection,
+}
+
+
+# Auto-generate the dataclass from the mapping VISION_FEATURE_MAP:
```
erlendvollset, line 527: I think this makes code completion engines useless for this class.

haakonvt: Probably 🪄  Did you see the old auto-generate version? It parses type hints that has to be on a very specific format (and that will break once we move to "future" versions of annotations). Should I just write them out in full in both places?

haakonvt: I'll write them both out in full, then make a test that ensures they have the same "members"

erlendvollset (approved):

PR #1117, opened 2023-01-18
Relax protobuf version from `^4.21` to `>=3.16.0`

haakonvt:
## Description
Ref [slack thread](https://cognitedata.slack.com/archives/CG10VQPFX/p1673970422103949), customers and internals wanting to upgrade to the SDK v5 has issues with the dependency `protobuf`. For example, Azure python function worker requires `protobuf` major version 3, while the SDK is pinned at `^4.21`. Same goes for a lot of popular python packages like `tensorflow` and `streamlit`. 

# ...story time! 🐇 
After a deep-dive down the rabbit hole, I found all kinds of fun, e.g. `protobuf` made a breaking change going from minor version `3.19.x` to `3.20`. Glad this was communicated well :trollface: .

Hurray, we can pin it to `3.20` or newer to support Azure & `streamlit`! "No", says `tensorflow`, which demands `<3.20`... but.. but... aren't both `tensorflow` and `protobuf` google libraries you say?? _Well, yes_. Their requirements file says "gonna need a quarted to update"... 

So back down the rabbit hole we go 🐰 . How do we get `protoc` to compile code for version `<3.20`? "Just brew install it", well, brew doesn't support installing an older version. Luckily, someone has made an older version available through `protobuf@3.6`. "No!" says homebrew, and adds: `"Error: protobuf@3.6 has been disabled because it is a versioned formula!"`. Then some confusion ensues, what is the major version after `3` for `protobuf`? `21`. Ay. _Twentyone_. That makes sense. Wonderful. Makes it so easy to guess which files to look for on the release page on Github. Finally, I get ahold of an older release. Unpacks and tries to run. "No" says my mac, and adds "I dont trust this executable, not one bit!". With some googling, I find out that the only way to allow it to run, is to find it in Finder, then ctrl+click it, choose "open" then "open" again with a very serious looking warning. I move it to `/usr/local/bin` and dumps a bunch in `/include/*`. 

Lets rock and roll. With the older `protoc` compiler ready, I generate the python code files from the `.proto` files then fire up some tests to see if the SDK still works. Yup. All good. What happens if I use the newer version (not major `v4`, but just `3.20` 😆 ) of `python-protobuf` with the "older" generated files. It fails. Of course it does. 😡 

Okey. Fine. I change from `>=3.16.0` to `>=3.16.0, <3.20`. All well and good now... let's run those tests again. OH WOULD YOU LOOK AT THAT, THEY RUN 5 TIMES SLOWER NOW!!!! What!! theeee f...... 🤦 

At this point, I decided, I need to rant about this. You are welcome. 

The story goes on from here, but to sum it up, these older versions of `protobuf` for python don't necessarily bundle compiled binaries (C code) when you pip install them. Thus the wonderful features of "faster than `JSON`" is now "much slower than `JSON`". Really happy we made the change.

What do we do with this info? We throw both old and new proto files into the SDK and make a nice, big ⚠️ warning ⚠️ to the user if their `protobuf` installation is missing binaries and is thus running at 🐌 -speed (aka "pure python" mode).

...and yes, it doesn't look pretty. No, no, I really don't care.

## Checklist:
- [ ] Tests added/updated.


cognite/client/_api/datapoints.py
```diff
@@ -117,6 +127,23 @@ def __init__(
...
         self.n_queries = len(all_queries)
 
+        # Fetching datapoints relies on protobuf, which, depending on OS and major version used
+        # might be running in pure python or compiled C code. We issue a warning if we can determine
+        # that the user is running in pure python mode (quite a bit slower...)
+        with contextlib.suppress(ImportError):
+            from google.protobuf.descriptor import _USE_C_DESCRIPTORS  # type: ignore [attr-defined]
+
+            if _USE_C_DESCRIPTORS is False:
+                warnings.warn(
```
erlendvollset, line 137: nice.

erlendvollset (approved):
LGTM

PR #1120, opened 2023-01-24
File Not Found OAuthInteractive Windows

### LINKED ISSUE #1116: OAuthInteractive as client provider does not work on windows
**System information (please complete the following information):**
 - OS: Windows 11
 - Python Version: Python 3.9.12
 - SDK Version: latest

**Describe the bug**
OAuthInteractive as client provider does not work on windows. Seems like the problem is this line, which only makes sense on linux/mac: https://github.com/cognitedata/cognite-sdk-python/blob/75cc08c491a5c588809e41ee279d7fbcf973a19b/cognite/client/credentials.py#L202 

I would suggest changing it to e.g. 
```python
Path(tempfile.gettempdir()) / f"cognitetokencache.{self.__client_id}.bin"
```

**To Reproduce**
Runnable code reproducing the error.
```python
from cognite.client import CogniteClient, ClientConfig
from cognite.client.credentials import OAuthInteractive
import os 

authority_url = os.getenv("COGNITE_AUTHORITY_URL")
client_id = os.getenv("COGNITE_CLIENT_ID")
project = os.getenv("COGNITE_PROJECT")
oauth_provider = OAuthInteractive(
    authority_url=authority_url,
    client_id=client_id,
    scopes=["https://api.cognitedata.com/.default"],
conf = ClientConfig(client_name="my_super_client", base_url="https://api.cognitedata.com", project=project, credentials=oauth_provider)
client = CogniteClient(config=conf)

client.assets.list()
```
**Expected behavior**
It should open an interactive login prompt.

**Screenshots**
image

doctrino:
## Description  
closes #1116 

**Problem**
When running the interactive login workflow on windows I get the following exception

```python
Exception ignored in atexit callback: <function _WithMsalSerializableTokenCache._create_serializable_token_cache.<locals>.__at_exit at 0x0000014A704B4D60>
  File "...\.venv\Lib\site-packages\cognite\client\credentials.py", line 108, in __at_exit
    with open(cache_path, "w") as fh:
         ^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '\\tmp\\cognitetokencache.123456789.bin'
```
**Solution**  
By allowing to pass in the path for the token cache, I can ensure that the folder exists.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [5.3.3] - 20-01-24
+### Changed
+- Allowed to pass in `token_cache_path` in `OAuthInteractive`.
+
 ## [5.3.2] - 20-01-24
 ### Changed
 - Update pytest and other dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Changes are grouped as follows
 
 ## [5.3.3] - 20-01-24
 ### Changed
-- Allowed to pass in `token_cache_path` in `OAuthInteractive`.
+- Allowed to pass in `token_cache_path` in `OAuthInteractive` and `OAuthDeviceCode`.
 
 ## [5.3.2] - 20-01-24
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ Changes are grouped as follows
 ### Changed
 - Allowed to pass in `token_cache_path` in `OAuthInteractive` and `OAuthDeviceCode`.
 
+### Fixed
+- Platform independent temp directory.
+
 ## [5.3.2] - 20-01-24
 ### Changed
 - Update pytest and other dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ Changes are grouped as follows
 - Allowed to pass in `token_cache_path` in `OAuthInteractive` and `OAuthDeviceCode`.
 
 ### Fixed
-- Platform independent temp directory.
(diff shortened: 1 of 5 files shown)


doctrino:
> How about instead changing the file mode to `w+` here, so the file is created if it doesn't exist? https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/credentials.py#L110

Agree, that is a good idea, but I also suggest keeping the extra parameter, as I prefer to control the location of the cache token.

erlendvollset:
SGTM. Could you add it to the DeviceLoginFlow aswell then - that also uses a persistent cache?

doctrino:
> d you add it to the DeviceLoginFlow aswell then - that also uses a persistent cache?

will do. Tested the solution above with w+, does not work, which I guess is because the folder is missing.

Auto create the folder if it is missing? I think it is best as many of our customers are on windows, and need as much help as possible.

erlendvollset:
Or looks like you can use tempfile.gettempdir() as a platform-agnostix approach.

doctrino:
> Or looks like you can use tempfile.gettempdir() as a platform-agnostic approach.

Yes I found the same thing

CHANGELOG.md
```diff
@@ -17,6 +17,13 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [5.3.3] - 20-01-24
+### Changed
+- Allowed to pass in `token_cache_path` in `OAuthInteractive` and `OAuthDeviceCode`.
+
+### Fixed
+- Platform independent temp directory.
```
doctrino, line 25:
```suggestion
- Platform independent temp directory for the caching of the token in tempfile.gettempdir().
```

erlendvollset (approved):
Good stuff!

PR #1122, opened 2023-01-25
Fix thread shutdown on PY39+ (fix #1121)

### LINKED ISSUE #1121: `atexit` raises `TypeError` after multiple calls to `client.datapoints` on PY>=39
Using SDK >= 5

```py
Exception ignored in: <module 'threading' from '/Users/foo-bar/.pyenv/versions/3.10.7/lib/python3.10/threading.py'>
  File "/Users/foo-bar/.pyenv/versions/3.10.7/lib/python3.10/threading.py", line 1537, in _shutdown
    atexit_call()
  File "/Users/foo-bar/.pyenv/versions/3.10.7/lib/python3.10/concurrent/futures/thread.py", line 29, in _python_exit
    q.put(None)
  File "/Users/foo-bar/.pyenv/versions/3.10.7/lib/python3.10/queue.py", line 150, in put
    self._put(item)
  File "/Users/foo-bar/.pyenv/versions/3.10.7/lib/python3.10/queue.py", line 236, in _put
    heappush(self.queue, item)
TypeError: '<' not supported between instances of 'NoneType' and 'tuple'
Exception ignored in atexit callback: <function python_exit at 0x10a8cbd90>
  File "/Users/foo-bar/.pyenv/versions/3.10.7/lib/python3.10/site-packages/cognite/client/utils/_priority_tpe.py", line 52, in python_exit
    queue.put(NULL_ENTRY)
  File "/Users/foo-bar/.pyenv/versions/3.10.7/lib/python3.10/queue.py", line 150, in put
    self._put(item)
  File "/Users/foo-bar/.pyenv/versions/3.10.7/lib/python3.10/queue.py", line 236, in _put
    heappush(self.queue, item)
TypeError: '<' not supported between instances of 'tuple' and 'NoneType'
```

haakonvt:
## Description
Starting in 3.9, ThreadPoolExecutor no longer uses daemon threads, but instead, an internal function hook very similar to atexit.register(), which gets called at 'threading' shutdown instead of interpreter shutdown. Check out https://github.com/python/cpython/issues/83993

Some links:

- https://www.andy-pearce.com/blog/posts/2022/Nov/whats-new-in-python-39-library-changes/#concurrent-execution 
- https://github.com/python/cpython/issues/83993
- https://github.com/python/cpython/pull/19149

This sounds like an argument for testing multiple Python versions (with `tox` perhaps)? 🤔 

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/utils/_priority_tpe.py
+++ b/cognite/client/utils/_priority_tpe.py
@@ -65,7 +65,6 @@ else:
     # function hook very similar to atexit.register(), which gets called at 'threading' shutdown
     # instead of interpreter shutdown. Check out https://github.com/python/cpython/issues/83993
     for i, exit_fn in enumerate(_threading_atexits):
-        print(i, exit_fn, exit_fn.func, _python_exit)
         if exit_fn.func is _python_exit:
             # TODO: Does this have unwanted side effects? If not, why no threading._unregister_atexit?
             _threading_atexits.pop(i)  # Note: Remove inplace
--- a/cognite/client/utils/_priority_tpe.py
+++ b/cognite/client/utils/_priority_tpe.py
@@ -1,5 +1,5 @@
 """
-This code has been modified from the original created by Oleg Lupats, 2019, under an MIT license:
+This code has been heavily modified from the original created by Oleg Lupats, 2019, under an MIT license:
 project = 'PriorityThreadPoolExecutor'
 url = 'https://github.com/oleglpts/PriorityThreadPoolExecutor'
 copyright = '2019, Oleg Lupats'
@@ -33,18 +33,18 @@ from __future__ import annotations
 import inspect
 import itertools
 import sys
-import threading
 import weakref
 from concurrent.futures import Future
 from concurrent.futures.thread import ThreadPoolExecutor, _base, _WorkItem
-from queue import PriorityQueue
+from queue import Empty, PriorityQueue
+from threading import Lock, Thread
 
-NULL_ENTRY = (-1, None, _WorkItem(None, None, (), {}))
+NULL_ENTRY = (-1, None, None)
 _THREADS_QUEUES = weakref.WeakKeyDictionary()
 _SHUTDOWN = False
 # Lock that ensures that new workers are not created while the interpreter is
 # shutting down. Must be held while mutating _THREADS_QUEUES and _SHUTDOWN.
-_GLOBAL_SHUTDOWN_LOCK = threading.Lock()
+_GLOBAL_SHUTDOWN_LOCK = Lock()
 
 
 def python_exit():
@@ -72,8 +72,8 @@ _register_atexit(python_exit)
 def _worker(executor_reference, work_queue):
     try:
         while True:
-            priority, _, work_item = work_queue.get(block=True)
-            if priority != -1:
+            work_item = work_queue.get(block=True)[-1]
+            if work_item is not None:
                 work_item.run()
                 del work_item
                 continue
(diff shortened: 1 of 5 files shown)


cognite/client/utils/_priority_tpe.py
```diff
@@ -54,8 +53,23 @@ def python_exit():
...
 
 
-atexit.unregister(_python_exit)
-atexit.register(python_exit)
+if sys.version_info[:2] < (3, 9):
+    import atexit
+
+    atexit.unregister(_python_exit)
+    atexit.register(python_exit)
+else:
```
erlendvollset, line 61: A comment explaining why we need the custom shutdown hook at all would be good.

erlendvollset (approved):
I don't have enough context here to give a proper review, but LGTM. Maybe @doctrino wants to have a look - I believe he added this module initially?

cognite/client/utils/_priority_tpe.py
erlendvollset: _Do_ we actually need it?

haakonvt: I've been down the rabbit hole now and yes, we need it, but also _there is a much better solution_ 😂

haakonvt:
~~~~~~~~
Edit: I was too fast. Need more time to investigate this 🤔  (base class uses the global variable `_threads_queues`, which makes it not trivial to separate in our tpe subclass)

haakonvt: I was the one to add the module btw, saw your comment just now. 

One thing that has bothered me with this approach is that we use the regular TPE extensively in all other parallelized calls than datapoints... and thus unregistering the `atexit`* handler seems plain wrong. The PTPE implementation we used was actually quite intertwined with TPE/threading etc. and was a replacement not really able to co-exist... 

Solution: Implement a PTPE-only `_THREADS_QUEUES` and register a separate, additional `atexit`*-handler for it.

(*`atexit` might be "atexit" module or `threading._register_atexit` depending on PY38 or above)

haakonvt: **TL;DR**: Now any usage of `ThreadPoolExecutor` together with `PriorityThreadPoolExecutor` is no longer an issue as both have separate cleanup functions registered, without any overlapping reference containers (for threads).

haakonvt:
This PR is probably a good change either way, but I found what seems to be the root cause of all this; `Future`s with an exception (i.e. ts not found) causes it to not be cleaned up and thus after the change from `atexit` to `threading....` the ref count is keeping it from being garbage collected. I added a super simple `future._exception = None` and 🪄 problem gone. 

I'll investigate some more tomorrow @erlendvollset

cognite/client/utils/_priority_tpe.py
```diff
@@ -30,32 +30,49 @@
...
+    items = list(_THREADS_QUEUES.items())
     for thread, queue in items:
         queue.put(NULL_ENTRY)
     for thread, queue in items:
         thread.join()
 
 
-atexit.unregister(_python_exit)
-atexit.register(python_exit)
+# We should register an at-exit-handler to clean up our threads gracefully at interpreter shutdown.
```
erlendvollset, line 63: I don't really think there is a need for "cleaning up the threads gracefully". Essentially you're just waiting for work to finish, and then tossing the results out, so the only thing you achieve is slower shutdown of the interpreter. Unless I'm missing something here, I think we can remove the atexit hook altogether.

haakonvt: With current tasks, yes, however, that gives us a "dependency" on the future use of the PTPE to never have tasks with side effects; not something I would want to perpetually make sure is the case.

haakonvt: I've gone down the deep rabbit hole here and I really think this implementation is the way way to do it. Separate thread pools, separate thread-safe queues, and separate cleanup functions registered that take the python version into account! 🍀

erlendvollset: Fair point - let's go for this approach then! 🤝

haakonvt: Lol YAGNI, learned something new today 😂 Btw, if you really want it removed why don’t we also remove from the standard library by monkeypatching? 😜 I feel good knowing we mimic the exact same approach as some code written by people way smarter than myself 😂

haakonvt: Also; the code do crash without it 😉

PR #1129, opened 2023-01-30
remove `tool.poetry.source` from `pyproject.toml`

haakonvt:
## Afaik, we don't need this 😄 
@erlendvollset this is mostly a question to you! Let me know if I'm wrong (just close if so!)

## Why remove?
It slows down development and hinders contributions (jfrog accounts love being blocked for some reason... 🤷 )


erlendvollset (approved):
Yeah that's not needed. Probably just a copy-paste blunder

PR #1130, opened 2023-01-30
Make SDK `pyodide`/browser compatible out-of-the-box

haakonvt:
The first file any import into `cognite.client` hits is `cognite.client.__init__`. Thus, this is where we add our patching function to make sure we can, before the user actually imports anything, monkey-patch all the corners of the SDK to make it `pyodide` compatible (i.e. running in a browser in WebAssembly).

This allows us also to make authenticating the `CogniteClient` ridiculously easy if-and-only-if we pick up that the user is running in Fusion (where we pass all necessary auth details for the user to not worry about anything):
```py
client = CogniteClient()
```
...the magic trick is executed as part of patching by:
```py
if os.getenv("COGNITE_FUSION_NOTEBOOK") is not None:
    cc.config.global_config.default_client_config = FusionNotebookConfig()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [5.5.3] - 16-02-23
+### Added
+- The SDK has been made `pyodide` compatible (to allow running natively in browsers). Missing features are `CredentialProvider`s with token refresh and `AssetsAPI.create_hierarchy`.
+
 ## [5.5.2] - 15-02-23
 ### Fixed
 - Fixed JSON dumps serialization error of instances of `ExtractionPipelineConfigRevision` and all subclasses (`ExtractionPipelineConfig`) as they stored a reference to the CogniteClient as a non-private attribute.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [5.5.3] - 16-02-23
+## [5.6.0] - 16-02-23
 ### Added
 - The SDK has been made `pyodide` compatible (to allow running natively in browsers). Missing features are `CredentialProvider`s with token refresh and `AssetsAPI.create_hierarchy`.
 
--- a/cognite/client/utils/_pyodide_helpers.py
+++ b/cognite/client/utils/_pyodide_helpers.py
@@ -9,14 +9,14 @@ import warnings
 from base64 import urlsafe_b64decode
 from concurrent.futures import CancelledError
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterator, Optional
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterator, Optional, Tuple
 
 from msal import PublicClientApplication
 
 import cognite.client as cc
 from cognite.client._http_client import _RetryTracker
 from cognite.client.config import ClientConfig
-from cognite.client.credentials import OAuthInteractive
+from cognite.client.credentials import CredentialProvider, OAuthInteractive
 from cognite.client.utils._concurrency import T_Result, TaskExecutor, TaskFuture
 
 if TYPE_CHECKING:
@@ -42,8 +42,15 @@ def patch_sdk_for_pyodide() -> None:
 
     # -----------------
     # Patch Cognite SDK
+    # - For good measure ;)
+    os.environ["COGNITE_DISABLE_PYPI_VERSION_CHECK"] = "1"
+
     # - Disable gzip, not supported:
     cc.config.global_config.disable_gzip = True
+
+    # - Use another HTTP adapter:
+    cc._http_client.HTTPClient._old__init__ = cc._http_client.HTTPClient.__init__  # type: ignore [attr-defined]
+    cc._http_client.HTTPClient.__init__ = http_client__init__  # type: ignore [assignment]
 
     # - If we are running inside of a JupyterLite Notebook spawned from Cognite Data Fusion, we set
     #   the default config to FusionNotebookConfig(). This allows the user to:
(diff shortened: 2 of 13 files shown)


cognite/client/utils/_pyodide_helpers.py
```diff
@@ -0,0 +1,230 @@
...
+
+    # - DatapointsAPI.retrieve/retrieve_arrays/retrieve_dataframe:
+    cc._api.datapoints.PriorityThreadPoolExecutor = SerialPriorityThreadPoolExecutor  # type: ignore [assignment, misc]
+    cc._api.datapoints.as_completed = serial_as_completed  # type: ignore [assignment]
+
+
+def pyodide_http_adapter_send(self: Any, request: PreparedRequest, **kwargs: Any) -> Response:
+    # 'self' is actually `pyodide_http._requests.PyodideHTTPAdapter`, but we don't want it as a
+    # dependency just for the sake of mypy; after all, we are patching like monkeys 🙈
+    (response := self._old_send(request, **kwargs)).raw.version = ""
```
andeplane, line 87: I think this can be simplified by changing how we test for raw.version instead of patching this.

haakonvt: I find one usage of it in `cognite/client/_api_client.py`:
`http_protocol_version = ".".join(list(str(res.raw.version)))`

It is used to print out a debug message, i.e. `HTTP/{http_protocol_version}` -> `HTTP/1.1`. Let's say I change it to a try-except, which protocol / what message should be logged instead?

haakonvt: I'll push a change, then feel free to correct the wording correctness 😉

cognite/client/utils/_pyodide_helpers.py
```diff
@@ -0,0 +1,218 @@
...
+    def __enter__(self) -> SerialPriorityThreadPoolExecutor:
+        return self
+
+    def __exit__(
+        self,
+        type: Optional[type[BaseException]],
+        value: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> None:
+        self.shutdown()
```
erlendvollset, line 141: I would suggest moving these into the concurrency module along with the PriorityThreadPoolExecutor, and toggling their usage with the same ConcurrencySettings option. Makes it easier to follow if it's done the same way for the different implementations.

cognite/client/utils/_pyodide_helpers.py
```diff
@@ -0,0 +1,218 @@
...
+        self._OAuthInteractive__redirect_port = redirect_port
+        self._OAuthInteractive__app = PublicClientApplication(client_id=client_id, authority=authority_url)
+
+    @staticmethod
+    def _load_payload_from_jwt(token: str) -> Dict[str, Any]:
+        """A JWT has the structure: `<header>.<payload>.<signature>`. This method returns payload as a dict"""
+        _, payload, _ = token.split(".")
+        if remainder := len(payload_bytes := payload.encode()) % 4:
+            payload_bytes += b"=" * (4 - remainder)
+        return json.loads(urlsafe_b64decode(payload_bytes))
```
erlendvollset, line 188: IMO if you're using the variable outside the scope of the if-statement, it's easier to follow if you define it upfront.
```suggestion
        payload_bytes = payload.encode()
        if remainder := len(payload_bytes) % 4:
            payload_bytes += b"=" * (4 - remainder)
        return json.loads(urlsafe_b64decode(payload_bytes))
```

cognite/client/utils/_pyodide_helpers.py
```diff
@@ -0,0 +1,218 @@
...
+        # Note: Death to name mangling:
+        self._OAuthCredentialProviderWithTokenRefresh__access_token = token
+        self._OAuthCredentialProviderWithTokenRefresh__token_refresh_lock = contextlib.nullcontext()
+        self._OAuthCredentialProviderWithTokenRefresh__access_token_expires_at = payload["exp"]
+
+        self._OAuthInteractive__authority_url = authority_url
+        self._OAuthInteractive__client_id = client_id
+        self._OAuthInteractive__scopes = [f"{base_url}/.default"]
+        self._OAuthInteractive__redirect_port = redirect_port
+        self._OAuthInteractive__app = PublicClientApplication(client_id=client_id, authority=authority_url)
```
erlendvollset, line 180: Not a big fan of this. 

Someone who goes and changes the implementation of OAuthCredentialProviderWithTokenRefresh or OAuthInteractive won't necessarily be aware of this class accessing their private attributes and making assumptions on their implementations. In this case I would say it's better to just copy the whole implementation into this module. 

Especially considering that we don't have tests covering this which would catch this thing breaking.

cognite/client/utils/_pyodide_helpers.py
andeplane: I think that was the one that crashed (hence the fix), yep

cognite/client/utils/_pyodide_helpers.py
haakonvt: Removed altogether while we wait on a solution for proper token refresh handling in coop with Fusion

cognite/client/utils/_pyodide_helpers.py
haakonvt: Entire class removed, see other comment

cognite/client/utils/_pyodide_helpers.py
haakonvt: Moved what could be moved without too much trouble. Patching now looks unified:

```py
# - Set all usage of thread pool executors to use dummy/serial-implementations:
cc.utils._concurrency.ConcurrencySettings.executor_type = "mainthread"
cc.utils._concurrency.ConcurrencySettings.priority_executor_type = "mainthread"
```

Skipped: Unable to use the same singleton pattern due to various logic calling on the pool to shut down explicitly. It seems like a significant risk to change that logic atm.

haakonvt: Also moved the usage of `concurrent.future.as_completed` to a pool method, that way we can forget about patching it 😄

haakonvt:
@erlendvollset is this one good to go? 😄

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [5.5.3] - 16-02-23
```
erlendvollset, line 20: Adding new functionality merits a minor bump
```suggestion
## [5.6.0] - 16-02-23
```

erlendvollset (approved):
Approved! But I think a minor bump is more appropriate.

PR #1131, opened 2023-01-31
abstract out `limit`-checks

haakonvt:
## Description
8 variances of "is limit `None`, -1 or `math.inf`" replaced with a utility function.

### DIFF
--- a/cognite/client/utils/_auxiliary.py
+++ b/cognite/client/utils/_auxiliary.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import functools
 import heapq
 import importlib
+import math
 import numbers
 import platform
 import random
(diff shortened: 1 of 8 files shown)


cognite/client/utils/_auxiliary.py
```diff
@@ -28,6 +29,10 @@
 THashable = TypeVar("THashable", bound=Hashable)
 
 
+def is_unlimited(limit: Any) -> bool:
```
erlendvollset, line 32:
```suggestion
def is_unlimited(limit: Optional[Union[float, int]]) -> bool:
```

haakonvt: Btw, what do you think about PEP-484 regarding "Union[float, int]"?

> Rather than requiring that users write import numbers and then use numbers.Float etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type float, an argument of type int is acceptable

erlendvollset: Yeah Optional[float] is sufficient

erlendvollset (approved):
Schemood 💯

PR #1133, opened 2023-01-31
Move validation of `token_custom_args` in `OAuthClientCredentials`

haakonvt:
## Description
Since SDK version `4.0.1` (#1003 ) we have accepted extra parameters to be included in the token request (passed along to the `fetch_token` function as `token_custom_args`). In the original (and current) implementation, in order to make sure none of these extra parameters conflict with a named parameter accepted by `OAuth2Session.fetch_token`, all were passed explicitly, with current values equal to their defaults, unless part of the few we actually need to pass (e.g. client ID + secret).

This PR future-proofs this validation process: Both the actual parameter names _and_ their default values might change in future versions of `requests_oauthlib` _and_ new ones might be added (/ removed). This would require manual inspection of `OAuth2Session.fetch_token` from time to time to ensure we are up-to-date _or_ an unnecessarily narrowly pinned dependency. Both options are undesirable. 

It also greatly improves the error message when a conflict arises:
```py
# Before
TypeError: fetch_token() got multiple values for keyword argument 'authorization_response'
# After
TypeError: The following reserved token custom arg(s) were passed: ['authorization_response']. 
    The full list of reserved custom args is: ['auth', 'authorization_response', 'body', 'cert', 'client_secret',
    'code', 'force_querystring', 'headers', 'include_client_id', 'method', 'password', 'proxies', 'timeout',
    'token_url', 'username', 'verify'].
```
...and it also moves the validation "earlier"; previously passing one or more reserved parameters would first raise when an actual request was made. Now, it raises during the instantiation of `OAuthClientCredentials`. 

### Additionally
Removed `client_id` from the passed args to `fetch_token` as the session already knows it. Did the same with `scopes` after making the change of passing it directly to `BackendApplicationClient` (in `OAuth2Session`).

### DIFF
--- a/cognite/client/credentials.py
+++ b/cognite/client/credentials.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import atexit
+import inspect
 import tempfile
 import threading
 from abc import abstractmethod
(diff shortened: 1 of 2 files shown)


erlendvollset (approved):
LGTM

PR #1134, opened 2023-02-01
Add support for aggregating metadata keys/values in AssetsAPI

erlendvollset:
## Description

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/assets.py
+++ b/cognite/client/_api/assets.py
@@ -302,7 +302,7 @@ class AssetsAPI(APIClient):
             partitions=partitions,
         )
 
-    def aggregate(self, filter: Union[AssetFilter, Dict] = None) -> List[AssetAggregate]:
+    def aggregate(self, filter: Union[AssetFilter, dict] = None) -> List[AssetAggregate]:
         """`Aggregate assets <https://docs.cognite.com/api/v1/#operation/aggregateAssets>`_
 
         Args:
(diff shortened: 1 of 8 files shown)


cognite/client/_api/assets.py
```diff
@@ -320,6 +321,47 @@ def aggregate(self, filter: Union[AssetFilter, Dict] = None) -> List[AssetAggreg
...
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> aggregate_by_prefix = c.assets.aggregate_metadata_keys(filter={"external_id_prefix": "prefix"})
+        """
+        return self._aggregate(filter=filter, aggregate="metadataKeys", cls=AggregateBucketResult)
+
+    def aggregate_metadata_values(
+        self, keys: Sequence[str], filter: Union[AssetFilter, Dict] = None
+    ) -> Sequence[AggregateBucketResult]:
```
haakonvt, line 345: Seems to return values case folded..? Could be quite surprising as the official documentation doesn't even mention it: https://docs.cognite.com/api/v1/#tag/Assets/operation/aggregateAssets

...however the docs under `events/aggregate` mention it explicitly:

_Also, in the case of text fields, the values are aggregated in a case-insensitive manner_

I think we should mention this in the docstring.

Screenshot 2023-02-01 at 22 19 31

haakonvt (commented):
A few observations after playing around with the API

cognite/client/data_classes/shared.py
```diff
@@ -34,8 +34,6 @@ def __init__(self, count: int = None, **kwargs: Any):
         self.count = count
         self.update(kwargs)
 
-    count = CognitePropertyClassUtil.declare_property("count")
```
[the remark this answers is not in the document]
erlendvollset: Good catch. The `declare_property` util is a remnant from when we autogenerated data classes, so I'd like to get rid of those. But we should pass the count argument to the dict's constructor, so that the value ends up in there too.

cognite/client/_api/assets.py
erlendvollset: Sad, but not something we should fix in the SDK. I'll create a ticket for searchberg on this.

erlendvollset: (also a good catch 💯)

cognite/client/data_classes/shared.py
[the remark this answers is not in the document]
erlendvollset: and no idea why the API returns both value and values.. I'll create a ticket on that too. But these aggregation endpoints will be replaced soon anyway, so I don't really think any of this will be fixed.

cognite/client/_api/assets.py
erlendvollset: And I'll update the docstring to mention this also 🤝

Issue #1112, opened 2023-01-11
Flaky test: `TestAssetPosterWorker.test.run`


haakonvt:
**Describe the bug**
In our build workflow, the test `TestAssetPosterWorker.test.run` may randomly fail (probably due to the magic and unpredictable world of threads...)

**Expected behavior**
Not be flaky.

**Screenshots**
<img width="989" alt="Screenshot 2023-01-11 at 15 17 46" src="https://user-images.githubusercontent.com/8521241/211831083-255877af-aaa9-4e64-aa35-d8b4f6cb5841.png">

haakonvt:
<img width="846" alt="image" src="https://user-images.githubusercontent.com/8521241/213824798-d0ea501e-1b2b-4d9e-a272-56a320e50c16.png">

erlendvollset:
Closing this as we think we've addressed the issue. Please reopen if the issue persists.

PR #1135, opened 2023-02-01
Fix bug in assets.create_hierarchy

erlendvollset:
Fix bug where AssetPosterWorker stops doing work after encountering an exception.

When a worker encountered an exception, it would just stop doing work. This means that 1) the method becomes slower as it encounters errors and 2) if it encounters more errors than there are workers, it would just stop progressing entirely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [5.4.1] - 02-02-23
+### Fixed
+- Bug where create_hierarchy would stop progressing after encountering more than `config.max_workers` failures.
+
 ## [5.4.0] - 02-02-23
 ### Added
 - Support for aggregating metadata keys/values for assets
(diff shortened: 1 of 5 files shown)


dmivankov (approved):

haakonvt:
The test `TestAssetPosterWorker.test_run` is flaky (#1112 ), and I suspect upping `time.sleep(0.1)` a tiny bit is enough. Maybe you could piggyback such a change on this PR 😄 

```py
class TestAssetPosterWorker:
    def test_run(self, cognite_client, mock_assets_response):
        q_req = queue.Queue()
        q_res = queue.Queue()

        w = _AssetPosterWorker(request_queue=q_req, response_queue=q_res, client=cognite_client.assets)
        w.start()
        q_req.put([Asset()])
        time.sleep(0.1)
        w.stop = True
        assert [Asset._load(mock_assets_response.calls[0].response.json()["items"][0])] == q_res.get()
        assert 1 == len(mock_assets_response.calls)
```

erlendvollset:
Will do 👌

haakonvt (approved):
King Kong! 🦍

PR #1136, opened 2023-02-02
Make `FunctionsAPI` use  `IAM.SessionsAPI` (remove separate implementation)

haakonvt:
### As posted on slack:
Hey there! I’m unable to call a function (it fails with `status="Failed"` and log message: `Could not fetch a valid token or a valid API key. Unable to instantiate a CogniteClient. Please verify your credentials`), depending on what type of `CredentialProvider` (from `cognite.client.credentials` ) I have used to instantiate my `CogniteClient`.

- OAuthClientCredentials ✅ 
- OAuthInteractive ❌ 
- OAuthDeviceCode ❌ 
- Token ✅ 
- User-made subclasses of `CredentialProvider` ❌ 

(Note: _All of these credentials are using the exact same app registration with all the required acls_)
Upon inspecting the code, it seems you only do an `insinstance`-check (on the ✅  ones only) and call the `SessionsAPI` _directly_ instead of using `client.iam.sessions[.create]` . I have ~~started a draft~~ [PR](https://github.com/cognitedata/cognite-sdk-python/pull/1136) to remove this “additional API implementation” and reduce coupling in the code base; a user is allowed to create his own `CredentialProvider` and that should work with Functions ootb imo 😄 

### Changes

- This PR makes both `OAuthInteractive`, `OAuthDeviceCode` (and any custom subclass impl.) "turn green" ✅ 
- The error message no longer includes whether "token exchange" or "client credentials" were used, but I think that's okay since we get the _actual_ error from the API instead. Probably that has some more concrete feedback to the user about what went wrong.
- Passing client credentials as a `ClientCredentials` object is now supported (extending the current behavior of passing a `dict` with two required keys)

## ~~TODO:~~
~~Tests are failing~~

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Changes are grouped as follows
 
 ## [5.9.1] - 20-03-23
 ### Fixed
-- `FunctionsAPI.call` now also works for client using auth flow `OAuthInteractive`, `OAuthDeviceCode` and any user-made subclass of `CredentialProvider`.
+- `FunctionsAPI.call` now also works for clients using auth flow `OAuthInteractive`, `OAuthDeviceCode`, and any user-made subclass of `CredentialProvider`.
 
 ### Improved
 - `FunctionSchedulesAPI.create` now also accepts an instance of `ClientCredentials` (used to be dictionary only).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,13 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [5.9.1] - 20-03-23
+### Fixed
+- `FunctionsAPI.call` now also works for client using auth flow `OAuthInteractive`, `OAuthDeviceCode` and any user-made subclass of `CredentialProvider`.
+
+### Improved
+- `FunctionSchedulesAPI.create` now also accepts an instance of `ClientCredentials` (used to be dictionary only).
+
 ## [5.9.0] - 20-03-23
 ### Added
 - New class `AssetHierarchy` for easy verification and reporting on asset hierarchy issues without explicitly trying to insert them.
--- a/cognite/client/_api/functions.py
+++ b/cognite/client/_api/functions.py
@@ -15,8 +15,9 @@ from zipfile import ZipFile
 from cognite.client import utils
 from cognite.client._api_client import APIClient
 from cognite.client._constants import LIST_LIMIT_CEILING, LIST_LIMIT_DEFAULT
-from cognite.client.credentials import OAuthClientCredentials, Token
+from cognite.client.credentials import APIKey
 from cognite.client.data_classes import (
+    ClientCredentials,
     Function,
     FunctionCall,
     FunctionCallList,
@@ -31,7 +32,6 @@ from cognite.client.data_classes import (
 )
 from cognite.client.data_classes.files import FileMetadata
 from cognite.client.data_classes.functions import FunctionCallsFilter, FunctionsStatus
-from cognite.client.exceptions import CogniteAPIError
 from cognite.client.utils._auxiliary import is_unlimited
 from cognite.client.utils._identifier import IdentifierSequence, SingletonIdentifierSequence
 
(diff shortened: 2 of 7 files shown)


ngerstle-cognite (commented):
Does this remove support for using token exchange based sessions with functions in the python sdk?

haakonvt:
> Does this remove support for using token exchange based sessions with functions in the python sdk?

@ngerstle-cognite quite the opposite in fact. With this change, all `CredentialProvider`s work, instead of just the 2 specific classes they checked for in the old code 😄

cognite/client/_api/functions.py
```diff
@@ -496,14 +489,15 @@ def _assert_exactly_one_of_folder_or_file_id_or_function_handle(
         folder: Optional[str], file_id: Optional[int], function_handle: Optional[Callable[..., Any]]
     ) -> None:
         source_code_options = {"folder": folder, "file_id": file_id, "function_handle": function_handle}
-        given_source_code_options = [key for key in source_code_options.keys() if source_code_options[key]]
+        given_source_code_options = [k for k, v in source_code_options.items() if v is not None]
         if len(given_source_code_options) < 1:
-            raise TypeError("Exactly one of the arguments folder, file_id and handle is required, but none were given.")
```
haakonvt, line 501: Note: `handle` is the wrong parameter name, thus changed to `function_handle`

CHANGELOG.md
```diff
@@ -17,6 +17,13 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [5.7.4] - 18-03-23
+### Fixed
+- `FunctionsAPI.call` now also works for client using auth flow `OAuthInteractive`, `OAuthDeviceCode` and any user-made subclass of `CredentialProvider`.
```
```suggestion
- `FunctionsAPI.call` now also works for clients using auth flow `OAuthInteractive`, `OAuthDeviceCode`, and any user-made subclass of `CredentialProvider`.
```

tests/tests_unit/test_api/test_functions.py
```diff
@@ -564,9 +567,8 @@ def test_function_call_timeout_from_api_key_flow(
 
     @pytest.mark.usefixtures("mock_sessions_with_client_credentials")
     def test_function_call_from_oidc_client_credentials_flow(
-        self, mock_functions_call_responses, cognite_client_with_client_credentials_flow
+        self, mock_functions_call_responses, cognite_client_with_client_credentials_flow, rsps
```
haakonvt, line 570:
```suggestion
        self, mock_functions_call_responses, cognite_client_with_client_credentials_flow
```

cognite/client/_api_client.py
```diff
@@ -79,7 +79,7 @@ def __init__(
         self._config = config
         self._api_version = api_version
         self._api_subversion = config.api_subversion
-        self._cognite_client = cognite_client
+        self._cognite_client = cast("CogniteClient", cognite_client)
```
erlendvollset, line 82: Not make it Optional instead? Do we ever not pass it?

tests/tests_integration/test_api/test_transformations/test_transformations.py
```diff
@@ -407,6 +407,7 @@ def test_update_instance_nodes(self, cognite_client, new_transformation):
             ViewInfo("myViewExternalId", "myViewVersion2", "test-space"), "test-space"
         )
 
+    @pytest.mark.skip  # TODO: view and edgeType are mutually exclusive (code: 400)
```
erlendvollset, line 410:
```suggestion
    @pytest.mark.xfail  # TODO: view and edgeType are mutually exclusive (code: 400)
```

cognite/client/_api_client.py
haakonvt: Agreed, seems like a much better solution. If this causes a lot of `mypy`-cascading issues, I'll put it in a separate PR.

tests/tests_integration/test_api/test_transformations/test_transformations.py
haakonvt: I guess this was already fixed in a separate PR. I'll verify.

cognite/client/_api_client.py
haakonvt: ...yup, definitely separate PR 😂 

image

haakonvt: Addressed in this PR: https://github.com/cognitedata/cognite-sdk-python/pull/1175

tests/tests_integration/test_api/test_transformations/test_transformations.py
haakonvt: Also this now: https://github.com/cognitedata/cognite-sdk-python/pull/1174

haakonvt:
@erlendvollset now that @geircognite has added certificate auth to the SDK, this PR suddenly doesn't add support for "all credential" types anymore (given that `SessionsAPI` does not support it). Should I add a check for it in `_create_session_and_return_nonce` (a function only used inside the `FunctionsAPI`) or just let it fail with the error message from the API (what I have done)?

Maybe you can say something about whether this might be supported soon/later/ever @ngerstle-cognite ? 😄

geircognite:
> 

@haakonvt My preference would be to just add a check and then return a sensible error message if it is certificate auth. Just adding this as a convenience mainly for extractors and I don't see why those would need to use the sessions API.

qTipTip (approved):
This looks great!

Thanks a lot for this! Very nice to get rid of the arguably slightly hacky code to determine what type of login-flow was being used.

haakonvt:
@erlendvollset I think I need an additional approval from you or someone "global" codeowners

erlendvollset (approved):

Issue #1013, opened 2022-08-27
Use singleton thread pool executor


erlendvollset:
Currently a new thread pool executor is spun up every time the SDK wants to perform some concurrent operations. We should instead use a singleton in order to reuse the thread pool across operations. This has the added benefit of making it possible for the SDK to respect to respect the `max_workers` param in all concurrent scenarios.

haakonvt:
Implemented by #1137

PR #1137, opened 2023-02-02
Use singleton threadpoolexecutor and implement an opt-in mainthread executor

erlendvollset:
Needed for running in pyodide. Also addresses #1013

### DIFF
--- a/cognite/client/utils/_concurrency.py
+++ b/cognite/client/utils/_concurrency.py
@@ -135,9 +135,9 @@ class SyncFuture(Future):
 
 class MainThreadExecutor(Executor):
     """
-    In order to support executing sdk methods in in the browser using pyodide (a port of CPython to webassembly),
+    In order to support executing sdk methods in the browser using pyodide (a port of CPython to webassembly),
     we need to be able to turn off the usage of threading. So we have this executor which implements the Executor
-    protocol which uses can use both but just executes everything serially in the main thread.
+    protocol but just executes everything serially in the main thread.
     """
 
     def submit(self, fn: Callable, *args: Any, **kwargs: Any) -> SyncFuture:
--- a/cognite/client/utils/_concurrency.py
+++ b/cognite/client/utils/_concurrency.py
@@ -112,17 +112,17 @@ def collect_exc_info_and_raise(
 T_Result = TypeVar("T_Result", covariant=True)
 
 
-class _Executor(Protocol):
-    def submit(self, fn: Callable[..., T_Result], *args: Any, **kwargs: Any) -> _Future[T_Result]:
+class TaskExecutor(Protocol):
+    def submit(self, fn: Callable[..., T_Result], *args: Any, **kwargs: Any) -> TaskFuture[T_Result]:
         ...
 
 
-class _Future(Protocol[T_Result]):
+class TaskFuture(Protocol[T_Result]):
     def result(self) -> T_Result:
         ...
 
 
-class SyncFuture(_Future):
+class SyncFuture(TaskFuture):
     def __init__(self, fn: Callable[..., T_Result], *args: Any, **kwargs: Any):
         self.__fn = fn
         self.__args = args
@@ -132,7 +132,7 @@ class SyncFuture(_Future):
         return self.__fn(*self.__args, **self.__kwargs)
 
 
-class MainThreadExecutor(_Executor):
+class MainThreadExecutor(TaskExecutor):
     """
     In order to support executing sdk methods in the browser using pyodide (a port of CPython to webassembly),
     we need to be able to turn off the usage of threading. So we have this executor which implements the Executor
@@ -151,7 +151,7 @@ class ConcurrencySettings:
     executor_type: Literal["threadpool", "mainthread"] = "threadpool"
 
 
-def get_executor(max_workers: int) -> _Executor:
+def get_executor(max_workers: int) -> TaskExecutor:
     global _THREAD_POOL_EXECUTOR_SINGLETON
 
     if max_workers < 1:
@@ -159,7 +159,7 @@ def get_executor(max_workers: int) -> _Executor:
 
     if ConcurrencySettings.executor_type == "threadpool":
         try:
-            executor: _Executor = _THREAD_POOL_EXECUTOR_SINGLETON
+            executor: TaskExecutor = _THREAD_POOL_EXECUTOR_SINGLETON
         except NameError:
             # TPE has not been initialized
             executor = _THREAD_POOL_EXECUTOR_SINGLETON = ThreadPoolExecutor(max_workers)
--- a/cognite/client/utils/_concurrency.py
+++ b/cognite/client/utils/_concurrency.py
@@ -160,27 +160,26 @@ def _shutdown_thread_pool_executor() -> None:
 atexit.register(_shutdown_thread_pool_executor)
 
 
-class Settings:
+class ConcurrencySettings:
     executor_type: Literal["threadpool", "mainthread"] = "threadpool"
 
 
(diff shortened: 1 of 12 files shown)


cognite/client/utils/_concurrency.py
```diff
@@ -109,38 +110,109 @@ def collect_exc_info_and_raise(
...
+        self.__args = args
+        self.__kwargs = kwargs
+
+    def result(self) -> T_Result:
+        return self.__fn(*self.__args, **self.__kwargs)
+
+
+class MainThreadExecutor(Executor):
+    """
+    In order to support executing sdk methods in in the browser using pyodide (a port of CPython to webassembly),
```
arnels, line 138:
```suggestion
    In order to support executing sdk methods in the browser using pyodide (a port of CPython to webassembly),
```

cognite/client/utils/_concurrency.py
```diff
@@ -109,38 +110,109 @@ def collect_exc_info_and_raise(
...
+    we need to be able to turn off the usage of threading. So we have this executor which implements the Executor
+    protocol which uses can use both but just executes everything serially in the main thread.
```
arnels, line 140:
```suggestion
    protocol but just executes everything serially in the main thread.
```

arnels (approved):

cognite/client/utils/_concurrency.py
```diff
@@ -109,38 +110,109 @@ def collect_exc_info_and_raise(
...
+
+
+class SyncFuture(Future):
+    def __init__(self, fn: Callable[..., T_Result], *args: Any, **kwargs: Any):
+        self.__fn = fn
+        self.__args = args
+        self.__kwargs = kwargs
+
+    def result(self) -> T_Result:
+        return self.__fn(*self.__args, **self.__kwargs)
```
haakonvt, line 133: Why not use _WorkItem from threading?

https://github.com/python/cpython/blob/c2c970fc26e6de5b9e338413eac71d46aea5ca5e/Lib/concurrent/futures/thread.py#L46

cognite/client/utils/_concurrency.py
```diff
@@ -109,38 +110,109 @@ def collect_exc_info_and_raise(
...
+    except NameError:
+        # TPE has not been initialized
+        pass
+
+
+atexit.register(_shutdown_thread_pool_executor)
+
+
+class Settings:
+    executor_type: Literal["threadpool", "mainthread"] = "threadpool"
```
haakonvt, line 164: Maybe `ExecutorSettings` or something a bit more descriptive?  🤔 

Is the plan that we import this and set `Settings.executor_type = "mainthread"` when running with pyodide?

cognite/client/utils/_concurrency.py
```diff
@@ -109,38 +110,109 @@ def collect_exc_info_and_raise(
...
+
+
+def get_executor(max_workers: int) -> Executor:
+    global _THREAD_POOL_EXECUTOR_SINGLETON
+    global _MAIN_THREAD_EXECUTOR_SINGLETON
+
```
haakonvt, line 170: No need to declare as globals unless we are changing them I think
```suggestion
```

cognite/client/utils/_concurrency.py
erlendvollset: Don't like using protected code

cognite/client/utils/_concurrency.py
```diff
@@ -109,38 +110,109 @@ def collect_exc_info_and_raise(
...
+    def __init__(self, fn: Callable[..., T_Result], *args: Any, **kwargs: Any):
+        self.__fn = fn
+        self.__args = args
+        self.__kwargs = kwargs
+
+    def result(self) -> T_Result:
+        return self.__fn(*self.__args, **self.__kwargs)
+
+
+class MainThreadExecutor(Executor):
```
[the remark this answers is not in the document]
erlendvollset: Yeah. But Protocols are kind of a mix between structural and nominal subtyping, so you can use them like interfaces aswell.

cognite/client/utils/_concurrency.py
erlendvollset: yes and yes.

cognite/client/utils/_concurrency.py
haakonvt: Fair enough! But we shouldn't (probably) overload typical class names from `threading` like `Future`, likely to become an import in a utils module named "concurrency" at some point.

erlendvollset: agree. Will rename Future and Executor to _Future and _Executor.

cognite/client/utils/_concurrency.py
```diff
@@ -109,38 +110,108 @@ def collect_exc_info_and_raise(
...
+def _shutdown_thread_pool_executor() -> None:
+    try:
+        tpe = _THREAD_POOL_EXECUTOR_SINGLETON
+        tpe.shutdown(wait=False)
+    except NameError:
+        # TPE has not been initialized
+        pass
+
+
+atexit.register(_shutdown_thread_pool_executor)
```
haakonvt, line 160: In PY39+ this will be executed after `ThreadPoolExecutor`'s own cleanup code is run.

erlendvollset: Ah that does the same thing - so this can be removed entirely afaict?

cognite/client/utils/_concurrency.py
```diff
@@ -109,38 +110,108 @@ def collect_exc_info_and_raise(
...
-                        failed_tasks.append(tasks[i])
-                    else:
-                        unknown_result_tasks.append(tasks[i])
-                else:
+    if ConcurrencySettings.executor_type == "threadpool":
+        try:
+            executor: _Executor = _THREAD_POOL_EXECUTOR_SINGLETON
+        except NameError:
+            # TPE has not been initialized
+            executor = _THREAD_POOL_EXECUTOR_SINGLETON = ThreadPoolExecutor(max_workers)
```
haakonvt, line 178: Is it really a singleton? ;P
```suggestion
            executor = _THREAD_POOL_EXECUTOR_SINGLETON = _THREAD_POOL_EXECUTOR_SINGLETON(max_workers)
```

erlendvollset: hmm, yes?

erlendvollset: note that l146 does not assign the type ThreadPoolExecutor to _THREAD_POOL_EXECUTOR_SINGLETON, it just declares the variable with a type annotation.

tests/tests_unit/test_utils/test_concurrency.py
```diff
@@ -0,0 +1,31 @@
...
+    get_executor,
+)
+
+
+class TestExecutor:
+    def test_set_and_get_executor(self) -> None:
+        executor = get_executor(1)
+        assert isinstance(executor, ThreadPoolExecutor)
+
+        ConcurrencySettings.executor_type = "mainthread"
```
haakonvt, line 16: Will this have a side effect for later tests? (`monkeypatch.setattr`). Btw, sorry for not using a review, thought I'd only have one comment 😂

cognite/client/utils/_concurrency.py
haakonvt: # 🤯 
Oh damn, you're right... And I read those lines several times without noticing

tests/tests_unit/test_utils/test_concurrency.py
erlendvollset: no worries 😄 and yes. that's a good catch.

PR #1057, opened 2022-10-03
Move extraction pipeline configs from experimental SDK

einarmo:
## Description
Please describe the change you have made.

## Checklist:
- [x] Tests added/updated.
- [x] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change. 
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring. 
- [x] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [x] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) per [semantic versioning](https://semver.org/).

PR #1143, opened 2023-02-09
Make `CogniteClient` private to avoid `JSON.dumps()` error

haakonvt:
## Description
Resource class `ExtractionPipelineConfigRevision` and all subclasses (`ExtractionPipelineConfig`) fails when printing (calls dump) and dumping. Found by @janne123456789 

Bug was introduced in #1057 which added the classes.

Code to reproduce error:
```py
>>> pipe_config = client.extraction_pipelines.config.retrieve(external_id="yo")
>>> print(pipe_config)  # or >>> pipe_config.dump()
TypeError: Object <unlocked _thread.lock object at 0x7fbd080b5420> of type <class '_thread.lock'> can't be serialized by the JSON encoder
```

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -54,6 +54,7 @@ class CogniteResponse:
         Returns:
             Dict[str, Any]: A dictionary representation of the instance.
         """
+        assert not hasattr(self, "cognite_client")
         return basic_instance_dump(self, camel_case=camel_case)
 
     @classmethod
@@ -100,6 +101,7 @@ class CogniteResource:
         Returns:
             Dict[str, Any]: A dictionary representation of the instance.
         """
+        assert not hasattr(self, "cognite_client")
         return basic_instance_dump(self, camel_case=camel_case)
 
     @classmethod
@@ -462,6 +464,7 @@ class CogniteFilter:
         Returns:
             Dict[str, Any]: A dictionary representation of the instance.
         """
+        assert not hasattr(self, "cognite_client")
         return basic_instance_dump(self, camel_case=camel_case)
 
 
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -54,7 +54,6 @@ class CogniteResponse:
         Returns:
             Dict[str, Any]: A dictionary representation of the instance.
         """
-        assert not hasattr(self, "cognite_client")
         return basic_instance_dump(self, camel_case=camel_case)
 
     @classmethod
@@ -101,7 +100,6 @@ class CogniteResource:
         Returns:
             Dict[str, Any]: A dictionary representation of the instance.
         """
-        assert not hasattr(self, "cognite_client")
         return basic_instance_dump(self, camel_case=camel_case)
 
     @classmethod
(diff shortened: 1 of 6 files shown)


cognite/client/data_classes/extractionpipelines.py
```diff
@@ -301,7 +301,7 @@ def __init__(
         self.revision = revision
         self.description = description
         self.created_time = created_time
-        self.cognite_client = cognite_client
+        self._cognite_client = cognite_client
```
haakonvt, line 304: I see that we do some magic in `__new__`, but this only sets `self._cognite_client` if `cognite_client` was given, it does not pop it from kwargs, and as such, subclasses are free to set it as non-private 🤔

einarmo:
I'm realizing I'm not actually 100% sure why we are sending the client into the DTO to begin with. I believe I probably just copied this from something else in the SDK...

haakonvt:
> I'm realizing I'm not actually 100% sure why we are setting it like this to begin with. It might have just been copied from something else.

It’s because certain of these resource objects have convenience methods that need a reference to the SDK to work, eg. `Asset.subtree()` or `TimeSeries.latest()`. Btw don’t bother with the extra tests; I updated the existing tests.

einarmo:
Don't think I can review this one, which I guess means that this module is under the wrong codeowners, it should be `@cognitedata/data-management`.

cognite/client/data_classes/_base.py
```diff
@@ -100,6 +101,7 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
         Returns:
             Dict[str, Any]: A dictionary representation of the instance.
         """
+        assert not hasattr(self, "cognite_client")
```
DD5HT, line 104: I'm not a big fan of asserts (also hasattr is relatively expensive) in non test code.

Maybe add a failing test instead?

haakonvt: How will that pick up when a new subclass is added to the SDK? I guess we could potentially make a recursive `cls.__subclasses()` test... 😬 🤔 Yeah, I think I like that approach 👍 

Btw, I agree that asserts are incorrectly used all over the SDK; this one, however, should be very much in line with their intended use-case (i.e. for the SDK developer, _not the user_ (`-O`)).

haakonvt: On second thought... no real way to verify all variants of subclasses. Some have required arguments, some are dataclasses with `__post_init__` method that do not allow inspection before instantiation... so I guess I'm back to wondering: _How will that pick up when a new subclass is added to the SDK?_

DD5HT: I'm one of the people who thinks classes where a mistake in programming :D 

There is also always the option to just enforce it via code review :)

haakonvt: # 🙈

haakonvt: I went down the rabbit hole - figured the base Cognite classes have a lot of improvement potential. I removed the `hasattr` check from this PR, and will add tests (solved the issue above) in a separate PR!

DD5HT (approved):

PR #1151, opened 2023-02-17
Draft: Improve base cognite classes

haakonvt:
### New implementation for `_cognite_client` attribute:

1. Removes custom `__new__` implementation which adds a 2x slowdown to object instantiation.
2. Removes custom `__getattribute__` implementation (hook before regular attribute lookup).
3. Adds `_cognite_client` as a property with a setter method that guarantees a valid value and a getter that raises `CogniteMissingClientError` in all cases except where the attribute is an instance of `CogniteClient`.

### New hierarchy to avoid code duplication

1. Introduces a `CogniteBase` class.
2. In progress

### New `_load` implementation for `dict` data

1. Removes the old `if hasattr -> setattr` implementation that called `to_snake_case` O(N*M)` times (N dicts, M elements per dict).
2. Adds `fast_dict_load` that uses `get_accepted_params` which caches the accepted parameters for each class it receives, returning a dict that maps `camelCase` to `snake_case` for efficient API response translation into our data classes.

## Checklist:
- [ ] Tests added/updated.


haakonvt:
Changed:
With `__new__` removed, subclasses of `CogniteResource` no longer has an implicit `__init__` that just swallows all args and all kwargs.

haakonvt:
# Interested in feedback on how to make the class hierarchy better!
- Type hints still feels lacking in many places, should we add `typing_extensions ^4` to get `Self` for instance?

PR #1153, opened 2023-02-18
Enforce that all mocked APIs exactly match available APIs

haakonvt:
## Description
In the past, the APIs available on `CogniteClientMock` has been a subset of the actual APIs available through the `CogniteClient`. This PR adds a test that enforces both that...

1. ...after a new API has been added to the SDK, the `CogniteClientMock` must be updated accordingly
2. ...once an API is removed from the SDK, the `CogniteClientMock` must be updated accordingly

## Checklist:
- [x] Tests added/updated.


erlendvollset (approved):
Nice 👏

PR #1158, opened 2023-02-23
Make asset creation user friendly with `AssetHierarchy` and `upsert`

haakonvt:
Link to slack discussion: https://cognitedata.slack.com/archives/C04R13T79N3

## Description
Adds a new data class for better asset hierarchy verification as requested by the Statnett project team. Able to make nice reports on any issues found (like the Java SDK, but more flexible 😉 ).

## Upsert? Upsert!
Creating asset hierarchies can be tiresome; when inserts partially fail, in order to clean up, you can't just run delete as with request sizes of max 1000 assets under the hood, the likelihood that a parent is still referenced is high... and so you must find all root assets and run a recursive delete.

Not anymore.

Introducing upsert with partial or full overwrite. Nice.

## Read more in `CHANGELOG.md`

### DIFF
--- a/tests/tests_integration/test_api/test_assets.py
+++ b/tests/tests_integration/test_api/test_assets.py
@@ -235,7 +235,7 @@ def create_hierarchy_with_cleanup(client, assets, upsert=False, upsert_mode=""):
 
 @pytest.fixture(scope="class")
 def set_create_lim(cognite_client):
-    with set_max_workers(2), pytest.MonkeyPatch.context() as mp:
+    with set_max_workers(cognite_client, 2), pytest.MonkeyPatch.context() as mp:
         # We set a low limit to hopefully detect bugs in how resources are split (+threading)
         # without unnecessarily overloading the API with many thousand assets/request:
         mp.setattr(cognite_client.assets, "_CREATE_LIMIT", 3)


cognite/client/data_classes/assets.py
```diff
@@ -449,3 +456,244 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
...
+
+            elif parent in has_cycles:
+                has_cycles.add(xid)
+
+            elif xid == parent:
+                has_cycles.add(xid)
+
+            else:
+                counter, seen = itertools.count(), {xid}
+                while next(counter) < 1_000_000:
```
nikokaoja, line 615: 1 000 000 could be replaced by parameter `max_hierarchy_depth`, set to 1 000 000 by default

cognite/client/data_classes/assets.py
```diff
@@ -449,3 +456,244 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
...
+                    xid, parent = parent, edges.get(parent)
+                    seen.add(xid)
+                    if parent in no_cycles:
+                        no_cycles |= seen
+                        break
+                    elif parent in seen:
+                        has_cycles |= seen
+                        break
+                else:  # escape hatch if we don't break out of while
+                    raise RuntimeError("Failed to locate root asset or a cycle after 1 mill. repeated parent lookups")
```
nikokaoja, line 625: RuntimeError(f"Failed to locate root asset or a cycle after {max_hierarchy_depth} repeated parent lookups")

cognite/client/data_classes/assets.py
```diff
@@ -449,3 +456,244 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
...
+    def _preconditions_for_cycle_check_are_met(self, on_error: OnErrorOptions) -> bool:
+        # Required before cycle-check can run
+        if not (not self._roots or self._orphans or self._missing or self._duplicates):
+            return True
+        self._on_error(on_error, "Unable to run cycle-check before basic issues are fixed.")
+        return False
+
+    def _locate_cycles(self, on_error: OnErrorOptions) -> Tuple[int, List[List[str]]]:
+        edges = dict(map(op.attrgetter("external_id", "parent_external_id"), self._assets))
+        has_cycles, no_cycles = set(), {None, *map(self._get_xid, self.roots)}
```
nikokaoja, line 601: Not sure what no_cycles contains at its start.

nikokaoja (commented):
LGTM

cognite/client/data_classes/assets.py
```diff
@@ -449,3 +456,244 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
...
+                no_cycles.add(xid)
+
+            elif parent in has_cycles:
+                has_cycles.add(xid)
+
+            elif xid == parent:
+                has_cycles.add(xid)
+
+            else:
+                counter, seen = itertools.count(), {xid}
```
DD5HT, line 614: This whole block should probably be a function.

cognite/client/data_classes/assets.py
DD5HT, line 615: why isn't that a for loop? counter isn't used anywhere else?
```
for i in range(100000)
```

haakonvt: That's a great idea 👌

cognite/client/data_classes/assets.py
haakonvt: Do you want me to add a comment? It's just an optimization to add all root asset external IDs to the set of assets that are not part of a cycle. I also add `None` to avoid `if x in no_cycles or x is None` below because whenever a parent external ID is `None`, you are for sure safe 😜

cognite/client/data_classes/assets.py
haakonvt: Started off as an infinite while loop then I started second-guessing myself and added the "escape hatch" 😂  I'll for sure swap it to for-loop instead 🚀

haakonvt: If you have any other thoughts/ideas, please join the discussion in slack! 😄 
https://cognitedata.slack.com/archives/C04R13T79N3

cognite/client/data_classes/assets.py
haakonvt: Yeah, probably! 🤷

cognite/client/utils/_text.py
```diff
@@ -0,0 +1,19 @@
...
+    HLINE = "\N{box drawings light horizontal}"
+    VLINE = "\N{box drawings light vertical}"
+    XLINE = "\N{box drawings light vertical and horizontal}"
+    TOPLINE = "\N{box drawings light down and horizontal}"
+
+
+def shorten(obj: Any, width: int = 20, placeholder: str = "...") -> str:
+    # 'textwrap.shorten' skips entire words... so we make our own:
+    if width < len(placeholder):
+        raise ValueError("Width must be strictly less than length of 'placeholder'")
```
haakonvt, line 14: Invert error message...

cognite/client/utils/_text.py
```diff
@@ -0,0 +1,19 @@
...
+
+    s = obj if isinstance(obj, str) else repr(obj)
+    if len(s) <= width:
+        return s
+    return f"{s[:width-3]}{placeholder}"
```
haakonvt, line 19: -3 should be -len(placeholder)

cognite/client/exceptions.py
```diff
@@ -211,6 +218,45 @@ class CogniteAPIKeyError(CogniteAuthError):
...
+            of assets that failed validation.
+    """
+
+    def __init__(self, message: str, hierarchy: AssetHierarchyValidator) -> None:
+        self.message = message
+        self._hierarchy = hierarchy
+
+    @property
+    def orphans(self) -> AssetList:
+        return self._hierarchy.orphans
```
haakonvt, line 241: Note to self: Consider implementing `__getattr__` to give direct access to `self.hierarchy` 🤔

cognite/client/data_classes/assets.py
```diff
@@ -449,3 +471,292 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
         if isinstance(self.labels, LabelFilter):
             result["labels"] = self.labels.dump(camel_case)
         return result
+
+
+OnErrorOptions = Literal["ignore", "warn", "raise"]
+
+
+class AssetHierarchyValidator:
```
erlendvollset, line 479: I'd suggest `ValidAssetHierarchy` or just `AssetHierarchy` instead.

cognite/client/_api/assets.py
```diff
@@ -412,19 +433,41 @@ def create(self, asset: Union[Asset, Sequence[Asset]]) -> Union[Asset, AssetList
...
+            1. All assets linking a parent by ID are assumed valid
+            2. All orphan assets are assumed valid. "Orphan" means the parent is not part of the given assets (category "orphans")
 
         Args:
-            assets (Sequence[Asset]]): List of assets to create. Requires each asset to have a unique external id.
+            assets (Sequence[Asset] | AssetHierarchyValidator): List of assets to create or an instance of AssetHierarchyValidator.
+            upsert (bool): Already existing assets will be updated instead of an exception being raised. You may control how updates
+                are applied with the 'upsert_mode' argument.
+            upsert_mode ("patch" | "replace"): Only applicable with upsert. Pass 'patch' to do incremental updates, or
+                'replace' to do full updates (default).
```
erlendvollset, line 467: It's not clear to me what an incremental update is. Perhaps this is more clear?
```suggestion
            upsert_mode ("patch" | "replace"): Only applicable with upsert. Pass 'patch' to only update fields with non-null values , or 'replace' to do full updates (default).
```

erlendvollset: I also think 'patch' should be default behaviour, since it's how the update() method behaves.

cognite/client/_api_client.py
```diff
@@ -804,16 +805,15 @@ def _convert_resource_to_patch_object(
...
                 patch_object["update"][key] = {"set": value}
         return patch_object
 
     @staticmethod
     def _status_ok(status_code: int) -> bool:
         return status_code in {200, 201, 202, 204}
 
-    @classmethod
-    def _raise_api_error(cls, res: Response, payload: Dict) -> NoReturn:
+    def _raise_api_error(self, res: Response, payload: Dict) -> NoReturn:
```
erlendvollset, line 816: This can remain a classmethod?

cognite/client/_api/assets.py
```diff
@@ -641,234 +706,169 @@ def _get_children(self, assets: List) -> List:
...
+        created_assets = AssetList([], cognite_client=self.assets_api._cognite_client)
+
+        # Kick things off with all...:
+        # 1. Root assets
+        # 2. Assets linking parent by ID
+        # 3. Orphans assets (if `hierarchy.ignore_orphans` is True)
+        to_create = insert_dct.pop(None)
+
+        # Optimization: Child assets can be created alongside parents -if- part of the same request.
+        self._extend_unblocked_from_subtree(to_create, insert_dct)
```
erlendvollset, line 731: Can you do a comparison of performance on the old and new implementations? I think losing the concurrency will have significant impact.

cognite/client/utils/_text.py
```diff
@@ -0,0 +1,39 @@
+import functools
```
erlendvollset, line 1: Nice change, but would be even nicer in a separate PR! The PR is not really reviewable commit-by-commit, so this causes noise when reviewing all files at once.

tests/utils.py
```diff
@@ -86,6 +87,23 @@ def random_gamma_dist_integer(inclusive_max, max_tries=100):
     raise RuntimeError(f"Max tries exceeded while generating a random integer in range [1, {inclusive_max}]")
 
 
+def generate_asset_tree(root_external_id: str, depth: int, children_per_node: int, current_depth=1):
```
erlendvollset, line 90: Let's keep this asset-specific test helper in `test_api/test_assets`.

tests/tests_unit/test_api/test_assets.py
```diff
@@ -314,237 +310,175 @@ def test_assets_update_object(self):
...
-        assert {a.external_id for a in e.value.successful} == {"0", "01"}
-
-    def test_post_with_all_failures_and_more_requests_than_workers(self, cognite_client, mock_post_assets_failures):
-        num_assets = cognite_client.config.max_workers * 10
-        assets = [Asset(name="400", external_id=str(i)) for i in range(num_assets)]
-        with pytest.raises(CogniteAPIError) as e:
-            cognite_client.assets.create_hierarchy(assets)
-
-        assert {a.external_id for a in e.value.failed} == {str(i) for i in range(num_assets)}
+# class TestAssetPoster:
```
erlendvollset, line 313: I think instead of revamping these unit tests, we should instead extend the integration tests for creating hierarchies. Will make it easier to evolve the implementation.

erlendvollset (commented):
Nice work @haakonvt! Left some comments but I think we're not too far off here. And I'll take it for a spin when I get some more time.

cognite/client/data_classes/assets.py
haakonvt: I'll go with `AssetHierarchy` then, I've been back and forth quite a few times already 😅

cognite/client/_api/assets.py
haakonvt: Should only be slightly worse in contrived scenarios: the typical hierarchy is growing geometrically with depth, thus any slowness in the start (to get the roots and first few levels unblocked) is the only time when the old double-concurrent solution is faster, as we get to the "lower levels", the number of assets >> number-of-workers and the waiting time is purely backend dependent. Tests also show this.

To combat some of that "initial slowness", `_extend_unblocked_from_subtree` will make sure to add at least one full request with roots and its children (>= 1000 assets), since these can be created simultaneously.

```
# Test 1:
- 262 142 assets
- hierarchy: networkx.balanced_tree(2, 17)
- 2: branching factor
- 17: depth

# Test 2:
- 50 100 assets
- hierarchy: 100 levels with 100 assets,
  each new level picks parents randomly from previous

# Test 1, new code:
- insert: 68.80 sec
# Test 1, old code:
- insert: 66.51 sec

# Test 2, old code
- insert: 24.89 sec
- insert: 17.92 sec
# Test 2, new code:
- insert: 29.41 sec
- insert: 26.99 sec
```

haakonvt: <= 1000 *

cognite/client/utils/_text.py
haakonvt: Will move to a separate PR then 😉

tests/utils.py
haakonvt: ...but it exists in two different places, with slightly different implementation (`name` is missing in one of them)

cognite/client/_api_client.py
haakonvt: Why? 🤔  It is not used anywhere outside of APIClient.

cognite/client/data_classes/assets.py
```diff
@@ -449,3 +471,292 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
...
+    def _preconditions_for_cycle_check_are_met(self, on_error: OnErrorOptions) -> bool:
+        if not (
+            self._invalid or self._unsure_parents or self._duplicates or (self._orphans and not self._ignore_orphans)
+        ):
+            return True
+        self._on_error(on_error, "Unable to run cycle-check before basic issues are fixed.")
+        return False
+
+    def _locate_cycles(self) -> Tuple[int, List[List[str]]]:
+        edges = dict(map(op.attrgetter("external_id", "parent_external_id"), self._assets))
```
spex66, line 645: @haakonvt 
isn't a dict-comprehension not easier -- more pythonic -- to read than this func-style?
```
edges = {asset["external_id"] : asset["parent_external_id"] for asset in self._assets}
```

haakonvt: Haha jajajaja I agree 😉 I'm pretty sure it was faster in the older days of Python, but that does not seem to be the case anymore (3.9.16 shows equal performance). Btw, I guess you meant:
```py
edges = {asset.external_id: asset.parent_external_id for asset in self._assets}
```

cognite/client/_api_client.py
erlendvollset: The code becomes easier to reason about when you can know that a method doesn't require access to any instance variables/methods just by looking at the definition. So IMHO it's good practice to default to using static/class methods.

cognite/client/_api/assets.py
erlendvollset: Nice!

tests/utils.py
erlendvollset: right, but if we won't be needing it in `test_unit/test_api/test_assets` anymore, then it will only be used in `test_integration/test_api/test_assets`

cognite/client/_api_client.py
haakonvt: I agree with most of that. But when a function like this is _only_ dependent on other `staticmethod`s, it feels like a design flaw with the class. What I mean is that I don't like using classes to group a set of functions instead of having the as simple standalone ones.

cognite/client/_api/assets.py
haakonvt: I was ofc not happy with the new code potentially being slower in synthetic scenarios ( 😂 ) so I've rewritten it 🤓 It is now has the exact same performance with the potential to be faster (queued requests now use `PriorityThreadPoolExecutor` with the size of the subtree as the priority) 🚀 .

cognite/client/utils/_concurrency.py
```diff
@@ -279,12 +292,18 @@ def execute_tasks(
...
     return TasksSummary(successful_tasks, unknown_result_tasks, failed_tasks, results, exceptions)
+
+
+def classify_error(err: Exception) -> Literal["failed", "unknown"]:
+    if isinstance(err, CogniteAPIError):
+        if err.code < 500:
+            return "failed"
+        else:
+            return "unknown"
+    return "failed"
```
haakonvt, line 309: Idk.
```suggestion
    if isinstance(err, CogniteAPIError) and err.code >= 500
        return "unknown"
    return "failed"
```

cognite/client/utils/_concurrency.py
```diff
@@ -237,6 +234,22 @@ def get_executor(max_workers: int) -> TaskExecutor:
     return executor
 
 
+def get_as_completed_fn(
```
erlendvollset, line 237: Why this rather than the static methods on the implementation?

cognite/client/_api/assets.py
```diff
@@ -412,32 +444,137 @@ def create(self, asset: Union[Asset, Sequence[Asset]]) -> Union[Asset, AssetList
...
         return self._create_multiple(list_cls=AssetList, resource_cls=Asset, items=asset)
 
-    def create_hierarchy(self, assets: Sequence[Asset]) -> AssetList:
-        """Create an asset hierarchy.
+    def create_hierarchy(
+        self,
+        assets: Union[Sequence[Asset], AssetHierarchy],
+        *,
+        upsert: bool = False,
+        upsert_mode: UpsertOptions = "patch",
```
erlendvollset, line 452: Let's be consistent with the naming
```suggestion
        upsert_mode: UpsertMode = "patch",
```

tests/tests_integration/test_api/test_assets.py
```diff
@@ -245,3 +176,203 @@ def test_filter_by_geo_location(self, cognite_client):
...
+
+
+@pytest.fixture(scope="class")
+def set_create_lim(cognite_client):
+    with pytest.MonkeyPatch.context() as mp:
+        # We set a low limit to hopefully detect bugs in how resources are split (+threading)
+        # without unnecessarily overloading the API with many thousand assets/request:
+        mp.setattr(cognite_client.assets, "_CREATE_LIMIT", 3)
+        mp.setattr(cognite_client.assets._config, "max_workers", 2)
+        yield
```
erlendvollset, line 231: we alread have the `set_request_limit` and `set_max_workers` ctx managers in tests/utils

cognite/client/_api/assets.py
haakonvt: Disagree, `upsert_mode` is a selection of one of the `UpsertOptions`.

cognite/client/utils/_concurrency.py
haakonvt: So I can definitely revert it 👍

tests/tests_integration/test_api/test_assets.py
erlendvollset: I agree it would be better to be more specific about which limits it changes.

cognite/client/_api/assets.py
haakonvt: Huh, weird, to me, it is literally (pun intended) just a collection of (specifically) allowed values. Not sure I follow the `Animal` example since a `Literal` allows wild combinations of types:
```py
# From PEP586
Literal[-4, "foo", b"foo", u"foo", True, Color.RED, None]
```
I ended up just spelling it out as `Literal["replace", "patch"]` in the type hints. I did the same for `OnErrorOptions = Literal["ignore", "warn", "raise"]` in `cognite/client/data_classes/assets.py`.

haakonvt: ...and yes, I would 10/10 times name that `AnimalOptions` 😜

cognite/client/utils/_concurrency.py
haakonvt: Reverted in #c44b555

erlendvollset (approved):
LGTM!

cognite/client/utils/_text.py
haakonvt: https://github.com/cognitedata/cognite-sdk-python/pull/1179

PR #1160, opened 2023-02-23
Make the SDK compatible with `pandas` major version 2 ahead of release

haakonvt:
## Description
Title says it all!

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [5.6.3] - 23-02-23
+### Added
+- Make the SDK compatible with `pandas` major version 2 ahead of release.
+
 ## [5.6.2] - 21-02-23
 ### Fixed
 - Fixed an issue where `Content-Type` was not correctly set on file uploads to Azure.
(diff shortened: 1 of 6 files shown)


erlendvollset (commented):
I guess we should change [this](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml#L24) version constraint to ">=1.4, <3" also then

haakonvt:
> I guess we should change [this](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml#L24) version constraint to ">=1.4, <3" also then

Haha I knew I had forgotten something 😂 🤦‍♂️

haakonvt:
Btw, I dropped the additional `<3` from the version requirement as I like libraries to be lenient and expect most upgrades to be fully compatible (we are not using any exotic part of the pandas API). An opinion adopter from: https://www.youtube.com/watch?v=WSVFw-3ssXM

erlendvollset:
I have a different position there! I like to not have my software randomly fail because major versions weren't sufficiently constrained. But pandas has historically had a lot of focus on not breaking their API and we use a relatively small surface area, so I'm fine with not putting a ceiling there 👍

erlendvollset (approved):

PR #1164, opened 2023-03-08
Fix identifier trouble in `ExtractionPipelineRun`s

haakonvt:
## Description
As discovered by @spex66 , `id` of ext. pipe runs was returned from the API, but hidden from users of the Python SDK.

Due to how the SDK is implemented (or more correctly, how the ext-pipe-API is inconsistent 😉 ), we can't make this change work without "sacrificing something". Here's a copy of note from the PR:
```
# Note: No way to make this id/xid API mixup completely correct. Either:
# 1. We use id / external_id for respecively "self id" / "ext.pipe external id"
#   - Problem: Only dataclass in the SDK where id and external_id does not point to same object...
# 2. We rename external_id to extpipe_external_id in the SDK only
#   - Problem: This dump method might be surprising to the user - if used (its public)...
# ...and 2 was chosen:
```

Slack discussion:
https://cognitedata.slack.com/archives/CG10VQPFX/p1677665680493719

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,12 +17,15 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [5.6.5] - 08-03-23
+## [5.7.0] - 08-03-23
 ### Removed
 - `ExtractionPipelineRunUpdate` was removed as runs are immutable.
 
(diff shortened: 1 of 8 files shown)


erlendvollset (approved):
Molto bene! 💯

PR #1167, opened 2023-03-10
Fix method dump of TransformationDestination

vuhainguyen:
## Description
We should ignore None in the method dump of TransformationDestination to avoid the error
`Result AttributeError("'NoneType' object has no attribute 'dump'")`

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/transformations/common.py
+++ b/cognite/client/data_classes/transformations/common.py
@@ -27,7 +27,8 @@ class TransformationDestination:
 
         needs_dump = set(iterable_to_case(("view", "edge_type"), camel_case))
         for k in needs_dump.intersection(ret):
-            ret[k] = ret[k].dump(camel_case=camel_case)
+            if ret[k] is not None:
+                ret[k] = ret[k].dump(camel_case=camel_case)
         return ret
 
     @staticmethod
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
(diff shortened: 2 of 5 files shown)


silvavelosa (approved):
LGTM ✔️

cognite/client/data_classes/transformations/common.py
```diff
@@ -27,7 +27,8 @@ def dump(self, camel_case: bool = False) -> Dict[str, Any]:
 
         needs_dump = set(iterable_to_case(("view", "edge_type"), camel_case))
         for k in needs_dump.intersection(ret):
-            ret[k] = ret[k].dump(camel_case=camel_case)
+            if ret[k] is not None:
+                ret[k] = ret[k].dump(camel_case=camel_case)
```
haakonvt, line 31: @vuhainguyen @silvavelosa This literally can't happen. `basic_obj_dump` will remove all keys where the corresponding value is `None`...

haakonvt: Just kidding, there are two "basic obj dump" functions 🤣 My bad... 🤦

haakonvt: I would recommend switching to from `_base.py`:

```py
def basic_instance_dump(obj: Any, camel_case: bool) -> Dict[str, Any]:
    # TODO: Consider using inheritance?
    dumped = {k: v for k, v in vars(obj).items() if v not in EXCLUDE_VALUE and not k.startswith("_")}
    if camel_case:
        return convert_all_keys_to_camel_case(dumped)
    return dumped
```

haakonvt: Clean-up PR incoming 🤓

PR #1169, opened 2023-03-14
Fix exc raised when user erroneously had too old `numpy` version

haakonvt:
## Description
- For users unknowingly using a too old version of `numpy` (against the SDK dependency requirements), an exception could be raised (`NameError: name 'np' is not defined`). This has been fixed.

https://cognitedata.slack.com/archives/CDJUS0FAB/p1678801763640389?thread_ts=1678800656.935999&cid=CDJUS0FAB


erlendvollset (approved):
LGTM

PR #1185, opened 2023-03-27
Bugfixes for `AssetList._retrieve_related_resources` convenience method

haakonvt:
## Description
The resource list that was returned after calling `.[time_series / events / files / ... etc]()` on an instance of `AssetList` was...

1. ...missing a `CogniteClient` reference (not a big deal yet, since neither of these have their own convenience methods).
2. ...missing the ID- and external ID lookup table, so `.get(...)` did not work.

This PR also changes:

- Makes use of the `split_into_chunks` convenience function.
- Instantiate only a single `CogniteResourceList` at the very end to ensure all lookups are built.
- ~~Disables future use of `extend` - at least until someone provides a working implementation that also updates the lookup tables~~
- Implemented `extend`.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [5.9.2] - 20-03-23
+### Fixed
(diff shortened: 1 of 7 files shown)


erlendvollset (approved):
LGTM

PR #1189, opened 2023-03-29
Timezone support

doctrino:
### Ready for Review 10/4
The main new method is `cognite_client.time_series.data.retrieve_dataframe_in_tz(...)` see documentation and changelog for more details.

A few comments

* I based the signature on `retrieve_dataframe` but did some simplifications. 

1. You can only passe either external_id or ids as a single entry or a list of entries. This is a simplification that can be extended, but suggest doing that later as this PR is a bit urgent.
2. start and end must be passed and be a `datetime`. My argument is that we need a time zone, and this is better than a separate timezone argument.

* I backported in `zoneinfo` which comes with Python `3.9` to handle time zones. It is an optional dependency and linked with pandas.
* When I read the code, I do some syntax fixer and refactorings. I have kept theses in separate commits.
* I created the utility function `cdf_aggregate`, Note that this utility function is for validation in the testing and is not directly used by the new `retrieve_dataframe_in_tz` methods.

### Update 7/4 -23
First version is done, only support for 1day granulrities, but all aggregations should be supported. 

### Original Plan
The plan is to add a limited scope timezone aggregation support to the SDK. This PR is set up for early feedback.
Scope:
* Aggregation average and sum.
* Granularity day, week, month, quarter, year.
* Timezone, Europe/Oslo.
* After 1970.

I included a suggested signature for the method, along with a few preliminary tests. 

The rough implementation plan is to find the splits between summertime and wintertime, and then to do utc offsetet calls for summer time and wintertime separately. Finally, do a call on the 23 and 25 hour day for sum and count to calculate average. Thus making 1-3 calls depending on which time range the user asks. 

## Checklist:
- [x] Tests added/updated.


tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -992,6 +1005,46 @@ def test_query_with_duplicates(self, retrieve_endpoints, one_mill_dps_ts, ms_bur
...
+    hourly_normal_dist: TimeSeries,
+    minutely_normal_dist: TimeSeries,
+):
+    # Arrange
+    time_series = {"hourly": hourly_normal_dist, "minutely": minutely_normal_dist}[raw_granularity]
+    raw_df = cognite_client.time_series.data.retrieve_dataframe(external_id=time_series.external_id)
+    pandas_agg = {"average": "mean"}.get(aggregation, aggregation)
+    pandas_gran = {"1d": "1D"}.get(granularity, granularity)
+    # Note pandas handles timezones correctly, which is why we use it for reference.
+    pandas_aggregate = raw_df.tz_localize("UTC").tz_convert(timezone).resample(pandas_gran).agg(pandas_agg)
```
haakonvt, line 1034: Yes, `pandas` handles timezones correctly for the raw data, but the aggregation function does not for `average`/`mean` (`sum` ofc is correct haha).

`pandas` mean:
$$\frac{1}{N} \sum_{i=1}^N x_i$$

`CDF` average: the integral over time:
$$\frac{1}{b-a} \int_{t=a}^{t=b}f(x)dx$$
...where `f(x)` is the function defined by either linear or step interpolating between the datapoints `x_i`, also considering the data point right before `t=a` and after `t=b`, unless we have data precisely at the boundary (`a, b`).

doctrino: I implemented a function for doing the aggregation correctly. Unfortunately, I had to make some assumptions, uniform index with known raw frequency. Not looked at step yet either. But it is a start.

haakonvt:
Forcing the use ZoneInfo seems like a great choice 👍 👍 

Btw, did you do this? (Have not inspected `pyproject` file :hide:)

image

doctrino:
> Forcing the use ZoneInfo seems like a great choice 👍 👍
> 
> Btw, did you do this? (Have not inspected `pyproject` file :hide:)
> 
> image

Thanks, I ran it to the same (read a lot on timezones in Python). Will clean it all up when I am done, and since I develop on a Windows system, it should be compatible with Windows :)

haakonvt (commented):
Some integration stuff 🤓

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +960,77 @@ def retrieve_dataframe(
...
+            limit=None,
+            include_aggregate_name=True,
+            column_names="external_id",
+            ignore_unknown_ids=ignore_unknown_ids,
+            **{id_name: queries},  # type:ignore
+        )
+        # The aggregation mean is arbitrary as the result will produce multiple columns
+        # for the same time series, with all values being nan except for one.
+        df = df.groupby(df.columns, axis=1).mean()
+        df = df.tz_localize("utc").tz_convert(tz.key)
```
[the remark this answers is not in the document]
doctrino: I was planning to ask you about this. Nice to get the answer before I even formulated the question  :)

cognite/client/utils/_time.py
```diff
@@ -173,6 +184,175 @@ def split_time_range(start: int, end: int, n_splits: int, granularity_in_ms: int
...
+        raw_freq (str): The frequency of the raw data. If it is not given, it is attempted inferred from raw_df.
+    """
+    if is_step:
+        raise NotImplementedError()
+
+    pd = cast(Any, local_import("pandas"))
+    granularity_pd = granularity.replace("m", "T")
+    grouping = raw_df.groupby(pd.Grouper(freq=granularity_pd))
+    if aggregate == "sum":
+        return grouping.sum()
```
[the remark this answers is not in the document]
doctrino: I plan to allow a lot more of them, but kept it in the original scope for now.

cognite/client/utils/_time.py
```diff
@@ -173,6 +184,175 @@ def split_time_range(start: int, end: int, n_splits: int, granularity_in_ms: int
...
+
+    # When the frequency of the data is 1 hour and above, the end point is excluded.
+    freq = pd.Timedelta(freq)
+    if freq >= pd.Timedelta("1hour"):
+        return grouping.apply(integrate_average)
+
+    def integrate_average_end_points(values: pandas.Series) -> float:
+        weights = np.full(len(values), 2)
+        weights[0] = weights[-1] = 1
+        return np.average(values.values, weights=weights, axis=0)
```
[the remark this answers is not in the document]
doctrino: Nice, that is a good improvement by doing a bit of computation tricks.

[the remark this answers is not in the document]
doctrino: Elegant, I have added it. However, I one of the bigger issues is the next lines:
```python
    step = pd.Timedelta(granularity_pd) // freq

    return (
        raw_df.rolling(window=pd.Timedelta(granularity_pd), closed="both")
        .apply(integrate_average_end_points)
        .shift(-step)
        .iloc[::step]
    )
```
This causes us to do the above calculation step times more than necessary. For example, if step is 2, then we are doing the above calculation 2x more than necessary. In addition, with the above code, the uniform index assumption is still required I think. A final note, this code is used for validation of the new `retrieve_dataframe_in_tz` so I have not spent too much time optimizing it thus far.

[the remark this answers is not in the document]
haakonvt: Remove rolling and use numpy directly? 🤷 Maybe not worth it as long as this is just to validate tests? Idk

haakonvt:
I toyed around with pandas as timezones in `date_range`. Seems you can get all the DST transitions "for free".

### Example with the sightly annoying granularity of 3 days:

```py
from itertools import cycle
import numpy as np
import pandas as pd
from zoneinfo import ZoneInfo
from cognite.client.utils._time import dst_transition_dates

utc = ZoneInfo("UTC")
oslo = ZoneInfo("Europe/Oslo")

idx = pd.date_range(start="2000", freq="3d", periods=1_000, tz=oslo)
didx = np.diff(idx)

mode = pd.value_counts(didx).nlargest(1).index.item()
starts = idx[:-1][didx != mode]
ends = idx[1:][didx != mode]

for s, e, is_fall in zip(starts, ends, cycle([False, True])):
    s = s.astimezone(utc)
    e = e.astimezone(utc)
    dsts = dst_transition_dates(ZoneInfo("Europe/Oslo"), s.year)
    dst = dsts[is_fall]
    print(f"{s}, {s<dst<e=}, {e}")
```

...with corresponding output:
```
2000-03-24 23:00:00+00:00, s<dst<e=True, 2000-03-27 22:00:00+00:00
2000-10-26 22:00:00+00:00, s<dst<e=True, 2000-10-29 23:00:00+00:00
2001-03-22 23:00:00+00:00, s<dst<e=True, 2001-03-25 22:00:00+00:00
2001-10-27 22:00:00+00:00, s<dst<e=True, 2001-10-30 23:00:00+00:00
2002-03-29 23:00:00+00:00, s<dst<e=True, 2002-04-01 22:00:00+00:00
2002-10-25 22:00:00+00:00, s<dst<e=True, 2002-10-28 23:00:00+00:00
2003-03-27 23:00:00+00:00, s<dst<e=True, 2003-03-30 22:00:00+00:00
2003-10-23 22:00:00+00:00, s<dst<e=True, 2003-10-26 23:00:00+00:00
2004-03-27 23:00:00+00:00, s<dst<e=True, 2004-03-30 22:00:00+00:00
2004-10-29 22:00:00+00:00, s<dst<e=True, 2004-11-01 23:00:00+00:00
2005-03-25 23:00:00+00:00, s<dst<e=True, 2005-03-28 22:00:00+00:00
2005-10-27 22:00:00+00:00, s<dst<e=True, 2005-10-30 23:00:00+00:00
2006-03-23 23:00:00+00:00, s<dst<e=True, 2006-03-26 22:00:00+00:00
2006-10-28 22:00:00+00:00, s<dst<e=True, 2006-10-31 23:00:00+00:00
2007-03-24 23:00:00+00:00, s<dst<e=True, 2007-03-27 22:00:00+00:00
2007-10-26 22:00:00+00:00, s<dst<e=True, 2007-10-29 23:00:00+00:00
```

haakonvt:
_...my point_, of which you obviously already are aware, was just that we can split super easily on timedelta between two consecutive elements not being equal to granularity. For short time scales, this is an awful algorithm having to build the index with potentially millions of timestamps that we just throw out "immediately", _but for this case_, with **day** being the shortest time span we are interested in, its merely a few thousand and quick AF 😅

doctrino:
> _...my point_, of which you obviously already are aware, was just that we can split super easily on timedelta between two consecutive elements not being equal to granularity. For short time scales, this is an awful algorithm having to build the index with potentially millions of timestamps that we just throw out "immediately", _but for this case_, with **day** being the shortest time span we are interested in, its merely a few thousand and quick AF 😅

A very good point, sometimes brute forcing it is the easiest. I have started to toy in my mind with supporting h, m, and s. But then I would not need this functionality anyways, as then you are still good.

cognite/client/utils/_time.py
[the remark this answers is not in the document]
doctrino: Thought about it, but need something like a np.strides, which is not worth it as this is only for testing

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+        column_names: Literal["id", "external_id"] = "external_id",
+    ) -> pd.DataFrame:
+        """Get datapoints directly in a pandas dataframe in the time zone given in start and end.
+
+        **Note** This is a convenicence method. It builds on top of the retrieve_array and retrieve_dataframe methods.
+        It enables you to do correct aggregates in a local time zone with daily, weekly, quartely, and yearly
+        aggregates, including accounting for DST (Daylight Saving Time).
+
+        In short, this method function as follows:
+            1. Get the time zone from start and end.
```
haakonvt, line 1026:
```suggestion
        In short, this method works as follows:
            1. Get the time zone from start and end (must be equal).
```

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+            2. Split the time range from start to end into intervals with fixed UTC offset.
+            3. Create a query for each interval and all the retrieve_array method.
```
haakonvt, line 1028:
```suggestion
            3. Create a query for each interval and pass all to the retrieve_arrays method.
```

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+            4. Stack the resulting arrays into a single column in the resulting DataFrame.
+
+        The queries to retrieve_arrays are translated to a multiple of hours. This means that time zones which
+        are not a whole hour offset from UTC are not supported.
```
haakonvt, line 1032: I suggest we make this into a red warning 😄 
```suggestion
        Warning:
            The queries to retrieve_arrays are translated to a multiple of hours. This means that time zones that
            are not a whole hour offset from UTC are not supported (yet). The same is true for time zones that observe DST
            with an offset from standard time that is not a multiple of 1 hour.
```

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+            uniform_index (bool): If querying aggregates and not use variable unit lengths (month, quarter, and year), specifying `uniform_index=True` will return a dataframe with an
+                equidistant datetime index from the earliest `start` to the latest `end` (missing values will be NaNs). If these requirements are not met, a ValueError is raised. Default: False
+            include_aggregate_name (bool): Include 'aggregate' in the column name, e.g. `my-ts|average`. Ignored for raw time series. Default: True
+            include_granularity_name (bool): Include 'granularity' in the column name, e.g. `my-ts|12h`. Added after 'aggregate' when present. Ignored for raw time series. Default: False
+            column_names ("id" | "external_id"): Use either ids or external ids as column names. Time series missing external id will use id as backup. Default: "external_id"
+
+        Returns:
+            pandas.DataFrame: A pandas DataFrame containing the requested time series with a DateTimeIndex localized
+            to the time zone from the start time.
```
haakonvt, line 1050:
```suggestion
            pandas.DataFrame: A pandas DataFrame containing the requested time series with a DatetimeIndex localized
            in the given time zone.
```

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+
+        Examples:
+
+            Get a pandas dataframe in the local timezone in Oslo, Norway:
```
haakonvt, line 1054:
```suggestion
            Get a pandas dataframe in the time zone of Oslo, Norway:
```

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+
+                >>> from cognite.client import CogniteClient
+                >>> try:
+                ...     from zoneinfo import ZoneInfo
+                ... except ModuleNotFoundError:
+                ...     from backports.zoneinfo import ZoneInfo
```
haakonvt, line 1060: To avoid all these backport annoyances, let's just do the import for the user?
```suggestion
                >>> # In Python >=3.9 you may import directly from `zoneinfo`
                >>> from cognite.client.utils import ZoneInfo
```

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+                >>> from cognite.client import CogniteClient
+                >>> from datetime import datetime
+                >>> try:
+                ...     from zoneinfo import ZoneInfo
+                ... except ModuleNotFoundError:
+                ...     from backports.zoneinfo import ZoneInfo
+                >>> client = CogniteClient()
+                >>> df = client.time_series.data.retrieve_dataframe(
+                ...     external_id=["foo", "bar"],
+                ...     aggregates=["sum"],
```
haakonvt, line 1082: Add an additional aggregate to further illustrate when to give a list vs a string directly? Would love one of the aggregates with an underscore 😉

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+                ...     granularity="1quarter",
+                ...     start=datetime(2020, 1, 1, tzinfo=ZoneInfo("Europe/Oslo")),
+                ...     end=datetime(2022, 12, 31, tzinfo=ZoneInfo("Europe/Oslo")),
+                ...     )
+        """
+        if (id is not None and external_id is not None) or (id is None and external_id is None):
+            raise ValueError("Either input ids or external ids")
```
haakonvt, line 1089: We have a utility function `exactly_one_is_not_none`:
```suggestion
        if not exactly_one_is_not_none(id, external_id):
            raise ValueError("Either input id(s) or external_id(s)")
```

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+                aggregates=aggregates,
+                granularity=granularity,
+                ignore_unknown_ids=ignore_unknown_ids,
+                uniform_index=uniform_index,
+                include_aggregate_name=include_aggregate_name,
+                include_granularity_name=include_granularity_name,
+                column_names=column_names,
+            )
+            df = df.tz_localize("utc").tz_convert(tz.key)
+            return df
```
haakonvt, line 1114:
```suggestion
            # For raw data, timestamps represent instants (not periods), so we can convert after fetching:
            return self.retrieve_dataframe(
                id=id,
                external_id=external_id,
                start=start,
                end=end,
                aggregates=aggregates,
                granularity=granularity,
                ignore_unknown_ids=ignore_unknown_ids,
                uniform_index=uniform_index,
                include_aggregate_name=include_aggregate_name,
                include_granularity_name=include_granularity_name,
                column_names=column_names,
            ).tz_localize("utc").tz_convert(tz.key)
```

haakonvt:
One last thing I forgot to comment on:
In order for `mypy` to accept your `ZoneInfo` imports, use `sys.version_info`:
```py
try:
     from zoneinfo import ZoneInfo
 except ImportError:
     from backports.zoneinfo import ZoneInfo
```

For some reason `mypy` understands this 🤷 (😂 ):
```py
if sys.version_info >= (3, 9):
    from zoneinfo import ZoneInfo
else:
    from backports.zoneinfo import ZoneInfo
```

scripts/create_ts_for_integration_tests.py
```diff
@@ -39,6 +43,7 @@ def create_dense_rand_dist_ts(xid, seed, n=1_000_000):
     np.random.seed(seed)
     idx = np.sort(
         np.random.randint(
+            # Windows throws a ValueError: low is out of bounds for int32 here.
```
haakonvt, line 46: Add `dtype=np.int64` to the parameters
```suggestion
```

scripts/create_ts_for_integration_tests.py
```diff
@@ -84,7 +90,7 @@ def create_edge_case_time_series(ts_api):
     print(f"Created {len(ts_lst)} sparse ts with data")
 
 
-def create_dense_time_series(ts_api):
+def create_dense_time_series() -> Tuple[List[TimeSeries], List[pd.DataFrame]]:
```
haakonvt, line 93: Why do you (randomly) add type hints here and not for any of the other functions? 👀 🤔

scripts/create_ts_for_integration_tests.py
```diff
@@ -176,25 +182,27 @@ def create_dense_time_series(ts_api):
...
         i += 5
 
-    ts_add(
-        TimeSeries(
-            name=NAMES[113],
-            external_id=NAMES[113],
-            is_string=False,
-            metadata={"offset": "n/a", "delta": "uniform random"},
+    if sys.platform != "win32":
+        # Windows does not support the low bound used for the index in create_dense_rand_dist_ts
```
haakonvt, line 186: Check the comment above. Numpy functions with default `dtype=int` is 32-bit on Windows for some reason (and not on Unix & Mac).

scripts/create_ts_for_integration_tests.py
```diff
@@ -206,10 +214,50 @@ def create_dense_time_series(ts_api):
...
+
+    ts_add(
+        TimeSeries(name=NAMES[117], external_id=NAMES[117], is_string=False, metadata={"offset": "n/a", "delta": "1M"})
+    )
+    minutely_idx = pd.date_range(start="2023-01-01", end="2023-12-31 23:59:59", freq="1min", tz="UTC")
+    df_add(pd.DataFrame(index=minutely_idx, data=np.random.normal(0, 1, len(minutely_idx)), columns=[NAMES[117]]))
+    return ts_lst, df_lst
+
+
+def create_if_not_exists(ts_api: TimeSeriesAPI, ts_list: List[TimeSeries], df_lst: List[pd.DataFrame]):
```
haakonvt, line 231: Missing `-> None`

scripts/create_ts_for_integration_tests.py
```diff
@@ -222,7 +270,11 @@ def create_dense_time_series(ts_api):
 
 if __name__ == "__main__":
     # To avoid accidental runs, please provide a valid config to CogniteClient:
-    client = CogniteClient(...)
+    from scripts import local_client
+
+    client = local_client.get_interactive()
```
haakonvt, line 275: If you want to keep this, then modify the comment above

scripts/create_ts_for_integration_tests.py
```diff
@@ -222,7 +270,11 @@ def create_dense_time_series(ts_api):
...
+
     delete_all_time_series(client.time_series)
-    create_dense_time_series(client.time_series)
+    ts_lst, df_lst = create_dense_time_series()
+    create_if_not_exists(client.time_series, ts_lst, df_lst)
     create_edge_case_time_series(client.time_series)
```
haakonvt, line 280: What about `create_edge_case_time_series` (w.r.t. "check if exist"?)

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -58,30 +68,32 @@
...
-            f"{prefix} 108: every millisecond, 1969-12-31 23:59:58.500 - 1970-01-01 00:00:01.500, numeric",
-            f"{prefix} 109: daily values, is_step=True, 1965-1975, numeric",
-            f"{prefix} 110: hourly values, is_step=True, 1969-10-01 - 1970-03-01, numeric",
-            f"{prefix} 111: every minute, is_step=True, 1969-12-31 - 1970-01-02, numeric",
-            f"{prefix} 112: every second, is_step=True, 1969-12-31 23:30:00 - 1970-01-01 00:30:00, numeric",
-            f"{prefix} 113: every millisecond, is_step=True, 1969-12-31 23:59:58.500 - 1970-01-01 00:00:01.500, numeric",
-            f"{prefix} 114: 1mill dps, random distribution, 1950-2020, numeric",
-            f"{prefix} 115: 1mill dps, random distribution, 1950-2020, string",
-            f"{prefix} 116: 5mill dps, 2k dps (.1s res) burst per day, 2000-01-01 12:00:00 - 2013-09-08 12:03:19.900, numeric",
+            f"{TEST_PREFIX}001: outside points, numeric",
```
haakonvt, line 78: This quote came to mind 😅 

image

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -58,30 +68,32 @@
...
+            f"{TEST_PREFIX}109: daily values, is_step=True, 1965-1975, numeric",
+            f"{TEST_PREFIX}110: hourly values, is_step=True, 1969-10-01 - 1970-03-01, numeric",
+            f"{TEST_PREFIX}111: every minute, is_step=True, 1969-12-31 - 1970-01-02, numeric",
+            f"{TEST_PREFIX}112: every second, is_step=True, 1969-12-31 23:30:00 - 1970-01-01 00:30:00, numeric",
+            f"{TEST_PREFIX}113: every millisecond, is_step=True, 1969-12-31 23:59:58.500 - 1970-01-01 00:00:01.500, numeric",
+            f"{TEST_PREFIX}114: 1mill dps, random distribution, 1950-2020, numeric",
+            f"{TEST_PREFIX}115: 1mill dps, random distribution, 1950-2020, string",
+            f"{TEST_PREFIX}116: 5mill dps, 2k dps (.1s res) burst per day, 2000-01-01 12:00:00 - 2013-09-08 12:03:19.900, numeric",
+            f"{TEST_PREFIX}119: hourly normally distributed (0,1) data, 2020-2024 numeric",
+            f"{TEST_PREFIX}120: minute normally distributed (0,1) data, 2023-01-01 00:00:00 - 2023-12-31 23:59:59, numeric",
```
haakonvt, line 96: I guess a good test time series would have e.g. only data during DST transitions (and oppositely maybe one year later)

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -992,6 +1023,311 @@ def test_query_with_duplicates(self, retrieve_endpoints, one_mill_dps_ts, ms_bur
             assert len(res_lst.get(external_id=ts_string.external_id)) == 2
 
 
+@pytest.fixture(scope="session")
+def hourly_2023(cognite_client, hourly_normal_dist) -> pd.DataFrame:
+    utc = zoneinfo.ZoneInfo("UTC")
+    # Adding a day to ensure we get the entire 2023 when converting to specific time zone later
```
haakonvt, line 1029: Adding? 🤔 Looking at you `start` over there

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -992,6 +1023,311 @@ def test_query_with_duplicates(self, retrieve_endpoints, one_mill_dps_ts, ms_bur
...
+        datetime(1970, 1, 1, 0, 0, 2, tzinfo=oslo),
+        "1second",
+        pd.DataFrame(
+            [1], index=[pd.Timestamp("1970-01-01 00:00:00", tz="Europe/Oslo")], columns=["count"], dtype="Int64"
+        ),
+        id="1 second granularity on minute raw data",
+    )
+
+
+class TestReprieveAggregateTimezoneDatapointsAPI:
```
haakonvt, line 1140:
```suggestion
class TestRetrieveAggregateTimezoneDatapointsAPI:
```

tests/tests_unit/test_data_classes/test_datapoints.py
```diff
@@ -0,0 +1,32 @@
...
+    @pytest.mark.dsl
+    @pytest.mark.parametrize("arrays, expected_array", list(factory_method_from_array_data()))
+    def test_factory_method_from_array(arrays: list[DatapointsArray], expected_array: DatapointsArray):
+        import numpy as np
+
+        actual_array = DatapointsArray.create_from_arrays(*arrays)
+
+        np.testing.assert_allclose(actual_array.average, expected_array.average)
+        # Comparing timestamps directly leads to an incorrect assertion fail.
+        np.testing.assert_allclose(actual_array.timestamp.astype("long"), expected_array.timestamp.astype("long"))
```
haakonvt, line 32: Hmm why? Are there Windows bugs in the `DatapointsArray` implementation?

tests/tests_unit/test_utils/test_time.py
```diff
@@ -261,3 +272,220 @@ def test_start_not_on_granularity_boundary(self, granularity, expected):
...
+    except ModuleNotFoundError:
+        return []
+    start, end = "2023-03-20", "2023-04-09 23:59:59"
+
+    yield pytest.param(
+        start,
+        end,
+        "1h",
+        "7d",
+        pd.DataFrame(index=pd.date_range(start, end, freq="7d"), data=[168, 168, 168], dtype="Int64"),
```
haakonvt, line 313: Why nullable int for pure int data?

cognite/client/data_classes/datapoints.py
```diff
@@ -225,6 +227,44 @@ def _load(  # type: ignore [override]
         dps_dct["timestamp"] = dps_dct["timestamp"].astype("datetime64[ms]").astype("datetime64[ns]")
         return cls(**convert_all_keys_to_snake_case(dps_dct))
 
+    @classmethod
+    def create_from_arrays(cls, *arrays: DatapointsArray) -> DatapointsArray:
+        sort_by_time = sorted((a for a in arrays if a.timestamp is not None), key=lambda a: a.timestamp[0])
+        first = sort_by_time[0]
+        if len(sort_by_time) == 1:
+            return first
+        # Explicit for readability
```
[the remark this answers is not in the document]
haakonvt: Disclaimer: Have not tested the code above

tests/tests_unit/test_data_classes/test_datapoints.py
haakonvt: Since numpy `datetime64` is stored as an integer under the hood, we should be doing an exact match:
```py
np.testing.assert_equal(actual_array.timestamp, expected_array.timestamp)
```

cognite/client/utils/_time.py
```diff
@@ -173,6 +387,222 @@ def split_time_range(start: int, end: int, n_splits: int, granularity_in_ms: int
...
+    # Pandas seems to have issues with ZoneInfo object, so removing the timezone and adding it back.
+    index = pd.date_range(start.replace(tzinfo=None), end.replace(tzinfo=None), freq=f"{freq}H").tz_localize(
+        start.tzinfo.key  # type: ignore
+    )
+    expected_freq = pd.Timedelta(f"{freq}H")
+    next_diff = index - index.to_series().shift(1)
+    last_diff = index - index.to_series().shift(-1)
+    index = index[(last_diff.abs() != expected_freq) | (next_diff.abs() != expected_freq)]
+
+    hour, zero = pd.Timedelta("1hour"), pd.Timedelta("0hour")
```
haakonvt, line 587: Consistency idk 🤷 
```suggestion
    hour, zero = pd.Timedelta(hours=1), pd.Timedelta(0)
```

cognite/client/utils/_time.py
```diff
@@ -173,6 +387,222 @@ def split_time_range(start: int, end: int, n_splits: int, granularity_in_ms: int
...
+
+    freq = granularity_in_hours(multiplier, unit)
+    # Pandas seems to have issues with ZoneInfo object, so removing the timezone and adding it back.
+    index = pd.date_range(start.replace(tzinfo=None), end.replace(tzinfo=None), freq=f"{freq}H").tz_localize(
+        start.tzinfo.key  # type: ignore
+    )
+    expected_freq = pd.Timedelta(f"{freq}H")
+    next_diff = index - index.to_series().shift(1)
+    last_diff = index - index.to_series().shift(-1)
+    index = index[(last_diff.abs() != expected_freq) | (next_diff.abs() != expected_freq)]
```
haakonvt, line 585: Perhaps swap subtraction on `last_` and get rid of all `abs` calls. 

Since calling `to_series()` on an index doubles the amount of data I would also suggest you only store the "mask" with boolean values to a variable:
```suggestion
    next_mask = (index - index.to_series().shift(1)) != expected_freq
    last_mask = (index.to_series().shift(-1) - index) != expected_freq
    index = index[last_mask | next_mask]
```

tests/tests_integration/test_api/test_datapoints.py
haakonvt: ...but seriously, at least keep the space before the number to improve readability

haakonvt (changes requested):
...and I'm through. Quite a lot of suggestions here 👍  Unsure if they are all worded as such 😅

cognite/client/_api/datapoint_tasks.py
```diff
@@ -135,20 +140,37 @@ def is_single_identifier(self) -> bool:
...
+        raise ValueError("Only ZoneInfo implementation of tzinfo is supported")
+
+    if start.tzinfo != end.tzinfo:
+        raise ValueError(
+            # Mypy fail to understand that we are guaranteed to have ZoneInfo objects, which have the key attribute.
+            f"start and end have different timezones, {start.tzinfo.key!r} and {end.tzinfo.key!r}. "  # type:ignore
+            "They must be the same."
         )
+
+    if any(timestamp.utcoffset().total_seconds() % 3600 != 0 for timestamp in (start, end)):  # type:ignore
```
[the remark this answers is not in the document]
doctrino: It was an attempt to do something for checking that case, but I agree doing it in the docstring is better.

cognite/client/_api/datapoint_tasks.py
```diff
@@ -164,6 +186,60 @@ def _ts_to_ms_frozen_now(self, ts: Union[int, str, datetime, None], default: int
...
+        self.defaults: Dict[str, Any] = dict(
+            start=user_query.start,
+            end=user_query.end,
+            limit=user_query.limit,
+            aggregates=user_query.aggregates,
+            granularity=user_query.granularity,
+            include_outside_points=user_query.include_outside_points,
+            ignore_unknown_ids=user_query.ignore_unknown_ids,
+        )
+        self.time_range_validator = _TimeRangeValidator()
```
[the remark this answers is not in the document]
doctrino: Sorry this is my mistake, I refactored it out for reuse. But later ended up not reusing it in `retrieve_dataframe_in_tz`

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
-            return df
-
         start = pd.Timestamp(min(q.start for q in fetcher.agg_queries), unit="ms")
         end = pd.Timestamp(max(q.end for q in fetcher.agg_queries), unit="ms")
         (granularity,) = grans_given
         # Pandas understand "Cognite granularities" except `m` (minutes) which we must translate:
         freq = cast(str, granularity).replace("m", "T")
         return df.reindex(pd.date_range(start=start, end=end, freq=freq, inclusive="left"))
 
+    Aggregates = Literal[
```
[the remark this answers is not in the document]
doctrino: I was hoping for some feedback on this one, will remove it, and instead introduce it in a separate PR across all the retrieve methods.

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+        Args:
+            start (datetime): Inclusive start, must be time zone aware.
+            end (datetime): Exclusive end, must be time zone aware and have the same time zone as start.
+            id (Union[int | Sequence[int] | None]): ID or list of IDs.
+            external_id (Union[int | Sequence[int] | None]): External ID or list of External IDs.
+            aggregates (Union[Literal, None]): Single aggregate or list of aggregates to retrieve. Default: None (raw datapoints returned)
+            granularity (str): The granularity to fetch aggregates at supported second, minute, hour, day, week, month, quarter and year. Default: None.
+            ignore_unknown_ids (bool): Whether to ignore missing time series rather than raising an exception. Default: False
+            uniform_index (bool): If querying aggregates and not use variable unit lengths (month, quarter, and year), specifying `uniform_index=True` will return a dataframe with an
+                equidistant datetime index from the earliest `start` to the latest `end` (missing values will be NaNs). If these requirements are not met, a ValueError is raised. Default: False
```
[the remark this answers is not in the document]
doctrino: It is a good point. Then, I am leaning towards dropping it. Not sure if it is unambiguous what a uniform monthly index is? But then you do not have then same length between all timestamps, so is the index really uniform?

Wdyt?

cognite/client/_api/datapoints.py
doctrino: Of course, that makes sense.

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+        multiplier, unit = get_granularity_multiplier_and_unit(granularity)  # type:ignore
+        if uniform_index and unit in VARIABLE_LENGTH_UNITS:
+            raise ValueError(
+                "Uniform index is not supported with a variable step length unit"
+                f" such as {', '.join(VARIABLE_LENGTH_UNITS)}."
+            )
+
+        intervals = to_fixed_utc_intervals(start, end, granularity)  # type:ignore
+
+        id_name, ids = ("external_id", external_id) if external_id else ("id", id)
```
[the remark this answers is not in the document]
doctrino: Good to know, doing the swap solution, assume 0 is not a valid id.

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+
+        The queries to retrieve_arrays are translated to a multiple of hours. This means that time zones which
+        are not a whole hour offset from UTC are not supported.
+
+        Args:
+            start (datetime): Inclusive start, must be time zone aware.
+            end (datetime): Exclusive end, must be time zone aware and have the same time zone as start.
+            id (Union[int | Sequence[int] | None]): ID or list of IDs.
+            external_id (Union[int | Sequence[int] | None]): External ID or list of External IDs.
+            aggregates (Union[Literal, None]): Single aggregate or list of aggregates to retrieve. Default: None (raw datapoints returned)
```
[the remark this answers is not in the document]
doctrino: replaced by string

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+                ...     from backports.zoneinfo import ZoneInfo
+                >>> client = CogniteClient()
+                >>> df = client.time_series.data.retrieve_dataframe(
+                ...     external_id=["foo", "bar"],
+                ...     aggregates=["sum"],
+                ...     granularity="1quarter",
+                ...     start=datetime(2020, 1, 1, tzinfo=ZoneInfo("Europe/Oslo")),
+                ...     end=datetime(2022, 12, 31, tzinfo=ZoneInfo("Europe/Oslo")),
+                ...     )
+        """
```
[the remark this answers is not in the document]
doctrino: What do you want to illustrate with such an example?

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+
+        # Aggregates
+        multiplier, unit = get_granularity_multiplier_and_unit(granularity)  # type:ignore
+        if uniform_index and unit in VARIABLE_LENGTH_UNITS:
+            raise ValueError(
+                "Uniform index is not supported with a variable step length unit"
+                f" such as {', '.join(VARIABLE_LENGTH_UNITS)}."
+            )
+
+        intervals = to_fixed_utc_intervals(start, end, granularity)  # type:ignore
```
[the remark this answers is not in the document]
doctrino: Thanks. My skepticism about the utility of mypy was renewed with this PR, very few bugs detected, lots of time spent satisfying it thirst for strictness and some overly verbose code as a result.

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+                uniform_index=uniform_index,
+                include_aggregate_name=include_aggregate_name,
+                include_granularity_name=include_granularity_name,
+                column_names=column_names,
+            )
+            df = df.tz_localize("utc").tz_convert(tz.key)
+            return df
+
+        # Aggregates
+        multiplier, unit = get_granularity_multiplier_and_unit(granularity)  # type:ignore
```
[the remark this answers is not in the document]
doctrino: Thanks, I will do. But I will add my frustration I do not like that `mypy` changes adds additional lines of code that are there only for mypy.

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +965,189 @@ def retrieve_dataframe(
...
+        queries = [
+            {id_name: id_, "aggregates": aggregates, **interval}
+            for id_, interval in itertools.product(ids, intervals)  # type:ignore
+        ]
+        arrays = self.retrieve_arrays(
+            limit=None,
+            ignore_unknown_ids=ignore_unknown_ids,
+            **{id_name: queries},  # type:ignore
+        )
+        arrays = DatapointsArrayList.create_with_unique_ids(arrays)  # type:ignore
```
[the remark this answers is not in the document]
doctrino: I think that is a good idea, I guess that could save some memory as well.

cognite/client/data_classes/datapoints.py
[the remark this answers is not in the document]
doctrino: I did not see the `_data_fields` method. That changes it, will do the implementation above. Wanted to avoid doing some dunder magic.

scripts/create_ts_for_integration_tests.py
doctrino: Inconsistent force of habit :)

tests/tests_integration/test_api/test_datapoints.py
doctrino: Thanks, I try to not go to overboard. This one comes from way to many bugs caused by misspelling a string, so better to have all strings as variables (also get easy lookup of their use as a bonus).

scripts/create_ts_for_integration_tests.py
doctrino: Laziness only commented it out. The create if not exist was to avoid deleting everything, when I has some windows issues with the creation. (Forgot that windows like np.float32). Fixing it.

tests/tests_unit/test_utils/test_time.py
```diff
@@ -261,3 +272,220 @@ def test_start_not_on_granularity_boundary(self, granularity, expected):
...
+
+        assert actual_start == expected_start, "Start is not aligning"
+        assert actual_end == expected_end, "End is not aligning"
+
+
+def cdf_aggregate_test_data():
+    try:
+        import pandas as pd
+    except ModuleNotFoundError:
+        return []
```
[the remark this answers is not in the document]
doctrino: I did not realize pytest.mark can be used on any function, not just test functions.

cognite/client/utils/_time.py
```diff
@@ -152,18 +164,220 @@ def convert_time_attributes_to_datetime(item: Union[Dict, List[Dict]]) -> Union[
...
+    def ceil(cls, date: datetime) -> datetime:
+        """
+        Ceils the date to the next monday
+        >>> WeekAligner.ceil(datetime(2023, 4, 9 ))
+        datetime.datetime(2023, 4, 10, 0, 0)
+        """
+        date = date.replace(hour=0, minute=0, microsecond=0)
+        if (weekday := date.weekday()) != 0:
+            return date + timedelta(days=7 - weekday)
+        return date
```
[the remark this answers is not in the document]
doctrino: I considered that, but my thinking is that can be extended later if there is a need.

cognite/client/utils/_time.py
```diff
@@ -152,18 +164,220 @@ def convert_time_attributes_to_datetime(item: Union[Dict, List[Dict]]) -> Union[
...
+        Floors the months to start of quarter.
+        >>> QuarterAligner.floor(datetime(2023, 3, 1))
+        datetime.datetime(2023, 1, 1, 0, 0)
+        >>> QuarterAligner.floor(datetime(2023, 8, 1))
+        datetime.datetime(2023, 7, 1, 0, 0)
+        >>> QuarterAligner.floor(datetime(2023, 12, 1))
+        datetime.datetime(2023, 10, 1, 0, 0)
+        """
+        month = 3 * ((date.month - 1) // 3) + 1
+        return date.replace(month=month, day=1, hour=0, minute=0, second=0, microsecond=0)
```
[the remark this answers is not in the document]
doctrino: and not have the opportunity do do some floor division? :)

cognite/client/utils/_time.py
```diff
@@ -173,6 +387,222 @@ def split_time_range(start: int, end: int, n_splits: int, granularity_in_ms: int
...
+            "end": end.to_pydatetime(),
+            "granularity": f"{(end-start).total_seconds()//3600:.0f}h",
+        }
+        for start, end in zip(index[:-1], index[1:])
+    ]
+
+
+def _to_fixed_utc_intervals_fixed_unit_length(
+    start: datetime, end: datetime, multiplier: int, unit: str, utc: zoneinfo.ZoneInfo
+) -> list[dict[str, datetime | str]]:
```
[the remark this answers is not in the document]
doctrino: Agree, but not sure how to do that, as I cannot import ZoneInfo on the on the modul-leve. Suggestions?

cognite/client/utils/_time.py
```diff
@@ -173,6 +387,222 @@ def split_time_range(start: int, end: int, n_splits: int, granularity_in_ms: int
...
+        .shift(-step)
+        .iloc[::step]
+    )
+
+
+def to_fixed_utc_intervals(start: datetime, end: datetime, granularity: str) -> list[dict[str, datetime | str]]:
+    try:
+        from zoneinfo import ZoneInfo  # type:ignore
+    except ImportError:
+        from backports.zoneinfo import ZoneInfo  # type:ignore
```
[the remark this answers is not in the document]
doctrino: Not allowed as this is a optional dependency until 3.9 is required. Or do you know a way to solve that?

cognite/client/utils/_time.py
```diff
@@ -173,6 +387,222 @@ def split_time_range(start: int, end: int, n_splits: int, granularity_in_ms: int
...
+        index = index[(index.month == 1) & (year_no % multiplier == 0)].tz_convert(utc)
+    else:
+        raise ValueError(f"Unit {unit} is not supported.")
+    return [
+        {
+            "start": start.to_pydatetime(),
+            "end": end.to_pydatetime(),
+            "granularity": f"{(end-start).total_seconds()//3600:.0f}h",
+        }
+        for start, end in zip(index[:-1], index[1:])
```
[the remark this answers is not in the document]
doctrino: That was my first implementation, and a little loud f**k when I realized it is only from 3.10

cognite/client/utils/_time.py
```diff
@@ -173,6 +387,222 @@ def split_time_range(start: int, end: int, n_splits: int, granularity_in_ms: int
...
+    elif unit in {"day", "days", "d"}:
+        return "day"
+    elif unit in {"weeks", "w", "week"}:
+        return "week"
+    elif unit in {"months", "month"}:
+        return "month"
+    elif unit in {"quarters", "quarter", "q"}:
+        return "quarter"
+    elif unit in {"year", "years", "y"}:
+        return "year"
```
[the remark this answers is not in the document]
doctrino: The problem I have is the choice of m for minutes, as (https://strftime.org/)[https://strftime.org/] uses m for month and M for minutes. We could allow Y, Q, S, and H. But not sure how to deal with m/M.

tests/tests_integration/test_api/test_datapoints.py
doctrino: What about: 
* 121 some burst of datapoints on transition dates
* 122 Burst scattered on all dates, but the transition dates.

cognite/client/utils/_time.py
```diff
@@ -173,6 +395,152 @@ def split_time_range(start: int, end: int, n_splits: int, granularity_in_ms: int
...
+
+    return start.tzinfo
+
+
+def to_pandas_freq(granularity: str, start: datetime) -> str:
+    multiplier, unit = get_granularity_multiplier_and_unit(granularity, standardize=True)
+
+    unit = {"s": "S", "m": "T", "h": "H", "d": "D", "w": "W-MON", "month": "MS", "quarter": "QS", "year": "AS"}.get(
+        unit, unit
+    )
```
haakonvt, line 536: Awesome!! This would allow us to more easily add week days later! 🚀

haakonvt (changes requested):
Last round of comments 🤩

haakonvt:
@doctrino one last _actual_ bug: 

This comment above, which was two different issues (sorry about that), was not resolved: https://github.com/cognitedata/cognite-sdk-python/pull/1189#discussion_r1163932032

Example that breaks:
```py
client.time_series.data.retrieve_dataframe_in_tz(
    external_id=xid,
    start=datetime(2020, 1, 1, tzinfo=utc),
    end=datetime(2024, 2, 1, tzinfo=utc),
    aggregates="average",
    granularity="12years",
)
```
This gives the error message from the API:
```
CogniteAPIError: Invalid aggregate size: 105192, limit is 100000 for HOUR | code: 400 | X-Request-ID: a161a30a-a736-9c53-8d64-b98bbd367938
```

cognite/client/utils/_identifier.py
```diff
@@ -31,24 +31,19 @@ def load(cls, id: Optional[int] = None, external_id: Optional[str] = None) -> Id
             return Identifier(external_id)
         raise ValueError("At least one of id and external id must be specified")
 
+    def name(self, camel_case: bool = False) -> str:
+        if isinstance(self.__value, int):
+            return "id"
+        return "externalId" if camel_case else "external_id"
```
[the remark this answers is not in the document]
doctrino: Thanks, decent golfing then :)

cognite/client/utils/_time.py
doctrino: Should it be applied in the old `retrieve_dataframe `as well? There there is a `.replace("m", "T") `to achieve the same.

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +967,183 @@ def retrieve_dataframe(
...
+                index given by the granularity datetime index from the earliest `start` to the latest `end` (missing values will be NaNs). Default: False
+            include_aggregate_name (bool): Include 'aggregate' in the column name, e.g. `my-ts|average`. Ignored for raw time series. Default: True
+            include_granularity_name (bool): Include 'granularity' in the column name, e.g. `my-ts|12h`. Added after 'aggregate' when present. Ignored for raw time series. Default: False
+            column_names ("id" | "external_id"): Use either ids or external ids as column names. Time series missing external id will use id as backup. Default: "external_id"
+
+        Returns:
+            pandas.DataFrame: A pandas DataFrame containing the requested time series with a DatetimeIndex localized
+            in the given time zone.
+
+        Examples:
```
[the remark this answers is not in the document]
doctrino: I wrote a Note section at the end. Furthermore, I throw a lower case on the unit, as I think this makes the granularity argument clearer to input, and simple say that month does not have a one letter equivalent.

cognite/client/_api/datapoints.py
```diff
@@ -959,27 +967,183 @@ def retrieve_dataframe(
...
+        if exactly_one_is_not_none(aggregates, granularity):
+            raise ValueError(
+                "Got only one of 'aggregates' and 'granularity'. "
+                "Pass both to get aggregates, or neither to get raw data"
+            )
+
+        tz = validate_timezone(start, end)
+        if aggregates is None and granularity is None:
+            # Raw Data only need to convert the timezone
+            return (
```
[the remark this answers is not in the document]
doctrino: It looks that is handled by the `retrieve_dataframe `method. Isn't it redundant to raise it here then?

cognite/client/utils/__init__.py
```diff
@@ -14,3 +16,11 @@
...
     timestamp_to_ms,
 )
+
+
+# Needed for doctest to pass.
+with contextlib.suppress(ImportError):
+    if sys.version_info >= (3, 9):
+        from zoneinfo import ZoneInfo
+    else:
+        from backports.zoneinfo import ZoneInfo
```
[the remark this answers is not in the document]
doctrino: went with the import_zoneinfo for clarity. But it is a good idea to introduce the function as it will replacing all these imports easier when we go to 3.9

cognite/client/utils/_time.py
haakonvt: Let's add that separately/later. The function will reject any "non-valid" granularities currently and I think it would be nice for us to add week/month/quarter/year

cognite/client/_api/datapoints.py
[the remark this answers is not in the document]
haakonvt: Lol, brainfart on my part. Agree 💯 %

cognite/client/_api/datapoints.py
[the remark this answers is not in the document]
haakonvt: Epic 👌

tests/utils.py
```diff
@@ -146,3 +151,71 @@ def set_request_limit(client, limit):
...
+        return grouping.sum()
+    elif aggregate == "count":
+        return grouping.count().astype("Int64")
+
+    # The average is calculated by the formula '1/(b-a) int_a^b f(t) dt' where f(t) is the continuous function
+    # This is weighted average of the sampled version of f(t)
+    np = cast(Any, local_import("numpy"))
+
+    def integrate_average(values: pandas.DataFrame) -> pandas.Series:
+        return pd.Series(((values.iloc[1:].values + values.iloc[:-1].values) / 2.0).mean(), index=values.columns)
```
haakonvt, line 196: Avoids runtimewarning from numpy:
```suggestion
    def integrate_average(values: pandas.DataFrame) -> pandas.Series:
        inner = ((values.iloc[1:].values + values.iloc[:-1].values) / 2.0)
        res = inner.mean() if inner.shape[0] else np.nan
        return pd.Series(res, index=values.columns)
```

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -992,6 +1023,361 @@ def test_query_with_duplicates(self, retrieve_endpoints, one_mill_dps_ts, ms_bur
...
+        granularity: str,
+        tz_name: str,
+        cognite_client: CogniteClient,
+        hourly_2023: pd.DataFrame,
+    ):
+        # Arrange
+        tz = ZoneInfo(tz_name)
+        start = datetime(2023, 1, 1, tzinfo=tz)
+        end = datetime(2023, 12, 31, 23, 0, 0, tzinfo=tz)
+        raw_df = hourly_2023.tz_convert(tz_name).loc[start:end].copy()
```
haakonvt, line 1238: Avoid ignored (but still warned) AttributeError from pandas cinit (doesn't like ZoneInfo in start/end). Could make an additional pd_start and pd_end, but a clever hack is just to make them strings:
```suggestion
        raw_df = hourly_2023.tz_convert(tz_name).loc[str(start):str(end)].copy()
```

haakonvt (changes requested):
Some warnings stuff++

cognite/client/_api/datapoints.py
```diff
@@ -1124,7 +1124,7 @@ def retrieve_dataframe_in_tz(
             raise ValueError(f"The following identifiers were not unique: {duplicated}")
 
         queries = [
-            {**ident_dct, **interval}  # type: ignore [arg-type]
+            {**ident_dct, "aggregates": aggregates, **interval}  # type: ignore [arg-type]
```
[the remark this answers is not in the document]
doctrino: Found it, already fixed.

cognite/client/_api/datapoints.py
```diff
@@ -592,6 +595,7 @@ def __init__(self, config: ClientConfig, api_version: Optional[str], cognite_cli
         self._DPS_INSERT_LIMIT = 100_000
         self._RETRIEVE_LATEST_LIMIT = 100
         self._POST_DPS_OBJECTS_LIMIT = 10_000
+        self._GRANULARITY_HOURS_LIMIT = 100_000.0
```
haakonvt, line 598: ...float? 🤔

cognite/client/_api/datapoints.py
```diff
@@ -1114,13 +1118,19 @@ def retrieve_dataframe_in_tz(
                 .tz_convert(tz.key)
             )
 
+        assert isinstance(granularity, str)  # mypy
+
+        if in_timedelta(granularity) / timedelta(hours=1) > self._GRANULARITY_HOURS_LIMIT:
+            multiplier, unit = get_granularity_multiplier_and_unit(granularity)
+            days = _unit_in_days(unit)
+            limit = math.floor(timedelta(hours=self._GRANULARITY_HOURS_LIMIT) / timedelta(days=days))
+            raise ValueError(f"Granularity above then maximum limit, {limit} {unit}s.")
```
haakonvt, line 1128:
```suggestion
            raise ValueError(f"Granularity above the maximum limit, {limit} {unit}s.")
```

cognite/client/utils/_time.py
```diff
@@ -556,3 +557,39 @@ def to_pandas_freq(granularity: str, start: datetime) -> str:
...
+        arg = {unit: 1}
+        return timedelta(**arg) / timedelta(hours=1)
+
+    if unit == "month":
+        days = 31.0 if ceil else 28.0
+    elif unit == "quarter":
+        days = 92.0 if ceil else 91.0
+    else:  # years
+        days = 366.0 if ceil else 365.0
+    return days
```
haakonvt, line 574: Is this only used for approximation on when to raise? (Since `month` typically is more likely to be 30 than 28)

If so (_should not be used for precise calculations_), let's state that explicitly in the docstring (or function name maybe)

----

Last option, just hardcode the max upper limits for week, month, quarter and year? 🤔 🤷

tests/tests_unit/test_api/test_datapoints.py
```diff
@@ -773,7 +774,7 @@ class TestRetrieveDataPointsInTz:
                 },
                 "Europe/Oslo",
                 "Europe/Oslo",
-                "Granularity above maximum aggregation limit of 12 years.",
+                "Granularity above then maximum limit, 11 years.",
```
haakonvt, line 777:
```suggestion
                "Granularity above the maximum limit, 11 years.",
```

tests/tests_unit/test_api/test_datapoints.py
```diff
@@ -786,7 +787,7 @@ class TestRetrieveDataPointsInTz:
                 },
                 "Europe/Oslo",
                 "Europe/Oslo",
-                "Granularity above maximum aggregation limit of 46 quarters.",
+                "Granularity above then maximum limit, 45 quarters.",
```
haakonvt, line 790:
```suggestion
                "Granularity above the maximum limit, 45 quarters.",
```

haakonvt (commented):
Last comments 🤞

cognite/client/_api/datapoints.py
doctrino: just to avoid mypy bugging me, but can make it into an int

cognite/client/utils/_time.py
doctrino: Made it private to avoid it being used, but will add that to the docstring as well

haakonvt (approved):
# 🕑 ▶️ 🕒 
Let's gooooo! 🥇

PR #1190, opened 2023-03-30
Update __init__.py

nimeshawij:
changed variable name TypeExtId to TypeExternalId, line number 61,66,71


erlendvollset (changes requested):
Please update the title and description of the PR. They should reflect what changes are being made and why, not which files or line numbers have been changed.

PR #1193, opened 2023-04-03
Retry CogniteConnectionRefused

akhramov:
## Description
Since CogniteConnectionRefused could be common, it makes sense to retry it.

This change

* Adds automatic retries for `CogniteConnectionRefused`
* Adds a user-friendly message for `CogniteConnectionRefused`

## Checklist:
- [x] Tests added/updated.


erlendvollset:
It was implemented in this way under the assumption that ECONNREFUSED is not a transient error. It is thrown if e.g. hostname is misspelled, which would never succeed no matter how many retries was applied. In what cases is it not a transient error?

erlendvollset:
Not that we retry all other connection errors, we only exclude ECONNREFUSED

erlendvollset (approved):
LGTM!

PR #1194, opened 2023-04-05
Add missing nullable int columns in `to_pandas`

haakonvt:
## Description
Quite a lot of nullable int columns were missing. This attempts to add all of them (while not touching ones like `id` which should always have a value (in the API response), except when the user instantiates them).

### Note
I'm not adding changelog entry & bumping version before this has been approved and is ready to merge to avoid the _never-ending merge conflicts with these three files_.

### DIFF
--- a/cognite/client/utils/_pandas_helpers.py
+++ b/cognite/client/utils/_pandas_helpers.py
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
     from cognite.client.data_classes._base import T_CogniteResource, T_CogniteResourceList
 
 
-NULLABLE_INT_COLS = [
+NULLABLE_INT_COLS = {
     "start_time",
     "end_time",
     "asset_id",
@@ -32,8 +32,8 @@ NULLABLE_INT_COLS = [
     "last_seen",
     "last_seen_time",
     "last_updated_time",
-]
-NULLABLE_INT_COLS_CC = list(map(to_camel_case, NULLABLE_INT_COLS))
+}
+NULLABLE_INT_COLS_CC = set(map(to_camel_case, NULLABLE_INT_COLS))
 
 
 def pandas_major_version() -> int:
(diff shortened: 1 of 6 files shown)


cognite/client/utils/_pandas_helpers.py
```diff
@@ -9,13 +9,33 @@
...
+    "schedule_id",
+    "session_id",
+    "deleted_time",
+    "last_success",
+    "last_failure",
+    "last_seen",
+    "last_seen_time",
+    "last_updated_time",
+}
+NULLABLE_INT_COLS_CC = set(map(to_camel_case, NULLABLE_INT_COLS))
```
erlendvollset, line 36:
```suggestion
NULLABLE_INT_COLS_CAMEL_CASE = set(map(to_camel_case, NULLABLE_INT_COLS))
```

cognite/client/utils/_pandas_helpers.py
```diff
@@ -37,6 +57,13 @@ def notebook_display_with_fallback(inst: Union[T_CogniteResource, T_CogniteResou
         return str(inst)
 
 
+def convert_nullable_int_cols(df: pd.DataFrame, camel_case: bool) -> pd.DataFrame:
```
erlendvollset, line 60:
```suggestion
def convert_nullable_int_cols_to_int64(df: pd.DataFrame, camel_case: bool) -> pd.DataFrame:
```

erlendvollset (approved):
LGTM, but wouldn't mind slightly more descriptive names!

cognite/client/utils/_pandas_helpers.py
haakonvt: This one is actually wrong; "int64" is "_not nullable int_", while "Int64" is nullable, so I'm keeping the original 🤙

PR #1201, opened 2023-04-17
V6

erlendvollset:
- Remove support for legacy auth
- Remove deprecated properties and parameters


qTipTip (approved):
LGTMlops 🚀

cognite/client/_cognite_client.py
haakonvt, line 1: Remove:
```
@property  # TODO (v6.0.0): Delete this whole property
    def extraction_pipeline_runs(self) -> ExtractionPipelineRunsAPI:
```

haakonvt, line 1: Remove:
```
@property  # TODO (v6.0.0): Delete this whole property
    def datapoints(self) -> DatapointsAPI:
        if int(self.version.split(".")[0]) >= 6:
```

CHANGELOG.md
haakonvt, line 1: Remove deprecated camelCase parameter support from:
```
cognite/client/_api/extractionpipelines.py (one use)

cognite/client/data_classes/shared.py (two uses)
```

haakonvt (changes requested):
There are a few properties & parameters that should be removed once moving to major version 6. I think it would make sense to hold off on the major version bump for as long as possible, and rather add warnings or similar?

We need to keep in mind each major version bump will keep a loooot of our customers from receiving updates to the SDK for a long time since we do not keep updating v5 with minor/patch releases after moving to 6 😬

haakonvt (changes requested):
Let's also add an entry in `MIGRATION_GUIDE.md` 🚀

haakonvt (approved):
LGTM. Could you fix the bad date on line 35 in the change log? It should not be November 😅 
```
## [5.11.0] - 17-11-23
```

PR #1203, opened 2023-04-18
Explicitly export types

erlendvollset:
- Use dmypy
- Explicitly export types
- Update changelog and bump version

We should ideally typecheck the test code base as well as that would help in catching more unexported types.


dbrattli (approved):
LGTM!

ponnapz (approved):

PR #1207, opened 2023-04-24
DatapointsAPI Signatures

### LINKED ISSUE #1199: Aggregates as a Literal
I like introducing the possible aggregates as a Literal, but I wonder if we should move it to a different PR?

We already have `ALL_SORTED_DP_AGGS` in `cognite/client/data_classes/datapoints.py`, and we could then change its definition to something like:
```
Aggregate = Literal[
    [
        "average",
        "continuous_variance",
        "count",
        ...,
    ]
]
ALL_SORTED_DP_AGGS = sorted(typing.get_args(Aggregate))
```
Regardless:

- In either case, remove the plural "s" from `Aggregates`.
- Accept camel case and snake case, but use snake case in the Literal to be consistent with the documentation on the other retrieve-X methods

_Originally posted by @haakonvt in https://github.com/cognitedata/cognite-sdk-python/pull/1189#discussion_r1162696663_

doctrino:
## Description

The goal is to make the signatures of the DatapointsAPI methods even more user friendly.

closes #1199 

* Replaced the `str` type hint with `Literal`
* Allow the `start` and `end` arguments to be `pandas.Timestamp` for the `retrieve_dataframe` methods
* Modernize the type hints, moving from `Union[...]` to `...|...`
* Support for giving `start` and `end` arguments in all `retrieve` methods as string.

## Checklist:
- [x] Tests added/updated.


haakonvt (commented):
Some thoughts

CHANGELOG.md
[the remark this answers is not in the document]
doctrino: That thought crossed my mind, and I think you are right. We just accept all `pd.Timestamp`. 

Will wait to do any changes until @erlendvollset has given his comments.

cognite/client/utils/_time.py
```diff
@@ -46,6 +46,14 @@ def import_zoneinfo() -> ZoneInfo:
         ) from e
 
 
+def import_zoneinfo_error() -> ZoneInfoNotFoundError:
+    if sys.version_info >= (3, 9):
+        from zoneinfo import ZoneInfoNotFoundError
+    else:
+        from backports.zoneinfo import ZoneInfoNotFoundError
+    return ZoneInfoNotFoundError  # type: ignore [return-value]
```
erlendvollset, line 54: This is redundant since we do the same conditional import on l18-l21, right?

cognite/client/_api/datapoints.py
```diff
@@ -600,16 +603,16 @@ def __init__(self, config: ClientConfig, api_version: Optional[str], cognite_cli
...
-        id: Union[None, int, Dict[str, Any], Sequence[Union[int, Dict[str, Any]]]] = None,
-        external_id: Union[None, str, Dict[str, Any], Sequence[Union[str, Dict[str, Any]]]] = None,
-        start: Union[int, str, datetime, None] = None,
-        end: Union[int, str, datetime, None] = None,
-        aggregates: Union[str, List[str], None] = None,
+        id: None | int | dict[str, Any] | Sequence[int | dict[str, Any]] = None,
+        external_id: None | str | dict[str, Any] | Sequence[str | dict[str, Any]] = None,
+        start: int | str | datetime | None = None,
+        end: int | str | datetime | None = None,
+        aggregates: AGGREGATE | List[AGGREGATE] | None = None,
```
erlendvollset, line 610: I agree that this is better, but changing this from `str` to `Literal` makes the type more strict and may break clients using a static type checker. If they've wrapped this method, and are passing a variable of type `str` it will start failing. e.g.
```
def my_dps_wrapper(id: int, ..., aggregate: str) -> Datapoints:
    return cognite_client.datapoints.retrieve(id=id, ..., aggregate=aggregate)

res = my_dps_wrapper(id=123, ..., aggregate="max")
```
is perfectly fine at runtime, but will not pass static type checking

cognite/client/_api/datapoints.py
```diff
@@ -600,16 +603,16 @@ def __init__(self, config: ClientConfig, api_version: Optional[str], cognite_cli
     def retrieve(
         self,
         *,
-        id: Union[None, int, Dict[str, Any], Sequence[Union[int, Dict[str, Any]]]] = None,
-        external_id: Union[None, str, Dict[str, Any], Sequence[Union[str, Dict[str, Any]]]] = None,
-        start: Union[int, str, datetime, None] = None,
-        end: Union[int, str, datetime, None] = None,
-        aggregates: Union[str, List[str], None] = None,
+        id: None | int | dict[str, Any] | Sequence[int | dict[str, Any]] = None,
```
erlendvollset, line 606: TIL! Didn't know pep604 and pep585 were backported to 3.8.

erlendvollset: We can make the pyupgrade linter in ruff apply this consistently accross the code base in a separate PR. (We may have to manually add `from __future__ import annotations` import statements where they're missing)

haakonvt: > TIL! Didn't know pep604 and pep585 were backported to 3.8.

Don't you remember reviewing this (`cognite/client/utils/_validation.py`) a few weeks back? 😜 

> We can make the pyupgrade linter in ruff apply this consistently across the code base in a separate PR

Although the future-import allows most of 604 and 585, there was a reason why I used the `pyupgrade` setting `--keep-runtime-typing` (to keep it from doing exactly this) as it ruined some of the `cast(<type(s)>, variable)`; these do not support... before:

```py
cast(list[str], 123)  # 3.9+
cast(str | list, 123)  # 3.10+
```

Maybe ruff is more clever and keeps these untouched?

> We may have to manually add from __future__ import annotations import statements where they're missing
There must be a code mod tool to auto-add the future-import, right 🤔  If not, back when I added all these, I wrote a quick script which we could add to pre-commit (since we already use a custom check that checks version).

CHANGELOG.md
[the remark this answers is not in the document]
doctrino: Did not know that` pd.Timestamp` is a subclass of datetime, and then I agree no need to state that explicitly.

cognite/client/_api/datapoints.py
doctrino: It is a good point, what about doing a union? Then, it is clear what aggregates exists, and at the same time we do not break static type checkers.
```suggestion
        aggregates: Union[AGGREGATE, str] | List[AGGREGATE | str] | None = None,
```

cognite/client/utils/_time.py
doctrino: No, this is the import at runtime, wile the l18-l21 is only type checking as far as I understand it.

CHANGELOG.md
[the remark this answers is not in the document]
doctrino: Added an example for other people like me that did not know.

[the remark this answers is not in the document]
doctrino: Fixed

erlendvollset (approved):
LGTM

cognite/client/_api/datapoints.py
```diff
@@ -947,13 +948,24 @@ def retrieve_dataframe(
...
-            starting Jan 1, 1970 all the way up to present, without having the aggregate name in the column names::
+            starting Jan 1, 1970 all the way up to present, without having the aggregate name in the column names:
 
                 >>> df = client.time_series.data.retrieve_dataframe(
                 ...     external_id=["foo", "bar"],
                 ...     aggregates=["average"],
                 ...     granularity="30d",
                 ...     include_aggregate_name=False)
+
+            Remember that pandas.Timestamp is is a subclass of datetime, so you can use a Timestamps as start and
```
haakonvt, line 959: is is

haakonvt: + Timestamp without s

haakonvt: Or remove the “a” before

PR #1208, opened 2023-04-25
Ambiguous time error

doctrino:
## Description
There is a bug when using start and end time between 1900 and 2023 with a granularity of month for Europe/Oslo
```bash
cognite\client\_api\datapoints.py:1135: in retrieve_dataframe_in_tz
    intervals = to_fixed_utc_intervals(start, end, granularity)
cognite\client\utils\_time.py:459: in to_fixed_utc_intervals
    return _to_fixed_utc_intervals_variable_unit_length(start, end, multiplier, unit)
cognite\client\utils\_time.py:468: in _to_fixed_utc_intervals_variable_unit_length
    index = pandas_date_range_tz(start, end, freq)
cognite\client\utils\_time.py:524: in pandas_date_range_tz
    return pd.date_range(start.replace(tzinfo=None), end.replace(tzinfo=None), freq=freq, inclusive=inclusive).tz_localize(start.tzinfo.key)  # type: ignore [union-attr]
..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\cognite-sdk-CLXl76zR-py3.8\lib\site-packages\pandas\core\indexes\datetimes.py:281: in tz_localize
    arr = self._data.tz_localize(tz, ambiguous, nonexistent)
..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\cognite-sdk-CLXl76zR-py3.8\lib\site-packages\pandas\core\arrays\_mixins.py:85: in method
    return meth(self, *args, **kwargs)
..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\cognite-sdk-CLXl76zR-py3.8\lib\site-packages\pandas\core\arrays\datetimes.py:987: in tz_localize
    new_dates = tzconversion.tz_localize_to_utc(

>   ???
E   pytz.exceptions.AmbiguousTimeError: Cannot infer dst time from 1916-10-01 00:00:00, try using the 'ambiguous' argument

pandas\_libs\tslibs\tzconversion.pyx:353: AmbiguousTimeError
```

### Result

After some digging, it seems this is a bug deep inside `pandas`, in the c-code. The c-class `MonthOffset` fails in the _apply function with Timestamp("1916-09-01 00:00:00+0200", tz="Europe/Oslo), i.e., [the DST transition in 1916](https://www.timeanddate.com/time/zone/norway/oslo?year=1916). Line 2578 in `pandas.core.arrays.datetime.py`. 

Suggested solution is to catch the error `AmbiguousTimeError` and give a more user friendly error to the user. 

In addition, bug report is submitted to `pandas`: https://github.com/pandas-dev/pandas/issues/52908

**The problem boils down to that in 1916, the DST transition was done at 00:00, instead of 02:00 as we do today.** 

## Checklist:
- [x] Tests added/updated.


haakonvt:
As briefly discussed on slack; "my 2 cents": _let it fail_. As far as I understand, the error message is correct and the _month start_ happened twice, so `1916-10-01` is not uniquely defined. Super interesting edge-case though.

CHANGELOG.md
```diff
@@ -17,6 +17,13 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.0.2] - 20-04-26
+### Fixed
+- Fixed a bug in `DatapointsAPI.retrieve_dataframe_in_tz` to raise an `AmbiguousTimeError` when getting an ambiguous timestamp.
+  An ambiguous timestamp is, for example, 1916-10-01 00:00:00 Europe/Oslo as this have two UTC times 1916-10-01 00:00:00+02:00
+  and 1916-10-01 00:00:00+01:00 depending on whether you are before or after the DST transition. This would occur if you used
+  a start and end time before and after 1916-10-01 with a `1month` granularity.
```
haakonvt, line 25: I suggest we don't elaborate too much in the changelog:
```suggestion
- Fixed a bug in `DatapointsAPI.retrieve_dataframe_in_tz` that could raise `AmbiguousTimeError` when subdividing the user-specified time range into UTC intervals (with fixed offset).
```

CHANGELOG.md
```diff
@@ -17,6 +17,13 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.0.2] - 20-04-26
```
haakonvt, line 20:
```suggestion
## [6.0.2] - 27-04-26
```

cognite/client/_api/datapoints.py
```diff
@@ -1026,6 +1026,11 @@ def retrieve_dataframe_in_tz(
             3. Create a query for each interval and pass all to the retrieve_arrays method.
             4. Stack the resulting arrays into a single column in the resulting DataFrame.
 
+        The splitting of time range into UTC intervals can lead to ambiguous timestamps. An ambiguous timestamp is,
+        for example, 1916-10-01 00:00:00 Europe/Oslo as this have two UTC times 1916-10-01 00:00:00+02:00
+        and 1916-10-01 00:00:00+01:00 depending on whether you are before or after the DST transition. In this case,
+        the splitting always chose the DST.
+
```
haakonvt, line 1033: I vote we just remove this for 2 reasons:

1. The error did not arise because of a user input error (start/end)...
2. ...as it is not possible to specify an ambiguous `datetime` (as far as I understand) since `datetime` uses `fold=0` by default (i.e. pick "first" / summer time). When/if a user would supply us with the same `datetime` but with `fold=1` it's also not a problem: With granularity of hour and below, we know when to start (pick "last" / winter time); with granularity of day and above, we always floor it, and the result of this is not ambiguous.

Maybe there's something I haven't considered 😅 

```suggestion
```

cognite/client/utils/_time.py
```diff
@@ -484,15 +484,13 @@ def _to_fixed_utc_intervals_fixed_unit_length(
...
 
     freq = multiplier * GRANULARITY_IN_HOURS[unit]
-    index = pandas_date_range_tz(start, end, f"{freq}H")
-    expected_freq = pd.Timedelta(hours=freq)
-    next_mask = (index - index.to_series().shift(1)) != expected_freq
-    last_mask = (index.to_series().shift(-1) - index) != expected_freq
-    index = index[last_mask | next_mask]
+    index = pandas_date_range_tz(start, end, to_pandas_freq(f"{multiplier}{unit}", start))
+    utc_offsets = index.to_series().apply(lambda t: t.utcoffset())
+    transition_raw = index[(utc_offsets != utc_offsets.shift(-1)) | (utc_offsets != utc_offsets.shift(1))]
```
haakonvt, line 489: Last thought: we now have every single UTC offset in the time range; should we raise if not all are whole-hour offsets from UTC? Seems like a great opportunity 🤩

tests/tests_unit/test_utils/test_time.py
```diff
@@ -576,3 +577,36 @@ def test_to_pandas_freq(granularity: str, start: str, expected_first_step: str):
...
+    )
+
+
+class TestPandasDateRangeTz:
+    @staticmethod
+    @pytest.mark.dsl
+    @pytest.mark.parametrize(
+        "start, end, granularity, expected_length",
+        list(pandas_date_range_tz_data()),
+    )
```
haakonvt, line 603: This parametrize has just a single test case. Could just merge the content of `pandas_date_range_tz_data` into the test? Then you also don't need the import guard since the test is marked `dsl`.

haakonvt (changes requested):
Mostly lgtm with a few discussion points.

cognite/client/utils/_time.py
doctrino: That is a great point, only thing is that it would actually not be raised if the granularity is variable (1 month or more) as this code is run for fixed cases. I will try to do some refactoring and then add the check, so it is raised in all circumstances.

cognite/client/_api/datapoints.py
doctrino: I think you are right. I originally thought of this as just a choice we make (which timestamp to chose), both being correct. And you are right there is a correct one, the one at the beginning of the month. Unfortunately, this is not what the code is always doing. Currently we are always choosing DST time, which is correct in the fall, but in the spring that means we lose one hour of the month. 

Can you check my thinking here? If it is true I might have to reconsider the implementation of the fix.

haakonvt: Actually, you don't - the one in spring is never ambiguous, but it can represent a non-existent time. I see that pandas provides the following options:
```
A nonexistent time does not exist in a particular timezone where clocks moved forward due to DST.
- ‘shift_forward’ will shift the nonexistent time forward to the closest existing time
- ‘shift_backward’ will shift the nonexistent time backward to the closest existing time
- ‘NaT’ will return NaT where there are nonexistent times
- timedelta objects will shift nonexistent times by the timedelta
- ‘raise’ will raise an NonExistentTimeError if there are nonexistent times.
```

I wonder if it is as simple as providing `"shift_backward"`, but it's to early in the morning for me to think this through 😂

cognite/client/utils/_time.py
haakonvt: Then I suggest leaving it for a future PR 👍

cognite/client/_api/datapoints.py
haakonvt: Found this: https://stackoverflow.com/a/18490073/3887338

Will play around with one of these that do not have "midnight".

cognite/client/utils/_time.py
doctrino: Good, will add it to my list

cognite/client/_api/datapoints.py
doctrino: Not sure if I follow you here. When you say `'Actually, you don't'`, you mean the spring one is not a problem as it is not in a month transition? 

The `nonexistent` as far as I understand it, is for the opposite problem, when you instead of having two UTC timestamps for a local time, it is when you try to use a local time that does not exists due to it is in a DST transition.

Anyways, so solution is do delete the above comment?

haakonvt: After some thinking: `"shift_forward"` is correct:

Screenshot 2023-04-27 at 09 55 49

doctrino: Updated it now. Hopefully good.

cognite/client/_api/datapoints.py
```diff
@@ -1017,8 +1017,8 @@ def retrieve_dataframe_in_tz(
             aggregates with automatic handling for daylight saving time (DST) transitions. If your time zone observes DST,
             and your query crosses at least one DST-boundary, granularities like "3 days" or "1 week", that used to represent
             fixed durations, no longer do so. To understand why, let's illustrate with an example: A typical time zone
-            that observes DST will skip one hour ahead during spring, leading to a day that is only 23 hours long, and oppositely
-            in the fall, turning back the clock one hour, yielding a 25-hour long day.
+            (above Equator) that observes DST will skip one hour ahead during spring, leading to a day that is only
```
haakonvt, line 1020: Seems capitalizing "equator" isn't strictly wrong, just not recommended. You choose 😉 
```suggestion
            (above the equator) that observes DST will skip one hour ahead during spring, leading to a day that is only
```

haakonvt (approved):
Ah, time zones are fun 💀

PR #1035, opened 2022-09-16
Run test-suite against a windows image

qTipTip:
- Only open the file handler once for writing.
- Parametrize `test_suite` to run on both ubuntu and windows.

## Description
Please describe the change you have made.

## Checklist:
- [ ] Tests added/updated.
- [ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change. 
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring. 
- [ ] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ ] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) per [semantic versioning](https://semver.org/).

PR #1213, opened 2023-05-02
Test Suite on Windows Image

doctrino:
## Description

Restarting the effort from https://github.com/cognitedata/cognite-sdk-python/pull/1035. Starting from scratch rather than continue on that one for two reasons:
1. No need to merge in all changes.
2. Fiona, as of [1.9a3 (2022-10-17)](https://github.com/Toblerity/Fiona/blob/master/CHANGES.txt#L93) includes GDAL in its wheel. This means that there is no need to download wheels manually for GDAL. 

**Note** I ended up splitting the `full_test_and_build` into `full_test` and `build`, so branch protection rules needs to be updated. 

### Improved
- The SDK has received several minor bugfixes to be more user-friendly on Windows.

### Fixed
- The utility function `cognite.client.utils.datetime_to_ms` now raises an understandable `ValueError` when unable to convert pre-epoch datetimes.
- Several functions reading and writing to disk now explicitly use UTF-8 encoding

## Progress
There are initially 10 tests failing on Windows:

- [x] `TestInsertDatapointsAPI.test_insert_before_epoch`
- [x] `TestFilesAPI.test_download_empty_file`
- [x] `TestGeospatialAPI.test_get_raster`
- [x] `TestGeospatialAPI.test_get_raster_with_transformation`
- [x] `TestSyntheticDatapointsAPI.test_query_with_start_before_epoch`
- [x] `TestAssetHierarchy.test_validate_asset_hierarchy_report_to_file` x 2
- [x] `TestAssetHierarchy.test_validate_asset_hierarchy_arbitrat_file_obj` x 2
- [x] `TestTimestampToMs.test_negative.t1`

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,18 @@ jobs:
         run: |
           pre-commit run --all-files
 
+      - name: Verify proto files
+        env:
+          TMPDIR: latest_proto
+          PROTO_URL: https://raw.githubusercontent.com/cognitedata/protobuf-files/master/v1/timeseries
+        run: |
+          mkdir $TMPDIR
+          curl --silent $PROTO_URL/data_points.proto --output $TMPDIR/data_points.proto
+          curl --silent $PROTO_URL/data_point_list_response.proto --output $TMPDIR/data_point_list_response.proto
+          diff $TMPDIR/data_points.proto cognite/client/_proto/data_points.proto
+          diff $TMPDIR/data_point_list_response.proto cognite/client/_proto/data_point_list_response.proto
+
+
   build_docs:
     runs-on: ubuntu-latest
     steps:
@@ -101,9 +113,6 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           file: ./coverage.xml
 
-      - name: Build package
-        run: poetry build
-
   build:
     runs-on: ubuntu-latest
     steps:
@@ -115,24 +124,6 @@ jobs:
       - name: Install full dependencies
         run: |
           python3 -m pip install --upgrade pip poetry
-          poetry config virtualenvs.create false
-          poetry install -E all
-
-      - name: Verify proto files
-        env:
-          TMPDIR: latest_proto
-          PROTO_URL: https://raw.githubusercontent.com/cognitedata/protobuf-files/master/v1/timeseries
-        run: |
-          mkdir $TMPDIR
-          curl --silent $PROTO_URL/data_points.proto --output $TMPDIR/data_points.proto
-          curl --silent $PROTO_URL/data_point_list_response.proto --output $TMPDIR/data_point_list_response.proto
-          diff $TMPDIR/data_points.proto cognite/client/_proto/data_points.proto
-          diff $TMPDIR/data_point_list_response.proto cognite/client/_proto/data_point_list_response.proto
-
-      - uses: codecov/codecov-action@v3
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
-          file: ./coverage.xml
 
       - name: Build package
         run: poetry build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,9 @@ on:
   pull_request:
     branches: [master]
 
+env:
+  PYTHON_VERSION: '3.8'
+
 jobs:
   lint:
     runs-on: ubuntu-latest
@@ -12,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install required dependencies
         run: |
@@ -30,7 +33,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install core dependencies
         run: |
@@ -47,7 +50,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install core dependencies
         run: |
@@ -61,13 +64,53 @@ jobs:
         run: pytest tests/tests_unit -n8 --dist loadscope --maxfail 10 -m 'not dsl'
           --test-deps-only-core
 
-  test_full_and_build:
+  test_full:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Install full dependencies
+        run: |
+          python3 -m pip install --upgrade pip poetry
+          poetry config virtualenvs.create false
+          poetry install -E all
+
+      - name: Test full
+        env:
+          LOGIN_FLOW: client_credentials
+          COGNITE_CLIENT_SECRET: ${{ secrets.COGNITE_CLIENT_SECRET }}
+          COGNITE_TOKEN_URL: https://login.microsoftonline.com/dff7763f-e2f5-4ffd-9b8a-4ba4bafba5ea/oauth2/v2.0/token
+          COGNITE_TOKEN_SCOPES: https://greenfield.cognitedata.com/.default
+          COGNITE_CLIENT_ID: 14fd282e-f77a-457d-add5-928ec2bcbf04
+          COGNITE_PROJECT: python-sdk-test
+          COGNITE_BASE_URL: https://greenfield.cognitedata.com
+          COGNITE_CLIENT_NAME: python-sdk-integration-tests
+          CI: 1
+        run: |
+          pytest tests --durations=10 --cov --cov-report xml:coverage.xml -n8 --dist loadscope --reruns 2
+
+      - uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          file: ./coverage.xml
+
+      - name: Build package
+        run: poetry build
+
+  build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install full dependencies
         run: |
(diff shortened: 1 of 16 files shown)


.github/workflows/build.yml
```diff
@@ -107,3 +103,36 @@ jobs:
 
       - name: Build package
```
erlendvollset, line 104: Remove build step from `test_full` job

.github/workflows/build.yml
```diff
@@ -107,3 +103,36 @@ jobs:
...
+          mkdir $TMPDIR
+          curl --silent $PROTO_URL/data_points.proto --output $TMPDIR/data_points.proto
+          curl --silent $PROTO_URL/data_point_list_response.proto --output $TMPDIR/data_point_list_response.proto
+          diff $TMPDIR/data_points.proto cognite/client/_proto/data_points.proto
+          diff $TMPDIR/data_point_list_response.proto cognite/client/_proto/data_point_list_response.proto
+
+      - uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          file: ./coverage.xml
```
erlendvollset, line 135: Remove codecov upload step from `build` job

.github/workflows/build.yml
erlendvollset: Also, will there be any issues with uploading the coverage report twice?

erlendvollset: Also, should test on windows on the master branch aswell. So you should make the same changes in the `release` workflow.

.github/workflows/build.yml
```diff
@@ -107,3 +103,36 @@ jobs:
...
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Install full dependencies
+        run: |
+          python3 -m pip install --upgrade pip poetry
+          poetry config virtualenvs.create false
+          poetry install -E all
+
+      - name: Verify proto files
```
erlendvollset, line 121: Maybe put this in the `lint` job instead?

.github/workflows/build.yml
doctrino: I think the coverage report will just overwrite the last one.

doctrino: Updated the release workflow now

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.1.2] - 02-05-23
+### Fixed
+- Handling negative epochs in the `DatapointsAPI` on Windows.
```
haakonvt, line 22: Since `datetime_to_ms` is exported as a public part of the SDK, I think it is more correct to name it specifically than the API which most heavily uses it. Up to you. 

Date at least needs +1 day:
```suggestion
## [6.1.2] - 03-05-23
### Fixed
- Handling `datetime`s in local time (naive) before epoch no longer raises on Windows (utility function `cognite.client.utils.datetime_to_ms`).
```

.github/workflows/release.yaml
```diff
@@ -44,31 +58,24 @@ jobs:
...
           --test-deps-only-core
 
-  test_full_and_build_and_release:
-    runs-on: ubuntu-latest
+  test_full:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
haakonvt, line 66: **[Comment not about this PR]** Since we require PRs to be up-to-date (with `master`) before merging, and `test_full` runs on PRs, it does not make sense to me to re-run the full test suite on merges to `master` / releases. It makes chain releases/merges twice as slow to complete and just more error-prone imo without being more secure. Wdyt @erlendvollset ?

haakonvt (changes requested):
Great changes, especially since a looot of our customer run on Windows 👌 👌

cognite/client/utils/_time.py
```diff
@@ -62,15 +62,19 @@ def get_utc_zoneinfo() -> ZoneInfo:
...
 
     Returns:
         ms: Milliseconds since epoch (negative for time prior to 1970-01-01)
     """
-    return int(1000 * dt.timestamp())
+    try:
+        return int(1000 * dt.timestamp())
+    except OSError:
+        # OSError is raised if dt.timestamp() is called before 1970-01-01 on Windows for naive datetime.
+        return int(1000 * (dt - datetime(1970, 1, 1)).total_seconds())
```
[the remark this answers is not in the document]
doctrino: This implementation does actually raise an error. 
```python
datetime(1969, 1, 1).astimezone(tz=timezone.utc)
  File "C:\Program Files\JetBrains\PyCharm 2022.3.2\plugins\python\helpers\pydev\pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
OSError: [Errno 22] Invalid argument
```

tests/tests_integration/test_api/test_geospatial.py
```diff
@@ -640,10 +643,8 @@ def test_get_raster(self, cognite_client, test_feature_type, test_feature_with_r
             raster_property_name="raster",
             raster_format="XYZ",
         )
-        raster_content = open(
-            "tests/tests_integration/test_api/geospatial_test_resources/raster-grid-example.xyz", "rb"
-        ).read()
-        assert res == raster_content
+        raster_content = (GEOSPATIAL_TEST_RESOURCES / "raster-grid-example.xyz").read_text()
+        assert res.decode(encoding="utf-8") == raster_content
```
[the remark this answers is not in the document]
doctrino: Haha, I tried to follow the advice of some past reviewer of not changing everything I see 😄. I changed this as I wanted to run the test from the local folder. Ideally I want to spit the test into two, as it is two tests and not a sequential double test.

Anyways, no `read_bytes` does not work. The returning bytes from the geospatial uses unix line endings, while reading bytes locally uses Windows line endings. The fix is to decode the bytes and compare the text instead, as you see here.

tests/tests_unit/test_utils/test_time.py
```diff
@@ -59,6 +59,7 @@ def test_aware_datetime_to_ms(self):
         assert datetime_to_ms(datetime(2018, 1, 31, tzinfo=utc)) == 1517356800000
         assert datetime_to_ms(datetime(2018, 1, 31, 11, 11, 11, tzinfo=utc)) == 1517397071000
         assert datetime_to_ms(datetime(100, 1, 31, tzinfo=utc)) == -59008867200000
+        assert datetime_to_ms(datetime(1900, 1, 1, tzinfo=utc)) == -2208988800000
```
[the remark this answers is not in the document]
doctrino: Yes, but I changed my mind while writing it, as the test is called  `test_aware_datetime_to_ms`, will setup a new test for naive `datetime_to_ms`.

cognite/client/utils/_time.py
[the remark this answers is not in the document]
doctrino: Furthermore, I am a bit skeptical of doing an implementation "just in case", I rather have it tested for the cases we are interested in.

tests/tests_integration/test_api/test_geospatial.py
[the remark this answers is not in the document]
haakonvt: > Haha, I tried to follow the advice of some past reviewer of not changing everything I see 

👀 ...but when you first make a change (and this is certainly a good one), I think you should fix "all of the occurrences"!

> Anyways, no read_bytes does not work

Ah I see 😞

cognite/client/utils/_time.py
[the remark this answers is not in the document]
haakonvt: > This implementation does actually raise an error.

Dammit Windows 😂 

> Furthermore, I am a bit skeptical of doing an implementation "just in case", I rather have it tested for the cases we are interested in.

...but your original code does datetime subtraction under the assumption that the only way this error triggers, is when "windows + naive". I think it is better to assume we don't know when an error might arise from e,g, an "exotic edge-device OS", and use code that uses the correct strategy for computing time spans (i.e. convert to utc first).

With that said, since `dt.astimezone(timezone.utc)` also fails on Windows (me -> 🤦‍♂️ x 1000), that strategy is out of the question, I vote to keep your existing solution. I would prefer if you added a check to ensure the timestamp is naive though and if not, reraise

haakonvt (commented):
(Added comment to avoid blocking PR)

tests/tests_unit/test_utils/test_time.py
```diff
@@ -45,12 +45,19 @@ class TestDatetimeToMs:
...
     )
-    def test_naive_datetime_to_ms(self, local_tz, expected_ms):
+    def test_naive_datetime_to_ms_unix(self, local_tz, expected_ms):
         with tmp_set_envvar("TZ", local_tz):
             time.tzset()
             assert datetime_to_ms(datetime(2018, 1, 31, tzinfo=None)) == expected_ms
             assert timestamp_to_ms(datetime(2018, 1, 31, tzinfo=None)) == expected_ms
 
+    @pytest.mark.skipif(
+        platform.system() != "Windows", reason="Windows assume naive dateimes are UTC, Unix assumes they are local"
```
haakonvt, line 55: That’s not correct afaik. Python naive datetimes are always assumed local time

doctrino: ?? Why are the tests passing then?

qTipTip: There is a warning in the datetime docs stating: 

```
Warning: Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC.
```

It is  not clear to me whether _many_ datetime method means them all, or if there are some that treats them differently.

cognite/client/utils/_time.py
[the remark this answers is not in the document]
doctrino: What about the ugly double try except?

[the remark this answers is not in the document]
doctrino: added it now

tests/tests_unit/test_utils/test_time.py
haakonvt: @qTipTip : _All methods_ assume naive = local time, except those prefixed with `utc`. This makes for example `datetime.utcnow()` particularly dangerous, as it returns a `datetime` in UTC - but naive - so that future method calls e.g. `.timestamp()` will give the wrong result - unless your local time is set to UTC ofc.

haakonvt: @doctrino : That must mean your system is using UTC or you have a setting somewhere that the Python interpreter picks up, for example, the environment variable `TZ` is set.

https://discuss.python.org/t/get-now-with-local-system-timezone-without-knowing-the-timezone-first/22665/12

haakonvt: image

haakonvt: I booted up my window pc to try this out, and I also got the subtraction of two naive datetimes to be computed in "absolute" time, i.e. same as if both were given UTC timezone info. What's interesting though is that this works the same on unix.

Suggested fix: remove the platform check and let it run on both win&unix OR lets revert this method entirely (back to just `.timestamp()`. What do you think?

haakonvt: ### My final opinion
`datetime_to_ms` needs to give the same result for the same input for any OS / platform. This is a higher requirement than supporting naive datetimes before epoch on Windows, so these necessarily need to fail.

cognite/client/utils/_time.py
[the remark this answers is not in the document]
haakonvt: Adressed in another thread

tests/tests_unit/test_utils/test_time.py
doctrino: Agreed to raise a ValueError instead with the message to not use native datetime.

PR #1215, opened 2023-05-05
Make `CogniteClient` picklable again

haakonvt:
## Description
The various `CredentialProvider`s have e.g. thread locks and `OAuth2Session` which hinders pickling/unpickling. This PR fixes this.

### Rationale
In the virtual flow meter service, we use `kedro`, and we need the client to be serializable to be passed between computation nodes.

## ~~TODO~~
- Add a test that does `pickle.loads(pickle.dumps(client))` then makes an API call.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.2.0] - 05-05-23
+### Fixed
+- Pickling a `CogniteClient` instance with certain `CredentialProvider`s no longer causes a `TypeError: cannot pickle ...` to be raised.
+
 ## [6.1.2] - 04-05-23
 ### Improved
 - The SDK has received several minor bugfixes to be more user-friendly on Windows.
--- a/tests/tests_integration/test_api/test_geospatial.py
+++ b/tests/tests_integration/test_api/test_geospatial.py
@@ -1,3 +1,4 @@
+import math
 import random
 import re
 import time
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.2.0] - 05-05-23
```
haakonvt, line 20: Update patch version instead?

tests/tests_integration/test_api/test_geospatial.py
```diff
@@ -706,8 +707,8 @@ def test_put_raster_custom_crs(self, cognite_client, test_feature_type, test_fea
         assert res.skew_x == 0.0
         assert res.skew_y == 0.0
         assert res.srid == 3857
-        assert res.upper_left_x == -0.5891363261459447
-        assert res.upper_left_y == -0.31623471547260973
+        assert math.isclose(res.upper_left_x, -0.5891363261459447)
+        assert math.isclose(res.upper_left_y, -0.31623471547260973, rel_tol=5e-7)
```
tuanng-cognite, line 711: we could change the expected value to the new one: -0.316234582133065 and use the default tolerance.

tuanng-cognite: the value is gotten from running test upon master branch

haakonvt: Why has the API response changed? If this is because the underlying test data has changed, I’ll happily update to a new value. But, if this is a calculated value that just suddenly changed, I’m quite worried tbh 😅

PR #1225, opened 2023-05-19
Data model centric support in transformation

nimeshawij:
## Description
Data model centric support in transformation.

## Checklist:
- [x] Tests added/updated.


nimeshawij:
@SunilVenekarKrishnamoorthy please check and approve.

tests/tests_integration/test_api/test_transformations/test_transformations.py
```diff
@@ -252,6 +254,37 @@ def test_create_instance_edges_type_transformation(self, cognite_client):
 
         cognite_client.transformations.delete(id=ts.id)
 
+    def test_create_instance_data_model_transformation(self, cognite_client):
```
SunilVenekarKrishnamoorthy, line 257: Can you rename this test to **test_create_instance_relationship_data_model_transformation**

And create one more **test_create_instance_type_data_model_transformation**

nimeshawij: test_create_instance_type_data_model_transformation and test_create_instance_relationship_data_model_transformation added

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.1.11] - 19-05-23
```
gregertw, line 20: Fix the date here.
```suggestion
## [6.1.11] - 22-05-23
```

gregertw: As we are adding functionality in a compatible way, I assume this should be 6.2.0?

nimeshawij: @SunilVenekarKrishnamoorthy

haakonvt:
Why do you implement a custom `__hash__`?

nimeshawij:
> Why do you implement a custom `__hash__`?

To provide efficient storage and retrieval capabilities. `__hash__` function also used in 'EdgeType' and 'ViewInfo' classes. So, I have used in the class 'DataModelInfo' as well.

nimeshawij:
> > To provide efficient storage and retrieval capabilities. (…)
> 
> Could you please share a short code snippet showcasing this?

@haakonvt  Do you mean code snippet like this? 

```
from typing import Optional

class DataModelInfo:
    def __init__(
        self,
        space: str,
        external_id: str,
        version: str,
        destination_type: str,
        destination_relationship_from_type: Optional[str] = None,
    ):
        self.space = space
        self.external_id = external_id
        self.version = version
        self.destination_type = destination_type
        self.destination_relationship_from_type = destination_relationship_from_type

    def __hash__(self) -> int:
        if self.destination_relationship_from_type is not None:
            return hash(
                (
                    self.space,
                    self.external_id,
                    self.version,
                    self.destination_type,
                    self.destination_relationship_from_type,
                )
            )
        else:
            return hash((self.space, self.external_id, self.version, self.destination_type))

    def __eq__(self, other) -> bool:
        if isinstance(other, DataModelInfo):
            return (
                self.space == other.space
                and self.external_id == other.external_id
                and self.version == other.version
                and self.destination_type == other.destination_type
                and self.destination_relationship_from_type == other.destination_relationship_from_type
            )
        return False

# Create instances of DataModelInfo
data_model1 = DataModelInfo("space1", "id1", "v1", "type1")
data_model2 = DataModelInfo("space2", "id2", "v2", "type2")
data_model3 = DataModelInfo("space1", "id1", "v1", "type1")

# Create a dictionary with DataModelInfo instances as keys
data = {
    data_model1: "Data 1",
    data_model2: "Data 2",
}

# Access data using DataModelInfo keys
print(data[data_model1])  # Output: Data 1
print(data[data_model2])  # Output: Data 2
print(data[data_model3])  # Output: KeyError: <DataModelInfo object at 0x...>
```

Result:

```
Data 1
Data 2
Data 1
```


When we try to access the data using 'data_model1' or 'data_model2' as keys, we get the expected output. However, when we try to access the data using 'data_model3', which has the same attribute values as data_model1, we get a KeyError. This is because 'data_model3' is a different object from data_model1 and, by default, dictionary lookups are based on object identity.

However, due to the implementation of the __hash__ method in the DataModelInfo class, 'data_model3' produces the same hash value as 'data_model1'. Therefore, when data_model3 is used as a key to access data in the dictionary, it should return the same value as 'data_model1'.

haakonvt:
@nimeshawij ...but `__eq__` is _not implemented_ so your example does not match this PR 😉 This PR still raises `KeyError` in your example code snippet.

nimeshawij:
> def __eq__(self, other) -> bool

yes, you are right.  ' __eq__' should implement in the class DataModelInfo, I think it should be include to all the other classes as well. What is your opinion? 

Ex:

```
class RawTable(TransformationDestination):
    def __init__(self, database: str = None, table: str = None):
        super().__init__(type="raw")
        self.database = database
        self.table = table

    def __hash__(self) -> int:
        return hash((self.type, self.database, self.table))
```

```
class SequenceRows(TransformationDestination):
    def __init__(self, external_id: str = None):
        super().__init__(type="sequence_rows")
        self.external_id = external_id

    def __hash__(self) -> int:
        return hash((self.type, self.external_id))

```

```
class ViewInfo:
    def __init__(self, space: str, external_id: str, version: str):
        self.space = space
        self.external_id = external_id
        self.version = version

    def __hash__(self) -> int:
        return hash((self.space, self.external_id, self.version))

    def dump(self, camel_case: bool = False) -> Dict[str, Any]:
        return basic_obj_dump(self, camel_case)
```


```
class EdgeType:
    def __init__(self, space: str, external_id: str):
        self.space = space
        self.external_id = external_id

    def __hash__(self) -> int:
        return hash((self.space, self.external_id))

    def dump(self, camel_case: bool = False) -> Dict[str, Any]:
        return basic_obj_dump(self, camel_case)
```

gregertw:
@haakonvt I'm struggling a bit to understand where you are guiding? Could you be more explicit? Don't you recommend implementing a hash function like this as it as breaking SDK behaviour? And it should throw a KeyError?

haakonvt:
> @haakonvt I'm struggling a bit to understand where you are guiding? Could you be more explicit? Don't you recommend implementing a hash function like this as it is breaking SDK behavior? And it should throw a KeyError?

Let me try to be more clear then! 😄 

1. We want maintainable code & a maintainable SDK; the less code we have, the simpler this is to accomplish.
2. When overriding the basic hash dunder method, it must be because we want to achieve something previously impossible; i.e., we want to allow users to do X with our objects because it is useful to how we want or expect them to interact with the objects.
3. To ensure the object behavior is as expected, we add tests that cover this (or these) object behavior(s) explicitly. In @nimeshawij example above, she wanted two different objects with the same attributes (to have the same hash) and evaluate equal ( `==` -> `__eq__`) so that either one could be used to look up the other one in a dictionary (which is backed by a hash table).

This is all well and good! ..._but in the current state_, only hash has been somewhat randomly by previous developers and 3) does not work as in the given example(!). There are also no tests for the wanted behavior(!) nor are there code examples in the docstring showcasing this(!).

All of this leads me back to thinking the custom hash methods was added without any particular reasoning and should be removed until a use case exists, and then step 1-2-3 should be thought through, planned, and implemented 🚀 

(Btw, I asked your "predecessors" (Emel and Vu Hai) the exact same question and neither of them had an answer to this question 😅 )

gregertw:
Great, thanks, makes a lot more sense to me now. I guess the question here is then if we must assume if somebody relies on using hashes in a way that will break without overriding the hash in this case, or not? I.e. better safe than sorry. If not, I agree with you that overriding the hash may not be necessary.

nimeshawij:
> If not, I agree with you that overriding the hash may not be necessary.

@gregertw  Do we need to clear `hash `method from all the classes or include `__eq__ ` to all?

haakonvt:
I won’t block this or any future PR based on this. I just want the transformations team to have a plan with how you choose to implement the methods and data classes in the SDK 😄

So, it is 100% up to you to decide what and how you want to do this. I’m just asking questions trying to understand so that I may review the code ☺️

nimeshawij:
> I won’t block this or any future PR based on this. I just want the transformations team to have a plan with how you choose to implement the methods and data classes in the SDK 😄
So, it is 100% up to you to decide what and how you want to do this. I’m just asking questions trying to understand so that I may review the code ☺️

@haakonvt I have followed the same pattern what Emel and Vu Hai did. I posted your question in data integration channel. As a quick fix or to approve PR and release the version, I can remove `__hash__` method for the data centric class.

nimeshawij:
@gregertw @haakonvt `hash `method is removed from the data model centric info class. Since, there is no proper use case planed for tested.

cognite/client/data_classes/transformations/common.py
```diff
@@ -140,6 +140,17 @@ def edges(
         """
         return Edges(view=view, instance_space=instance_space, edge_type=edge_type)
 
+    @staticmethod
+    def instances(data_model: Optional[DataModelInfo] = None, instance_space: Optional[str] = None) -> Instances:
+        """
+        Args:
+            dataModel (DataModelInfo): information of the Data Model.
```
haakonvt, line 147: Please always use snake case 😄 
```suggestion
            data_model (DataModelInfo): information of the Data Model.
```

cognite/client/data_classes/transformations/common.py
```diff
@@ -306,9 +355,10 @@ def _load_destination_dct(
     if destination_type in simple:
         return simple[destination_type](**snake_dict)
 
-    nested: Dict[str, Union[type[Nodes], type[Edges]]] = {
+    nested: Dict[str, Union[type[Nodes], type[Edges], type[Instances]]] = {
```
haakonvt, line 358: You may consider using the pipe operator for unions:
```suggestion
    nested: Dict[str, type[Nodes] | type[Edges] | type[Instances]] = {
```

pyproject.toml
```diff
@@ -1,7 +1,8 @@
 [tool.poetry]
 name = "cognite-sdk"
 
-version = "6.2.0"
+version = "6.1.11"
```
haakonvt, line 4: Incorrect version

haakonvt (changes requested):
Good test coverage 😄  Very close to being ready to merge

pyproject.toml
nimeshawij: version - fixed

cognite/client/_version.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import annotations
 
-__version__ = "6.2.0"
+__version__ = "6.1.11"
```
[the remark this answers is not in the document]
nimeshawij: version - fixed

cognite/client/data_classes/transformations/common.py
nimeshawij: noted 👍

nimeshawij:
The` build / test_full (windows-latest) (pull_request)` failed due to  `tests/tests_integration/test_api/test_data_modeling/test_data_models.py`  failing, which is not related to this PR. 


```
=========================== short test summary info ===========================
ERROR tests/tests_integration/test_api/test_data_modeling/test_data_models.py::TestDataModelsAPI::test_list - AssertionError: Please create at least one data model in CDF.
assert 0 > 0
 +  where 0 = len([])
= 1758 passed, 17 skipped, 9 xpassed, 22 warnings, 1 error, 2 rerun in 204.23s (0:03:24) =
Error: Process completed with exit code 1.
```

haakonvt:
> The` build / test_full (windows-latest) (pull_request)` failed due to `tests/tests_integration/test_api/test_data_modeling/test_data_models.py` failing, which is not related to this PR.
> 
> ```
> =========================== short test summary info ===========================
> ERROR tests/tests_integration/test_api/test_data_modeling/test_data_models.py::TestDataModelsAPI::test_list - AssertionError: Please create at least one data model in CDF.
> assert 0 > 0
>  +  where 0 = len([])
> = 1758 passed, 17 skipped, 9 xpassed, 22 warnings, 1 error, 2 rerun in 204.23s (0:03:24) =
> Error: Process completed with exit code 1.
> ```

This is not the test that failed, you are looking at the wrong run.

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.2.1] - 2023-05-26
+- Data model centric support in transformation
```
haakonvt, line 21:
```suggestion
## [6.2.1] - 2023-05-26
### Added
- Data model centric support in transformation
```

cognite/client/_version.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import annotations
 
-__version__ = "6.2.0"
+__version__ = "6.2.1"
+
```
haakonvt, line 4:
```suggestion
__version__ = "6.2.1"
```

haakonvt, line 4:
```suggestion
__version__ = "6.2.1"
```

haakonvt (approved):

PR #1230, opened 2023-05-23
Make custom pre-commit checks extendable + new version check

haakonvt:
## Description
1. Makes it easy to extend our `pre-commit` setup with new custom checks. Each check is run independently of the others and all that fail are listed in the output / feedback to the user, see screenshot below.
2. Adds an additional test that makes sure the version in `CHANGELOG.md` matches the two other versions given by `_version.py` and in `pyproject.toml`.
3. Also adds a test on the date format as this has been in the wrong format a looot of times. Also reverts the 'year' specifier to 4 digits. This was not an intended change in `v5` (I'm to blame hah!).

image


scripts/custom_checks/version.py
```diff
@@ -0,0 +1,48 @@
...
+            f"cognite/client/_version.py: ({__version__}). Did you forgot to add a new entry? "
+            "Or maybe you haven't followed the required format?"
+        )
+    return None
+
+
+def changelog_entry_date() -> Optional[str]:
+    match = next(_parse_changelog())
+    try:
+        datetime.strptime(date := match.group(2), "%Y-%m-%d")
```
erlendvollset, line 45: Could check that it's equal to today's date as well?

erlendvollset (approved):
Good stuff!

PR #1231, opened 2023-05-24
DMS Support for Spaces

doctrino:
## Description

Space Support
------------------------
Setup SDK support for spaces in the DMS, plan is to add views, containers, data models, and instances later. Spaces is urgent for use in Bootstrap CLI.

* Refactored the `Identifier` class to a core class, and then `Identifies` and `DataModeldentifier` implementation.
* Note I selected to keep the `Identifier `and `IdentifierSequence `names to have minimum impact on the code base. 
* This enabled using the `APIClient `in the implementation of `SpacesAPI`. 

I went with a DataModelingAPI as a container for the SpacesAPI such that it match the underlying REST API:
```python
from cognite.client import CogniteClient

client = CogniteClient()
client.data_modeling.spaces.apply(Space(space="mySpace", name="My Space", description="Description of my Space").

```

Docs Reorg
-----------------
* Reorganized documentation to match API documentation.
* Split docs into multiple files for developer ease.
image

## Checklist:
- [x] Tests added/updated.
- [x] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [x] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [x] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #1234, opened 2023-05-24
Mocked APIs that change attribute path will now fail in tests

haakonvt:
## Description
Mocked APIs that change attribute path will now fail in tests.

Discovered here: https://github.com/cognitedata/cognite-sdk-python/pull/1231#discussion_r1204758219 

This will now fail in tests:
`Failed: Attribute 'models' for API (or sub-API): <class 'cognite.client._api.data_modeling.DataModelingAPI'> available on 'CogniteClientMock', was not found on the 'CogniteClient' object, has it been misspelled?`

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/tests/tests_unit/test_testing.py
+++ b/tests/tests_unit/test_testing.py
@@ -1,3 +1,5 @@
+from operator import attrgetter
+
 import pytest
 
 from cognite.client import ClientConfig, CogniteClient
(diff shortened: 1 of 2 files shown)


erlendvollset (approved):
🙌

PR #1240, opened 2023-05-31
CDF-18779: Update Assets/Events documentation with the latest `partitions` limits.

voron86:
## Description
`partitions` argument description is updated for Assets and Events


haakonvt:
Just my 2 cents: I think you go into too much detail. Instead of explaining how silent repartitioning works, just say “using more than 10 will actually decrease throughout and in the future possibly return an error from the API”.

PR #1241, opened 2023-05-31
 Adding Support for Instances.

doctrino:
## Description
Part 1 of InstanceAPI, support for CRUD operations. 

## Checklist:
- [x] Tests added/updated.


erlendvollset:
A few preliminary notes on the interface:

1) On instances.apply/delete it's a point that we should allow operations on nodes and edges together transactionally, so the overloads for those should accept `instance: Sequence[NodeApply | EdgeApply]`
2) I prefer `list_nodes(...)` and `list_edges(...)` to `list("node", ...)` and `list("edge", ...)`

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -0,0 +1,486 @@
...
+            ]
+            if self.properties
+            else None,
+        )
+
+    def as_id(self) -> NodeId:
+        return NodeId(space=self.space, external_id=self.external_id)
+
+
+class NodeUpdate(InstanceUpdate):
```
erlendvollset, line 257: I suggest `NodeApplyResult` rather than `NodeUpdate`. (And NodeApplyResultList, EdgeApplyResult, EdgeApplyResultList aswell)

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -0,0 +1,486 @@
...
+            sources=[
+                NodeOrEdgeData(source=source, properties=space_properties)  # type: ignore[arg-type]
+                for space_properties in self.properties.values()
+            ]
+            if self.properties
+            else None,
+        )
+
+
+class InstanceUpdate(InstanceCore):
```
erlendvollset, line 152: InstanceApplyResult

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -0,0 +1,486 @@
...
+        **_: dict,
+    ):
+        super().__init__(space, external_id, instance_type)
+        self.existing_version = existing_version
+        self.sources = sources
+
+    def dump(self, camel_case: bool = False) -> dict[str, Any]:
+        output = super().dump(camel_case)
+        if self.sources:
+            output["sources"] = [source.dump(camel_case) for source in self.sources]
```
erlendvollset, line 89: Should also include "existing_version"?

cognite/client/data_classes/data_modeling/ids.py
```diff
@@ -69,15 +69,55 @@ def load(
...
+        raise NotImplementedError
+
 
 @dataclass
 class ViewId(VersionedDataModelingId):
     _type = "view"
 
+    @property
+    def identifier(self) -> str:
+        return f"{self.external_id}/{self.version}"
```
erlendvollset, line 119: Hmm. Not so sure about these `identifier` properties on ID objects. Not very intuitive to me what that would return. Perhaps drop the property on ContainerId and just be super explicit in the ViewId naming. smth like 
```
@property
def external_id_slash_version(self) -> str:
    return f"{self.external_id}/{self.version}"
```

erlendvollset: Or I see it's being used to created property references. So perhaps have something like this instead:

```
class PropertyRef:
    @overload
    @staticmethod
    def of(view_id: ViewId, property_identifier: str) -> tuple[str, str, str]:
        ...

    @overload
    @staticmethod
    def of(container_id: ContainerId, property_identifier: str) -> tuple[str, str, str]:
        ...

    @staticmethod
    def of(view_or_container_id: ContainerId | ViewId, property_identifier: str) -> tuple[str, str, str]:
        if isinstance(view_or_container_id, ViewId):
            external_id_maybe_versioned = f"{view_or_container_id.external_id}/{view_or_container_id.version}"
        elif isinstance(view_or_container_id, ContainerId):
            external_id_maybe_versioned = view_or_container_id.external_id
        else:
            raise ValueError
        return (view_or_container_id.space, external_id_maybe_versioned, property_identifier)

```

cognite/client/data_classes/data_modeling/instances.py
erlendvollset: Ah right, the implementation in the super class already handles it. Never mind then 👍

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -0,0 +1,485 @@
...
+                        container the container(s) making up this node.
+    """
+
+    def __init__(
+        self,
+        space: str,
+        external_id: str,
+        existing_version: int = None,
+        sources: list[NodeOrEdgeData] = None,
+        **_: dict,
```
erlendvollset, line 205: Why do we have these? They make it much easier to introduce bugs by e.g. passing "exsiting_version=1"

erlendvollset: I see we have them for the schema data classes as well. Let's try to get rid of them.

erlendvollset: We can and should find a way to avoid that without swallowing kwargs

doctrino: as agreed, only doing it on the read classes now.

cognite/client/data_classes/data_modeling/ids.py
doctrino: What about renaming it and making it a method? As far as I can see the PropertyRef.of is not the same as what I need.

```python
@dataclass
class ViewId(VersionedDataModelingId):
    _type = "view"

    def as_property_identifier(self) -> str:
        return f"{self.external_id}/{self.version}"
```

PR #1242, opened 2023-06-01
Refactor function schedule create after API-key deprecation

haakonvt:
## Description
After API keys were phased out in major version 6, `FunctionSchedulesAPI.create` should also have been updated, but that's too late now (and will have to wait for v7). Either way, this PR does a few things:

1. Makes it so that schedules can be created as specified by the official documentation:

> Nonce retrieved from sessions API when creating a session. This will be used to bind the session before executing the function. The corresponding access token will be passed to the function and used to instantiate the client of the handle() function. You can create a session via the Sessions API. **_When using the Python SDK, the session will be created behind the scenes when creating the schedule_**.

This used to only happen when passing explicit client credentials. It now also works for current user credentials which used to fail at runtime with "Could not fetch a valid token (...)" because a session was never created for the user it was just assumed that an API key was used to create the function.
 
2. Remove the `ValueError` as this is already enforced directly by the API:
`CogniteAPIError: When creating a schedule with OIDC-tokens, you must use 'function_id' and not 'function_external_id'  | code: 400 | X-Request-ID: 57f30ed5-eb68-955d-90c3-1fd0e0d73b97`

3. Improve documentation with an extra example that uses current user credentials. Also moves the warning on the `data` argument to a separate box to make it more clear:

Screenshot 2023-05-31 at 21 46 54

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [6.2.2] - 2023-06-01
+## [6.2.2] - 2023-06-05
 ### Fixed
 - Creating function schedules with current user credentials now works (used to fail at runtime with "Could not fetch a valid token (...)" because a session was never created.)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.2.2] - 2023-06-01
+### Fixed
+- Creating function schedules with current user credentials now works (used to fail at runtime with "Could not fetch a valid token (...)" because a session was never created.)
+
 ## [6.2.1] - 2023-05-26
 ### Added
 - Data model centric support in transformation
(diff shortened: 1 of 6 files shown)


erlendvollset (approved):
LGTM!

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.2.2] - 2023-06-01
```
haakonvt, line 20:
```suggestion
## [6.2.2] - 2023-06-05
```

cognite/client/_api/functions.py
```diff
@@ -931,41 +932,53 @@ def create(
...
+                >>> schedule = c.functions.schedules.create(
+                ...     name="My schedule",
+                ...     function_id=456,
+                ...     cron_expression="*/5 * * * *",
+                ...     description="A schedule just used for some temporary testing.",
+                ... )
 
+        """
+        _get_function_identifier(function_id, function_external_id)
+        nonce = _create_session_and_return_nonce(self._cognite_client, client_credentials)
```
HMEiding, line 981: Most likely not a big issue, but previously, a user on GCP could create a schedule linked with `function_external_id` and without client credentials. The use case for this was primarily for functions with API-keys attached, but it could also be functions without API-keys attached (that just did some logic without needing to authenticate to CDF). Functions of the latter kind can no longer be created after this change.

PR #1243, opened 2023-06-02
Remove class variables `_LIST_CLASS`

haakonvt:
## Description
`_LIST_CLASS` is a remnant from an earlier SDK version when the resource and list class were inferred automatically / implicitly used. These days we favor explicitness 😉 

Also adds `_RESOURCE_PATH` to `FunctionCallsAPI`.

### DIFF
--- a/cognite/client/_api/functions.py
+++ b/cognite/client/_api/functions.py
@@ -70,7 +70,6 @@ def _get_function_identifier(function_id: Optional[int], function_external_id: O
 
 class FunctionsAPI(APIClient):
     _RESOURCE_PATH = "/functions"
-    _LIST_CLASS = FunctionList
 
     def __init__(self, config: ClientConfig, api_version: Optional[str], cognite_client: CogniteClient) -> None:
         super().__init__(config, api_version, cognite_client)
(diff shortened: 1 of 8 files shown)


cognite/client/_api/functions.py
```diff
@@ -788,9 +789,9 @@ def get_response(
         """
         identifier = _get_function_identifier(function_id, function_external_id)
         function_id = _get_function_internal_id(self._cognite_client, identifier)
-        url = f"/functions/{function_id}/calls/{call_id}/response"
-        res = self._get(url)
-        return res.json().get("response")
+
+        resource_path = self._RESOURCE_PATH_RESPONSE.format(function_id, call_id)
+        return self._get(resource_path).json().get("response")
```
haakonvt, line 794: (`interpolate_and_url_encode` is not needed as only integers are ever used)

cognite/client/_api/functions.py
```diff
@@ -768,7 +769,7 @@ def get_response(
             function_external_id (str, optional): External ID of the function on which the call was made.
 
         Returns:
-            Response from the function call.
+            Dict[str, Any] | None: Response from the function call.
```
haakonvt, line 772: @HMEiding note that the official docs list the `"response" (object)` in the API repsonse as required, but it is actually missing when the function returns `None` (i.e. this returns `None`). https://developer.cognite.com/api/v1/#tag/Function-calls/operation/getFunctionCallResponse

erlendvollset (approved):
Lgtm 😎

PR #1244, opened 2023-06-06
Fix unrecoverable state for request retries during token refresh

haakonvt:
## Description
During a "token refresh cycle", it is quite likely, _especially in concurrent scenarios_, that worker threads get stuck with an outdated token. This is an **unrecoverable** state and thus always leads to an exception being raised once the "retry budget is spent".

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,15 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [6.3.2] - 2023-06-07
+## [6.3.2] - 2023-06-08
 ### Fixed
 - Requests being retried around a token refresh cycle, no longer risk getting stuck with an outdated token.
+
+### Added
+- `CredentialProviders` subclassing `_OAuthCredentialProviderWithTokenRefresh`, now accepts a new parameter, `token_expiry_leeway_seconds`, controlling how early a token refresh request should be initiated (before it expires).
+
+### Changed
+- `CredentialProviders` subclassing `_OAuthCredentialProviderWithTokenRefresh` now uses a safer default of 15 seconds (up from 3 sec) to control how early a token refresh request should be initiated (before it expires).
 
 ## [6.3.1] - 2023-06-07
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,13 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.2.3] - 2023-06-06
```
haakonvt, line 20: Likely 😉 
```suggestion
## [6.2.3] - 2023-06-07
```

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
...
+### Fixed
+- Requests being retried around a token refresh cycle, no longer gets stuck with an outdated token.
```
```suggestion
## [6.2.3] - 2023-06-07
### Fixed
- Requests being retried around a token refresh cycle, no longer risk getting stuck with an outdated token.
```

cognite/client/_http_client.py
```diff
@@ -97,13 +97,16 @@ def __init__(
         config: HTTPClientConfig,
         session: requests.Session,
         retry_tracker_factory: Callable[[HTTPClientConfig], _RetryTracker] = _RetryTracker,
+        refresh_auth_header: Callable[[MutableMapping[str, Any]], None] | None = None,
```
erlendvollset, line 100: Can do this and drop the null check on L131.
```suggestion
        refresh_auth_header: Callable[[MutableMapping[str, Any]], None] = lambda headers: None,
```

erlendvollset (changes requested):

cognite/client/_http_client.py
haakonvt: Having an anonymous function as the default argument would break the "pickleability" of any `CogniteClient` with using it. I'd rather change it to a positional arg and rework all the tests 😓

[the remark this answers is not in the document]
haakonvt: Good catch, fully agree, not sure where to begin to set up pyodide testing.

cognite/client/credentials.py
```diff
@@ -15,6 +15,8 @@
 
 from cognite.client.exceptions import CogniteAuthError
 
+_TOKEN_EXPIRY_LEEWAY_SECONDS = 15
```
erlendvollset, line 18:
```suggestion
_TOKEN_EXPIRY_LEEWAY_SECONDS_DEFAULT = 15
```

erlendvollset (approved):
LGTM 💯 Consider changing the constant name.

PR #1248, opened 2023-06-09
Make CogniteResourceList generic on inner type

erlendvollset:
Inner types would be inferred as Any previous to this. Now we get correct type information when accessing inner elements.
```

c = CogniteClient()

a = c.data_modeling.container.list()[0]
reveal_type(a) # reveals Container

for a in c.data_modeling.container.list():
    reveal_type(a) # reveals Container
```


doctrino (approved):
LGTM. I guess not worth a bump?

arnels (approved):

PR #1254, opened 2023-06-19
annotations reverse lookup with test

olacognite:
## Description
Added a reverse lookup method to the annotations client, and a resource type "ResourceReference", since this is what the endpoint returns. (Objects only containing id/external_id. Could maybe consider combining with byIds endpoints and resolve the resources at some point)

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/contextualization.py
+++ b/cognite/client/data_classes/contextualization.py
@@ -844,3 +844,16 @@ class VisionExtractJob(VisionJob):
         raise CogniteException(
             "Extract job is not completed. If the job is queued or running, wait for completion and try again"
         )
+
+
+class ResourceReference(CogniteResource):
+    def __init__(
+        self, id: int | None = None, external_id: str | None = None, cognite_client: None | CogniteClient = None
+    ):
+        self.id = id
+        self.external_id = external_id
+        self._cognite_client: CogniteClient = cast("CogniteClient", None)  # Read only
+
+
+class ResourceReferenceList(CogniteResourceList[ResourceReference]):
+    _RESOURCE = ResourceReference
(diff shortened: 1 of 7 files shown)


cognite/client/data_classes/contextualization.py
```diff
@@ -844,3 +844,16 @@ def save_predictions(
         raise CogniteException(
             "Extract job is not completed. If the job is queued or running, wait for completion and try again"
         )
+
+
+class ResourceReference(CogniteResource):
```
erlendvollset, line 849: Why does this need to be a CogniteResource?

tests/tests_integration/test_api/test_annotations.py
```diff
@@ -73,6 +74,32 @@ def file_id(cognite_client: CogniteClient) -> int:
     cognite_client.files.delete(id=file.id)
 
 
+@pytest.fixture
```
erlendvollset, line 77: Let's make this session-scoped

tests/tests_integration/test_api/test_annotations.py
```diff
@@ -86,6 +113,28 @@ def base_suggest_annotation(base_annotation: Annotation) -> Annotation:
     return ann
 
 
+@pytest.fixture
```
erlendvollset, line 116: Also session-scoped

erlendvollset (changes requested):

cognite/client/data_classes/contextualization.py
olacognite: The generic _list returns a CogniteResourceList, might get around that maybe. The smoothest experience would probably be to return the actual resources, and use files/3D apis to retrieve them.

erlendvollset: Ah now I see.

erlendvollset (approved):

PR #1257, opened 2023-06-20
Some minor breaking changes to DMS client

erlendvollset:
- Rename ConstraintIdentifier, IndexIdentifier, ContainerPropertyIdentifier
- Remove implementation of  Instance:as_apply
- Make VersionedDAtaModelingId:load also accdept tuple[str, str]
- Make Instance:properties easier to consume
- Rename as_reference() to as_id() on ContainerCore and ContainerView
- Remove redundant EdgeApply:create method
- Bump version and changelog

We’ll try hard not to make any more breaking changes, but I think we should accept the cost of breaking a few consumers now early on to really nail the user experience.


doctrino (commented):
The last example in the `instances.apply `needs to be updated as that use 

```python
                >>> person_to_actor = dm.EdgeApply.create(space="mySpace",
                ...                                       external_id="relation:arnold_schwarzenegger:actor",
                ...                                       type="Person.roles",
                ...                                       start_node="person:arnold_schwarzenegger",
                ...                                       end_node="actor:arnold_schwarzenegger",
                ... )
                >>> res = c.data_modeling.instances.apply([person, actor], [person_to_actor])
```

doctrino (approved):
LGTM, although your tests are failing.

PR #1258, opened 2023-06-21
Support for Datapoints Subscription

doctrino:
### Added
- Support for Data Point Subscription, `client.time_series.subscriptions`. Note this is an experimental feature.

## Checklist:
- [x] Tests added/updated.


cognite/client/data_classes/filters.py
```diff
@@ -113,6 +115,13 @@ def load(cls, filter_: dict[str, Any]) -> Filter:
     def _filter_body(self) -> list | dict:
         ...
 
+    def _filter_names(self) -> list[str]:
```
erlendvollset, line 118:
```suggestion
    def _involved_filter_types(self) -> set[type[Filter]]:
```
I'd suggest using the filter types here instead.

cognite/client/_api/datapoints_subscriptions.py
```diff
@@ -0,0 +1,259 @@
...
+            items=update,
+            list_cls=DataPointSubscriptionList,
+            resource_cls=DatapointSubscription,
+            update_cls=DataPointSubscriptionUpdate,
+        )
+
+    def list_data(
+        self,
+        external_id: str,
+        partitions: Sequence[tuple[int, str] | int | DataPointSubscriptionPartition],
```
erlendvollset, line 192: Homogeneous collections are easier to reason about. Is there really a need for mixing them?
```suggestion
        partitions: Sequence[tuple[int, str]] | Sequence[int] | Sequence[DataPointSubscriptionPartition],
```

cognite/client/_api/datapoints_subscriptions.py
```diff
@@ -0,0 +1,259 @@
...
+    def retrieve(
+        self, external_id: list[str] | tuple[str], ignore_unknown_ids: bool = False
+    ) -> DataPointSubscriptionList:
+        ...
+
+    @overload
+    def retrieve(
+        self, external_id: str | Sequence[str], ignore_unknown_ids: bool = False
+    ) -> Optional[DatapointSubscription] | DataPointSubscriptionList:
+        ...
```
erlendvollset, line 114:
```suggestion
    @overload
    def retrieve(self, external_id: str, ignore_unknown_ids: bool = False) -> Optional[DatapointSubscription]:
        ...
```

cognite/client/_api/datapoints_subscriptions.py
```diff
@@ -0,0 +1,259 @@
...
+            warn(self._warning_message, FutureWarning)
+
+        return self._update_multiple(
+            items=update,
+            list_cls=DataPointSubscriptionList,
+            resource_cls=DatapointSubscription,
+            update_cls=DataPointSubscriptionUpdate,
+        )
+
+    def list_data(
```
erlendvollset, line 189: I'm thinking that this should be a generator instead, and that the cursoring logic can be handled behind the scenes. i.e. `Batch:partitions` is never exposed.

cognite/client/data_classes/datapoints_subscriptions.py
```diff
@@ -0,0 +1,419 @@
...
+
+class _Metadata:
+    def __call__(self, key: str) -> list[str]:
+        return ["metadata", key]
+
+
+class DataPointSubscriptionFilterProperties:
+    description = ["description"]
+    external_id = ["externalId"]
+    metadata = _Metadata()
```
erlendvollset, line 409:
```suggestion
def _metadata(self, key: str) -> list[str]:
    return ["metadata", key]


class DataPointSubscriptionFilterProperties:
    description = ["description"]
    external_id = ["externalId"]
    metadata = _metadata
```

scripts/create_datapoint_subscription_test_data.py
```diff
@@ -0,0 +1,69 @@
...
+    client.time_series.subscriptions.create(sub2)
+    print("Created datapoint subscriptions sub2")
+    sub3 = DataPointSubscriptionCreate(
+        external_id="PYSDKDataPointSubscriptionTest3",
+        name="PYSDKDataPointSubscriptionTest3_10ts",
+        time_series_ids=TIMESERIES_EXTERNAL_IDS,
+        partition_count=2,
+    )
+
+    client.time_series.subscriptions.create(sub3)
```
erlendvollset, line 60: Ideally these are idempotent. If only sub2 happens to get deleted, we can't recreate it with this script since `.create(sub1)` would fail.

erlendvollset (approved):
LGTM!

PR #1259, opened 2023-06-21
Fix data_modeling handling non 200 status for apply endpoints.

doctrino:
### Fixed
- Handling non 200 responses in `data_modeling.spaces.apply`, `data_modeling.data_models.apply`, 
  `data_modeling.views.apply` and `data_modeling.containers.apply`

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/data_modeling/instances.py
+++ b/cognite/client/data_classes/data_modeling/instances.py
@@ -66,8 +66,13 @@ class NodeOrEdgeData:
 
     def dump(self, camel_case: bool = False) -> dict:
         output = asdict(self)
-        if self.source and isinstance(self.source, (ContainerId, ViewId)):
-            output["source"] = self.source.dump(camel_case)
+        if self.source:
+            if isinstance(self.source, (ContainerId, ViewId)):
+                output["source"] = self.source.dump(camel_case)
+            elif isinstance(self.source, dict):
+                output["source"] = self.source
+            else:
+                raise TypeError(f"source must be ContainerId, ViewId or a dict, but was {type(self.source)}")
         return convert_all_keys_to_camel_case_recursive(output) if camel_case else output
 
 
--- a/cognite/client/data_classes/data_modeling/instances.py
+++ b/cognite/client/data_classes/data_modeling/instances.py
@@ -66,7 +66,7 @@ class NodeOrEdgeData:
 
     def dump(self, camel_case: bool = False) -> dict:
         output = asdict(self)
-        if self.source:
+        if self.source and isinstance(self.source, (ContainerId, ViewId)):
             output["source"] = self.source.dump(camel_case)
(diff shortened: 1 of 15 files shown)


tests/tests_integration/test_api/test_data_modeling/test_containers.py
```diff
@@ -101,4 +98,53 @@ def test_retrieve_non_existent(self, cognite_client: CogniteClient):
...
+            cognite_client.data_modeling.containers.apply([valid_container, invalid_container])
+
+        # Assert
+        assert "One or more spaces do not exist" in error.value.message
+        assert error.value.code == 400
+        assert len(error.value.successful) == 1
+        assert len(error.value.failed) == 1
+
+        # Cleanup
+        cognite_client.data_modeling.instances.delete(valid_container.as_id())
```
erlendvollset, line 150: Cleanups need to either go in a fixture, or go in the `finally` block of a `try`. Otherwise cleanup won't be run if any part of the test fails.

tests/tests_integration/test_api/test_data_modeling/test_containers.py
```diff
@@ -101,4 +98,53 @@ def test_retrieve_non_existent(self, cognite_client: CogniteClient):
...
+                    properties={"name": dm.ContainerProperty(type=dm.Text())},
+                    used_for="node",
+                )
+            )
+
+        # Assert
+        assert error.value.code == 400
+        assert "One or more spaces do not exist" in error.value.message
+
+    def test_apply_failed_and_successful_task(
```
erlendvollset, line 118: Let's make sure we have these tests for all the resource types.

doctrino: Working on it :)

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -66,7 +66,7 @@ def load(cls, data: dict) -> NodeOrEdgeData:
 
     def dump(self, camel_case: bool = False) -> dict:
         output = asdict(self)
-        if self.source:
+        if self.source and isinstance(self.source, (ContainerId, ViewId)):
             output["source"] = self.source.dump(camel_case)
```
erlendvollset, line 70: When is source not `ContainerId` or `ViewId`?

When doing instance checks like this it's good practice to handle all known types and throw a type error when hitting an unexpected type. Makes debugging issues at runtime much easier.
```suggestion
        if self.source:
		if isinstance(self.source, (ContainerId, ViewId)):
	        	output["source"] = self.source.dump(camel_case)
	        else:
	        	raise TypeError(f"source must be ContainerId or ViewId, was {type(self.source)}")
```

doctrino: The motivation for this was a traceback I got passed, in which source was a dict which is acceptable. But yes will update it to follow this practice.

PR #1261, opened 2023-06-22
Use correct ID types and improve tests for dms

erlendvollset:
- Don't use ID base classes directly
- Be more thorough when testing. Don't just assert that collecgtion sizes are the same

## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/data_modeling/ids.py
+++ b/cognite/client/data_classes/data_modeling/ids.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import ABC
 from dataclasses import asdict, dataclass, field
-from typing import ClassVar, Literal, Optional, Sequence, Tuple, Type, TypeVar, Union, cast
+from typing import Any, ClassVar, Literal, Optional, Sequence, Tuple, Type, TypeVar, Union, cast
 
 from cognite.client.utils._auxiliary import rename_and_exclude_keys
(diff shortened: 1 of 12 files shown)


cognite/client/data_classes/data_modeling/ids.py
```diff
@@ -1,15 +1,24 @@
...
+from typing import Any, ClassVar, Literal, Optional, Sequence, Tuple, Type, TypeVar, Union, cast
 
 from cognite.client.utils._auxiliary import rename_and_exclude_keys
 from cognite.client.utils._identifier import DataModelingIdentifier, DataModelingIdentifierSequence
 from cognite.client.utils._text import convert_all_keys_recursive, convert_all_keys_to_snake_case
 
 
 @dataclass(frozen=True)
-class DataModelingId:
+class AbstractDataclass(ABC):
```
doctrino, line 13: What is the advantage of a shared base class?

doctrino: Ensure frozen?

doctrino (approved):
Just a question. Nice to see the cleanup and improvement.

cognite/client/data_classes/data_modeling/ids.py
erlendvollset: Ensuring that the direct subclasses of this class aren't instantiated directly. For Python to validate this at runtime without this, you need at least one abstract method defined (which we don't have here).

PR #1265, opened 2023-06-26
Run mypy on dms test code base aswell

erlendvollset:
There are a number of issues which tend to not be surfaced until you actually run mypy on the invoking code. If you wanna know more, check this out: https://sethmlarson.dev/tests-arent-enough-case-study-after-adding-types-to-urllib3. 

A few minor things got uncovered here, and it also reduces likelihood of introducing bugs in the future.

### DIFF
--- a/cognite/client/data_classes/data_modeling/views.py
+++ b/cognite/client/data_classes/data_modeling/views.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import json
-from abc import ABC
+from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass
 from typing import Any, Dict, Literal, Optional, Union, cast
 
@@ -18,7 +18,6 @@ from cognite.client.data_classes.data_modeling.data_types import (
 )
 from cognite.client.data_classes.data_modeling.filters import Filter
 from cognite.client.data_classes.data_modeling.ids import ContainerId, ViewId
-from cognite.client.utils._auxiliary import rename_and_exclude_keys
 from cognite.client.utils._text import (
     convert_all_keys_to_camel_case_recursive,
     convert_all_keys_to_snake_case,
@@ -105,7 +104,7 @@ class ViewApply(ViewCore):
     def load(cls, resource: dict | str) -> ViewApply:
         data = json.loads(resource) if isinstance(resource, str) else resource
         if "properties" in data and isinstance(data["properties"], dict):
-            data["properties"] = {k: ViewPropertyDefinition.load(v) for k, v in data["properties"].items()} or None
+            data["properties"] = {k: ViewPropertyApply.load(v) for k, v in data["properties"].items()} or None
 
         return cast(ViewApply, super().load(data))
 
@@ -173,7 +172,7 @@ class View(ViewCore):
     def load(cls, resource: dict | str) -> View:
         data = json.loads(resource) if isinstance(resource, str) else resource
         if "properties" in data and isinstance(data["properties"], dict):
-            data["properties"] = {k: ViewPropertyDefinition.load(v) for k, v in data["properties"].items()} or None
+            data["properties"] = {k: ViewProperty.load(v) for k, v in data["properties"].items()} or None
 
         return cast(View, super().load(data))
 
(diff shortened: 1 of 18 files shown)


tests/tests_integration/test_api/test_data_modeling/test_containers.py
```diff
@@ -1,23 +1,33 @@
...
-import cognite.client.data_classes.data_modeling as dm
 from cognite.client import CogniteClient
+from cognite.client.data_classes.data_modeling import (
+    ContainerApply,
+    ContainerId,
+    ContainerList,
+    ContainerProperty,
+    Int32,
+    Space,
+    Text,
```
doctrino, line 13: I will just note that I have the opposite personal preference. By importing `from cognite.client import data_modeling as dm` it becomes clear where the classes comes from. In addition, you avoid having a lot imports at the top like this.

tests/tests_integration/test_api/test_data_modeling/test_data_models.py
```diff
@@ -1,23 +1,35 @@
...
+    ContainerId,
+    DataModelApply,
+    DataModelId,
+    DataModelList,
+    MappedApplyPropertyDefinition,
+    Space,
+    View,
+    ViewApply,
+    ViewId,
+)
```
doctrino, line 16: Same here

tests/tests_unit/test_data_classes/test_data_models/test_data_types.py
```diff
@@ -24,22 +23,14 @@ class TestPropertyType:
...
 
         assert data == ContainerDirectRelation.load(data).dump(camel_case=True)
-
-    def test_load_dump_view_direct_relation(self):
-        data = {
-            "type": "direct",
-            "source": {"space": "mySpace", "externalId": "myId", "version": "myVersion", "type": "view"},
-        }
-
-        assert data == ViewDirectRelation.load(data).dump(camel_case=True)
```
doctrino, line 45: Why are you removing this test? Is it no longer valid?

tests/tests_unit/test_data_classes/test_data_models/test_views.py
```diff
@@ -1,33 +1,47 @@
...
+            "name": "fullName",
+            "nullable": False,
+            "autoIncrement": False,
+            "defaultValue": None,
+        }
+        actual = ViewProperty.load(input)
+
+        assert actual.dump() == {
+            "auto_increment": False,
+            "container": {"external_id": "myExternalId", "space": "mySpace"},
```
doctrino, line 20: Why is the `"type": container"` no longer in the dump? Should .load().dump() result in the same data?

cognite/client/data_classes/data_modeling/views.py
```diff
@@ -256,114 +255,90 @@ def __init__(
         self.include_global = include_global
 
 
-@dataclass
-class ViewDirectRelation(DirectRelation):
-    source: Optional[ViewId] = None
```
doctrino, line 261: Was this not in use?

tests/tests_unit/test_data_classes/test_data_models/test_views.py
doctrino: I see now that `ViewProperty` is an abstract class and does do not produce the same subclass.

cognite/client/data_classes/data_modeling/views.py
doctrino: I use this in `pygen`, so that will be a breaking change for that package.

doctrino (commented):
So for my understanding there are two changes here?
* Introduction of `mypy` on the data modeling tests, which discovered issues in particular with the views;
* The ViewProperties are not split into read and write classes. 

I guess you are planning to add it, but this definitely deserves a bump. I am fairly certain that this is going to break `pygen` which I released yesterday, so need a new version.

tests/tests_unit/test_data_classes/test_data_models/test_data_types.py
doctrino: I see you removed the class.

tests/tests_integration/test_api/test_data_modeling/test_containers.py
erlendvollset: Two reasons:
1) Consistency with rest of code base
2) It's important to me to have the imports at the top so that it is easy to see exactly what a given module depends on.

tests/tests_integration/test_api/test_data_modeling/test_data_models.py
erlendvollset: answered above.

cognite/client/data_classes/data_modeling/views.py
erlendvollset: We don't need ContainerDirectRelation or ViewDirectRelation. Those are just DirectRelation really. Added some type aliases for backwards compatiblity.

erlendvollset: Slight breaking change in that `source` is moved from `ViewDirectRelation` to `MappedProperty`, so that it's on the same level on input and output. This is just plaing wrong in the API, and I'd like it to make sense in the SDK.

tests/tests_integration/test_api/test_data_modeling/test_containers.py
erlendvollset: Also when i use the `dm.BlaBla` notation, intellij doesn't flag type errors in docstring examples.

erlendvollset: Making that entirely consistent in https://github.com/cognitedata/cognite-sdk-python/pull/1266. Also running docstring tests so we catch more issues there.

doctrino (approved):

PR #1270, opened 2023-06-27
feat: support for data model retrieve inline-views

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.


cognite/client/_api/data_modeling/data_models.py
```diff
@@ -90,11 +90,14 @@ def __iter__(self) -> Iterator[DataModel]:
         """
         return cast(Iterator[DataModel], self())
 
-    def retrieve(self, ids: DataModelIdentifier | Sequence[DataModelIdentifier]) -> DataModelList:
+    def retrieve(
+        self, ids: DataModelIdentifier | Sequence[DataModelIdentifier], inline_views: bool = False
+    ) -> DataModelList:
```
erlendvollset, line 95: Let's make `DataModel` generic on the return type of `DataModel:views` ( i.e. `list[View]` or `list[ViewId]`), then add overloads here on the value of `inline_views`

tests/tests_integration/test_api/test_data_modeling/test_data_models.py
```diff
@@ -24,6 +25,13 @@ def cdf_data_models(cognite_client: CogniteClient) -> DataModelList:
     return data_models
 
 
+@pytest.fixture(scope="function")
+def movie_model(cdf_data_models: DataModelList) -> DataModel:
+    movie_model = next(m for m in cdf_data_models if m.external_id == "Movie")
```
haakonvt, line 30: `assert` won't be reached as `next` raises `StopIteration` when not found. How about:
```suggestion
    movie_model = cdf_data_models.get(external_id="Movie")
```

tests/tests_integration/test_api/test_data_modeling/test_spaces.py
```diff
@@ -33,6 +33,11 @@ def test_list(self, cognite_client: CogniteClient, cdf_spaces: SpaceList) -> Non
 
         assert _dump(actual_space_in_cdf) == _dump(cdf_spaces)
 
+    def test_list_include_global(self, cognite_client: CogniteClient, cdf_spaces: SpaceList) -> None:
+        spaces = cognite_client.data_modeling.spaces.list(include_global=True)
+
+        assert len(spaces) > len(cdf_spaces)
```
erlendvollset, line 39: These length checks are a recurring theme, and are very brittle. Please be more specific about the assertions. You're also making strong assumptions on the nature of the data produced by the `cdf_spaces` fixture. It's important to design your tests in such a way that they don't risk false successes in the future if there are changes to the data. So one should be extra careful when using widely scoped fixtures.

I can think of several ways this can go wrong. E.g. we decide to filter down or limit the spaces in the `cdf_spaces` (for making the tests independent of other things going on the project for example), then include_global is broken so that it no longer returns global spaces, then this test will still pass.

A better implementation is to explicitly list all spaces with and without `include_global` and then checking that the first collection contains spaces with `is_global=True` and the other doesn't.

tests/tests_integration/test_api/test_data_modeling/test_views.py
```diff
@@ -79,6 +79,17 @@ def test_delete_non_existent(self, cognite_client: CogniteClient, integration_te
             == []
         )
 
+    def test_retrieve_without_inherited_properties(self, cognite_client: CogniteClient, cdf_views: ViewList) -> None:
+        assert len(cdf_views) >= 1, "Please add at least one view to the test environment"
```
erlendvollset, line 83: Let's put this assertion in the fixture instead? It will be entirely useless if doesn't return any data. It would be even better to retrieve them by ids there and raise if any are missing, then the fixture becomes self-documenting aswell.

tests/tests_integration/test_api/test_data_modeling/test_views.py
```diff
@@ -79,6 +79,17 @@ def test_delete_non_existent(self, cognite_client: CogniteClient, integration_te
...
+        # Arrange
+        view = cdf_views[0]
+
+        # Act
+        retrieved = cognite_client.data_modeling.views.retrieve(view.as_id(), include_inherited_properties=False)[0]
+
+        # Assert
+        assert retrieved is not None
```
erlendvollset, line 91: Above would raise KeyError if missing, so this won't be reached.
But let's actually assert that at least one of the properties are inherited (there is a flag on the property def which indicates if it's inherited or not) and vice versa when include_inherited_properties is False.

tests/tests_integration/test_api/test_data_modeling/test_data_models.py
```diff
@@ -115,6 +123,14 @@ def test_retrieve_multiple(self, cognite_client: CogniteClient, cdf_data_models:
         # Assert
         assert [dm.as_id() for dm in retrieved] == ids
 
+    def test_retrieve_with_inline_views(self, cognite_client: CogniteClient, movie_model: DataModel) -> None:
+        # Act
+        retrieved = cognite_client.data_modeling.data_models.retrieve(movie_model.as_id(), inline_views=True)[0]
+
+        # Assert
+        assert retrieved.views is not None
```
erlendvollset, line 133: Let's make DataModel.views a `List ` instead of `Optional[List]` so this check isn't necessary.

tests/tests_integration/test_api/test_data_modeling/test_data_models.py
```diff
@@ -115,8 +125,16 @@ def test_retrieve_multiple(self, cognite_client: CogniteClient, cdf_data_models:
         # Assert
         assert [dm.as_id() for dm in retrieved] == ids
 
+    def test_retrieve_with_inline_views(self, cognite_client: CogniteClient, movie_model: DataModel) -> None:
+        # Act
+        retrieved = cognite_client.data_modeling.data_models.retrieve(movie_model.as_id(), inline_views=True)[0]
+
+        # Assert
```
erlendvollset, line 132: Check the case when the flag is False also.

erlendvollset (changes requested):
include_global is supported on all the schema resources. Let's add it across the board.

tests/tests_integration/test_api/test_data_modeling/test_spaces.py
doctrino: Good point, a bit quick on writing this a similar tests.

tests/tests_integration/test_api/test_data_modeling/test_views.py
doctrino: I cannot see that flag in the spec. Which flag are you referring to?

tests/tests_integration/test_api/test_data_modeling/test_data_models.py
doctrino: I agree this is better, but just for my understanding we are breaking the spec (or the spec is documented wrongly) as views are not listed as required.

tests/tests_integration/test_api/test_data_modeling/test_views.py
erlendvollset: Hmm. Doesn't look like that's actually exposed through DMS, only in the internal API. Then this is fine for now.

cognite/client/data_classes/data_modeling/data_models.py
```diff
@@ -152,7 +155,14 @@ def dump(self, camel_case: bool = False) -> dict[str, Any]:
     def as_apply(self) -> DataModelApply:
         views: Optional[List[Union[ViewId, ViewApply]]] = None
         if self.views:
```
erlendvollset, line 157: This is redundant now since it can't be None

tests/tests_integration/test_api/test_data_modeling/test_data_models.py
erlendvollset: Still think you need to handle that. So Optional[List] in the parameters and then
```
self.views: List[T_View] = views or []
```
in the constructor body.

erlendvollset (approved):

PR #1271, opened 2023-06-27
Data Modeling Validate Only Write Classes

doctrino:
## Description

We had some old containers, and views which we are unable to delete as space and external id is validated upon creating the read classes. Moved that validation to only apply for the write classes, so that we can download these old containers, and deletl them. 

## Checklist:
- [ ] Tests added/updated.


erlendvollset (approved):
LGTM. Can consider adding a unit test

doctrino:
> 

I have more bugs to fix, so will include it in a later PR, as this is fairly urgent to get out.

PR #1279, opened 2023-07-07
Factory method for client config and credentials

doctrino:
## Description
I see these URLs copied around all over in solutions. Suggest to instead just include them in the SDK. 

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_cognite_client.py
+++ b/cognite/client/_cognite_client.py
@@ -28,6 +28,7 @@ from cognite.client._api.transformations import TransformationsAPI
 from cognite.client._api.vision import VisionAPI
 from cognite.client._api_client import APIClient
 from cognite.client.config import ClientConfig, global_config
+from cognite.client.credentials import CredentialProvider, OAuthClientCredentials, OAuthInteractive
 
 
 class CogniteClient:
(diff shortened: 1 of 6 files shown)


cognite/client/_cognite_client.py
```diff
@@ -111,3 +112,45 @@ def config(self) -> ClientConfig:
...
         return self._config
+
+    @classmethod
+    def default(
+        cls,
+        project: str,
+        cdf_cluster: str,
+        tenant_id: str,
+        client_id: str,
+        client_secret: str | None = None,
```
erlendvollset, line 123: Let's make this accept a credential provider instead, and rather rely on factory methods for those to ease instantiation. We can rather add `default_oauth_interactive` and `default_oauth_client_credentials` in addition.

cognite/client/config.py
```diff
@@ -101,3 +102,29 @@ def __str__(self) -> str:
 
     def _repr_html_(self) -> str:
         return str(self)
+
+    @classmethod
+    def create_default(
```
erlendvollset, line 107: Just `default`, to stay consitent with the method on CogniteClient?
```suggestion
    def default(
```

erlendvollset (approved):
LGTM

PR #1287, opened 2023-07-13
Upsert

doctrino:
## Description
Done ready for full review. 

Found a issue with `Sequence.columns` added an issue on it here; https://github.com/cognitedata/cognite-sdk-python/issues/1292 

But saving that implementation for later. 

## Checklist:
- [x] Tests added/updated.


cognite/client/_api_client.py
```diff
@@ -797,6 +802,88 @@ def _update_multiple(
...
+                    successful_resources = list_cls([successful_resources], cognite_client=self._cognite_client)
+
+            result = list_cls(
+                (successful_resources or []) + (created or []) + (updated or []), cognite_client=self._cognite_client
+            )
+
+            # Reorder to match the order of the input items
+            result_by_identifier = {identifier: item for item in result for identifier in [item.external_id, item.id]}
+            result = list_cls(
+                [result_by_identifier[item.external_id or item.id] for item in items],
```
haakonvt, line 879: You can't use `item.external_id or item.id` as `""` is a valid external ID.

cognite/client/_api_client.py
```diff
@@ -797,6 +802,88 @@ def _update_multiple(
...
+            to_create = [
+                items_by_external_id[external_id]
+                for external_id in not_found_error.failed
+                if external_id in missing_external_ids
+            ]
+
+            # Updates can have either external id or id. If they have an id, they must exist in CDF.
+            to_update = [
+                items_by_external_id[identifier] if isinstance(identifier, str) else items_by_id[identifier]
+                for identifier in not_found_error.failed
```
haakonvt, line 831: This list lookup (O(N^2)) should probably be rewritten to linear time complexity

cognite/client/_api/events.py
```diff
@@ -443,3 +443,38 @@ def search(self, description: str = None, filter: Union[EventFilter, Dict] = Non
...
+
+    @overload
+    def upsert(self, item: Sequence[Event]) -> EventList:
+        ...
+
+    @overload
+    def upsert(self, item: Event) -> Event:
+        ...
+
+    def upsert(self, item: Event | Sequence[Event]) -> Event | EventList:
```
haakonvt, line 455: Since this method doesn't support `EventUpdate`s, the upsert mode is a weird mix between overwrite and "leave as is":

```py
>>> ev = Event(external_id="x", metadata={"a": "1"}, type="foo")
>>> client.events.upsert(ev)
>>> ev.type = None
>>> ev.metadata = {"b": "2"}
>>> client.events.upsert(ev).dump()
{'type': 'foo',
 'metadata': {'b': '2'},
 ...}
```
Here, the `a` key in `metadata` is overwritten, but the event `type` is left as is. There would be no way for the user of the `upsert` method to remove an attribute i.e. set it to null.

I'm not sure what I think is best, but I think either the mode should be "overwrite all attributes", i.e. _here's how I want my event to look like_, or all updates should be "additive" (as in patching), i.e. if I send metadata, don't replace, but add such that existing keys/values are left untouched (this is also an issue for any attached `labels`).

What I did with `AssetsAPI.create_hierachy` was to have an `upsert_mode` keyword argument that could be set to either replace or patch.

tests/tests_integration/test_api/test_events.py
```diff
@@ -1,12 +1,15 @@
+from __future__ import annotations
+
 from datetime import datetime
 from unittest import mock
 
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
```
haakonvt, line 7: I think this works:
```suggestion
from pytest import MonkeyPatch
```

tests/tests_integration/test_api/test_events.py
```diff
@@ -109,3 +112,213 @@ def test_delete_with_nonexisting(self, cognite_client):
...
+            end_time=1,
+            subtype="mySubType2",
+        )
+        preexisting_update = Event._load(preexisting.dump(camel_case=True))
+        preexisting_update.subtype = "mySubType1"
+
+        try:
+            created_existing = cognite_client.events.create(preexisting)
+            assert created_existing is not None
+            monkeypatch.setattr(cognite_client.events, "_UPDATE_LIMIT", 1)
```
haakonvt, line 235: I would add an assert that verifies that you send multiple API requests just in case we rewrite the current thread-pool-executor-monster 😅

cognite/client/_api/events.py
doctrino: At least the many use cases I know about, would like this to be either create this event if it does not exists, or overwrite it with this new information if it does.

cognite/client/_api_client.py
```diff
@@ -797,6 +802,88 @@ def _update_multiple(
...
+                if identifier not in missing_external_ids or isinstance(identifier, int)
+            ]
```
haakonvt, line 833: Disregarding my previous comment about the loop, my brain wasn't working 😂

tests/tests_integration/test_api/test_events.py
doctrino: Do you have a simple code snippet for doing that?

cognite/client/_api/events.py
```diff
@@ -445,3 +446,50 @@ def search(
...
+        """Upsert events, i.e., update if it exists, and create if it does not exist.
+         Note this is a convenience method that handles the upserting for you by first calling update on all items,
+         and if any of them fail because they do not exist, it will create them instead.
+
+        Args:
+            item (Event | Sequence[Event]): Event or list of events to upsert.
+            mode (Literal['patch', "replace"]): Whether to patch or replace in the case the events are existing.
+
+        Returns:
+            EventList: List of upserted events.
```
erlendvollset, line 468:
```suggestion
            Event | EventList: The upserted event(s).
```

cognite/client/_api_client.py
```diff
@@ -797,6 +807,98 @@ def _update_multiple(
...
+            result_by_identifier = {
+                identifier: item
+                for item in result
+                for identifier in [item.external_id, item.id]
+                if identifier is not None
+            }
+            result = list_cls(
+                [result_by_identifier[item.external_id if item.external_id is not None else item.id] for item in items],
+                cognite_client=self._cognite_client,
+            )
```
haakonvt, line 896: The `result` variable already has the mapping stored on the object (to power the `.get` method), so you just need the reorder-code:

Here's a suggestion 🤓 
```suggestion
            # Reorder to match the order of the input items
            result.data = [
                result.get(**Identifier.load(item.id, item.external_id).as_dict(camel_case=False)) for item in items
            ]
```

cognite/client/_api_client.py
```diff
@@ -835,25 +937,51 @@ def _prepare_item_chunks(
...
+        update_attribute_by_name = {prop.name: prop for prop in update_attributes}
         for key, value in dumped_resource.items():
-            if to_snake_case(key) in update_attributes:
-                patch_object["update"][key] = {"set": value}
+            if (snake := to_snake_case(key)) not in update_attribute_by_name:
+                continue
+            prop = update_attribute_by_name[snake]
+            if prop.name == "columns":
+                continue
+                # Update columns properties in Sequence is not supported
```
haakonvt, line 966:
```suggestion
            if prop.name == "columns":
                # Update columns properties in Sequence is not supported
                continue
```

cognite/client/data_classes/assets.py
```diff
@@ -338,6 +339,22 @@ def labels(self) -> _LabelAssetUpdate:
...
 
+    @classmethod
+    def _get_update_properties(cls) -> list[PropertySpec]:
+        return [
+            # External ID is nullable, but is used in the upsert logic and thus cannot be nulled out.
+            PropertySpec("external_id", is_nullable=False),
+            PropertySpec("name", is_nullable=False),
+            PropertySpec("description"),
+            PropertySpec("data_set_id"),
+            PropertySpec("metadata", is_list=True),
```
haakonvt, line 350: The `is_list` name feels odd. What about `is_object` or `is_container`?

cognite/client/data_classes/_base.py
```diff
@@ -375,8 +383,9 @@ def dump(self, camel_case: bool = True) -> Dict[str, Any]:
         return dumped
 
     @classmethod
-    def _get_update_properties(cls) -> List[str]:
-        return [key for key in cls.__dict__ if not (key.startswith("_") or key == "columns")]
+    def _get_update_properties(cls) -> list[PropertySpec]:
+        # Default implementation should be overridden in subclasses
+        return [PropertySpec(key) for key in cls.__dict__ if not (key.startswith("_") or key == "columns")]
```
haakonvt, line 388: Is this ever used now? If no, should we make it raise an exception?

cognite/client/data_classes/files.py
```diff
@@ -280,6 +281,28 @@ def geo_location(self) -> _PrimitiveFileMetadataUpdate:
...
+            PropertySpec("metadata", is_list=True),
+            PropertySpec("asset_ids", is_list=True),
+            PropertySpec("source_created_time"),
+            PropertySpec("source_modified_time"),
+            PropertySpec(
+                "data_set_id",
+            ),
+            PropertySpec(
+                "security_categories",
+            ),
```
haakonvt, line 301:
```suggestion
            PropertySpec("data_set_id"),
            PropertySpec("security_categories"),
```

tests/tests_integration/test_api/test_events.py
haakonvt: Yup, check out the `post_spy`, which works like a "man in the middle" for you:
```py
@pytest.fixture
def post_spy(cognite_client):
    dps_api = cognite_client.time_series.data
    with patch.object(dps_api, "_post", wraps=dps_api._post):
        yield
```

haakonvt (commented):
This is good stuff, yo

cognite/client/_api/assets.py
```diff
@@ -671,6 +671,47 @@ def update(self, item: Union[Asset, AssetUpdate, Sequence[Union[Asset, AssetUpda
...
+        ...
+
+    def upsert(self, item: Asset | Sequence[Asset], mode: Literal["patch", "replace"]) -> Asset | AssetList:
+        """Upsert assets, i.e., update if it exists, and create if it does not exist.
+         Note this is a convenience method that handles the upserting for you by first calling update on all items,
+         and if any of them fail because they do not exist, it will create them instead.
+
+        Args:
+            item (Asset | Sequence[Asset]): Asset or list of assets to upsert.
+            mode (Literal['patch', "replace"]): Whether to patch or replace in the case the assets are existing.
```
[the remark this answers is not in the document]
haakonvt: (Multiple occurrences)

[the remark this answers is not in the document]
doctrino: Black seems to not care about docstrings :(

cognite/client/_api/files.py
```diff
@@ -465,6 +466,49 @@ def update(
...
+    @overload
+    def upsert(self, item: FileMetadata, mode: Literal["patch", "replace"]) -> FileMetadata:
+        ...
+
+    def upsert(
+        self, item: FileMetadata | Sequence[FileMetadata], mode: Literal["patch", "replace"]
+    ) -> FileMetadata | FileMetadataList:
+        """Upsert files, i.e., update if it exists, and create if it does not exist.
+         Note this is a convenience method that handles the upserting for you by first calling update on all items,
+         and if any of them fail because they do not exist, it will create them instead.
```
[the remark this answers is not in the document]
doctrino: I guess the use case would be if you want to update the metadata for a file, and not necessarily the file content. @erlendvollset Wdyt?

cognite/client/data_classes/assets.py
doctrino: I think `is_object` is misleading. Not a strong preference, although I lean towards staying `is_list`. @erlendvollset Thought?

erlendvollset (commented):
Let's add a section to the docs describing the patch vs. replace behaviour.

erlendvollset:
We should also make it _very_ explicit that this upsert operation is not atomic.

cognite/client/_api_client.py
```diff
@@ -835,25 +942,51 @@ def _prepare_item_chunks(
...
+        update_attribute_by_name = {prop.name: prop for prop in update_attributes}
         for key, value in dumped_resource.items():
-            if to_snake_case(key) in update_attributes:
-                patch_object["update"][key] = {"set": value}
+            if (snake := to_snake_case(key)) not in update_attribute_by_name:
+                continue
+            prop = update_attribute_by_name[snake]
+            if prop.name == "columns":
+                # Update column property in Sequence is not supported
+                continue
```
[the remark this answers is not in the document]
doctrino: Yes, it is leftover code.

cognite/client/data_classes/assets.py
doctrino: I think `is_primitive` would be the opposite of `is_list`? In that case I will go with `is_container`

erlendvollset (approved):
LGTM!

PR #1291, opened 2023-07-18
Dump of relationships

doctrino:
## Description
If you try to dump a relationship as into a yaml file, you get the following:
```bash
yaml.representer.RepresenterError: ('cannot represent an object', {'externalId': 'relationship_to.bid_matrix_generator_config_sequence'})
```

This fixes that bug.

## Checklist:
- [ ] Tests added/updated.


erlendvollset (approved):

haakonvt:
> > Should this be done in parent class or is this only an issue for this particular class?
> 
> I do not think so, as the properties of the class are not in the parent class. Starting to use something like `hasattr` seems hacky to me. But I will likely add some tests at some point that finds all subclasses of CogniteResource, create an instance, dump, serialize with json and yaml, and loads. So we can catch all the places this is an issue.

Nice, that would be a great test 👌 There's already code for finding the subclasses btw (which I added to ensure all APIs are available on the mocked client):
```py
from tests.utils import all_subclasses
```

PR #1299, opened 2023-07-24
feature: flag to expand metadata to columns in `to_pandas`

HaydenCognite:
Bumped `pyyaml` as it did not work to install the poetry environment otherwise.

## Description

Added the functionality for expanding metadata columns as discussed on Slack. Appreciate all feedback on the code.

Also, please advise what this consists of as far as version change goes. Seems like it is a minor change according to the semantic versioning link below.

## Checklist:
- [X] Tests added/updated.
- [X] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.


cognite/client/data_classes/_base.py
```diff
@@ -136,7 +136,10 @@ def _load(
         raise TypeError(f"Resource must be json str or dict, not {type(resource)}")
 
     def to_pandas(
-        self, expand: Sequence[str] = ("metadata",), ignore: Optional[List[str]] = None, camel_case: bool = False
+        self,
+        expand: Sequence[str] = ("metadata",),
+        ignore: Optional[List[str]] = None,
+        camel_case: bool = False,
```
doctrino, line 142: why did you add the extra comma here, you did no changes to this function? 
```suggestion
        camel_case: bool = False
```

cognite/client/data_classes/_base.py
```diff
@@ -145,6 +148,8 @@ def to_pandas(
                 Will expand metadata by default.
             ignore (List[str]): List of row keys to not include when converting to a data frame.
             camel_case (bool): Convert column names to camel case (e.g. `externalId` instead of `external_id`)
+            expand_metadata (bool): Expand the metadata column into separate columns.
+            metadata_prefix (str): Prefix to use for metadata columns.
```
doctrino, line 152: Copy pasted in the wrong place?

cognite/client/data_classes/_base.py
```diff
@@ -281,14 +285,32 @@ def get(self, id: Optional[int] = None, external_id: Optional[str] = None) -> Op
...
         """
         pd = cast(Any, utils._auxiliary.local_import("pandas"))
         df = pd.DataFrame(self.dump(camel_case=camel_case))
+
+        if expand_metadata and "metadata" in df.columns:
+            meta_df = pd.json_normalize(df["metadata"])
+            meta_df_prefix_cols = [f"{metadata_prefix}{col}" for col in meta_df.columns]
+            if common_cols := set(meta_df_prefix_cols).intersection(set(df.columns)):
+                raise ValueError(f"Metadata contains columns that are already present in the dataframe: {common_cols}")
+            df = pd.concat([df.drop("metadata", axis=1), meta_df.add_prefix(metadata_prefix)], axis=1)
```
doctrino, line 312: Some 🏌
```suggestion
            meta_df = pd.json_normalize(df["metadata"]).add_prefix(metadata_prefix)
            if common_cols := set(meta_df.columns).intersection(set(df.columns)):
                raise ValueError(f"Metadata contains columns that are already present in the dataframe: {common_cols}")
            df = pd.concat([df.drop("metadata", axis=1), meta_df], axis=1)
```

tests/tests_unit/test_base.py
```diff
@@ -239,6 +240,36 @@ def test_to_pandas_no_camels(self):
...
+        event_list = EventList(
+            [
+                Event(external_id="ev1", metadata={"value1": 1, "value2": "hello"}),
+                Event(external_id="ev2", metadata={"value1": 2, "value2": "world"}),
+            ]
+        )
+
+        expected_df = pd.DataFrame(
+            data={"external_id": ["ev1", "ev2"], "metadata.value1": [1, 2], "metadata.value2": ["hello", "world"]},
+            # index=[0, 1],
```
doctrino, line 256:
```suggestion
```

doctrino (commented):
A few minor things.

cognite/client/data_classes/_base.py
HaydenCognite: I accidentally put my code in `CogniteResource` instead of `CogniteResourceList`, that is why. 😅

cognite/client/data_classes/_base.py
HaydenCognite: Hang on, don't we want to leave the choice in the hands of the user @doctrino? Should we not keep the `if expand_metadata` bit in? Otherwise this might be a breaking change, right?

doctrino: Just a bad copy from me, I lost the first line. Fixed it now.

doctrino: the main argument for doing this change is to avoid doing the same operation (adding a prefix) in two different ways. If there were a bug in `[f"{metadata_prefix}{col}" for col in meta_df.columns]` it could pass the ValueError check, even though it should not have.

HaydenCognite: Yes, good point. 👍

cognite/client/data_classes/_base.py
```diff
@@ -281,14 +280,31 @@ def get(self, id: Optional[int] = None, external_id: Optional[str] = None) -> Op
...
             pandas.DataFrame: The dataframe.
         """
         pd = cast(Any, utils._auxiliary.local_import("pandas"))
         df = pd.DataFrame(self.dump(camel_case=camel_case))
+
+        if expand_metadata and "metadata" in df.columns:
+            meta_df = pd.json_normalize(df["metadata"]).add_prefix(metadata_prefix)
+            if common_cols := set(meta_df.columns).intersection(set(df.columns)):
+                raise ValueError(f"Metadata contains columns that are already present in the dataframe: {common_cols}")
+            df = pd.concat([df.drop("metadata", axis=1), meta_df], axis=1)
```
haakonvt, line 306: Question: Using `join` already gives a quite clear error message when columns overlap, example:
```py
pd.DataFrame({"id": [1,2,3], "foo": list("asd")}).join(
    pd.DataFrame({"foo": list("xyz")})
)
# --> Raises ValueError ...
#     ...columns overlap but no suffix specified: Index(['foo'], dtype='object')
```
Maybe we don't have to write our own error handling? 🤔

cognite/client/data_classes/_base.py
```diff
@@ -281,14 +280,31 @@ def get(self, id: Optional[int] = None, external_id: Optional[str] = None) -> Op
...
+        camel_case: bool = False,
+        expand_metadata: bool = False,
+        metadata_prefix: str = "metadata.",
+    ) -> pandas.DataFrame:
         """Convert the instance into a pandas DataFrame.
 
+        Args:
+            camel_case (bool): Convert column names to camel case (e.g. `externalId` instead of `external_id`)
+            expand_metadata (bool): Expand the metadata column into separate columns.
+            metadata_prefix (str): Prefix to use for metadata columns.
```
haakonvt, line 294: Should we add a field to the docstring explaining that this function raises when the metadata column names (after adding prefix) collide with existing columns?

haakonvt (commented):
This seems like a great quality-of-life upgrade 👌

cognite/client/data_classes/_base.py
HaydenCognite: Good point! With the concatenate function, this can be achieved using the `verify_integrity` flag. See attached. What are your opinions @haakonvt and @doctrino? I am in favour of taking away the freedom from the user here and forcing the verify integrity, and then removing the error checking to simplify the codebase/addition.

Screenshot 2023-07-25 at 12 28 18
Figure 1: Without verify_integrity flag.

Screenshot 2023-07-25 at 12 29 39
Figure 2: With verify_integrity flag.

cognite/client/data_classes/_base.py
```diff
@@ -281,14 +280,31 @@ def get(self, id: Optional[int] = None, external_id: Optional[str] = None) -> Op
...
+
         Returns:
             pandas.DataFrame: The dataframe.
         """
         pd = cast(Any, utils._auxiliary.local_import("pandas"))
         df = pd.DataFrame(self.dump(camel_case=camel_case))
+
+        if expand_metadata and "metadata" in df.columns:
+            meta_df = pd.json_normalize(df["metadata"]).add_prefix(metadata_prefix)
+            if common_cols := set(meta_df.columns).intersection(set(df.columns)):
```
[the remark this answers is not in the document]
HaydenCognite: Nice one!

cognite/client/data_classes/_base.py
HaydenCognite: Would you still do it if we remove the check as per my suggestion? My thoughts are that if it is not an error raised in the function itself, it should not be mentioned in the docstring. Only in the functions that they are raised in.

cognite/client/data_classes/_base.py
HaydenCognite: Nevermind, I read these comments in reverse order. 😅 I see now that the concat function would be replaced by the join by using the speedier implementation.

cognite/client/data_classes/_base.py
```diff
@@ -281,14 +280,31 @@ def get(self, id: Optional[int] = None, external_id: Optional[str] = None) -> Op
...
+                raise ValueError(f"Metadata contains columns that are already present in the dataframe: {common_cols}")
+            df = pd.concat([df.drop("metadata", axis=1), meta_df], axis=1)
+
         return convert_nullable_int_cols(df, camel_case)
```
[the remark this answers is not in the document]
HaydenCognite: No worries, I will reassign `df` to the last line and move it such that it happens before.

cognite/client/data_classes/_base.py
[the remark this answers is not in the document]
HaydenCognite: Is this faster because there is no prefix added to the metadata? I guess that would be an O(n) operation, if we let it fail (since we can do that with `join`), then we can keep the speed benefit. Right? This makes sense to me, but then I need to make a minor adjustment to the function and its test.

cognite/client/data_classes/_base.py
doctrino: I do not understand what you are referencing here Hayden. A line in the docstring explaining the possible ValueError is good.

haakonvt:
Remember to bump version & add a changelog entry. I'll have a look at your other comments later today 😄

cognite/client/data_classes/_base.py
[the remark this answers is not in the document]
doctrino: This makes `common_cols` an index so you would have to do this
not (common_cols := meta_df.columns.intersection(df.columns)).empty

which I think is less readable than the original implementation.

cognite/client/data_classes/_base.py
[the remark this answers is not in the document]
doctrino: In the code you have Håkon you lose the prefix. I guess the full code would be

```python
        if expand_metadata and "metadata" in df.columns:
            meta_series = (meta := df.pop("metadata"))[meta.notna()]
            meta_df = pd.DataFrame(meta_series.values.tolist(), index=meta_series.index).add_prefix(metadata_prefix)
            df = df.join(meta_df)
```

Here I also remove the error check, as `join` takes care of it.

cognite/client/data_classes/_base.py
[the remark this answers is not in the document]
doctrino: Will anyways be obsolete when we go for the `join`.

cognite/client/data_classes/_base.py
HaydenCognite: In general I agree @doctrino, but maybe consider this example, as this might illustrate better what I am trying to say. Let us say you have a function which calls several other functions. Maybe something like this:
```
def func(...):
    func_a(...)
    func_b(...)
    func_c(...)
```
Let us propose that each of these sub-functions all have their own error-raising capabilities and have been documented which includes documenting the errors and that no error-raising happens in `func()`. Would you repeat those errors in the docstring for `func()`? My thoughts are no, because they should already be documented in `func_a`, `func_b`, and `func_c`. @haakonvt wanted to change it such that we use the join instead, which would make this check obsolete since this type of check is already performed in the `join` method. At the end of the day, I will happily update the docs to include it if a majority wants it, but I just don't see the the point if we are using another function (join) which checks the error and no error raising happens in the one being written (to_pandas). 😄

cognite/client/data_classes/_base.py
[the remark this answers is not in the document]
HaydenCognite: @doctrino My question above is wondering whether the lost prefix was a deliberate choice/the source of the speedup. Maybe I am bad at wording things, but the prefix insertion is a linear operation, so maybe that is where the speed gains are happening. We could also just assume the end-user will not have issues with overlapping columns. It will make the code faster by not having the prefixes, or maybe doing a check in the code for the length of the prefix, so that it does not unnecessarily do the prefix operation.

CHANGELOG.md
haakonvt, line 1: Please revert `CHANGELOG.md` changes. If you want to add a linter for markdown, do so in a separate PR 🤙

CHANGELOG.md
```diff
@@ -17,152 +19,222 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.12.0] - 2023-07-26
+
+### Added
+
+* Added option in `.to_pandas()` method to include expanding metadata columns into a dataframe as well as an option to prefix those columns. Default is to prepend ".metadata" and `expand_metadata` defaults to `False` in order to maintain backwards compatibility.
+* Updated installation process for contributing.
```
haakonvt, line 27: Perhaps use `-` as all other entries do this 😅

haakonvt, line 27: The dot should come after metadata I think. I also took the liberty of rewriting slightly 😉 
```suggestion
- Added option `expand_metadata` to `.to_pandas()` method for list resource types which converts the metadata (if any) into separate columns in the returned dataframe. Also added `metadata_prefix` to control the naming of these columns (default is "metadata.").
```

tests/tests_unit/test_base.py
haakonvt, line 1: Lets add a test for empty metadata (all nulls) just in case.

haakonvt (changes requested):
A few more comments from me 😄

cognite/client/data_classes/_base.py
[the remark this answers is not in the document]
haakonvt: Agree, I guess only remove the last set then 😅

cognite/client/data_classes/_base.py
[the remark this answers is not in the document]
haakonvt: Yeah I dropped it from both the old and new to simplify the runtime examples as it has ~no effect

CHANGELOG.md
HaydenCognite: I did not see those changes. It must have happened in my editor. I will fix.

CHANGELOG.md
HaydenCognite: Again, must have been my editor. I fix.

CHANGELOG.md
haakonvt: No stress. Seems like a good change if you can add it to pre-commit 🚀

tests/tests_unit/test_base.py
HaydenCognite: Do you mean where there is no metadata initialised, or if there is an empty dictionary? Appears they are different when I try these two.

haakonvt: That's a good point; if none of the resource elements (Asset/Event etc) have metadata, then that column won't even exist, so "all null" can't happen. Then, just make sure one of the tests has some nulls 👍 (Maybe this is already the case)

HaydenCognite: Will do!

cognite/client/data_classes/_base.py
HaydenCognite: Discussion was resolved on Slack. We decided to add a comment in the docstrings as this would be useful for end users (especially new end users).

tests/tests_unit/test_base.py
```diff
@@ -239,6 +240,35 @@ def test_to_pandas_no_camels(self):
...
+                Event(external_id="ev2", metadata={"value1": 2, "value2": "world"}),
+            ]
+        )
+
+        expected_df = pd.DataFrame(
+            data={"external_id": ["ev1", "ev2"], "metadata.value1": [1, 2], "metadata.value2": ["hello", "world"]},
+        )
+
+        actual_df = event_list.to_pandas(expand_metadata=True)
+        pd.testing.assert_frame_equal(expected_df, actual_df, check_like=True)
```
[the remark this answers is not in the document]
HaydenCognite: Easiest rewrite ever (just had to remove it).

tests/tests_unit/test_base.py
```diff
@@ -239,6 +240,42 @@ def test_to_pandas_no_camels(self):
...
+        event_list = EventList(
+            [Event(external_id="ev1", metadata={"val1": 1}), Event(external_id="ev2", metadata={"val2": 2})]
+        )
+        expected_df = pd.DataFrame(
+            data={
+                "external_id": ["ev1", "ev2"],
+                "metadata.val1": [1, None],
+                "metadata.val2": [None, 2],
+            }
+        )
```
HaydenCognite, line 274: I assume this is what you meant with some nulls @haakonvt?

haakonvt: Looks good!

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.12.0] - 2023-07-26
+### Added
+- Added option `expand_metadata` to `.to_pandas()` method for list resource types which converts the metadata (if any) into separate columns in the returned dataframe. Also added `metadata_prefix` to control the naming of these columns (default is "metadata.").
+- Updated poetry installation process for contributing.
```
haakonvt, line 23: Please remove the “updated poetry…”

doctrino (approved):

cognite/client/data_classes/_base.py
```diff
@@ -281,15 +278,33 @@ def get(self, id: Optional[int] = None, external_id: Optional[str] = None) -> Op
...
+            pandas.DataFrame: The Cognite resource as a dataframe.
         """
         pd = cast(Any, utils._auxiliary.local_import("pandas"))
         df = pd.DataFrame(self.dump(camel_case=camel_case))
-        return convert_nullable_int_cols(df, camel_case)
+        df = convert_nullable_int_cols(df, camel_case)
+
+        if expand_metadata and "metadata" in df.columns:
+            # Equivalent to pd.json_normalize(df["metadata"]) but is a faster implementation.
+            meta_series = (meta := df.pop("metadata"))[meta.notna()]
```
haakonvt, line 304: Last comment, not sure what I was thinking here lol
```suggestion
            meta_series = df.pop("metadata").dropna()
```

haakonvt (approved):
One last comment then... 

# 🚀

cognite/client/data_classes/_base.py
haakonvt: ...also not sure you need the comment regarding pd.json_normalize

HaydenCognite: Well @doctrino mentioned it would be nice to add, and I thought it was fine to add too. 😄 🤷‍♂️ You are a wizard @haakonvt. 🧙‍♂️  So it is good to make sure people understand the where/why of the wizardry. In the (maybe distant) future, someone might replace this more syntactically-complex but algorithmically-more-efficient solution with a less-verbose but more-inefficient solution like `pd.json_normalize` without realising it is less efficient.

PR #1315, opened 2023-08-14
Fix bug view apply

doctrino:
### Fixed
- Fixed bug in `ViewApply.properties` had type hint `ConnectionDefinition` instead of `ConnectionDefinitionApply`.
- Fixed bug in `dump` methods of `ViewApply.properties` causing the return code `400` with message
  `Request had 1 constraint violations. Please fix the request and try again. [type must not be null]` to be returned
  from the CDF API.

In addition, fixing failing subscription test on `master`. 

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,14 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.13.3] - 2023-08-14
+### Fixed
+- Fixed bug in `ViewApply.properties` had type hint `ConnectionDefinition` instead of `ConnectionDefinitionApply`.
+- Fixed bug in `dump` methods of `ViewApply.properties` causing the return code `400` with message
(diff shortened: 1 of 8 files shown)


erlendvollset (approved):
LGTM

CHANGELOG.md
```diff
@@ -17,6 +17,14 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.13.3] - 2023-08-14
+### Fixed
+- Fixed bug in `ViewApply.properties` had type hint `ConnectionDefinition` instead of `ConnectionDefinitionApply`.
```
haakonvt, line 22: What was the bug? I (personally) don't think we need to update the changelog with updates/fixes to type annotations 😄

doctrino: I think it is still nice to include, as it makes it clear for posterity that this was a mistake.

haakonvt: ...then the add `pydoclint` PR will be 1000+ lines in the changelog 😂

PR #1317, opened 2023-08-16
File download keep folders

CogniteJohan:
## Description
Added parameter `keep_folder_structure`to `client.files.download` to allow downloading files to a folder structure matching the one in CDF.
This also keeps all files when downloading, and avoid the problem of files being ignored if they have the same name in different folders as @padalby made me aware of.

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/files.py
```diff
@@ -767,12 +769,30 @@ def download(
...
+                if metadata.directory:
+                    cdf_directory = Path(metadata.directory[1:])  # Making the directory relative
+                else:
+                    cdf_directory = Path("")
+                file_folder = directory / cdf_directory
+
+                file_folder.mkdir(parents=True, exist_ok=True)
+                self._download_files_to_directory(
+                    directory=file_folder,
+                    all_ids=[item for item in all_ids if item.get("id") == metadata.id],
```
erlendvollset, line 791: This breaks if external ids are used.

Also, this doesn't need to be quadratic in time. Build a identifier -> metadata map upfront.

cognite/client/_api/files.py
```diff
@@ -767,12 +769,30 @@ def download(
...
             directory = Path(directory)
+        assert directory.is_dir(), f"{directory} is not a directory"
+
         all_ids = IdentifierSequence.load(id, external_id).as_dicts()
         id_to_metadata = self._get_id_to_metadata_map(all_ids)
-        assert directory.is_dir(), f"{directory} is not a directory"
-        self._download_files_to_directory(directory, all_ids, id_to_metadata)
+
+        if keep_folder_structure:
+            for metadata in id_to_metadata.values():
```
erlendvollset, line 781: Doing these one by one is going to be significantly slower than the other implementation. Let's download them in parallel, and then create the necessary subdural, ane then move the files to the appropriate path instead.

tests/tests_unit/test_api/test_files.py
```diff
@@ -131,6 +131,35 @@ def download_link_callback(request):
...
+        rsps.POST,
+        cognite_client.files._get_base_url_with_base_path() + "/files/byids",
+        status=200,
+        json={"items": [{"id": 1, "name": "file_a", "directory": "/rootdir/subdir"}, {"id": 2, "name": "file_a"}]},
+    )
+
+    def download_link_callback(request):
+        identifier = jsgz_load(request.body)["items"][0]
+        response = {}
+        if "id" in identifier and identifier.get("id") == 1:
```
erlendvollset, line 146: You don't need to check for the existense of the key when using .get()

tests/tests_unit/test_api/test_files.py
```diff
@@ -491,6 +520,19 @@ def test_download(self, cognite_client, mock_file_download_response):
             with open(fp2, "rb") as fh:
                 assert b"content2" == fh.read()
 
+    def test_download_with_folder_structure(self, cognite_client, mock_file_download_response_with_folder_structure):
+        with TemporaryDirectory() as dir:
+            cognite_client.files.download(directory=dir, id=[1, 2], keep_folder_structure=True)
```
erlendvollset, line 525: Let's use a mix of internal and external ids.

cognite/client/_api/files.py
```diff
@@ -744,17 +744,19 @@ def download(
...
         id: Optional[Union[int, Sequence[int]]] = None,
         external_id: Optional[Union[str, Sequence[str]]] = None,
+        keep_folder_structure: bool = False,
     ) -> None:
         """`Download files by id or external id. <https://developer.cognite.com/api#tag/Files/operation/downloadLinks>`_
 
         This method will stream all files to disk, never keeping more than 2MB in memory per worker.
         The files will be stored in the provided directory using the name retrieved from the file metadata in CDF.
-
+        You can choose to keep the folder structure from CDF or not.
```
erlendvollset, line 753: Some more details would be good. Specify that the `directory` field will be respected for each file relative to the provided root directory.

erlendvollset (changes requested):

CogniteJohan:
@erlendvollset  the build/lint fails because of some problems in docstrings outside of my changes. The file downloads has been rewritten to accommodate for concurrent downloads.

haakonvt:
> @erlendvollset the build/lint fails because of some problems in docstrings outside of my changes. The file downloads have been rewritten to accommodate for concurrent downloads.

I think this is a caching issue of some sort; running the same pre-commit locally does not give any raised problems/errors - and it didn't do so in past runs in Github either. Seems that an older version is installed for some reason, veeeery weird, as it is fixed in the pre-commit config file. Anyway, I bumped it again in #1337 and it now seems to run smoothly again.

cognite/client/_api/files.py
haakonvt, line 1: You still overwrite files btw 😉  All files that are downloaded with both `name` and `directory` attributes matching --> only one file will come out victorious.

I suggest you actually raise when this is the case (as people would use this new argument and _trust_ that all files were downloaded imo)

cognite/client/_api/files.py
```diff
@@ -600,17 +600,19 @@ def download(
...
         external_id: str | Sequence[str] | None = None,
+        keep_folder_structure: bool = False,
     ) -> None:
         """`Download files by id or external id. <https://developer.cognite.com/api#tag/Files/operation/downloadLinks>`_
 
         This method will stream all files to disk, never keeping more than 2MB in memory per worker.
-        The files will be stored in the provided directory using the name retrieved from the file metadata in CDF.
-
+        The files will be stored in the provided directory using the file name retrieved from the file metadata in CDF.
+        You can also choose to keep the folder structure from CDF, where the files will be stored in subfolders under directory named after the directory field in CDF.
```
haakonvt, line 609: Since the original text (and the field name used by CDF) uses _directory_ (for folder), I suggest we stick to this name (also in the argument name, i.e. `keep_directory_structure` or `keep_dir_structure`

cognite/client/_api/files.py
```diff
@@ -620,12 +622,34 @@ def download(
...
+        else:
+            ids: list[dict[str, int | str]] = []
+            file_folders: list[Path] = []
+            for _id, _metadata in id_to_metadata.items():
+                if isinstance(_id, int):
+                    if _metadata.directory:
+                        cdf_directory = Path(_metadata.directory[1:])  # Making the directory relative
+                    else:
+                        cdf_directory = Path("")
+                    file_folder = directory / cdf_directory
```
haakonvt, line 644: Just some 🏌️ 
```suggestion
                    file_folder = directory
                    if _metadata.directory:
                        file_folder /= _metadata.directory[1:]  # Making the directory relative
```

cognite/client/_api/files.py
```diff
@@ -620,12 +622,34 @@ def download(
...
+        assert directory.is_dir(), f"{directory} is not a directory"
+
         all_ids = IdentifierSequence.load(id, external_id).as_dicts()
         id_to_metadata = self._get_id_to_metadata_map(all_ids)
-        assert directory.is_dir(), f"{directory} is not a directory"
-        self._download_files_to_directory(directory, all_ids, id_to_metadata)
+
+        if not keep_folder_structure:
+            self._download_files_to_directory(directory, all_ids, id_to_metadata)
+        else:
```
haakonvt, line 635: If you return here, you may de-indent the whole rest of the function 🚀 
```py
if not keep_folder_structure:
    self._download_files_to_directory(directory, all_ids, id_to_metadata)
    return

```

haakonvt (commented):
A little feedback 😄 Great functionality! Surprised this hasn't been raised / wished for earlier 😅

cognite/client/_api/files.py
CogniteJohan: As agreed on Slack, in case of duplicates we send a warning but let the download go through. We can consider an new PR adding a flag where duplicates can add external id at end of file name so that we can get all files locally

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.16.0] - 2023-08-28
+### Added
+- Added parameter `keep_folder_structure`to `client.files.download` to allow downloading files to a folder structure matching the one in CDF.
+This also keeps all files when downloading, and avoid the problem of files being ignored if they have the same name in different folders.
```
haakonvt, line 23:
```suggestion
- Added parameter `keep_directory_structure`to `client.files.download` to allow downloading files to a folder structure matching the one in CDF.

### Improved
- Using `client.files.download` will still skip files with the same name when writing to disk, but now a `UserWarning` is raised, specifying which files are concerned.
```

cognite/client/_api/files.py
```diff
@@ -620,12 +626,53 @@ def download(
...
-        assert directory.is_dir(), f"{directory} is not a directory"
-        self._download_files_to_directory(directory, all_ids, id_to_metadata)
+
+        if not keep_directory_structure:
+            all_file_names = [
+                cast(str, _metadata.name) for _id, _metadata in id_to_metadata.items() if isinstance(_id, int)
+            ]
+            duplicate_names = [name for name, count in collections.Counter(all_file_names).items() if count > 1]
+            if duplicate_names:
+                duplicate_names.sort()
```
haakonvt, line 643: We have a utils fn for this already:
```py
from cognite.client.utils._auxiliary import find_duplicates

duplicate_names = sorted(find_duplicates(all_file_names))
if duplicate_names:
```

cognite/client/_api/files.py
```diff
@@ -620,12 +626,53 @@ def download(
...
+                warning_message = f"""There are {len(duplicate_names)} duplicate file names.
+    Only the contents of one of the files with the same name will be downloaded to the same directory.
+    This concerns: {duplicate_names}"""
```
haakonvt, line 646: Using implicit string concat. is preferred:
```suggestion
warning_message = (
                    f"There are {len(duplicate_names)} duplicate file names. Only the contents of one of the files "
                    f"with the same name will be downloaded to the same directory. This concerns: {duplicate_names}"
                )
```

cognite/client/_api/files.py
```diff
@@ -620,12 +626,53 @@ def download(
...
+                file_directories.append(file_directory)
+                full_file_names.append(str((file_directory / cast(str, _metadata.name)).resolve()))
+
+        full_duplicate_names = [name for name, count in collections.Counter(full_file_names).items() if count > 1]
+        if full_duplicate_names:
+            full_duplicate_names.sort()
+            warning_message = f"""There are {len(full_duplicate_names)} duplicate file names.
+    Only the contents of one of the files with the same name will be downloaded to the same directory.
+    This concerns: {full_duplicate_names}"""
+            warnings.warn(message=warning_message, stacklevel=2)
```
haakonvt, line 670: Much of this feels duplicated, can you split it out into a separate function?

tests/tests_unit/test_api/test_files.py
```diff
@@ -491,6 +525,22 @@ def test_download(self, cognite_client, mock_file_download_response):
...
+            )
+
+            fp1 = os.path.join(dir, "rootdir/subdir/file_a")
+            fp2 = os.path.join(dir, "file_a")
+            assert os.path.isfile(fp1)
+            assert os.path.isfile(fp2)
+            with open(fp1, "rb") as fh:
+                assert b"contentSubDir" == fh.read()
+            with open(fp2, "rb") as fh:
+                assert b"contentNoDir" == fh.read()
```
haakonvt, line 542: Using `pathlib.Path` is preferred, but not required 😄

cognite/client/_api/files.py
```diff
@@ -620,12 +626,53 @@ def download(
                 >>> c = CogniteClient()
                 >>> c.files.download(directory="my_directory", id=[1,2,3], external_id=["abc", "def"])
         """
+
         if isinstance(directory, str):
             directory = Path(directory)
+        assert directory.is_dir(), f"{directory} is not a directory"
```
haakonvt, line 632: Asserts are not meant for the end user, raise an exception instead (`IsADirectoryError`)

haakonvt: Sorry this was a stupid suggestion, the problem is that it's not a directory 😂 This should of course be `NotADirectoryError ` :hide: @CogniteJohan

haakonvt (approved):
LGTM, good work! Please allow @erlendvollset the opportunity for one last look before merging 😄

erlendvollset (approved):
LGTM

PR #1323, opened 2023-08-20
Make dps test data always deterministic in xdist spawned workers

haakonvt:
## Description
A lot of the test data for DatapointsAPI are based on randomness, but when parallelizing workers with `pytest-xdist` we need to use a seed that makes this generation exactly equal across the 8(?) or so workers. Previously this was done by rounding `time.time()`, but ofc, that breaks every so often... https://github.com/cognitedata/cognite-sdk-python/actions/runs/5918418047/job/16047728671?pr=1322

This however, is a thing of the past as we change to using a unique identifier for each test run, provided by `pytest` itself.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/tests/tests_integration/test_api/test_datapoints.py
+++ b/tests/tests_integration/test_api/test_datapoints.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 import itertools
 import random
 import re
-import time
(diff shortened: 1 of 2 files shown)


erlendvollset (approved):
LGTM

PR #1324, opened 2023-08-22
Make `DatapointSubscriptionFilterProperties` properties immutable

haakonvt:
## Description
A newer version of ruff forces these to be annotated with ClassVar... because they are mutable, which feels wrong to me, so I made them into properties.

### DIFF
--- a/cognite/client/data_classes/datapoints_subscriptions.py
+++ b/cognite/client/data_classes/datapoints_subscriptions.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import json
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Literal, Optional, Type
+from enum import auto
+from typing import TYPE_CHECKING, Any, Optional, Type
 
 from cognite.client.data_classes import Datapoints, filters
 from cognite.client.data_classes._base import (
@@ -11,6 +12,7 @@ from cognite.client.data_classes._base import (
     CogniteResource,
     CogniteResourceList,
     CogniteUpdate,
+    EnumProperty,
     PropertySpec,
     T_CogniteResource,
 )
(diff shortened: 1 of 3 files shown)


cognite/client/data_classes/datapoints_subscriptions.py
```diff
@@ -371,21 +371,58 @@ class DatapointSubscriptionList(CogniteResourceList[DatapointSubscription]):
...
-    unit = ["unit"]
-    asset_id = ["assetId"]
-    asset_root_id = ["assetRootId"]
-    created_time = ["createdTime"]
-    data_set_id = ["dataSetId"]
-    id = ["id"]
-    last_updated_time = ["lastUpdatedTime"]
-    is_step = ["isStep"]
-    is_string = ["isString"]
+class _DatapointSubscriptionFilterProperties:
```
erlendvollset, line 374: Does a frozen data class not work?

haakonvt:
> Could also consider the enum approach Anders introduces in the advanced filtering PR

@erlendvollset Excellent suggestions, really liked this approach to it 👌 Have updated the PR with some additional magic that can be reverted, lmk!

cognite/client/data_classes/datapoints_subscriptions.py
haakonvt: Went with enum instead 👍

cognite/client/data_classes/datapoints_subscriptions.py
```diff
@@ -369,17 +371,17 @@ def _metadata(key: str) -> list[str]:
     return ["metadata", key]
 
 
-class DatapointSubscriptionFilterProperties:
-    description = ["description"]
-    external_id = ["externalId"]
+class DatapointSubscriptionFilterProperties(EnumProperty):
+    description = auto()
+    external_id = auto()
     metadata = _metadata
```
erlendvollset, line 377: I think this breaks filtering on metadata properties. Since this is an EnumProperty, a few helpers will invoke `.as_reference()` on it, which will result in `[["metadata", "<key>"]]` now rather than `["metadata", "<key>"]` when using `DatapointSubscriptionFilterProperties.metadata("<key>")`

erlendvollset (changes requested):

cognite/client/data_classes/datapoints_subscriptions.py
haakonvt: No, it's just a string, so it will pass through as expected (I've tested)

erlendvollset (approved):

PR #1328, opened 2023-08-25
Specify all supported Python versions in Black config

MortGron:
## Description
Black documentation says: 

> -t, --target-version
Python versions that should be supported by Black’s output. You should include all versions that your code supports.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,12 +21,12 @@ repos:
       - id: black
         exclude: ^cognite/client/_proto.*$
         args:
-          - --line-length
-          - '120'
-          - --target-version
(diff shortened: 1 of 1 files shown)


erlendvollset (approved):
LGTM. But I don't understand the point of specifying the upper bound. Won't this in practice have to be the same as specifying only py38?

MortGron:
> LGTM. But I don't understand the point of specifying the upper bound. Won't this in practice have to be the same as specifying only py38?

Might be because new minor versions of Python are not fully backward compatible but that is just speculation from my side on why Black wants it that way.

PR #1333, opened 2023-08-27
Annotations cleanup

haakonvt:
## Description

- fix `Function.retrieve_call` return annot
- Fix wrong annots for `Sequence.rows`
- Fix retrieve (single) methods in `FunctionsAPI`


erlendvollset (approved):
LGTM! Running mypy on the test code will help us catch these types of issues statically. It should be a long term goal to run mypy on everything, but we can introduce it incrementally to the test code base.

PR #1334, opened 2023-08-27
Make `limit` optional for auto-paging endpoints

haakonvt:
## Description
We need to clearly distinguish between endpoints that auto-pages for the user (i.e. support "unlimited") and i.e. search endpoints that have a clearly defined max `limit`.

This PR tries to...

- Annotate limit as int OR None for those supporting unlimited
- Remove docstring saying "pass -1, None... to fetch all" where not possible
- Changing default limit away from a constant, and back to a specific number (how basic resource types are implemented)

## Checklist:
- [ ] Tests added/updated.


cognite/client/_api/annotations.py
```diff
@@ -173,12 +173,12 @@ def reverse_lookup(self, filter: AnnotationReverseLookupFilter, limit: int | Non
...
         )
 
-    def list(self, filter: AnnotationFilter | dict, limit: int = LIST_LIMIT_DEFAULT) -> AnnotationList:
+    def list(self, filter: AnnotationFilter | dict, limit: int | None = LIST_LIMIT_DEFAULT) -> AnnotationList:
         """`List annotations. <https://developer.cognite.com/api#tag/Annotations/operation/annotationsFilter>`_
 
         Args:
             filter (AnnotationFilter | dict): Return annotations with parameter values that matches what is specified. Note that annotated_resource_type and annotated_resource_ids are always required.
-            limit (int): Maximum number of annotations to return. Defaults to 25.
+            limit (int | None): Maximum number of annotations to return. Defaults to 25.
```
doctrino, line 181:
```suggestion
            limit (int | None): Maximum number of annotations to return. Defaults to 25. Set to -1, float("inf") or None to return all items.
```

cognite/client/_api/data_modeling/instances.py
```diff
@@ -532,7 +532,7 @@ def search(
         instance_type: Literal["node"] = "node",
         properties: list[str] | None = None,
         filter: Filter | dict | None = None,
-        limit: int = INSTANCES_LIST_LIMIT_DEFAULT,
+        limit: int = 1000,
```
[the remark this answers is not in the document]
doctrino: I agree with Erlend here. I think I did a mistake here by introducing `INSTANCES_LIST_LIMIT_DEFAULT = 1000`, I think instead this one should be set to `LIST_LIMIT_DEFAULT`  thus 25. And then let the user have to explicitly set limit if they want more.

cognite/client/_api/data_modeling/instances.py
```diff
@@ -555,7 +555,7 @@ def search(
         instance_type: Literal["node", "edge"] = "node",
         properties: list[str] | None = None,
         filter: Filter | dict | None = None,
-        limit: int = INSTANCES_LIST_LIMIT_DEFAULT,
+        limit: int = 1000,
```
doctrino, line 558:
```suggestion
        limit: int = LIST_LIMIT_DEFAULT,
```
Suggest doing the same here, with same argument as above.

cognite/client/_api/data_modeling/instances.py
```diff
@@ -857,7 +857,7 @@ def list(
         instance_type: Literal["node"] = "node",
         include_typing: bool = False,
         sources: ViewIdentifier | Sequence[ViewIdentifier] | View | Sequence[View] | None = None,
-        limit: int = INSTANCES_LIST_LIMIT_DEFAULT,
+        limit: int | None = INSTANCES_LIST_LIMIT_DEFAULT,
```
doctrino, line 860:
```suggestion
        limit: int | None = LIST_LIMIT_DEFAULT,
```
Suggest removing the `INSTANCES_LIST_LIMIT_DEFAULT` with same argument as above.

doctrino (commented):
I will repeat my main issue here:

I think the introduction of the constant `INSTANCES_LIST_LIMIT_DEFAULT` was a mistake by me. The SDK is used for exploration and then I think a limit of 25 aka `LIST_LIMIT_DEFAULT` is much better, and instead force the user to explicitly request more in case they need it.

cognite/client/_api/data_modeling/instances.py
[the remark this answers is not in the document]
haakonvt: > Why are we changing these to not be constants?

Because that is the de facto standard in the SDK:
```
# Defaults:
[('AssetsAPI    , list limit: 25, search limit: 100',),
 ('DocumentsAPI , list limit: 25, search limit: 25',),
 ('EventsAPI    , list limit: 25, search limit: 100',),
 ('FilesAPI     , list limit: 25, search limit: 100',),
 ('SequencesAPI , list limit: 25, search limit: 100',),
 ('TimeSeriesAPI, list limit: 25, search limit: 100',)]
```
All 25's are `LIST_LIMIT_DEFAULT` and all 100s are literals.

> The default should not be the max btw - it encourages abuse of the API. Most use cases only need the first few results

Very much agree, but my change left the value unchanged, `INSTANCES_LIST_LIMIT_DEFAULT` is set to 1000.

cognite/client/_api/annotations.py
haakonvt: Nice catch 👌

cognite/client/_api/data_modeling/instances.py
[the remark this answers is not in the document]
haakonvt: I guess my real question here is: why shouldn't just all endpoints default to something small like 25? 🤔 What is the logic behind 100 for search and 25 for list? 🤔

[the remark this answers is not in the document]
haakonvt: Personally, when using search, I typically need/use 1-5 top entries

cognite/client/_api/iam.py
```diff
@@ -97,11 +97,11 @@ def delete(self, id: int | Sequence[int]) -> None:
 class SecurityCategoriesAPI(APIClient):
     _RESOURCE_PATH = "/securitycategories"
 
-    def list(self, limit: int = LIST_LIMIT_DEFAULT) -> SecurityCategoryList:
+    def list(self, limit: int | None = LIST_LIMIT_DEFAULT) -> SecurityCategoryList:
         """`List security categories. <https://developer.cognite.com/api#tag/Security-categories/operation/getSecurityCategories>`_
 
         Args:
-            limit (int): Max number of security categories to return. Defaults to 25.
+            limit (int | None): Max number of security categories to return. Defaults to 25.
```
haakonvt, line 104:
```suggestion
            limit (int | None): Max number of security categories to return. Defaults to 25. Set to -1, float("inf") or None to return all items.
```

cognite/client/_api/labels.py
```diff
@@ -60,7 +60,7 @@ def list(
             external_id_prefix (str | None): filter label definitions with external ids starting with the prefix specified
             data_set_ids (int | Sequence[int] | None): return only labels in the data sets with this id / these ids.
             data_set_external_ids (str | Sequence[str] | None): return only labels in the data sets with this external id / these external ids.
-            limit (int): Maximum number of label definitions to return.
+            limit (int | None): Maximum number of label definitions to return.
```
haakonvt, line 63:
```suggestion
            limit (int | None): Maximum number of label definitions to return. Defaults to 25. Set to -1, float("inf") or None to return all items.
```

erlendvollset (approved):

PR #1335, opened 2023-08-27
Add `UserProfilesAPI`

haakonvt:
## Description
Added the user profiles API.

See discission around `/byids` not returning in the same order: https://cognitedata.slack.com/archives/C85QG3UBU/p1693303250158219

# TODO:
- Documentation not added

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/user_profiles.py
+++ b/cognite/client/_api/user_profiles.py
@@ -1,0 +1,44 @@
+from __future__ import annotations
+
+from typing import Sequence
+
+from cognite.client._api_client import APIClient
+from cognite.client._constants import LIST_LIMIT_DEFAULT
+from cognite.client.data_classes.user_profiles import UserProfile, UserProfileList
+from cognite.client.utils._identifier import UserIdentifierSequence
+
+
+class UserProfilesAPI(APIClient):
+    _RESOURCE_PATH = "/profiles"
+
+    def me(self) -> UserProfile:
+        return UserProfile._load(self._get(self._RESOURCE_PATH + "/me").json())
+
+    def list(self, limit: int | None = LIST_LIMIT_DEFAULT, initial_cursor: str | None = None) -> UserProfileList:
+        return self._list(
+            "GET",
+            list_cls=UserProfileList,
+            resource_cls=UserProfile,
+            limit=limit,
+            initial_cursor=initial_cursor,
+        )
+
+    def retrieve(self, user_identifier: str) -> UserProfile | None:
+        identifier = UserIdentifierSequence.load(user_identifier).as_singleton()
+        return self._retrieve_multiple(
+            list_cls=UserProfileList,
+            resource_cls=UserProfile,
+            identifiers=identifier,
+        )
+
+    def retrieve_multiple(self, user_identifiers: Sequence[str]) -> UserProfileList:
+        identifiers = UserIdentifierSequence.load(user_identifiers)
+        return self._retrieve_multiple(
+            list_cls=UserProfileList,
+            resource_cls=UserProfile,
+            identifiers=identifiers,
+        )
+
+    def search(self, name: str, limit: int = 25) -> UserProfileList:
+        # prefix on name
+        return
--- a/cognite/client/_api/user_profiles.py
+++ b/cognite/client/_api/user_profiles.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import List, Sequence, cast
+from typing import List, MutableSequence, cast, overload
 
 from cognite.client._api_client import APIClient
 from cognite.client._constants import LIST_LIMIT_DEFAULT
@@ -32,100 +32,53 @@ class UserProfilesAPI(APIClient):
         """
         return UserProfile._load(self._get(self._RESOURCE_PATH + "/me").json())
 
-    def list(self, limit: int | None = LIST_LIMIT_DEFAULT, initial_cursor: str | None = None) -> UserProfileList:
-        """`List user profiles <https://developer.cognite.com/api#tag/User-profiles/paths/~1profiles/get>`_
-
-        List all user profiles in the current CDF project. The results are ordered alphabetically by name.
-
-        Args:
-            limit (int | None): Maximum number of user profiles to return. Defaults to 25. Set to -1, float("inf") or None to return all.
-            initial_cursor (str | None): Start fetching from a specific cursor.
-
-        Returns:
-            UserProfileList: List of user profiles.
-
-        Examples:
-
-            List all user profiles:
-
-                >>> from cognite.client import CogniteClient
-                >>> c = CogniteClient()
-                >>> res = c.iam.user_profiles.list(limit=None)
-        """
-        return self._list(
-            "GET",
-            list_cls=UserProfileList,
-            resource_cls=UserProfile,
-            limit=limit,
-            initial_cursor=initial_cursor,
-        )
-
+    @overload
     def retrieve(self, user_identifier: str) -> UserProfile | None:
-        """`Retrieve a single user profile by user identifier. <https://developer.cognite.com/api#tag/User-profiles/paths/~1profiles~1byids/post>`_
+        ...
 
-        Retrieves one user profile as indexed by the user identifier in the same CDF project.
+    @overload  # Note, can't use Sequence[str], as str itself matches this requirement...
+    def retrieve(self, user_identifier: MutableSequence[str] | tuple[str, ...]) -> UserProfileList:
+        ...
 
-        Args:
-            user_identifier (str): The user identifier of the user profile to retrieve.
-
-        Returns:
-            UserProfile | None: The requested user profile or None if it doesn't exist.
-
-        Examples:
-
-            Get user profile by user identifier:
-
-                >>> from cognite.client import CogniteClient
-                >>> c = CogniteClient()
-                >>> res = c.iam.user_profiles.retrieve("foo")
-        """
-        identifier = UserIdentifierSequence.load(user_identifier).as_singleton()
-        return self._retrieve_multiple(
-            list_cls=UserProfileList,
-            resource_cls=UserProfile,
-            identifiers=identifier,
-        )
-
-    def retrieve_multiple(self, user_identifiers: Sequence[str]) -> UserProfileList:
-        """`Retrieve multiple user profiles by user identifier. <https://developer.cognite.com/api#tag/User-profiles/paths/~1profiles~1byids/post>`_
+    def retrieve(
+        self, user_identifier: str | MutableSequence[str] | tuple[str, ...]
+    ) -> UserProfile | UserProfileList | None:
+        """`Retrieve user profiles by user identifier. <https://developer.cognite.com/api#tag/User-profiles/paths/~1profiles~1byids/post>`_
 
         Retrieves one or more user profiles indexed by the user identifier in the same CDF project.
 
         Args:
-            user_identifiers (Sequence[str]): The list of user identifiers to retrieve.
+            user_identifier (str | MutableSequence[str] | tuple[str, ...]): The single user identifier (or sequence of) to retrieve profile(s) for.
 
         Returns:
-            UserProfileList: The requested user profiles.
+            UserProfile | UserProfileList | None: UserProfileList if a sequence of user identifier were requested, else UserProfile. If a single user identifier is requested and it is not found, None is returned.
 
         Raises:
-            CogniteNotFoundError: One or more user identifiers does not exist.
+            CogniteNotFoundError: A sequences of user identifiers were requested, but one or more does not exist.
 
         Examples:
 
-            Get user profiles by user identifier:
+            Get a single user profile:
 
                 >>> from cognite.client import CogniteClient
                 >>> c = CogniteClient()
-                >>> res = c.iam.user_profiles.retrieve_multiple(["foo", "bar"])
-        """
-        if isinstance(user_identifiers, str):
-            user_identifiers = [user_identifiers]
+                >>> res = c.iam.user_profiles.retrieve("foo")
 
-        identifiers = UserIdentifierSequence.load(user_identifiers)
-        unsorted_profiles = cast(
-            UserProfileList,
-            self._retrieve_multiple(
-                list_cls=UserProfileList,
-                resource_cls=UserProfile,
-                identifiers=identifiers,
-            ),
+            Get multiple user profiles:
+
+                >>> res = c.iam.user_profiles.retrieve(["bar", "baz"])
+        """
+        identifiers = UserIdentifierSequence.load(user_identifier)
+        profiles = self._retrieve_multiple(
+            list_cls=UserProfileList,
+            resource_cls=UserProfile,
+            identifiers=identifiers,
         )
+        if identifiers.is_singleton():
+            return profiles
         # TODO: The API does not guarantee any ordering (against style guidelines, no timeline for fix)
         #       so we sort manually for now:
-        return UserProfileList(
-            cast(List[UserProfile], [unsorted_profiles.get(user) for user in user_identifiers]),
-            cognite_client=unsorted_profiles._cognite_client,
-        )
+        return UserProfileList(cast(List[UserProfile], [profiles.get(user) for user in user_identifier]))
 
     def search(self, name: str, limit: int = 100) -> UserProfileList:
         """`Search for user profiles <https://developer.cognite.com/api#tag/User-profiles/paths/~1profiles~1search/post>`_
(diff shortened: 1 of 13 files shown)


tests/tests_integration/test_api/test_transformations/test_schedules.py
```diff
@@ -72,7 +72,7 @@ def other_schedule(cognite_client, other_transformation):
 
 
 @pytest.mark.skipif(
-    os.environ.get("LOGIN_FLOW") != "client_credentials", reason="This test requires client_credentials auth"
+    os.getenv("LOGIN_FLOW") != "client_credentials", reason="This test requires client_credentials auth"
```
doctrino, line 75: 🙏I have had this on my list for a long time. Thank you for adding this one.

cognite/client/data_classes/user_profiles.py
```diff
@@ -0,0 +1,67 @@
...
+            user. Use the user_identifier property instead.
+        display_name (str | None): The display name for the user.
+        job_title (str | None): The user's job title.
+        last_updated_time (int | None): The number of milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds.
+        cognite_client (CogniteClient | None): No description.
+    """
+
+    def __init__(
+        self,
+        user_identifier: str | None = None,
```
doctrino, line 30: Looking at the docs, this is required so suggest making it required here as well. 
```suggestion
        user_identifier: str,
```

cognite/client/data_classes/user_profiles.py
```diff
@@ -0,0 +1,67 @@
...
+        given_name: str | None = None,
+        surname: str | None = None,
+        email: str | None = None,
+        display_name: str | None = None,
+        job_title: str | None = None,
+        last_updated_time: int | None = None,
```
doctrino, line 36: Same for this one.
```suggestion
        last_updated_time: int,
```

cognite/client/data_classes/user_profiles.py
```diff
@@ -0,0 +1,67 @@
...
+        self.given_name = given_name
+        self.surname = surname
+        self.email = email
+        self.display_name = display_name
+        self.job_title = job_title
+        self.last_updated_time = last_updated_time
+        self._cognite_client = cast("CogniteClient", cognite_client)
+
+
+class UserProfileList(CogniteResourceList[UserProfile]):
```
doctrino, line 49: Should we split the `CogniteResourceList `into a `CogniteResourceListCore `and `CogniteResourceList`? My thinking is that then `UserProfileList` could inherit from the `Core` version that should be independent of `external_id` and `id`. I think the data modeling classes would similarly benefit from using a version of `CogniteResourceListCore` without `external_id` and `id`.

cognite/client/utils/_identifier.py
```diff
@@ -226,3 +239,31 @@ class SingletonIdentifierSequence(IdentifierSequenceCore[Identifier]):
...
+        elif isinstance(user_identifiers, Sequence):
+            return cls(identifiers=list(map(UserIdentifier, map(str, user_identifiers))), is_singleton=False)
+
+        raise TypeError(f"user_identifiers must be of type str or Sequence[str]. Found {type(user_identifiers)}")
+
+    def chunked(self, chunk_size: int) -> list[UserIdentifierSequence]:  # type: ignore [override]
+        return [
+            UserIdentifierSequence(chunk, is_singleton=self.is_singleton())
+            for chunk in split_into_chunks(self._identifiers, chunk_size)
+        ]
```
doctrino, line 259: If we change the parent implementation to 
```python
    def chunked(self, chunk_size: int) -> list[Self]:
        return [
            type(self)(chunk, is_singleton=self.is_singleton())
            for chunk in split_into_chunks(self._identifiers, chunk_size)
        ]
```
I don't think this will be necessary.

cognite/client/utils/_identifier.py
```diff
@@ -226,3 +239,31 @@ class SingletonIdentifierSequence(IdentifierSequenceCore[Identifier]):
...
+
+    def assert_singleton(self) -> None:
+        if not self.is_singleton():
+            raise ValueError("Exactly one user identifier (string) must be specified")
+
+    def as_dicts(self) -> list[dict[str, str]]:  # type: ignore [override]
+        return [identifier.as_dict() for identifier in self._identifiers]
+
+    def as_primitives(self) -> list[str]:  # type: ignore [override]
+        return [identifier.as_primitive() for identifier in self._identifiers]
```
doctrino, line 269: If we introduce a new type in the core like this:
```python
T_Identifier_Type = TypeVar("T_Identifier_Type")

class IdentifierSequenceCore(Generic[T_Identifier, T_Identifier_Type], ABC):
```
and change the implementation in the parent class to 
```python
    def as_dicts(self) -> list[dict[str, T_Identifier_Type]]:
        return [identifier.as_dict() for identifier in self._identifiers]

    def as_primitives(self) -> list[T_Identifier_Type]:
        return [identifier.as_primitive() for identifier in self._identifiers]
```
Then, these methods do not need to be reimplemented. 
```suggestion
```

cognite/client/utils/_identifier.py
```diff
@@ -226,3 +239,31 @@ class SingletonIdentifierSequence(IdentifierSequenceCore[Identifier]):
...
+
+    def chunked(self, chunk_size: int) -> list[UserIdentifierSequence]:  # type: ignore [override]
+        return [
+            UserIdentifierSequence(chunk, is_singleton=self.is_singleton())
+            for chunk in split_into_chunks(self._identifiers, chunk_size)
+        ]
+
+    def assert_singleton(self) -> None:
+        if not self.is_singleton():
+            raise ValueError("Exactly one user identifier (string) must be specified")
```
doctrino, line 263: If the parent class take in `identifier name` which cold be used in this error message, then that would improve all the other child classes as well, and avoid this reimplementation.

cognite/client/_api/user_profiles.py
```diff
@@ -0,0 +1,154 @@
...
+
+            Get your own user profile:
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> res = c.iam.user_profiles.me()
+        """
+        return UserProfile._load(self._get(self._RESOURCE_PATH + "/me").json())
+
+    def list(self, limit: int | None = LIST_LIMIT_DEFAULT, initial_cursor: str | None = None) -> UserProfileList:
```
doctrino, line 35: I do not think we should expose the cursor here. I think that is a detail that should be hidden by the SDK. This is what we have done for all other listing methods I am familiar with. 
```suggestion
    def list(self, limit: int | None = LIST_LIMIT_DEFAULT) -> UserProfileList:
```

cognite/client/_api/user_profiles.py
```diff
@@ -0,0 +1,154 @@
...
+                >>> res = c.iam.user_profiles.retrieve("foo")
+        """
+        identifier = UserIdentifierSequence.load(user_identifier).as_singleton()
+        return self._retrieve_multiple(
+            list_cls=UserProfileList,
+            resource_cls=UserProfile,
+            identifiers=identifier,
+        )
+
+    def retrieve_multiple(self, user_identifiers: Sequence[str]) -> UserProfileList:
```
doctrino, line 89: Suggest using a single method with overloads instead of the split. This is what we have done for all newer implementations in the SDK. 
```python
retrieve(self, user_identifier: str | Sequence[str]) -> UserProfile | UserProfileList | None:
```

cognite/client/_api/user_profiles.py
```diff
@@ -0,0 +1,154 @@
...
+            ),
+        )
+        # TODO: The API does not guarantee any ordering (against style guidelines, no timeline for fix)
+        #       so we sort manually for now:
+        return UserProfileList(
+            cast(List[UserProfile], [unsorted_profiles.get(user) for user in user_identifiers]),
+            cognite_client=unsorted_profiles._cognite_client,
+        )
+
+    def search(self, name: str, limit: int = 100) -> UserProfileList:
```
doctrino, line 130: Why do you change the default limit? Suggest keeping it in line with all other listing type endpoints. 
```suggestion
    def search(self, name: str, limit: int = LIST_LIMIT_DEFAULT) -> UserProfileList:
```

doctrino, line 130: I see the API calls this name, however, that seems misleading to me. What about calling it `name_prefix`?
```suggestion
    def search(self, name_prefix: str, limit: int = 100) -> UserProfileList:
```

tests/tests_integration/test_api/test_user_profiles.py
```diff
@@ -0,0 +1,53 @@
...
+    profile = cognite_client.iam.user_profiles.list(limit=1)[0]
+    assert isinstance(profile.user_identifier, str)
+
+    profile_retrieve = cognite_client.iam.user_profiles.retrieve(profile.user_identifier)
+    assert profile_retrieve is not None
+    assert profile == profile_retrieve
+
+
+def test_user_profiles_api__retrieve_not_exist(cognite_client):
+    assert cognite_client.iam.user_profiles.retrieve(random_string(10)) is None
```
doctrino, line 37: Just because I am not a big fan of randomness in tests, and to make it even more explicit. 
```suggestion
    assert cognite_client.iam.user_profiles.retrieve("Not existing user profile") is None
```

tests/tests_integration/test_api/test_user_profiles.py
```diff
@@ -0,0 +1,53 @@
...
+
+    assert isinstance(profiles, UserProfileList)
+    assert isinstance(profiles_retrieve, UserProfileList)
+    assert profiles == profiles_retrieve
+
+
+def test_user_profiles_api__retrieve_multiple_not_exist(cognite_client):
+    # Endpoint does not support 'ignore unknown ids'
+    with pytest.raises(CogniteNotFoundError) as err:
+        cognite_client.iam.user_profiles.retrieve_multiple([user_id := random_string(10)])
```
doctrino, line 52: What about testing with a combination of unknown and known user profiles?

tests/tests_integration/test_api/test_user_profiles.py
```diff
@@ -0,0 +1,53 @@
...
+    assert cognite_client.iam.user_profiles.retrieve(random_string(10)) is None
+
+
+def test_user_profiles_api__retrieve_multiple(cognite_client):
+    profiles = cognite_client.iam.user_profiles.list(limit=5)
+    profiles_retrieve = cognite_client.iam.user_profiles.retrieve_multiple([p.user_identifier for p in profiles])
+
+    assert isinstance(profiles, UserProfileList)
+    assert isinstance(profiles_retrieve, UserProfileList)
+    assert profiles == profiles_retrieve
```
doctrino, line 46: Could benefit from arrange, act, assert comments 🧌

doctrino (commented):
🤯 I was not aware that we had this API, very cool. Great initiative for adding SDK support for it.

cognite/client/data_classes/user_profiles.py
haakonvt: Can't unfortunately, as it breaks how we load data into CogniteResource classes

haakonvt: I.e. this monstrosity which I'm about to remove in an upcoming PR:
```py
elif isinstance(resource, dict):
    instance = cls(cognite_client=cognite_client)  # <--
    for key, value in resource.items():
        snake_case_key = to_snake_case(key)
        if hasattr(instance, snake_case_key):
            setattr(instance, snake_case_key, value)
    return instance
```

cognite/client/data_classes/user_profiles.py
haakonvt: Can't for the same reason as above

cognite/client/data_classes/user_profiles.py
haakonvt: Agree with the incentive --> already its kind of broken for example DatabaseList, TableList which use `name`. These could then also make good use of a separate `.get` which used name instead 👌

cognite/client/utils/_identifier.py
haakonvt: This was my first version actually, but our (old) version of `mypy` doesn't work with `Self` (it seems):
```
...error: Variable "typing_extensions.Self" is not valid as a type  [valid-type]
        def chunked(self, chunk_size: int) -> list[Self]:
```

cognite/client/utils/_identifier.py
haakonvt: Will try this one out (tomorrow) 🚀

cognite/client/_api/user_profiles.py
haakonvt: I was very in doubt on this one 😅 But when I saw that `_list` already had support for `initial_cursor` I thought "why not" 🤷‍♂️

cognite/client/_api/user_profiles.py
haakonvt: Sure, I never liked typing `_multiple` anyway 😂

cognite/client/_api/user_profiles.py
haakonvt: See other PR

haakonvt: Considered it, but the API logic might change, so better keep the info tight imo. The link to the docs is always there.

tests/tests_integration/test_api/test_user_profiles.py
haakonvt: I guess you hate the `hypothesis` package then 😉

tests/tests_integration/test_api/test_user_profiles.py
haakonvt: Yes. Very yes

tests/tests_integration/test_api/test_user_profiles.py
haakonvt: Haha, you wish :trollface:

cognite/client/_api/user_profiles.py
haakonvt: Removed `initial_cursor `

cognite/client/data_classes/user_profiles.py
haakonvt: ...but I think this should be a separate PR, so holding off on making this change here.

cognite/client/utils/_identifier.py
haakonvt: Made `mypy` happy by changing to:
```py
    def chunked(self: T_IdentifierSequenceCore, chunk_size: int) -> list[T_IdentifierSequenceCore]:
        (...)

T_IdentifierSequenceCore = TypeVar("T_IdentifierSequenceCore", bound=IdentifierSequenceCore)

File: cognite/client/utils/_identifier.py (hunk 6 above)
Comment (haakonvt) [reply cognite/client/utils/_identifier.py:269]: Raises an error from typing extensions: 🤔 
```
cognite/client/data_classes/_base.py:30: in <module>
    from cognite.client.utils._identifier import IdentifierSequence
cognite/client/utils/_identifier.py:185: in <module>
    class IdentifierSequence(IdentifierSequenceCore[Identifier]):
../../.pyenv/versions/3.8.16/lib/python3.8/typing.py:261: in inner
    return func(*args, **kwds)
../../.pyenv/versions/3.8.16/lib/python3.8/typing.py:897: in __class_getitem__
    _check_generic(cls, params)
.venv/lib/python3.8/site-packages/typing_extensions.py:164: in _check_generic
    raise TypeError(f"Too {'many' if alen > elen else 'few'} parameters for {cls};"
E   TypeError: Too few parameters for <class 'cognite.client.utils._identifier.IdentifierSequenceCore'>; actual 1, expected 2
```

Comment (haakonvt) [reply cognite/client/utils/_identifier.py:269]: When I think about it, this doesnt make that much sense; we want it to a Generic of IdentifierCore (our protocol with as_dict/as_primitive), and adding this unbounded, would mean we could make it a generic of "whatever".

File: cognite/client/utils/_identifier.py (hunk 7 above)
Comment (haakonvt) [reply cognite/client/utils/_identifier.py:263]: How would you write the error message to be generic?
```py
raise ValueError("Exactly one of id or external id must be specified")
raise ValueError("Exactly one user identifier (string) must be specified")
```

cognite/client/utils/_identifier.py
haakonvt: I still removed these method overrides - at the cost of `mypy` wrongly inferring their return types to also include `int`.

cognite/client/utils/_identifier.py
doctrino: I guess it would be the entire message, but it is more explicit to just overwrite.

doctrino (approved):
This looks good to me now :)

PR #1337, opened 2023-08-28
Make Iterator[foo] default return annot (again) for generators

haakonvt:
## Description
With the release of `pydoclint` v0.3.0, Iterators are back in the game.


erlendvollset (approved):
LGTM

PR #1340, opened 2023-08-30
[DEGR-2757] Rewrite function validation using AST parsing. 

qTipTip:
## Description

In response to [an issue](https://cognitedata.atlassian.net/browse/DEGR-2757) pertaining to how we validate function code before deploying, I've rewritten the code to use AST parsing for basic validation instead of relying on `importlib`.

We also now do the full `importlib` validation, but now in a subprocess which should isolate it from everything else going on.
Thus, one function validation should not interfere with another function validation. 

For users running the SDK through Pyodide, where `subprocess / multiprocess` is not available, we fall back to the old importlib method, and clean up loaded modules afterwords as best we can. 
 
## Checklist:
- [x] Tests added/updated. These should capture the desired output. 


cognite/client/_api/functions.py
```diff
@@ -525,6 +523,29 @@ def convert_file_path_to_module_path(file_path: str) -> str:
     return ".".join(Path(file_path).with_suffix("").parts)
 
 
+def _get_handle_function_node(handler_path: Path) -> ast.FunctionDef | None:
```
qTipTip, line 526: Is the `|` typing okay here? I am developing on Python 3.10.6. Ruff auto-corrected it to `|` from `Optional`.

haakonvt:
Have you considered/tested `importlib.util.find_spec`?
https://docs.python.org/3/library/importlib.html#importlib.util.find_spec

cognite/client/_api/functions.py
```diff
@@ -521,44 +525,94 @@ def _create_session_and_return_nonce(
...
-    cached_handler_module = sys.modules.get("handler")
-    if cached_handler_module:
-        del sys.modules["handler"]
+    _validate_function_from_ast(handle_function_node)
 
-    module_path = convert_file_path_to_module_path(function_path)
-    handler = importlib.import_module(module_path)
+    # Opt-in import checks
+    if check_imports:
+        with _import_lock:
```
qTipTip, line 587: @haakonvt 

Is this along the lines of what you envisioned? Not used thread locks before.

tests/tests_unit/test_api/test_functions.py
```diff
@@ -405,13 +405,27 @@ class TestFunctionsAPI:
...
+    )
+    def test_validate_folder(self, function_folder, function_path, exception):
+        folder = os.path.join(os.path.dirname(__file__), "function_test_resources", function_folder)
+        if exception is None:
+            validate_function_folder(folder, function_path)
+        else:
+            with pytest.raises(exception):
+                validate_function_folder(folder, function_path)
+
+    @pytest.mark.skip(reason="We no longer import the function code, so unable to verify imports")
```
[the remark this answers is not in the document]
qTipTip: Question no longer relevant.

cognite/client/_api/functions.py
qTipTip: Removed threadlock.

qTipTip:
> Have you considered/tested/evaluated `importlib.util.find_spec`? https://docs.python.org/3/library/importlib.html#importlib.util.find_spec
> 
> Edit: Took the discussion to slack

@haakonvt 

I decided on using a separate process for the import-check. That should isolate it from whatever else might be going on. I tried monkey patching out the `sys.modules` with a new dictionary, but it had too many sideeffects.

haakonvt (approved):
Nice! 💯

PR #1341, opened 2023-08-30
Fix `FilesAPI.upload` for pyodide users

haakonvt:
## Description
- Uploading files using `client.files.upload` now works when running with `pyodide`. Previously, the file handle would just be cast to string and uploaded.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/files.py
+++ b/cognite/client/_api/files.py
@@ -17,7 +17,7 @@ from typing import (
 
 from cognite.client import utils
 from cognite.client._api_client import APIClient
-from cognite.client._constants import LIST_LIMIT_DEFAULT
+from cognite.client._constants import _RUNNING_IN_BROWSER, LIST_LIMIT_DEFAULT
 from cognite.client.data_classes import (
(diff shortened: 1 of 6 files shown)


cognite/client/_api/files.py
```diff
@@ -484,7 +485,13 @@ def upload(
         raise ValueError(f"The path '{path}' does not exist")
 
     def _upload_file_from_path(self, file: FileMetadata, file_path: str, overwrite: bool) -> FileMetadata:
+        from cognite.client import _RUNNING_IN_BROWSER
```
erlendvollset, line 488: I'd suggest placing this constant somewhere else than `cognite/client/__init__.py` so you don't need to use local imports every time you need to use it.

erlendvollset (approved):
Approved, but consider moving the constant!

cognite/client/_api/files.py
haakonvt: Yes, very yes 😅

PR #1331, opened 2023-08-27
Various SDK tweaks

haakonvt:
## Description
    * Make `_RETRYABLE_POST_ENDPOINT_REGEX_PATTERNS` into a `frozenset` 
    * simplify `EntityMatchingAPI.list` return logic
    * direct type comparison now done using identity

PR #1348, opened 2023-09-03
Speed up basic API response translation (`_load`)

haakonvt:
### New faster `_load` implementation

1. Removes the old `if hasattr -> setattr` implementation that called `to_snake_case` `O(N*M)` times (N dicts, M elements per dict).
2. Changed to using the newly added `fast_dict_load` function, which uses `get_accepted_params` that caches the accepted parameters for each class it receives, returning a dict that maps `camelCase` to `snake_case` for very efficient API response translation into our data classes.

### This change gives a nice 45 % improvement in loading/conversion time:

image

----

### Bugfix
The old version accepted both snake_case and camelCase, leading to potential key collisions (without raising an error):
```py
Asset._load({"parent_id": 1, "parentId": 2})  # which one is loaded?
```

Split from #1151 .

➡️  Depends on test fixes in #1331 .

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/transformations/__init__.py
+++ b/cognite/client/data_classes/transformations/__init__.py
@@ -57,14 +57,11 @@ class SessionDetails:
 
     @classmethod
     def _load(cls, resource: dict[str, Any]) -> SessionDetails:
-        to_load = {}
-        if "sessionId" in resource:
-            to_load["session_id"] = resource["sessionId"]
-        if "clientId" in resource:
-            to_load["client_id"] = resource["clientId"]
-        if "projectName" in resource:
-            to_load["project_name"] = resource["projectName"]
-        return cls(**to_load)
+        return cls(
+            session_id=resource.get("sessionId"),
+            client_id=resource.get("clientId"),
+            project_name=resource.get("projectName"),
+        )
 
     def dump(self, camel_case: bool = False) -> dict[str, Any]:
         """Dump the instance into a json serializable Python data type.
--- a/cognite/client/data_classes/transformations/__init__.py
+++ b/cognite/client/data_classes/transformations/__init__.py
@@ -27,7 +27,7 @@ from cognite.client.data_classes.transformations.jobs import TransformationJob, 
 from cognite.client.data_classes.transformations.schedules import TransformationSchedule
 from cognite.client.data_classes.transformations.schema import TransformationSchemaColumnList
 from cognite.client.exceptions import CogniteAPIError
-from cognite.client.utils._text import convert_all_keys_to_camel_case, convert_all_keys_to_snake_case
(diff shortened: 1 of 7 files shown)


tests/tests_unit/test_base.py
```diff
@@ -139,8 +139,8 @@ def test_dump_camel_case(self):
 
     def test_load(self):
         assert MyResource(1).dump() == MyResource._load({"varA": 1}).dump()
-        assert MyResource(1, 2).dump() == MyResource._load({"var_a": 1, "var_b": 2}).dump()
-        assert {"var_a": 1} == MyResource._load({"var_a": 1, "var_c": 1}).dump()
```
doctrino, line 143: bug in the test? Strange that this passed at all.

cognite/client/data_classes/transformations/__init__.py
```diff
@@ -49,6 +49,17 @@ def __init__(
...
+    @classmethod
+    def _load(cls, resource: dict[str, Any]) -> SessionDetails:
+        to_load = {}
+        if "sessionId" in resource:
+            to_load["session_id"] = resource["sessionId"]
+        if "clientId" in resource:
+            to_load["client_id"] = resource["clientId"]
+        if "projectName" in resource:
+            to_load["project_name"] = resource["projectName"]
+        return cls(**to_load)
```
doctrino, line 61: Why not just?
```suggestion
        return cls(
            session_id=resource.get("sessionId"),
            client_id=resource.get("clientId"),
            project_name=resource.get("projectName"),
        )
```

doctrino (commented):
Maybe worth a bump? It it is an improvement in speed.

tests/tests_unit/test_base.py
haakonvt: The old code would try every single key, while mine only grabs the specific camelCased keys we expect to load, hence the old also accepted snake_case (which was a serious bug, passing `asset_id=1` and `assetId=2` would not raise an error but give varying results depending on how it was serialized/deserialized).

cognite/client/data_classes/transformations/__init__.py
erlendvollset: Why would we want to use default values for the parameters when loading the response from the API?

erlendvollset: All keys should be present in responses.

erlendvollset: Also thinking that none of these attributes should actually be nullable.

haakonvt: Aight, will change. None were listed as required in the official API docs hence the checks

erlendvollset (approved):
LGTM!

PR #1349, opened 2023-09-04
[DEGR-2279] Remove runtime py37 from Functions docs

HMEiding:
## Description
- Remove runtime `"py37"`.
- Was announced on [Cognite Hub](https://hub.cognite.com/product-updates/deprecation-of-python-3-7-on-cognite-functions-1797) three months ago.


cognite/client/data_classes/functions.py
```diff
@@ -36,7 +36,7 @@ class Function(CogniteResource):
         env_vars (dict | None): User specified environment variables on the function ((key, value) pairs).
         cpu (Number | None): Number of CPU cores per function. Defaults to 0.25. Allowed values are in the range [0.1, 0.6].
         memory (Number | None): Memory per function measured in GB. Defaults to 1. Allowed values are in the range [0.1, 2.5].
-        runtime (str | None): Runtime of the function. Allowed values are ["py37", "py38", "py39"]. The runtime "py38" resolves to the latest version of the Python 3.8 series. Will default to "py38" if not specified.
+        runtime (str | None): Runtime of the function. Allowed values are ["py38", "py39","py310"]. The runtime "py38" resolves to the latest version of the Python 3.8 series. Will default to "py38" if not specified.
```
haakonvt, line 39: "Check the latest documentation for an up-to-date list on supported runtimes"? 🤷

haakonvt (approved):
👍

cognite/client/data_classes/functions.py
HMEiding: Good suggestion, thanks! 👍

HMEiding: That is a good suggestion. One downside though, is that if we don't explicitly list the options here, the user will not know the format of the variable (is it `"3.8"`, `"py3.8"`, `"py38"` etc.). So I actually prefer stating the options here.

PR #1354, opened 2023-09-05
Various additions

haakonvt:
## Description

1. Make `ALLOWED_HANDLE_ARGS` into a module-level constant
2. Add `pyodide` fix missing `multiprocessing` implementation with `_run_import_check_backup`
3. Bugfix `_validate_function_handle` that allowed `def handle(*, foo)` to pass (now uses `inspect.signature`)
4. Simplify errors raised from `check_imports`
5. Remove `validate_function_args`, now `_validate_function_handle` handles both cases

### DIFF
--- a/cognite/client/_api/functions.py
+++ b/cognite/client/_api/functions.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
 import ast
+import importlib
 import os
 import re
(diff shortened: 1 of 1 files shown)


qTipTip (approved):
Looks good!

PR #1355, opened 2023-09-06
Modify docstring formatter be lenient with "short" docstrings

haakonvt:
## Description

- Don't format short docstrings
- skip docfmt on non-stringified annots
- revert a few docstrings to short format


erlendvollset (approved):
Nice! Thank you!

PR #1356, opened 2023-09-06
Add some more ergonomics to instance syncing along with sqlite example

erlendvollset:
- Add backoff utilitiy
- Add callback-based method for subscribing to a given dms query
- Add docs along with example on how to sync to local sqlite

### DIFF
--- a/cognite/client/_api/data_modeling/instances.py
+++ b/cognite/client/_api/data_modeling/instances.py
@@ -1,8 +1,24 @@
 from __future__ import annotations
 
 import json
+import logging
+import random
+import time
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, Iterator, List, Literal, Sequence, Union, cast, overload
+from datetime import datetime, timezone
+from threading import Thread
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Iterator,
+    List,
+    Literal,
+    Sequence,
+    Union,
+    cast,
+    overload,
+)
 
 from cognite.client._api_client import APIClient
 from cognite.client._constants import DEFAULT_LIMIT_READ
@@ -37,6 +53,7 @@ from cognite.client.data_classes.data_modeling.instances import (
     NodeApplyResult,
     NodeApplyResultList,
     NodeList,
+    SubscriptionContext,
 )
 from cognite.client.data_classes.data_modeling.query import (
     Query,
@@ -45,6 +62,8 @@ from cognite.client.data_classes.data_modeling.query import (
 from cognite.client.data_classes.data_modeling.views import View
 from cognite.client.data_classes.filters import Filter, _validate_filter
 from cognite.client.utils._identifier import DataModelingIdentifierSequence
+from cognite.client.utils._retry import Backoff
+from cognite.client.utils._text import random_string
 
 from ._data_modeling_executor import get_data_modeling_executor
 
@@ -69,6 +88,8 @@ _DATA_MODELING_SUPPORTED_FILTERS: frozenset[type[Filter]] = frozenset(
         filters.Overlaps,
     }
 )
+
+_LOGGER = logging.getLogger(__name__)
 
 
 class _NodeOrEdgeList(CogniteResourceList):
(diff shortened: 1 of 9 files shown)


docs/source/data_modeling.rst
```diff
@@ -126,6 +126,96 @@ Query instances
...
+                    (node.space, node.external_id)
+                    for node in result["nodes"]
+                    if node.deleted_time is not None
+                ]
+                # Updates must be done in the same transaction as persisting the cursor.
+                # A transaction is implicitly started by sqlite here.
+                connection.executemany(
+                    "INSERT INTO instance VALUES (?, ?, ?)", inserts
+                )
+                connection.executemany(
```
alexbrasetvik, line 202: Delete must happen before insert. The same `space, external_id` may appear as several tombstones and then a new instance, which must result in the instance being saved.

erlendvollset: Ah right. TY 🙏

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -809,3 +813,11 @@ class InstancesDeleteResult:
 
     nodes: list[NodeId]
     edges: list[EdgeId]
+
+
+@dataclass
+class SubscriptionContext:
+    _canceled: bool = False
```
erlendvollset, line 820: Note to self: add some heartbeat timestamps like last_successful_sync and last_successful_callback, to make it easy to check for progress.

cognite/client/_api/data_modeling/instances.py
```diff
@@ -394,6 +414,80 @@ def delete(
...
+                    if any(len(instances) > 0 for instances in result.data.values()):
+                        callback(result)
+                        subscription_context.last_successful_callback = datetime.now(tz=timezone.utc)
+                        # only progress the cursor if the callback executed successfully
+                        cursors = result.cursors
+                        backoff.reset()
+                        continue
+                    else:
+                        # No data was returned, so we back off before we try again.
+                        time.sleep(next(backoff))
```
alexbrasetvik, line 480: I think polling delay/interval and error backoff should be separate concepts.

If there's a poll interval, I'd delay the first one with `random.uniform(0, poll_interval)` to smear out a bit, but let the subsequent ones go with a consistent delay.

If `self.sync` throws, we shouldn't need to re-query before the next attempt.

docs/source/data_modeling.rst
```diff
@@ -126,6 +126,102 @@ Query instances
...
+    def sync_space_to_sqlite(
+        db_name: str, space_to_sync: str
+    ) -> SubscriptionContext:
+        with sqlite_connection(db_name) as connection:
+            connection.execute(
+                "CREATE TABLE IF NOT EXISTS instance (space TEXT, external_id TEXT, data JSON)"
+            )
+            connection.execute(
+                "CREATE UNIQUE INDEX IF NOT EXISTS space_extid_idx ON instance (space, external_id)"
+            )
```
alexbrasetvik, line 167:
```suggestion
            connection.execute(
                "CREATE TABLE IF NOT EXISTS instance (space TEXT, external_id TEXT, data JSON, PRIMARY KEY(space, external_id))"
            )
```

docs/source/data_modeling.rst
```diff
@@ -126,6 +126,102 @@ Query instances
...
+
+
+    def sqlite_connection(db_name: str) -> sqlite3.Connection:
+        return sqlite3.connect(db_name, check_same_thread=False)
+
+
+    def sync_space_to_sqlite(
+        db_name: str, space_to_sync: str
+    ) -> SubscriptionContext:
+        with sqlite_connection(db_name) as connection:
```
alexbrasetvik, line 161: Maybe split up into clearer setup vs continuous sync functions?

alexbrasetvik (changes requested):
Looks generally good. I think the error vs polling delays should be separate.

The SQLite example can split up setup vs continuous sync a bit, so we can more clearly excessively comment the latter.

cognite/client/_api/data_modeling/instances.py
alexbrasetvik: Sorry, I meant if `callback` throws.

arnels (approved):
LGTM

alexbrasetvik (approved):
LGTM.

Let's follow up with future PRs that make the SDK logic not log full traces for intermittent errors like connectivity blips, 429s, etc., but this looks like a good start.

PR #1363, opened 2023-09-11
Upgrade `mypy` to ^1

haakonvt:
## Description
See title.

### Why?
I want to use `typing_extensions.Self` 🤓


cognite/client/data_classes/data_modeling/data_types.py
```diff
@@ -35,7 +34,8 @@ def as_tuple(self) -> tuple[str, str]:
         return self.space, self.external_id
 
 
-class PropertyType(ABC):
+@dataclass
+class PropertyType:
```
erlendvollset, line 38: Why not abstract anymore?

haakonvt: It lacks any abstract methods, so it can be instantiated. Also, `mypy` complained about `output = asdict(self)` (I'll see if I can make this work without a type ignore comment)

haakonvt:
```
cognite/client/data_classes/data_modeling/data_types.py:43: error: No overload
variant of "asdict" matches argument type "PropertyType"  [call-overload]
            output = asdict(self)
```

haakonvt: The issue magically vanished now 🥷

cognite/client/data_classes/aggregations.py
```diff
@@ -213,10 +213,6 @@ class AggregationFilter(ABC):
     def dump(self) -> dict[str, Any]:
         return {self._filter_name: self._filter_body()}
 
-    @classmethod
-    def load(cls, filter_: dict[str, Any]) -> AggregationFilter:
```
doctrino, line 217: Why removing this?

haakonvt: It is not implemented in any subclasses (and thus invalid for subclasses using `@final`).

```
cognite/client/data_classes/aggregations.py:217: error: Missing return
statement  [empty-body]
        def load(cls, filter_: dict[str, Any]) -> AggregationFilter:
        ^
cognite/client/data_classes/aggregations.py:217: note: If the method is meant to be abstract, use @abc.abstractmethod
```
****
#### Also:

image

doctrino (approved):

cognite/client/data_classes/aggregations.py
doctrino: Makes sense

PR #1370, opened 2023-09-14
Resolve duplicate file names

CogniteJohan:
## Description
Added parameter `resolve_duplicate_file_names` to `client.files.download`. 
This will keep all the files when downloading to local machine, even if they have the same name.

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/files.py
```diff
@@ -612,19 +612,55 @@ def retrieve_download_urls(
...
+            if file_name not in file_names_unique:
+                file_names_unique.append(file_name)
+                continue
+
+            file_base = file_name
+            file_ext = ""
+            if "." in file_name:
+                file_name_split = file_name.split(".")
+                file_base = ".".join(file_name_split[:-1])
+                file_ext = f".{file_name_split[-1]}"
```
haakonvt, line 634: I think you are better off using `pathlib` for this with suffix (lots of good methods on `Path`) 😄

haakonvt (commented):
Some initial feedback

cognite/client/_api/files.py
```diff
@@ -612,19 +612,55 @@ def retrieve_download_urls(
         results = tasks_summary.joined_results(unwrap_fn=lambda res: res.json()["items"])
         return {result.get("id") or result["externalId"]: result["downloadUrl"] for result in results}
 
+    @staticmethod
+    def _create_unique_file_names(file_names_in: list[str] | list[Path]) -> list[str]:
+        """
+        Create unique file names by appending a number to the base file name.
+        """
```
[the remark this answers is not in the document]
CogniteJohan: Changed the logic

cognite/client/_api/files.py
CogniteJohan: Pathlib it is :)

cognite/client/_api/files.py
```diff
@@ -612,19 +612,52 @@ def retrieve_download_urls(
         results = tasks_summary.joined_results(unwrap_fn=lambda res: res.json()["items"])
         return {result.get("id") or result["externalId"]: result["downloadUrl"] for result in results}
 
+    @staticmethod
+    def _create_unique_file_names(file_names_in: list[str] | list[Path]) -> list[str]:
```
haakonvt, line 616: This produces weird results when there's multiple suffixes:
```
["archive.tar.gz", "archive.tar.gz"]
--> file.tar(1).gz
```

cognite/client/_api/files.py
```diff
@@ -612,19 +612,52 @@ def retrieve_download_urls(
...
+        """
+        Create unique file names by appending a number to the base file name.
+        """
+        file_names: list[str] = [str(file_name) for file_name in file_names_in]
+        file_names_unique_original = set(file_names)
+        file_names_unique_created: list = []
+        file_names_original_count: dict = {}
```
haakonvt, line 623: Although _explicit is better than implicit_, with this much overlap between variable names, the overall readability of the function goes down (in my opinion 😉 ). I think that in a function dealing with file names, it would be fine to use variables like `count`, `original` and `newly_created` or perhaps just `changed`.

cognite/client/_api/files.py
```diff
@@ -612,19 +612,52 @@ def retrieve_download_urls(
...
+
+            file_base = f"{Path(file_name).parent / Path(file_name).stem}"
+            file_ext = Path(file_name).suffix
+
+            new_name = file_name
+            while (new_name in file_names_unique_created) or (new_name in file_names_unique_original):
+                if file_name not in file_names_original_count:
+                    file_names_original_count[file_name] = 1
+
+                new_name = f"{file_base}({file_names_original_count.get(file_name)}){file_ext}"
```
haakonvt, line 638: Use normal getitem (`[...]`) unless you plan for keys to be missing.

cognite/client/_api/files.py
```diff
@@ -748,15 +786,14 @@ def _process_file_download(
...
         identifier: dict[str, int | str],
         id_to_metadata: dict[str | int, FileMetadata],
+        file_path: Path,
     ) -> None:
-        id = IdentifierSequence.unwrap_identifier(identifier)
-        file_metadata = id_to_metadata[id]
-        file_path = (directory / cast(str, file_metadata.name)).resolve()
-        file_is_in_download_directory = directory.resolve() in file_path.parents
+        file_path_absolute = file_path.resolve()
+        file_is_in_download_directory = directory.resolve() in file_path_absolute.parents
```
haakonvt, line 792: Probably not worth the trouble here, but `PurePath.is_relative_to` could maybe have been used.

haakonvt: Nevermind, its a 3.9 feature 😅

tests/tests_unit/test_api/test_files.py
```diff
@@ -525,12 +525,45 @@ def test_download(self, cognite_client, mock_file_download_response):
...
+            "a(2).txt",
+            "a(1).txt",
+        ]
+        assert cognite_client.files._create_unique_file_names(["a.txt", "file", "a(1).txt", "a.txt", "file"]) == [
+            "a.txt",
+            "file",
+            "a(1).txt",
+            "a(2).txt",
+            "file(1)",
+        ]
```
haakonvt, line 541: Additional test cases:
- windows style paths
- paths with some directories first

CogniteJohan:
@haakonvt Windows-folder tests finally work :)

cognite/client/_api/files.py
haakonvt: I unresolved this one as it still gives me weird results. If you disagree, let me know 😉

CogniteJohan: I do actually disagree here, as this will occur very infrequently and when it does the soluton still works ok. If we need to define file extensions differently than the pathlib suffix function, I think the complexity of the codebase becomes unnecessary complex and not yield much value. But if you have a clever solution, I would love to see it :)

haakonvt: Maybe this can be used?
```
>>> Path("archive.tar.gz").suffixes
['.tar', '.gz']
```

CogniteJohan: Good call! I'll be back...

tests/tests_unit/test_api/test_files.py
```diff
@@ -525,12 +526,58 @@ def test_download(self, cognite_client, mock_file_download_response):
             with open(fp2, "rb") as fh:
                 assert b"content2" == fh.read()
 
+    def test_create_unique_file_names(self, cognite_client):
```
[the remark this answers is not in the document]
CogniteJohan: Agreed :)

cognite/client/_api/files.py
```diff
@@ -611,19 +612,50 @@ def retrieve_download_urls(
...
 
+    @staticmethod
+    def _create_unique_file_names(file_names_in: list[str] | list[Path]) -> list[str]:
+        """
+        Create unique file names by appending a number to the base file name.
+        """
+        file_names: list[str] = [str(file_name) for file_name in file_names_in]
+        unique_original = set(file_names)
+        unique_created: list = []
+        original_count: dict = defaultdict(int)
```
haakonvt, line 623: Does `mypy` complain if you don't type annotate these? Just using list and dict means items are `Any`, would be better to either remove and let it infer, or if that fails, add subtype.

Also, there's `typing.DefaultDict` for a more precise annotation.

cognite/client/_api/files.py
```diff
@@ -611,19 +612,50 @@ def retrieve_download_urls(
...
+
+        for file_name in file_names:
+            if file_name not in unique_created:
+                unique_created.append(file_name)
+                continue
+
+            file_suffixes = Path(file_name).suffixes
+            file_postfix = "".join(file_suffixes)
+            file_base = file_name[: -len(file_postfix)] if file_postfix else file_name
```
haakonvt, line 632: Just a suggestion:
```suggestion
            # Awaiting .removesuffix in 3.9:
            file_base = file_name[: file_name.index(file_postfix) or None]
```

cognite/client/_api/files.py
CogniteJohan: MyPy accepts inference on the list, but not the default-dict. Ruff changed the typing.DefaultDict to use the defaultdict-class directly...

cognite/client/_api/files.py
CogniteJohan: That is much more readable :) Thanks

cognite/client/_api/files.py
haakonvt: Nice! `ruff` ❤️

haakonvt (approved):
Lgtm, great additions to the usability of FilesAPI 🚀

PR #1373, opened 2023-09-19
Update apply_dml example

cognite-fholm:
Update apply_dml example to avoid confusion between id and name

## Description
Updated apply_dml example to avoid confusion between id and name


muradsater (approved):

erlendvollset (approved):
Lgtm

PR #1379, opened 2023-09-24
Add mypy flags `--warn-redundant-casts` and  `--warn-unused-ignores`

haakonvt:
## Description
- Adds a few extra `mypy` warning flags and update accordingly. 
- Removes 14 unneeded `cast`s
- Removes 15 unneeded `# type: ignore [...]` comments

#### Should make the code base easier to "keep clean" as we use newer- and more of v1 `mypy` features.


doctrino (approved):
Nice

erlendvollset (approved):
Love to see this!

PR #1384, opened 2023-09-28
fix: When token cache is invalid, don't raise `TypeError`

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/credentials.py
+++ b/cognite/client/credentials.py
@@ -207,15 +207,15 @@ class OAuthDeviceCode(_OAuthCredentialProviderWithTokenRefresh, _WithMsalSeriali
         return self.__scopes
 
     def _refresh_access_token(self) -> tuple[str, float]:
-        # First check if there is a serialized token cached on disk.
+        # First check if a token cache exists on disk. If yes, find and use:
+        # - A valid access token.
+        # - A valid refresh token, and if so, use it automatically to redeem a new access token.
+        credentials = None
         if accounts := self.__app.get_accounts():
-            credentials = self.__app.acquire_token_silent_with_error(scopes=self.__scopes, account=accounts[0])
-        else:
-            credentials = {}
+            credentials = self.__app.acquire_token_silent(scopes=self.__scopes, account=accounts[0])
 
-        # If not, we acquire a new token using device code auth flow:
-        # or the token might be expired, so we try to refresh it.
-        if not accounts or "error" in credentials:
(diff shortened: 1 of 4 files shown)


cognite/client/credentials.py
```diff
@@ -288,7 +288,10 @@ def scopes(self) -> list[str]:
     def _refresh_access_token(self) -> tuple[str, float]:
         # First check if there is a serialized token cached on disk.
         if accounts := self.__app.get_accounts():
-            credentials = self.__app.acquire_token_silent(scopes=self.__scopes, account=accounts[0])
+            credentials = self.__app.acquire_token_silent_with_error(scopes=self.__scopes, account=accounts[0])
+            if "error" in credentials:
```
erlendvollset, line 292: Perhaps a retry loop with exponential backoff? There is some retryutil added recently

erlendvollset: Or maybe we should make this exception something the existing retry logic can.

erlendvollset: And I would also suggest being more specific about which error to retry. Is there more info in the error object?

erlendvollset: I see now this is only when using the persisted token, so I guess a single retry should suffice and that you want to do it regardless of the error mode. LGTM!

erlendvollset: But we should use the same for logic for the other login flows

doctrino: Done

cognite/client/credentials.py
```diff
@@ -209,9 +209,13 @@ def scopes(self) -> list[str]:
...
         if accounts := self.__app.get_accounts():
-            credentials = self.__app.acquire_token_silent(scopes=self.__scopes, account=accounts[0])
-        # If not, we acquire a new token using device code auth flow:
+            credentials = self.__app.acquire_token_silent_with_error(scopes=self.__scopes, account=accounts[0])
         else:
+            credentials = {}
+
+        # If not, we acquire a new token using device code auth flow:
+        # or the token might be expired, so we try to refresh it.
+        if not accounts or "error" in credentials:
```
haakonvt: MSAL docs says this can still return `None` when:
-> None when there is simply no token in the cache.

haakonvt (approved):
Pushed a few suggested commits after 🆗 -ing on slack.

PR #1385, opened 2023-09-29
attempt for doing camel case check in all classes

HaydenCognite:
## Description
Dodgy [attempt](https://cognitedata.atlassian.net/jira/software/projects/DOGE/boards/871?selectedIssue=DOGE-11) to address this one. Please correct me if I have misunderstood the task. :)

## Checklist:
- [X] Tests added/updated.

### DIFF
--- a/tests/tests_unit/test_meta.py
+++ b/tests/tests_unit/test_meta.py
@@ -1,7 +1,9 @@
 import inspect
 from pathlib import Path
 
-from cognite.client.data_classes._base import CogniteResource
+import pytest
+
+from cognite.client.data_classes._base import CogniteResource, CogniteResourceList
 from tests.utils import all_subclasses
 
 ALL_FILEPATHS = Path("cognite/client/").rglob("*.py")
@@ -32,8 +34,12 @@ def test_ensure_all_files_use_future_annots():
             assert file.readline() == "from __future__ import annotations\n", err_msg.format(filepath)
 
 
-def test_ensure_all_tests_use_camel_case_except_dump():
-    err_msg = "Class: '{}' contains camel_case=True as default."
-    for cls in all_subclasses(CogniteResource):
-        if param := inspect.signature(cls.to_pandas).parameters.get("camel_case"):
-            assert param.default is False, err_msg.format(cls.__name__)
+@pytest.mark.parametrize("cls", [CogniteResource, CogniteResourceList])
+def test_ensure_all_tests_use_camel_case_except_dump(cls):
+    method = "to_pandas"
+    err_msg = "Class: '{}' for method {} does not default to False."
+    for cls in all_subclasses(cls):
+        if cls_method := getattr(cls, method, False):
+            continue
+        if param := inspect.signature(cls_method).parameters.get("camel_case"):
+            assert param.default is False, err_msg.format(cls.__name__, method)
--- a/tests/tests_unit/test_meta.py
+++ b/tests/tests_unit/test_meta.py
@@ -35,11 +35,10 @@ def test_ensure_all_files_use_future_annots():
 
 
 @pytest.mark.parametrize("cls", [CogniteResource, CogniteResourceList])
-def test_ensure_all_tests_use_camel_case_except_dump(cls):
-    method = "to_pandas"
-    err_msg = "Class: '{}' for method {} does not default to False."
-    for cls in all_subclasses(cls):
-        if cls_method := getattr(cls, method, False):
+def test_ensure_all_to_pandas_methods_use_snake_case(cls):
+    err_msg = "Class: '{}' for method to_pandas does not default camel_case parameter to False."
+    for sub_cls in all_subclasses(cls):
+        if not (cls_method := getattr(sub_cls, "to_pandas", False)):
             continue
         if param := inspect.signature(cls_method).parameters.get("camel_case"):
-            assert param.default is False, err_msg.format(cls.__name__, method)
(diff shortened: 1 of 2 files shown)


tests/tests_unit/test_meta.py
```diff
@@ -19,10 +21,20 @@ def keep(path):
...
             assert file.readline() == "from __future__ import annotations\n", err_msg.format(filepath)
+
+
+def test_ensure_all_tests_use_camel_case_except_dump():
+    err_msg = "File: '{}' contains camel_case=True and not a dump at line {}."
+    for filepath in ALL_FILEPATHS:
+        if "tests/" in str(filepath):
+            with filepath.open("r") as file:
+                for line_num, line in enumerate(file.readlines()):
+                    if ".dump(camel_case" in line:
```
haakonvt, line 37: This would skip files with arguments splatted to multiple lines. I would attack this from either AST parsing, or using the inspect module on the class objects, something along these lines:
```py
from tests.utils import all_subclasses

for cls in all_subclasses(CogniteResource):
    if method := getattr(cls, "to_pandas"):
        if param := inspect.signature(method).parameters.get("camel_case"):
            assert param.default is False
```
+ add more base classes that we want to be easily dataframe-convertible.

haakonvt (changes requested):
Nice attempt! 🥈 Let's get this over the finish line!

tests/tests_unit/test_meta.py
haakonvt: I guess we don't need `getattr` when we know all subclasses has this method... lol whoopsie. But for a general class, it would be beneficial ofc (with an additional third-argument `None` to not error when missing).

tests/tests_unit/test_meta.py
```diff
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+ALL_FILEPATHS = Path("cognite/client/").glob("**/*.py")
```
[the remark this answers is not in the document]
HaydenCognite: Ooo! Like in [this link](https://stackoverflow.com/questions/2186525/how-to-use-to-find-files-recursively)? There is a comment in there about how apparently `os.walk` is faster (although this person also had ~10K files to go through. Makes me wonder what the differences are under the hood and when the payoff is worth it. But yes, will update to rglob. I can search for other globs that can be replaced rglob if you like? I will replaced any patterns like `.glob("**/*.py")` with `.rglob("*.py")`, right?

tests/tests_unit/test_meta.py
HaydenCognite: Yeah, we talked about some of this offline and on Slack (although good to document here). Just a bit busy with customer stuff but will get to it soon.

tests/tests_unit/test_meta.py
```diff
@@ -19,10 +27,19 @@ def keep(path):
...
-    for filepath in filter(keep, all_filepaths):
+    for filepath in filter(keep, ALL_FILEPATHS):
         with filepath.open("r") as file:
             # We just read the first line from each file:
             assert file.readline() == "from __future__ import annotations\n", err_msg.format(filepath)
+
+
+@pytest.mark.parametrize("cls", [CogniteResource, CogniteResourceList])
+def test_ensure_all_tests_use_camel_case_except_dump(cls):
+    method = "to_pandas"
```
haakonvt, line 39: This seems a bit excessive?

tests/tests_unit/test_meta.py
```diff
@@ -19,10 +27,19 @@ def keep(path):
...
+    err_msg = "Class: '{}' for method {} does not default to False."
+    for cls in all_subclasses(cls):
```
haakonvt, line 41: Perhaps rename the base class so that the variable names are different

tests/tests_unit/test_meta.py
```diff
@@ -19,10 +27,19 @@ def keep(path):
...
+        if cls_method := getattr(cls, method, False):
+            continue
```
haakonvt, line 43: You need to check the logic on this one 😉

tests/tests_unit/test_meta.py
```diff
@@ -19,10 +27,19 @@ def keep(path):
...
-
-    for filepath in filter(keep, all_filepaths):
+    for filepath in filter(keep, ALL_FILEPATHS):
         with filepath.open("r") as file:
             # We just read the first line from each file:
             assert file.readline() == "from __future__ import annotations\n", err_msg.format(filepath)
+
+
+@pytest.mark.parametrize("cls", [CogniteResource, CogniteResourceList])
+def test_ensure_all_tests_use_camel_case_except_dump(cls):
```
haakonvt, line 38:
```suggestion
def test_ensure_all_to_pandas_methods_use_snake_case(cls):
```

tests/tests_unit/test_meta.py
HaydenCognite: Yeah, I think I noticed it before pushing, but then I got distracted and forgot to update. 😅

tests/tests_unit/test_meta.py
HaydenCognite: Sorry, yes, my bad.

tests/tests_unit/test_meta.py
HaydenCognite: I make a variable for (almost) anything that is needed more than once. It is used three times. 🤷

haakonvt (approved):
Lgtm!

PR #1386, opened 2023-09-29
Minor cleanups of some python antipatterns

HaydenCognite:
## Description

Take on the range/len cleanup described [here](https://cognitedata.atlassian.net/jira/software/projects/DOGE/boards/871?selectedIssue=DOGE-26). Took the liberty of cleaning up a few other things like some f-strings, list comprehensions, etc.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/assets.py
+++ b/cognite/client/_api/assets.py
@@ -1009,17 +1009,14 @@ class AssetsAPI(APIClient):
     def _get_asset_subtree(self, assets: list, current_depth: int, depth: int | None) -> list:
         subtree = assets
         if depth is None or current_depth < depth:
-            children = self._get_children(assets)
-            if children:
+            if children := self._get_children(subtree):
                 subtree.extend(self._get_asset_subtree(children, current_depth + 1, depth))
         return subtree
 
     def _get_children(self, assets: list) -> list:
         ids = [a.id for a in assets]
-        tasks = []
         chunk_size = 100
-        for i in range(0, len(ids), chunk_size):
-            tasks.append({"parent_ids": ids[i : i + chunk_size], "limit": -1})
+        tasks = [{"parent_ids": ids[i : i + chunk_size], "limit": -1} for i in range(len(ids), chunk_size)]
         tasks_summary = utils._concurrency.execute_tasks(self.list, tasks=tasks, max_workers=self._config.max_workers)
         tasks_summary.raise_compound_exception_if_failed_tasks()
         res_list = tasks_summary.results
--- a/cognite/client/_api/assets.py
+++ b/cognite/client/_api/assets.py
@@ -1015,8 +1015,7 @@ class AssetsAPI(APIClient):
 
     def _get_children(self, assets: list) -> list:
         ids = [a.id for a in assets]
-        chunk_size = 100
-        tasks = [{"parent_ids": chunk} for chunk in split_into_chunks(ids, chunk_size)]
+        tasks = [{"parent_ids": chunk, "limit": -1} for chunk in split_into_chunks(ids, 100)]
         tasks_summary = utils._concurrency.execute_tasks(self.list, tasks=tasks, max_workers=self._config.max_workers)
         tasks_summary.raise_compound_exception_if_failed_tasks()
         res_list = tasks_summary.results
--- a/cognite/client/_api/assets.py
+++ b/cognite/client/_api/assets.py
@@ -49,7 +49,7 @@ from cognite.client.data_classes.assets import AssetPropertyLike, AssetSort, Sor
 from cognite.client.data_classes.filters import Filter, _validate_filter
 from cognite.client.data_classes.shared import AggregateBucketResult
 from cognite.client.exceptions import CogniteAPIError
-from cognite.client.utils._auxiliary import split_into_n_parts
+from cognite.client.utils._auxiliary import split_into_chunks, split_into_n_parts
 from cognite.client.utils._concurrency import classify_error, get_priority_executor
 from cognite.client.utils._identifier import IdentifierSequence
 from cognite.client.utils._text import to_camel_case
@@ -1016,7 +1016,7 @@ class AssetsAPI(APIClient):
     def _get_children(self, assets: list) -> list:
         ids = [a.id for a in assets]
         chunk_size = 100
-        tasks = [{"parent_ids": ids[i : i + chunk_size], "limit": -1} for i in range(len(ids), chunk_size)]
+        tasks = [{"parent_ids": chunk} for chunk in split_into_chunks(ids, chunk_size)]
         tasks_summary = utils._concurrency.execute_tasks(self.list, tasks=tasks, max_workers=self._config.max_workers)
         tasks_summary.raise_compound_exception_if_failed_tasks()
         res_list = tasks_summary.results
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -1520,7 +1520,7 @@ class DatapointsPoster:
 
     def _bin_datapoints(self, dps_object_list: list[dict[str, Any]]) -> list[list[dict[str, Any]]]:
         for dps_object in dps_object_list:
-            for i in range(0, len(dps_object["datapoints"]), self.limit):
+            for i in range(len(dps_object["datapoints"]), self.limit):
                 dps_object_chunk = {k: dps_object[k] for k in ["id", "externalId"] if k in dps_object}
                 dps_object_chunk["datapoints"] = dps_object["datapoints"][i : i + self.limit]
                 for bin in self.bins:
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -1520,7 +1520,7 @@ class DatapointsPoster:
 
     def _bin_datapoints(self, dps_object_list: list[dict[str, Any]]) -> list[list[dict[str, Any]]]:
         for dps_object in dps_object_list:
-            for i in range(len(dps_object["datapoints"]), self.limit):
+            for i in range(0, len(dps_object["datapoints"]), self.limit):
                 dps_object_chunk = {k: dps_object[k] for k in ["id", "externalId"] if k in dps_object}
                 dps_object_chunk["datapoints"] = dps_object["datapoints"][i : i + self.limit]
                 for bin in self.bins:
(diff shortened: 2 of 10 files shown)


cognite/client/_api/assets.py
```diff
@@ -1012,17 +1012,14 @@ def retrieve_subtree(
...
                 subtree.extend(self._get_asset_subtree(children, current_depth + 1, depth))
         return subtree
 
     def _get_children(self, assets: list) -> list:
         ids = [a.id for a in assets]
-        tasks = []
         chunk_size = 100
-        for i in range(0, len(ids), chunk_size):
-            tasks.append({"parent_ids": ids[i : i + chunk_size], "limit": -1})
+        tasks = [{"parent_ids": ids[i : i + chunk_size], "limit": -1} for i in range(len(ids), chunk_size)]
```
haakonvt, line 1022: Please instead use `split_into_chunks` from utils.

cognite/client/_api/datapoints.py
```diff
@@ -1520,7 +1520,7 @@ def _validate_and_format_datapoints(
 
     def _bin_datapoints(self, dps_object_list: list[dict[str, Any]]) -> list[list[dict[str, Any]]]:
         for dps_object in dps_object_list:
-            for i in range(0, len(dps_object["datapoints"]), self.limit):
+            for i in range(len(dps_object["datapoints"]), self.limit):
```
haakonvt, line 1523: This does not work, you are now trying to loop from `len(...)` to `self.limit` instead of jumping `self.limit` at the time.

cognite/client/_api/datapoints.py
```diff
@@ -1537,9 +1537,7 @@ def _bin_datapoints(self, dps_object_list: list[dict[str, Any]]) -> list[list[di
         return binned_dps_object_list
 
     def _insert_datapoints_concurrently(self, dps_object_lists: list[list[dict[str, Any]]]) -> None:
-        tasks = []
-        for dps_object_list in dps_object_lists:
-            tasks.append((dps_object_list,))
+        tasks = [(dps_object_list,) for dps_object_list in dps_object_lists]
```
haakonvt, line 1540: :nice:

tests/tests_unit/test_api/test_sequences.py
```diff
@@ -135,7 +135,7 @@ def mock_get_sequence_data_two_col(rsps, cognite_client):
     }
     rsps.add(
         rsps.POST,
-        cognite_client.sequences._get_base_url_with_base_path() + "/sequences/data/list",
+        f"{cognite_client.sequences._get_base_url_with_base_path()}/sequences/data/list",
```
haakonvt, line 138: I do think I prefer the original here when the variable has such a long dotted name.

tests/tests_unit/test_api/test_sequences.py
```diff
@@ -558,7 +558,7 @@ def test_retrieve_dataframe_columns_mixed_with_zero(
     def test_retrieve_dataframe_columns_many_extid(self, cognite_client, mock_get_sequence_data_many_columns):
         data = cognite_client.sequences.data.retrieve(external_id="foo", start=1000000, end=1100000)
         assert isinstance(data, SequenceData)
-        assert ["ceid" + str(i) for i in range(200)] == list(data.to_pandas().columns)
+        assert [f"ceid{i!s}" for i in range(200)] == list(data.to_pandas().columns)
```
haakonvt, line 561:
```suggestion
        assert [f"ceid{i}" for i in range(200)] == list(data.to_pandas().columns)
```

haakonvt (changes requested):
I know this is still in draft so maybe you have already addressed some of these comments; but I would just like to give some feedback after the workshop session 😄 

Biggest remark: you can not remove the initial zero in `range` when used with three arguments.

cognite/client/_api/datapoints.py
HaydenCognite: Whoops! Thanks for the catch, I forgot that you can set negative indices and that's why it won't run through any of it. Like below.

```
>>> for x in range(89, 2):
...     print(x)
...
>>> for x in range(89, 2, -1):
...     print(x)
...
89
88
87
86
85
...
```

cognite/client/_api/assets.py
```diff
@@ -1012,17 +1012,14 @@ def retrieve_subtree(
...
                 subtree.extend(self._get_asset_subtree(children, current_depth + 1, depth))
         return subtree
 
     def _get_children(self, assets: list) -> list:
         ids = [a.id for a in assets]
-        tasks = []
         chunk_size = 100
-        for i in range(0, len(ids), chunk_size):
-            tasks.append({"parent_ids": ids[i : i + chunk_size], "limit": -1})
+        tasks = [{"parent_ids": chunk} for chunk in split_into_chunks(ids, chunk_size)]
```
haakonvt, line 1022: Do we need the `chunk_size` or can we remove the variable?

tests/tests_integration/test_api/test_geospatial.py
```diff
@@ -277,12 +276,12 @@ def test_update_multiple_features(self, cognite_client, allow_crs_transformation
             feature_type_external_id=test_feature_type.external_id,
             feature=[
                 Feature(external_id=test_features[idx].external_id, temperature=6.237, pressure=12.21, volume=34.43)
-                for idx in range(0, len(test_features))
+                for idx in range(len(test_features))
```
haakonvt, line 279: Look at how can you remove `range(len`

tests/tests_integration/test_api/test_geospatial.py
```diff
@@ -277,12 +276,12 @@ def test_update_multiple_features(self, cognite_client, allow_crs_transformation
...
             ],
             allow_crs_transformation=allow_crs_transformation,
             chunk_size=2,
         )
-        for idx in range(0, len(test_features)):
+        for idx in range(len(test_features)):
```
haakonvt, line 284: Same here, try to remove the range-len-construct

tests/tests_integration/test_api/test_templates.py
```diff
@@ -75,9 +77,15 @@ def new_template_instance(cognite_client, new_template_group_version):
...
-        events.append(Event(external_id="test_evt_templates_1_" + str(i), type="test_templates_1", start_time=i * 1000))
+    len_new_view = 1000
+    events = [
+        Event(
+            external_id=f"test_evt_templates_1_{i}",
+            type="test_templates_1",
+            start_time=i * len_new_view,
+        )
+        for i in range(len_new_view + 1)
+    ]
```
haakonvt, line 88:
```suggestion
    events = [
        Event(external_id=f"test_evt_templates_1_{i}", type="test_templates_1", start_time=i * len_new_view)
        for i in range(1001)
    ]
```

tests/tests_unit/test_api/test_sequences.py
```diff
@@ -113,7 +113,7 @@ def mock_get_sequence_data_many_columns(rsps, cognite_client):
     json = {
         "id": 0,
         "externalId": "eid",
-        "columns": [{"externalId": "ceid" + str(i)} for i in range(0, 200)],
+        "columns": [{"externalId": "ceid" + str(i)} for i in range(200)],
```
haakonvt, line 116: You changed this to `f"ceid{i}"` somewhere similar

tests/tests_integration/test_api/test_templates.py
HaydenCognite: This won't work because now `len_new_view` is undefined.

cognite/client/_api/assets.py
haakonvt, line 1022: You lost the `"limit": -1` part it seems

cognite/client/_api/datapoints.py
haakonvt, line 1540: Looks like `list(zip(dps_object_lists))` would work 😜 Not that I think you should change it.

cognite/client/_api/assets.py
HaydenCognite: Ah yes, well spotted! Thank you! 🙏

cognite/client/_api/datapoints.py
HaydenCognite: I could go for both, honestly. The reader knows that this is "type/variable/data massaging" in both I guess. 🤷

cognite/client/_api/assets.py
HaydenCognite: OK, going to push now. 😓

tests/tests_integration/test_api/test_templates.py
haakonvt: Just use a literal, that was my point 😄

cognite/client/_api/synthetic_time_series.py
```diff
@@ -74,10 +74,8 @@ def query(
             expressions if (isinstance(expressions, Sequence) and not isinstance(expressions, str)) else [expressions]
         )
 
-        for i in range(len(expressions_to_iterate)):
-            expression, short_expression = self._build_expression(
-                expressions_to_iterate[i], variables, aggregate, granularity
-            )
+        for exp in expressions_to_iterate:
+            expression, short_expression = self._build_expression(exp, variables, aggregate, granularity)
```
haakonvt, line 78: Nice, this is exactly what I had in mind

tests/tests_integration/test_api/test_templates.py
HaydenCognite: Yeah but with mine you only have one place to update it in case it needs to be changed.

haakonvt: But they are not the same variable! First says do 1000 loops, the other says change start time to be seconds, not ms

HaydenCognite: Ah whoops! Misread it then. Good thing I already made the change though.

tests/tests_integration/test_api/test_templates.py
```diff
@@ -75,9 +77,14 @@ def new_template_instance(cognite_client, new_template_group_version):
 
 @pytest.fixture
 def new_view(cognite_client, new_template_group_version):
-    events = []
-    for i in range(0, 1001):
-        events.append(Event(external_id="test_evt_templates_1_" + str(i), type="test_templates_1", start_time=i * 1000))
+    events = [
```
haakonvt, line 80: One more idea, make this session scoped to just run once

haakonvt (approved):
One last comment (unrelated to your changes, just made me aware of a bad fixture)

PR #1392, opened 2023-10-01
Support Units: Timeseries/Datapoint Extension

doctrino:
## Description

## Checklist:
- [x] Tests added/updated.


cognite/client/_api_client.py
```diff
@@ -354,7 +357,10 @@ def _retrieve_multiple(
             for id_chunk in identifiers.chunked(self._RETRIEVE_LIMIT)
         ]
         tasks_summary = utils._concurrency.execute_tasks(
-            self._post, tasks, max_workers=self._config.max_workers, executor=executor
+            functools.partial(self._post, api_subversion=api_subversion),
```
evertoncolling, line 360: Interesting solution using `functools.partial` here 🙂

docs/source/core_data_model.rst
```diff
@@ -154,6 +154,18 @@ Events Data classes
 
 Data points
 -----------
+.. warning::
+    TimeSeries Unit is a new feature:
+      * The API specification is in beta.
+      * The SDK implementation is in alpha.
+
+    Unit is implemented in the Datapoints APIs with the parameters `target_unt` in
```
evertoncolling, line 162:
```suggestion
    Unit conversion is implemented in the Datapoints APIs with the parameters `target_unit` and `target_unit_system` in
```

docs/source/core_data_model.rst
```diff
@@ -154,6 +154,18 @@ Events Data classes
...
+    the retrieve methods below. It is only the use of this arguments that is in alpha. Using the methods below
```
evertoncolling, line 163:
```suggestion
    the retrieve methods below. It is only the use of these arguments that is in alpha. Using the methods below
```

docs/source/core_data_model.rst
```diff
@@ -154,6 +154,18 @@ Events Data classes
 
 Data points
 -----------
+.. warning::
+    TimeSeries Unit is a new feature:
```
evertoncolling, line 158:
```suggestion
    TimeSeries unit support is a new feature:
```

cognite/client/_api_client.py
haakonvt: Probably because my monster code for fetching datapoints wasn’t super easy to extend… 😂

evertoncolling:
One thing that is missing is the updated spec for the `DataPointListItem`. It might not be strictly necessary, but wanted to highlight it.

For TimeSeries without a `unitExternalId` set, the response will be of this form:
```json
{
    "id": int,
    "externalId": string,
    "isString": bool,
    "isStep": bool,
    "unit": string,
    ...,
    "nextCursor": string
}
```
But for TimeSeries with a `unitExternalId` set, the response will change a little bit:
```json
{
    "id": int,
    "externalId": string,
    "isString": bool,
    "isStep": bool,
    "unit": string, // only present when request was made without `targetUnit` or `targetUnitSystem`
    "unitExternalId": string, // always present. reflects the actual `unitExternalId` associated to the datapoints
    ...,
    "nextCursor": string
}
```
It seems we don't use most of the TimeSeries information that comes with a datapoints response anyway, but I'm not sure if we need to update the Protobuf schemas with this new field. Also, in the future it would be nice to have an option to append the unit of the datapoints to the pandas column name in a similar way as we do for aggregates and granularity.

cognite/client/_api_client.py
haakonvt: A simplification is on its way actually…

evertoncolling (approved):
LGTM 👍! I added some suggestions and a small disclaimer on the updated spec for `DataPointListItem`, but I leave it for the Protobuf experts to evaluate if it's a must have or it can be added later on.

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -1032,6 +1089,43 @@ def test_query_with_duplicates(self, retrieve_endpoints, one_mill_dps_ts, ms_bur
             assert len(res_lst.get(external_id=ts_numeric.external_id)) == 3
             assert len(res_lst.get(external_id=ts_string.external_id)) == 2
 
+    def test_retrieve_datapoints_in_target_unit(
```
haakonvt, line 1092: None of these tests uses `target_unit_system`, so I suggest you do a comparison between:
```
target_unit_system="Imperial"
# and
target_unit="temperature:deg_f"
```
and check for equality

tests/tests_integration/test_api/test_time_series.py
```diff
@@ -112,6 +112,24 @@ def test_list_with_filters(self, cognite_client, post_spy):
         assert 0 == len(res)
         assert 1 == cognite_client.time_series._post.call_count
 
+    def test_list_timeseries_with_target_unit(self, cognite_client: CogniteClient) -> None:
+        ts1 = TimeSeries(external_id="test_list_timeseries_with_target_unit:1", unit_external_id="temperature:deg_c")
+        ts2 = TimeSeries(external_id="test_list_timeseries_with_target_unit:2", unit_external_id="temperature:deg_f")
```
haakonvt, line 117: You need a random string at the end to avoid windows-ubuntu-runner-race-conditions

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -313,6 +313,27 @@ def parametrized_values_uniform_index_fails(testrun_uid):
...
+        external_id="test_retrieve_datapoints_in_target_unit",
+        name="test_retrieve_datapoints_in_target_unit",
+        is_string=False,
+        unit_external_id="temperature:deg_c",
+    )
+    created_timeseries: TimeSeries | None = None
+    cognite_client.time_series.delete(external_id=timeseries.external_id, ignore_unknown_ids=True)
+
+    try:
+        created_timeseries = cognite_client.time_series.create(timeseries)
```
haakonvt, line 328: To avoid windows-ubuntu-runner-race-conditions, let's do `upsert` here, without deleting after.

cognite/client/_api/datapoints.py
```diff
@@ -604,6 +623,8 @@ def retrieve(
         end: int | str | datetime | None = None,
         aggregates: Aggregate | str | list[Aggregate | str] | None = None,
         granularity: str | None = None,
+        target_unit: str | None = None,
+        target_unit_system: str | None = None,
```
haakonvt, line 627: To mimic the API, these should actually be passed as part of individual settings, not at the top level. I still agree with adding here, as the default for any non-specified. But! Passing as part of the individual settings doesn't work:
```py
>>> client.time_series.data.retrieve(
>>>     id={"id": created_timeseries.id, "target_unit": "temperature:k"})
KeyError: "Dict provided by argument `id` included key(s) not understood: ['target_unit']. Required key: `id`. Optional: ['start', 'end', 'granularity', 'aggregates', 'include_outside_points', 'ignore_unknown_ids', 'limit']."

```

cognite/client/_api/datapoints.py
```diff
@@ -769,18 +792,29 @@ def retrieve(
...
         elif not dps_lst and ignore_unknown_ids:
             return None
         return dps_lst[0]
 
+    def _get_api_subversion(self, target_unit: str | None, target_unit_system: str | None) -> str | None:
+        api_subversion = None
+        if target_unit is not None or target_unit_system is not None:
+            api_subversion = "beta"
+            self._unit_warning.warn()
+        return api_subversion
```
haakonvt, line 816: See prev. comment for context. 

This would need to be more involved and check the per-time-series settings.

cognite/client/_api/datapoints.py
```diff
@@ -624,6 +645,8 @@ def retrieve(
             end (int | str | datetime | None): Exclusive end. Default: "now"
             aggregates (Aggregate | str | list[Aggregate | str] | None): Single aggregate or list of aggregates to retrieve. Default: None (raw datapoints returned)
             granularity (str | None): The granularity to fetch aggregates at. e.g. '15s', '2h', '10d'. Default: None.
+            target_unit (str | None): The unit externalId of the data points returned. If the time series does not have a unitExternalId that can be converted to the targetUnit, an error will be returned. Cannot be used with targetUnitSystem.
+            target_unit_system (str | None): The unit system of the data points returned. Cannot be used with targetUnit.
```
haakonvt, line 649: We should add at least one example of how to use this.

cognite/client/_api/datapoints.py
```diff
@@ -179,21 +180,26 @@ def _finalize_tasks(self, ordered_results: list[BaseConcurrentTask], resource_ls
...
 
-    def _make_dps_request_using_protobuf(self, payload: DatapointsPayload) -> bytes:
+    def _make_dps_request_using_protobuf(self, payload: DatapointsPayload, api_subversion: str | None) -> bytes:
         return self.dps_client._do_request(
             json=payload,
             method="POST",
             url_path=f"{self.dps_client._RESOURCE_PATH}/list",
             accept="application/protobuf",
             timeout=self.dps_client._config.timeout,
+            api_subversion=api_subversion,
```
haakonvt, line 190: We pass `api_subversion` via several functions; should we set it on the instance instead to simplify?

cognite/client/_api/datapoints.py
haakonvt: (Let me know if you want me to do this part)

haakonvt (commented):
There's something that I don't like about the mix of beta and "normal" for TimeSeries (very poorly explained but read on lol); an unsuspecting user calling `upsert` with `mode=replace` will now get a warning seemingly out of nowhere - they might unknowingly wipe the "proper unit" of their time series etc. 

How can we make this better? I think we need to make this opt-in.

cognite/client/_api/time_series.py
```diff
@@ -586,13 +641,19 @@ def upsert(
                 >>> new_time_series = TimeSeries(external_id="new_timeSeries", description="New timeSeries")
                 >>> res = c.time_series.upsert([existing_time_series, new_time_series], mode="replace")
         """
+        api_subversion = self._get_subapiversion_update(item)
+        if mode == "replace":
+            api_subversion = "beta"
+            self._unit_warning.warn()
```
haakonvt, line 647: It will be awkward to implement, but I think `replace` without using the new features should not wipe units.

Maybe this could be an "easy" solution; adding something like `in_beta` to `PropertySpec`:

```py
@dataclass
class PropertySpec:
    name: str
    is_container: bool = False
    is_nullable: bool = True
    in_beta: bool = False
```

cognite/client/_api/time_series.py
```diff
@@ -122,15 +130,19 @@ def __call__(
...
+
+        api_subversion = self._get_api_subversion(filter.unit_external_id, filter.unit_quantity)
+
         return self._list_generator(
             list_cls=TimeSeriesList,
             resource_cls=TimeSeries,
             method="POST",
             chunk_size=chunk_size,
-            filter=filter,
+            filter=filter.dump(camel_case=True),
```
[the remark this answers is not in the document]
doctrino: My guess is that it didn't, and it was never triggered.

cognite/client/_api/time_series.py
```diff
@@ -168,7 +180,14 @@ def retrieve(self, id: int | None = None, external_id: str | None = None) -> Tim
...
         """
         identifiers = IdentifierSequence.load(ids=id, external_ids=external_id).as_singleton()
-        return self._retrieve_multiple(list_cls=TimeSeriesList, resource_cls=TimeSeries, identifiers=identifiers)
+
+        return self._retrieve_multiple(
+            list_cls=TimeSeriesList,
+            resource_cls=TimeSeries,
+            identifiers=identifiers,
+            # Unit external id is only supported in beta, this can be removed when it is released in GA.
+            api_subversion="beta",
```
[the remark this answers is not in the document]
doctrino: You could pass flag into the retrieve methods, something like
```python
def retrieve(self, id: int | None = None, external_id: str | None = None, include_beta_properties: bool = False) -> TimeSeries | None:
```
However, when it gets into GA you would have to have a deprecation warning before removing it. So currently, I am leaning towards doing this silently as it is just adding an optional field.

cognite/client/_api/datapoints.py
haakonvt: Pushed a fix in 9b6accd.

cognite/client/_api/datapoints.py
haakonvt: Also fixed in 9b6accd.

cognite/client/_api/datapoints.py
haakonvt: Also fixed in 9b6accd.

cognite/client/_api/datapoints.py
```diff
@@ -784,6 +784,18 @@ def retrieve(
...
+
+            If we have created a timeseries set with 'unit_external_id' we can use the 'target_unit' parameter to convert the datapoints to the desired unit.
+            In the example below, we assume that the timeseries is set with unit_external_id 'temperature:deg_c' and id='42'.
+
+                >>> client.time_series.data.retrieve(
+                ...   id=42, start="2w-ago", limit=None, target_unit="temperature:deg_f")
+
+            Or alternatively, we can use the 'target_unit_system' parameter to convert the datapoints to the desired unit system.
+
+                >>> client.time_series.data.retrieve(
```
haakonvt, line 789: Very nice 👌

cognite/client/_api/time_series.py
doctrino: Fixed.

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -577,6 +591,51 @@ def test_retrieve_nothing__multiple(
...
+        retrieve_method = getattr(cognite_client.time_series.data, retrieve_method_name)
+
+        res = retrieve_method(external_id=timeseries.external_id, end=3, **kwargs)
+
+        if isinstance(res, pd.DataFrame):
+            res = DatapointsArray(value=res.values)
+
+        assert math.isclose(res.value[0], -40, abs_tol=0.5)
+        assert math.isclose(res.value[1], 32, abs_tol=0.5)
+        assert math.isclose(res.value[2], 212, abs_tol=0.5)
```
haakonvt, line 618: I would prefer going with the defaults, unless we suspect a larger relative error than 1e-9 (the default), which I don't see for a multiply+add operation.
```suggestion
        assert math.isclose(res.value[0], -40)
        assert math.isclose(res.value[1], 32)
        assert math.isclose(res.value[2], 212)
```

haakonvt (commented):
Great addition of `convert_resource_to_patch_object_test_cases` 🚀 

A few things missing still missing:
- Updated proto files
- `Datapoints` (+array) objects are missing `unit_external_id` (I pushed a failing test for this in 08d8019)

tests/tests_integration/test_api/test_time_series.py
```diff
@@ -129,12 +129,14 @@ def test_list_timeseries_with_target_unit(self, cognite_client: CogniteClient) -
...
-            unit_external_id="temperature:deg_c", external_id_prefix="test_list_timeseries_with_target_unit", limit=2
+            unit_external_id="temperature:deg_c",
+            external_id_prefix=f"test_list_timeseries_with_target_unit:{system}",
+            limit=2,
         )
 
         assert len(listed) == 1
         assert listed[0].unit_external_id == "temperature:deg_c"
-        assert listed[0].external_id == "test_list_timeseries_with_target_unit:1"
+        assert listed[0].external_id == f"test_list_timeseries_with_target_unit:{system}:1"
```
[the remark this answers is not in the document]
doctrino: if you create a random string, you end up recreating these every time. I don't think that is necessary as you are testing the list method, so you should only create them once and then continue to use them.

tests/tests_integration/test_api/test_time_series.py
```diff
@@ -129,12 +129,14 @@ def test_list_timeseries_with_target_unit(self, cognite_client: CogniteClient) -
             cognite_client.time_series.create(new_ts)
 
         listed = cognite_client.time_series.list(
-            unit_external_id="temperature:deg_c", external_id_prefix="test_list_timeseries_with_target_unit", limit=2
+            unit_external_id="temperature:deg_c",
+            external_id_prefix=f"test_list_timeseries_with_target_unit:{system}",
+            limit=2,
```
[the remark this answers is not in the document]
doctrino: Not necessary, you are only expected to fetch one. The limit is there only to ensure you don't get a lot if the test fails.

tests/tests_integration/test_api/test_time_series.py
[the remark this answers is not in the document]
haakonvt: I see. Why do you append the system platform?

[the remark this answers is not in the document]
doctrino: Because you asked for some uniqueness, and I added it without checking the test to carefully, will remove it again.

docs/source/core_data_model.rst
```diff
@@ -154,6 +154,18 @@ Events Data classes
...
 Data points
 -----------
+.. warning::
+    TimeSeries unit support is a new feature:
+      * The API specification is in beta.
+      * The SDK implementation is in alpha.
+
+    Unit conversion is implemented in the Datapoints APIs with the parameters `target_unit` and `target_unit_system` in
+    the retrieve methods below. It is only the use of these arguments that is in alpha. Using the methods below
+    without this arguments is stable.
```
haakonvt, line 164:
```suggestion
    the retrieve methods below. It is only the use of these arguments that is in alpha. Using the methods below
    without these arguments is stable.
```

docs/source/core_data_model.rst
```diff
@@ -370,6 +382,19 @@ Calculate the result of a function on time series
...
 Time series
 -----------
+
+.. warning::
+    TimeSeries unit support is a new feature:
+      * The API specification is in beta.
+      * The SDK implementation is in alpha.
+
+    Unit is implemented in the TimeSeris APIs with the parameters `unit_external_id` and `unit_quantity` in
+    the methods below. It is only the use of these arguments that is in alpha. Using the methods below
```
haakonvt, line 392:
```suggestion
    Unit is implemented in the TimeSeries APIs with the parameters `unit_external_id` and `unit_quantity` in
    the methods below. It is only the use of these arguments that is in alpha. Using the methods below
```

haakonvt (approved):
Let' go! 🙌

PR #1395, opened 2023-10-02
fix dynamic task

VerstraeteBert:
## Description

## Checklist:
- [ ] Tests added/updated.


cognite/client/data_classes/workflows.py
```diff
@@ -175,7 +176,7 @@ def _load(cls, resource: dict | str, cognite_client: CogniteClient | None = None
             external_id=function["externalId"],
             data=function.get("data"),
             # Allow default to come from the API.
-            is_async_complete=resource.get("isAsyncComplete"),  # type: ignore[arg-type]
+            is_async_complete=resource.get("isAsyncComplete", False),
```
haakonvt, line 179: Load methods should not use defaults (if missing in the API response then it should be `None` here as well). I got schooled by Erlend on this a few PRs ago 😅 
```suggestion
            is_async_complete=resource.get("isAsyncComplete"),
```

tests/tests_unit/test_data_classes/test_workflows.py
```diff
@@ -71,3 +72,58 @@ class TestWorkflowIds:
...
     def test_load(self, resource: Any, expected: WorkflowIds):
         assert WorkflowIds._load(resource) == expected
+
+
+class TestWorkflowExecutionDetailed:
+    @cached_property
+    def test_data(self) -> dict:
+        test_data = Path(__file__).parent / "data/workflow_execution.json"
+        with test_data.open() as f:
+            return json.load(f)
```
haakonvt, line 82: Please use a session-scoped fixture.

cognite/client/data_classes/workflows.py
```diff
@@ -272,29 +273,59 @@ def _load(cls: type[Self], resource: dict | str, cognite_client: CogniteClient |
...
+
+    def dump(self, camel_case: bool = False) -> dict[str, Any]:
+        return {
+            self.task_type: {
+                "tasks": self.tasks if isinstance(self.tasks, str) else [task.dump(camel_case) for task in self.tasks]
+            }
+        }
+
+    def __init__(self, tasks: list[WorkflowTask] | str) -> None:
+        self.tasks = tasks
```
doctrino, line 328: It is common to have the `__init__` method first in the class. Can you move it?

cognite/client/data_classes/workflows.py
```diff
@@ -317,6 +349,7 @@ def __init__(
         self,
         external_id: str,
         parameters: WorkflowTaskParameters,
+        type: Literal["function", "transformation", "cdf", "dynamic"] | None = None,
```
doctrino, line 352: This is double information that does not have to be exposed to the user. The type is kept as the class type

cognite/client/data_classes/workflows.py
```diff
@@ -535,6 +553,7 @@ def __init__(
         start_time: int | None = None,
         end_time: int | None = None,
         reason_for_incompletion: str | None = None,
+        type: str | None = None,
```
doctrino, line 556: Same here, type is already kept in the class type.

cognite/client/data_classes/workflows.py
```diff
@@ -474,37 +501,27 @@ def load(cls, data: dict[str, Any]) -> CDFTaskOutput:
...
-        output = data["output"]
-        return cls([WorkflowTask._load(task) for task in output["dynamicTasks"]])
+        return cls()
 
     def dump(self, camel_case: bool = False) -> dict[str, Any]:
-        return {
-            "output": {"dynamicTasks": [task.dump(camel_case) for task in self.dynamic_tasks]},
-            ("taskType" if camel_case else "task_type"): self.task_type,
-        }
+        return {}
```
doctrino, line 524: Doesn't dynamic task have output?

doctrino (changes requested):
My main issue with the is the introduction of type, this is an API parameter that should not be exposed as an attribute but rather kept as the class. This is similar to how we, for example, solved the filtering in FDM.

cognite/client/data_classes/workflows.py
doctrino: Can't you just use a property then? 
```python
@property
def type(self) -> str:
    self.paramterers.task_type
```

The way the class is set up now you can create a `WorkflowTask` which has `parameters` of type `FunctionTaskParamters` while you set the type to `transformation`. Having an option to create an inconsistent instance is a code smell to me, as it means you have to ensure you input the same information twice.

cognite/client/data_classes/workflows.py
doctrino: Same solution her, I think just a property will give you the same connivence without relying on the user of the class to ensure they input a type that is consistent with the `output` parameter.

doctrino (approved):
Looks good, great job

PR #1398, opened 2023-10-03
Bug wrong encoding when uploading Nordic characters through files API

doctrino:
## Description
Reference https://github.com/psf/requests/issues/5560

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/files.py
+++ b/cognite/client/_api/files.py
@@ -545,6 +545,9 @@ class FilesAPI(APIClient):
                 >>> c = CogniteClient()
                 >>> res = c.files.upload_bytes(b"some content", name="my_file", asset_ids=[1,2,3])
         """
+        if isinstance(content, str):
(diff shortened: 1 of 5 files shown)


erlendvollset (approved):

cognite/client/_api/files.py
```diff
@@ -545,6 +545,9 @@ def upload_bytes(
                 >>> c = CogniteClient()
                 >>> res = c.files.upload_bytes(b"some content", name="my_file", asset_ids=[1,2,3])
         """
+        if isinstance(content, str):
+            content = content.encode("utf-8")
```
haakonvt, line 549: Is this Windows `requests` bug? Great catch anyway!

doctrino: actually, based on the bug it seems to also happen on unix systems.

PR #1406, opened 2023-10-05
Session Retrieve

doctrino:
## Description
Problem some of the workflow tests have started to fail with due to maximum number of sessions `CogniteAPIError: There can be only 10000 sessions`.

Realized that `retrieve` was not implemented, so did that. 

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/iam.py
```diff
@@ -187,17 +187,49 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
+
+                >>> from cognite.client import CogniteClient
+                >>> c = CogniteClient()
+                >>> with c.iam.sessions() as session:
+                ...     print(session.status)
+        """
+
+        return CDFSession(self, client_credentials)
+
+    def create(self, client_credentials: ClientCredentials | dict | None = None) -> CreatedSession:
```
erlendvollset, line 216: Let's add a note about discouraging usage of this method since revoking is not handled. Should use the ctx manager instead unless you know what you're doing.

cognite/client/_api/iam.py
```diff
@@ -187,17 +187,49 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
         super().__init__(config, api_version, cognite_client)
         self._LIST_LIMIT = 100
 
-    def create(self, client_credentials: ClientCredentials | None = None) -> CreatedSession:
+    def __call__(self, client_credentials: ClientCredentials | dict | None = None) -> CDFSession:
```
erlendvollset, line 190: Let's not use `__call__` for this. All other APIClients use this for iterating over the resource, so we should be consistent.

Some naming suggestions: `session_ctx`, `create_ctx`, `create_and_revoke`.

cognite/client/_api/iam.py
```diff
@@ -214,16 +246,62 @@ def revoke(self, id: int | Sequence[int]) -> SessionList:
...
+            status (str | SessionStatus | None): If given, only sessions with the given status are returned.
 
         Returns:
             SessionList: a list of sessions in the current project.
         """
         filter = {"status": status} if status is not None else None
         return self._list(list_cls=SessionList, resource_cls=Session, method="GET", filter=filter)
+
+
+class CDFSession:
```
erlendvollset, line 292:
```suggestion
class SessionContextManager:
```

cognite/client/_api/iam.py
erlendvollset, line 190: This can also be implemented using the @contextmanager decorator, and a try:finally around the `yield`. Personally, I like that syntax better. But this works too!

cognite/client/_api/functions.py
```diff
@@ -1024,26 +1023,26 @@ def create(
...
-                    "cronExpression": cron_expression,
-                    "nonce": nonce,
-                }
-            ]
-        }
 
-        if data:
-            body["items"][0]["data"] = data
+        with self._cognite_client.iam.sessions(client_credentials) as session:
+            body: dict[str, list[dict[str, str | int | None | dict]]] = {
```
haakonvt, line 1028: This doesn't look right to me; aren't you killing the schedule right after it has been created? 🤔

cognite/client/_api/functions.py
```diff
@@ -394,16 +393,16 @@ def call(
         """
         identifier = IdentifierSequence.load(ids=id, external_ids=external_id).as_singleton()[0]
         id = _get_function_internal_id(self._cognite_client, identifier)
-        nonce = create_session_and_return_nonce(self._cognite_client, api_name="Functions API")
 
-        if data is None:
-            data = {}
-        url = f"/functions/{id}/call"
-        res = self._post(url, json={"data": data, "nonce": nonce})
+        with self._cognite_client.iam.sessions() as session:
```
haakonvt, line 397: A similar concern here, when `wait=False`, don't you risk killing the CDF connection of the currently running function? The official docs say:

> Revoke access to a session. Revocation of a session may in some cases take up to 1 hour to take effect.

Since a function can run for up to 15 minutes (on some cloud vendors), this is not a risk I would be willing to take 😬

CHANGELOG.md
```diff
@@ -17,6 +17,14 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.31.0] - 2023-10-06
+### Added
+- Retrieve method for session, `client.iam.session.retrieve`
+- Added option for getting a session as a context objet with `client.iam.session()`
+### Fixed
+- Used the context session to revoke the session when used in `Functions` and `Workflows`. This is to avoid hitting
+  the limit for sessions would could occure when triggering a lot of workflows or functions.
```
haakonvt, line 26: I don't think this should be in the change log as it mainly concerns how we run integration tests.

CHANGELOG.md
```diff
@@ -17,6 +17,14 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.31.0] - 2023-10-06
+### Added
+- Retrieve method for session, `client.iam.session.retrieve`
+- Added option for getting a session as a context objet with `client.iam.session()`
```
haakonvt, line 23: I don't see how or when "normal" users would need this (the context manager). I think this just could be a part of our test helper functions.

cognite/client/_api/iam.py
haakonvt: ➕1 vote in favor of the decorator 🎉

cognite/client/_api/iam.py
haakonvt: Perhaps my understanding is wrong, but I don't see why I would use the auto-revoking context manager. If I create a session, it is because I want to pass the nonce to some service that binds it (whatever that means) and continues running on my behalf; like a transformation or a function schedule. Wouldn't the act of revoking effectively stop the service shortly after? 🤔

cognite/client/_api/iam.py
```diff
@@ -214,16 +246,62 @@ def revoke(self, id: int | Sequence[int]) -> SessionList:
...
+    def __enter__(self) -> CreatedSession:
+        self._created_session = self._session_api.create(self._client_credentials)
+        return self._created_session
+
+    def __exit__(self, exc_type: Any = None, exc_value: Any = None, traceback_: Any = None) -> bool:
+        suppress_exception = False
+        if self._created_session is not None:
+            self._session_api.revoke(self._created_session.id)
+
+        return suppress_exception
```
haakonvt, line 307: `suppress_exception`?

cognite/client/_api/iam.py
haakonvt: Well, clearly I’m misunderstanding what revoking means; just had a function schedule complete successfully 8 hours after revoking the session it was created from… 🤔 

So what exactly does revoking imply I wonder?

erlendvollset: Right, so this is only the desired behaviour in integration tests then. I've probably misunderstood how sessions were being used - thought you would create one session per function run, but it's one per function?

doctrino: My understanding of it from workflows is that the session is used to trigger a workflow, but once it is triggered you don't need the session to be valid to continue running the workflow. I would guess it is similar for Functions.

cognite/client/_api/functions.py
doctrino: I think it is for you use the session for creating the schedule, but lets check with the functions team.

cognite/client/_api/iam.py
doctrino: Completely forgot about it, changed the implementation now.

cognite/client/_api/iam.py
haakonvt: Aight, I've checked a bit more into this now (with some help from @ngerstle-cognite ❤️ ) and can confirm that revoking the session *will inevitably kill the access to CDF*. This can take up to 45 min / 1 hour. The reason my function kept running nicely was that I actually didn't make any CDF calls in my function... Yes, huge 🤦 on my part 😂 

So, changes we need to make to this PR:
- Remove the context manager from the public methods and move to test tooling.
- Then revert the changes to `FunctionSchedulesAPI.create`...
- ...and `FunctionsAPI.call`. For this one, we could keep the changes for when `wait` is passed as `True`, since this would not kill the session before the function was finished. However, I think a better solution is to not add this kind of complexity into the SDK.

cognite/client/_api/iam.py
```diff
@@ -447,14 +448,39 @@ def revoke(self, id: int | Sequence[int]) -> SessionList:
...
+            resource_cls=Session,
+            identifiers=identifiers,
+        )
+
+    def list(self, status: str | SessionStatus | None = None) -> SessionList:
         """`List all sessions in the current project. <https://developer.cognite.com/api#tag/Sessions/operation/listSessions>`_
 
         Args:
-            status (str | None): If given, only sessions with the given status are returned.
+            status (str | SessionStatus | None): If given, only sessions with the given status are returned.
```
haakonvt, line 480: I vote to just rip off the band-aid 😅 
```suggestion
            status (SessionStatus | None): If given, only sessions with the given status are returned.
```

cognite/client/data_classes/iam.py
```diff
@@ -336,23 +336,27 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
         }
 
 
+SessionStatus: TypeAlias = Literal["ready", "active", "cancelled", "expired", "revoked", "access_lost"]
```
haakonvt, line 339: I think we should upper case these:

```suggestion
SessionStatus: TypeAlias = Literal["READY", "ACTIVE", "CANCELLED", "EXPIRED", "REVOKED", "ACCESS_LOST"]
```

cognite/client/data_classes/iam.py
```diff
@@ -336,23 +336,27 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+SessionType: TypeAlias = Literal["CLIENT_CREDENTIALS", "TOKEN_EXCHANGE"]
```
haakonvt, line 340: Looks like there might be a new one now, `ONESHOT_TOKEN_EXCHANGE`:
```suggestion
SessionType: TypeAlias = Literal["CLIENT_CREDENTIALS", "TOKEN_EXCHANGE", "ONESHOT_TOKEN_EXCHANGE"]
```

tests/tests_integration/test_api/test_iam.py
```diff
@@ -59,12 +59,15 @@ def test_create_and_delete(self, cognite_client):
...
-    )
-    def test_create_and_revoke(self, cognite_client):
+    def test_create_retrieve_and_revoke(self, cognite_client: CogniteClient) -> None:
         res = cognite_client.iam.sessions.create()
+
+        retrieved = cognite_client.iam.sessions.retrieve(res.id)
+
+        assert retrieved.id == res.id
         assert res.id in {s.id for s in cognite_client.iam.sessions.list("READY")}
         assert res.id in {s.id for s in cognite_client.iam.sessions.revoke(res.id) if s.status == "REVOKED"}
```
haakonvt, line 73: Line 72 and 73 feels like mixtures of setup, teardown and asserts 😅 Consider if its worth an improvement (idea: should we revoke session in a `finally` clause?)

tests/tests_integration/test_api/test_workflows.py
```diff
@@ -206,13 +206,30 @@ def workflow_execution_list(
...
 
+@pytest.fixture()
+def clean_created_sessions(cognite_client: CogniteClient) -> None:
+    existing_active_sessions = cognite_client.iam.sessions.list(status="active")
+    yield None
+    current_sessions = cognite_client.iam.sessions.list(status="active")
+    existing_ids = {session.id for session in existing_active_sessions}
+    to_revoked = [session.id for session in current_sessions if session.id not in existing_ids]
+    for session_id in to_revoked:
+        cognite_client.iam.sessions.revoke(session_id)
```
haakonvt, line 230:
```suggestion
    cognite_client.iam.sessions.revoke(to_revoked)
```

tests/tests_integration/test_api/test_workflows.py
```diff
@@ -368,9 +385,13 @@ def test_retrieve_non_existing_workflow_execution(self, cognite_client: CogniteC
...
 
-    @pytest.mark.skip("Bad test: CogniteAPIError: There can be only 10000 sessions")
     def test_trigger_retrieve_detailed_update_update_task(
-        self, cognite_client: CogniteClient, add_multiply_workflow: WorkflowVersion
+        self,
+        cognite_client: CogniteClient,
+        add_multiply_workflow: WorkflowVersion,
+        # Each trigger creates a new execution, so we need to clean up after each test to avoid
+        # running out of quota
+        clean_created_sessions: None,
```
haakonvt, line 394: Optionally, you may decorate the test with a pytest mark:
```py
@pytest.mark.usefixtures("clean_created_sessions")
```

haakonvt (commented):
The PR; it's alive, IT'S ALIVE

cognite/client/_api/iam.py
```diff
@@ -447,14 +448,39 @@ def revoke(self, id: int | Sequence[int]) -> SessionList:
...
+        """
+
+        identifiers = IdentifierSequence.load(ids=id, external_ids=None)
+        return self._retrieve_multiple(
+            list_cls=SessionList,
+            resource_cls=Session,
+            identifiers=identifiers,
+        )
+
+    def list(self, status: str | SessionStatus | None = None) -> SessionList:
```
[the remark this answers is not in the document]
haakonvt: Very optional: _SessionList object has no attribute 'as_ids'_

tests/tests_integration/test_api/test_iam.py
doctrino: What about now? Also discovered that session.revoked was not overloaded correctly.

cognite/client/data_classes/iam.py
doctrino: Nice, forgot to check the spec again.

tests/tests_integration/test_api/test_iam.py
```diff
@@ -64,10 +64,15 @@ def test_create_and_delete(self, cognite_client):
...
+            created = cognite_client.iam.sessions.create()
 
-        retrieved = cognite_client.iam.sessions.retrieve(res.id)
+            retrieved = cognite_client.iam.sessions.retrieve(created.id)
 
-        assert retrieved.id == res.id
-        assert res.id in {s.id for s in cognite_client.iam.sessions.list("READY")}
-        assert res.id in {s.id for s in cognite_client.iam.sessions.revoke(res.id) if s.status == "REVOKED"}
+            assert retrieved.id == created.id
+            assert created.id in {s.id for s in cognite_client.iam.sessions.list("READY")}
```
haakonvt, line 74: Potentially the need for a limit arg here

tests/tests_integration/test_api/test_iam.py
haakonvt: Nice 👌

haakonvt (approved):
`lgtm.upper()`

PR #1407, opened 2023-10-05
`isinstance` checks should not use generics from typing

HaydenCognite:
## Description
See title, stolen from [doge34](https://cognitedata.atlassian.net/browse/DOGE-34). @haakonvt to check if I missed anything or misunderstood something. I also made a few other changes that may or may not be welcome. :)

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/raw.py
+++ b/cognite/client/_api/raw.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Iterator, List, Sequence, cast, overload
+from typing import TYPE_CHECKING, Any, Iterator, Sequence, cast, overload
 
 from cognite.client import utils
 from cognite.client._api_client import APIClient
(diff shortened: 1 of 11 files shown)


cognite/client/_api/raw.py
```diff
@@ -495,12 +495,9 @@ def retrieve(self, db_name: str, table_name: str, key: str) -> Row | None:
...
         if columns is None:
             return None
-        if not isinstance(columns, List):
+        if not isinstance(columns, list):
             raise ValueError("Expected a list for argument columns")
-        if len(columns) == 0:
-            return ","
-        else:
-            return ",".join([str(x) for x in columns])
+        return "," if len(columns) == 0 else ",".join([str(x) for x in columns])
```
HaydenCognite, line 500: I know not everyone is a fan of this so let me know if it should be reverted.

haakonvt, line 500: Lgtm, I would switch it around though, also don't need to build a temporary list:
```suggestion
        return ",".join(str(x) for x in columns) if columns else ","
```

erlendvollset (approved):
LGTM

haakonvt (approved):
Lgtm, just had one comment

cognite/client/_api/raw.py
haakonvt: As a side note, I generally dislike conditional expressions (except when they are super-short) 😉

HaydenCognite: OK, I will take a mix of your answer and @haakonvt s answer by reverting it, but also removing the list creation. Hope that is fine with both of ye.

HaydenCognite: No worries, I will take a mix of both of your comments (removing the list instantiation and returning to the if-else branch).

PR #1411, opened 2023-10-06
add ability to easily toggle debug logging on/off

ddonukis:
## Description
Made it easy to toggle debug logging on and off.
```python
client.config.debug = True  # globally enable debugging for cognite sdk
client.config.debug = False  # disable
client.config.debug  # return global state of congite sdk logging
```
In addition, added a docs section explaining how to enable/disable debug logs.

## Checklist:
- [x] Tests added/updated.


docs/source/cognite_client.rst
```diff
@@ -5,12 +5,11 @@ CogniteClient
     :members:
     :member-order: bysource
 
+.. _class_client_ClientConfig:
```
ddonukis, line 8: I added an explicit label as it appears to be impossible to link to autogenerated classes. The link works well in my locally built docs.

cognite/client/utils/_logging.py
```diff
@@ -62,11 +64,22 @@ def format(self, record: logging.LogRecord) -> str:
 
 
 def _configure_logger_for_debug_mode() -> None:
-    logger = logging.getLogger("cognite-sdk")
+    logger = logging.getLogger(COGNITE_LOGGER_NAME)
     logger.setLevel("DEBUG")
```
haakonvt, line 68: Could change this one to `logger.setLevel(logging.DEBUG)` while you're at it

cognite/client/data_classes/transformations/__init__.py
```diff
@@ -27,13 +27,14 @@
...
 from cognite.client.exceptions import CogniteAPIError
+from cognite.client.utils._logging import COGNITE_LOGGER_NAME
 from cognite.client.utils._text import convert_all_keys_to_camel_case
 
 if TYPE_CHECKING:
     from cognite.client import CogniteClient
 
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger(COGNITE_LOGGER_NAME)
```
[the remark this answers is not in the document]
haakonvt: This one should remain as is - the warning should always be emitted.

haakonvt, line 37:
```suggestion
logger = logging.getLogger(__name__)
```

haakonvt (commented):
Lgtm! But we need @erlendvollset to also take a look at this before merging 😄

cognite/client/data_classes/transformations/__init__.py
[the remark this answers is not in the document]
ddonukis: Aha, I see. Should the warnings module be used instead to be explicit?
https://docs.python.org/3/library/warnings.html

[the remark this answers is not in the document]
haakonvt: Yes, I think that would be nice 👌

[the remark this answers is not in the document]
ddonukis: Took a closer look at the code.
are you talking about this message that it should always be emitted?
```python
logger.debug(
  f"Unable to create a session and get a nonce towards {project=} using the provided "
  f"{credentials_name} credentials: {err!r}"
)
```
Not sure what you mean by "emitted" but as far as I understand it will not be shown by by default, because the default logger level is INFO. And this won't be shown even if we enable debug logging because this is not using a cognite sdk logger or a child logger 🤷

cognite/client/utils/_logging.py
```diff
@@ -2,6 +2,8 @@
 
 import logging
 
+COGNITE_LOGGER_NAME = "cognite-sdk"
```
erlendvollset, line 5: Let's drop this and just use `logging.getLogger(__name__)` everywhere

cognite/client/utils/_logging.py
```diff
@@ -62,11 +64,22 @@ def format(self, record: logging.LogRecord) -> str:
 
 
 def _configure_logger_for_debug_mode() -> None:
-    logger = logging.getLogger("cognite-sdk")
+    logger = logging.getLogger(COGNITE_LOGGER_NAME)
```
erlendvollset, line 67: Then this becomes just logging.getLogger("cognite.client") as all modules' `__name__` variable will start with "cognite.client".

cognite/client/data_classes/transformations/__init__.py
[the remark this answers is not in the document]
haakonvt: I wrote that log message - and I regret not using a `warning` instead. I don't see how or when a user specifying to use nonce credentials would not want to know about OIDC credentials being used instead (passed to, and stored in the transformation backend).

I vote for `UserWarning` 👍 

Please also make it a bit more verbose, perhaps adding an extra sentence
`The provided OIDC credentials will be passed on to the transformation service.`

cognite/client/config.py
```diff
@@ -6,6 +6,11 @@
 
 from cognite.client._version import __api_subversion__
 from cognite.client.credentials import CredentialProvider
+from cognite.client.utils._logging import (
+    _configure_logger_for_debug_mode,
+    _disable_debug_logging,
+    _is_debug_logging_enabled,
+)
```
ddonukis, line 13: utils import on top is causing the following error:
```
ImportError: cannot import name 'global_config' from partially initialized module 'cognite.client.config' (most likely due to a circular import) (/Users/dmytrod/dev/cognite-sdk-python/cognite/client/config.py)
```
If I move the imports `global_config = GlobalConfig()` the error disappears.

Any recommendations on how to solve this issue @erlendvollset ?

cognite/client/data_classes/transformations/__init__.py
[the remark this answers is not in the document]
ddonukis: updated to use user warning

cognite/client/config.py
haakonvt: Easy option is to do local imports

cognite/client/data_classes/transformations/__init__.py
```diff
@@ -267,10 +264,12 @@ def _try_get_or_create_nonce(
                 # This is fine, we might be missing SessionsACL. The OIDC credentials will then be passed
                 # directly to the backend service. We do however not catch CogniteAuthError as that would
                 # mean the credentials are invalid.
-                logger.debug(
+                msg = (
                     f"Unable to create a session and get a nonce towards {project=} using the provided "
                     f"{credentials_name} credentials: {err!r}"
+                    "\nProvided OIDC credentials will be passed on to the transformation service."
                 )
+                warnings.warn(msg, UserWarning)
```
haakonvt, line 272: Nice, much more explicit what happens under the hood 🚀

cognite/client/config.py
ddonukis: I see that it used to be a local import before my changes... maybe that's the reason why 😝

ddonukis:
All the comments are addressed now and the imports fixed. This should be good to go.

cognite/client/_api_client.py
```diff
@@ -65,7 +65,7 @@
     from cognite.client import CogniteClient
     from cognite.client.config import ClientConfig
 
-log = logging.getLogger("cognite-sdk")
+log = logging.getLogger(__name__)
```
erlendvollset, line 68: We should make an announcement about this change in the sdk and python channels.

erlendvollset (approved):
Excellent!

PR #1418, opened 2023-10-10
Simplify conditional expressions with True/False as output

HaydenCognite:
## Description
Simplifying some bits and bobs in conditional statements.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/functions.py
+++ b/cognite/client/_api/functions.py
@@ -32,7 +32,7 @@ from cognite.client.data_classes import (
     TimestampRange,
 )
 from cognite.client.data_classes.functions import FunctionCallsFilter, FunctionsStatus
-from cognite.client.utils._auxiliary import assert_type, exactly_one_is_not_none, is_unlimited, local_import
+from cognite.client.utils._auxiliary import assert_type, is_unlimited, local_import
 from cognite.client.utils._identifier import Identifier, IdentifierSequence
 from cognite.client.utils._session import create_session_and_return_nonce
 
@@ -475,10 +475,11 @@ class FunctionsAPI(APIClient):
         folder: str | None, file_id: int | None, function_handle: Callable[..., Any] | None
     ) -> None:
         source_code_options = {"folder": folder, "file_id": file_id, "function_handle": function_handle}
-        # TODO: Fix
-        # exactly_one_is_not_none(source_code_options)
+        # TODO: Fix to use exactly_one_is_not_none function
         given_source_code_options = [key for key in source_code_options if source_code_options[key]]
-        if not exactly_one_is_not_none(given_source_code_options):
+        if not given_source_code_options:
+            raise TypeError("Exactly one of the arguments folder, file_id and handle is required, but none were given.")
+        elif len(given_source_code_options) > 1:
             raise TypeError(
                 "Exactly one of the arguments folder, file_id and handle is required, but "
                 + ", ".join(given_source_code_options)
--- a/cognite/client/_api/functions.py
+++ b/cognite/client/_api/functions.py
@@ -32,7 +32,7 @@ from cognite.client.data_classes import (
     TimestampRange,
 )
 from cognite.client.data_classes.functions import FunctionCallsFilter, FunctionsStatus
-from cognite.client.utils._auxiliary import assert_type, is_unlimited, local_import
+from cognite.client.utils._auxiliary import assert_type, exactly_one_is_not_none, is_unlimited, local_import
 from cognite.client.utils._identifier import Identifier, IdentifierSequence
 from cognite.client.utils._session import create_session_and_return_nonce
 
@@ -475,10 +475,10 @@ class FunctionsAPI(APIClient):
         folder: str | None, file_id: int | None, function_handle: Callable[..., Any] | None
     ) -> None:
         source_code_options = {"folder": folder, "file_id": file_id, "function_handle": function_handle}
+        # TODO: Fix
+        # exactly_one_is_not_none(source_code_options)
         given_source_code_options = [key for key in source_code_options if source_code_options[key]]
-        if not given_source_code_options:
-            raise TypeError("Exactly one of the arguments folder, file_id and handle is required, but none were given.")
-        elif len(given_source_code_options) > 1:
+        if not exactly_one_is_not_none(given_source_code_options):
             raise TypeError(
                 "Exactly one of the arguments folder, file_id and handle is required, but "
                 + ", ".join(given_source_code_options)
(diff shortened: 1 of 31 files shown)


cognite/client/_api/functions.py
```diff
@@ -192,11 +192,10 @@ def create(
         sleep_time = 1.0  # seconds
         for i in range(MAX_RETRIES):
             file = self._cognite_client.files.retrieve(id=file_id)
-            if file is None or not file.uploaded:
-                time.sleep(sleep_time)
-                sleep_time *= 2
-            else:
+            if file is not None and file.uploaded:
```
erlendvollset, line 195: Can be further simplified to
```suggestion
            if file and file.uploaded:
```

erlendvollset (approved):
LGTM!

cognite/client/_api/functions.py
HaydenCognite: Nice one, thanks Erlend!

cognite/client/_api_client.py
```diff
@@ -261,9 +261,7 @@ def _is_retryable(self, method: str, path: str) -> bool:
 
         if method in ["GET", "PUT", "PATCH"]:
             return True
-        if method == "POST" and self._url_is_retryable(path):
-            return True
-        return False
+        return bool(method == "POST" and self._url_is_retryable(path))
```
haakonvt, line 264: We can drop the `bool`. We can also add in the first `if` check:
```py
return method in ["GET", "PUT", "PATCH"] or method == "POST" and self._url_is_retryable(path)
```
No, you don't need parentheses as `and` binds stronger than `or` 🚀

tests/tests_integration/test_api/test_events.py
```diff
@@ -215,7 +215,7 @@ def test_aggregate_has_type(self, cognite_client: CogniteClient, event_list: Eve
 
         count = cognite_client.events.aggregate_count(EventProperty.type, advanced_filter=is_integration_test)
 
-        assert count >= sum(1 for e in event_list if e.type)
+        assert count >= sum(bool(e.type) for e in event_list)
```
haakonvt, line 218: Perhaps using `len` makes it more readable? (I know sum was used in the original) I leave it up to you
```suggestion
        assert count >= len([e in event_list if e.type])
```
If you decide to change it, there's also this one later:
`assert count >= sum(s.asset_id is not None for s in sequence_list)`

tests/tests_integration/test_api/test_sequences.py
```diff
@@ -48,8 +48,9 @@ def root_asset(cognite_client: CogniteClient) -> Asset:
         name="integration_test:root_asset",
         external_id="integration_test:root_asset",
     )
-    retrieved = cognite_client.assets.retrieve_multiple(external_ids=[root_asset.external_id], ignore_unknown_ids=True)
-    if retrieved:
+    if retrieved := cognite_client.assets.retrieve_multiple(
+        external_ids=[root_asset.external_id], ignore_unknown_ids=True
+    ):
```
haakonvt, line 53: This change requires an extra line compared to not using the walrus 🤔

haakonvt (commented):
Good stuff!

tests/tests_integration/test_api/test_events.py
HaydenCognite: I am kind of 50-50 on this. 🤷‍♂️  Won't calculating len on an iterable increase cost compared to a generator? Readability and performance are both important. Maybe a comment to clarify?

tests/tests_integration/test_api/test_sequences.py
HaydenCognite: I find this one more readable than the original. 🤷 

Screenshot 2023-10-11 at 10 26 28

Figure 1: After with walrus.

Screenshot 2023-10-11 at 10 27 44

Figure 2: Before with line stretching out further.

🤷‍♂️

cognite/client/_api_client.py
HaydenCognite: Love it! The AND operator always ought to have priority when no bracketing, since as soon as one is not true, you can short-circuit that check and move along! 🚀 ⚡

cognite/client/_api/functions.py
```diff
@@ -476,8 +475,8 @@ def _assert_exactly_one_of_folder_or_file_id_or_function_handle(
         folder: str | None, file_id: int | None, function_handle: Callable[..., Any] | None
     ) -> None:
         source_code_options = {"folder": folder, "file_id": file_id, "function_handle": function_handle}
-        given_source_code_options = [key for key in source_code_options.keys() if source_code_options[key]]
-        if len(given_source_code_options) < 1:
+        given_source_code_options = [key for key in source_code_options if source_code_options[key]]
+        if not given_source_code_options:
```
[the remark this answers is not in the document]
HaydenCognite: So if I make use of `exactly_one_is_not_none`, then I would have to combine these messages since they are currently splitting the space up into three options (not saying this is a bad idea). Maybe saying something like `"Exactly one of the arguments folder, file_id and handle is required, but {given_source_code_options} was provided"`. Does that sound reasonable to you?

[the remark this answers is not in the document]
haakonvt: Btw, this sounds like a separate PR, let's move the discussion there if you chose to continue with this 👍

tests/tests_integration/test_api/test_events.py
haakonvt: Honestly, this is a test, and readability matters most. I think the main issue with your approach is the `bool(e.type)` which frankly looks weird to the untrained eye. I think I favor the original one, or `len`, just for its clarity.

If this had been a performance-critical part of the SDK, then I'd favor whatever would run the fastest 😅 But yeah, let's keep in mind this is after all just a test.

tests/tests_integration/test_api/test_sequences.py
haakonvt: Up to you 😉

cognite/client/_api/functions.py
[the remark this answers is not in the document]
HaydenCognite: No worries, I will put it on the board.

tests/tests_integration/test_api/test_events.py
HaydenCognite: Fair enough, I think the tests take long enough already, but I guess developers are not running them that often, so probably good to prioritise readability here. 👍

tests/tests_integration/test_api/test_sequences.py
HaydenCognite: Alright, I'll make an executive decision here. 😛

PR #1421, opened 2023-10-12
Update data_modeling.instances.apply examples

erlendvollset:
This makes more sense, as 'actor' and 'person' properties would be populated on the same node. Not two different nodes, with an edge between them.

## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/data_modeling/instances.py
+++ b/cognite/client/_api/data_modeling/instances.py
@@ -572,44 +572,61 @@ class InstancesAPI(APIClient):
                 >>> nodes = [NodeApply("mySpace", "myNodeId")]
                 >>> res = c.data_modeling.instances.apply(nodes)
 
-            Create two nodes with data with an one to many edge, and a one to one edge
+            Create two nodes with data with a one-to-many edge
 
                 >>> from cognite.client import CogniteClient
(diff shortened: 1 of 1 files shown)


cognite/client/_api/data_modeling/instances.py
```diff
@@ -572,44 +572,61 @@ def apply(
...
+                ...             ViewId("mySpace", "MovieView", "v1"),
+                ...             {"title": "Terminator", "releaseYear": 1984}
+                ...         )
+                ...     ]
+                ... )
+                ... # This is one-to-many edge, in this case from a person to a movie
+                >>> actor_to_movie = EdgeApply(
+                ...     space="actors",
+                ...     external_id="relation:arnold_schwarzenegger:terminator",
+                ...     type=("types", "acts-in"),
```
doctrino, line 607: How will the hyphen be shown in `graphQL`?
```suggestion
                ...     type=("types", "acts-in"),
```

erlendvollset: As specified in the "relation" directive here https://developer.cognite.com/dm/graphql/modeling#specification-of-directives

PR #1425, opened 2023-10-16
Simplify datapoints queries with limit

haakonvt:
## Description
The current code is 100% unmaintainable. Period. I know this because I wrote it 😉 

This PR simplifies one thing: fetching datapoints with a non-infinite limit is no longer concurrent. Yes this is a slow-down, but it's entirely worth it.

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/datapoint_tasks.py
```diff
@@ -529,17 +537,17 @@ def __init__(self, *, limit: int, **kwargs: Any) -> None:
         assert isinstance(limit, int)
 
     @property
-    def ts_task_type(self) -> type[ParallelLimitedAggTask]:
-        return ParallelLimitedAggTask
+    def get_task_orchestrator(self) -> type[SerialLimitedAggTaskOrchestrator]:
```
erlendvollset, line 540: just task_orchestrator since it's a property?

erlendvollset (approved):
Nothing controversial here. I'm all for simplifying! Nice work!

cognite/client/_api/datapoint_tasks.py
haakonvt: Aye, agreed

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.33.2] - 2023-10-16
+### Fixed
+- When fetching datapoints from "a few time series" (implementation detail), all missing, non-ignorable time series
+  are now raised together in a `CogniteNotFoundError` rather than only the first encountered.
+
```
haakonvt, line 24:
```suggestion
## [6.33.2] - 2023-10-16
### Fixed
- When fetching datapoints from "a few time series" (implementation detail), all missing, non-ignorable time series
  are now raised together in a `CogniteNotFoundError` rather than only the first encountered.

### Improved
- Datapoints fetching has a lower peak memory consumption when fetching from multiple time series simultaneously.
```

PR #1426, opened 2023-10-16
Support data_set_id in `functions.create()`

ddonukis:
## Description
- `functions.create()` now supports `data_set_id` parameter that allows you to explicitly set the data set the function code will be uploaded to for deployment. Specifying a data set can help resolve the error 403 that happens due to missing permissions to create files not scoped to a data set.
- wrap a confusing `Resource not found.` error when failing to create a file

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/files.py
```diff
@@ -562,10 +562,17 @@ def upload_bytes(
...
         )
+        try:
+            res = self._post(
+                url_path=self._RESOURCE_PATH, json=file_metadata.dump(camel_case=True), params={"overwrite": overwrite}
+            )
+        except CogniteAPIError as e:
+            if e.code == 403 and "insufficient access rights" in e.message:
+                dsid_notice = " Try to provide a data_set_id." if data_set_id is None else ""
+                msg = f"Could not upload a file due to insufficient access rights.{dsid_notice}"
+                raise CogniteFileUploadError(message=msg, code=e.code) from e
```
erlendvollset, line 573: This is a breaking change. Also not convinced that this should be a CogniteFileUploadError, so don't think it merits a major bump. You can raise a subclass of CogniteAPIError instead, so not to break the API.

tests/tests_unit/test_api/test_functions.py
```diff
@@ -1147,35 +1165,53 @@ def validate_file_upload_call(*args, **kwargs):
...
     mock.files.upload_bytes.assert_called_once()
     call = mock.files.upload_bytes.call_args
     assert len(call.args) == 1 and type(call.args[0]) is bytes  # noqa: E721
-    assert call.kwargs == {"name": "name.zip", "external_id": xid, "overwrite": overwrite}
+    assert call.kwargs == {"name": "name.zip", "external_id": xid, "overwrite": overwrite, "data_set_id": data_set_id}
 
 
-@pytest.mark.parametrize("xid, overwrite", ((None, False), ("xid", True)))
-def test__zip_and_upload_folder__zip_file_content(fns_api_with_mock_client, xid, overwrite):
+@pytest.mark.parametrize(
```
erlendvollset, line 1203: A dedicated test for this would be better. If we extend these for every new feature we add, this test is going to have very many reasons to fail, which makes it less valuable when it _does_ fail.

erlendvollset (commented):
Nice! Some small comments

cognite/client/_api/files.py
ddonukis: Good point, I didn't realized it's a breaking change! Replaced with a subclass of API error.

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.33.2] - 2023-10-16
+### Added
+- `functions.create()` now supports `data_set_id` parameter that allows you to explicitly set the data set the function code will be uploaded to for deployment. Specifying a data set can help resolve the error 403 that happens due to missing permissions to create files not scoped to a data set.
```
```suggestion
## [6.33.2] - 2023-10-18
### Added
- `functions.create()` now accepts a `data_set_id` parameter. Note: This is not for the Cognite function, but for the zipfile containing
  the source code files that is uploaded on the user's behalf (from which the function is then created). Specifying a data set may
  help resolve the error 'Resource not found' (403) that happens when a user is not allowed to create files outside a data set.
```

cognite/client/_api/files.py
```diff
@@ -562,10 +562,17 @@ def upload_bytes(
...
+        try:
+            res = self._post(
+                url_path=self._RESOURCE_PATH, json=file_metadata.dump(camel_case=True), params={"overwrite": overwrite}
+            )
+        except CogniteAPIError as e:
+            if e.code == 403 and "insufficient access rights" in e.message:
+                dsid_notice = " Try to provide a data_set_id." if data_set_id is None else ""
+                msg = f"Could not create a file due to insufficient access rights.{dsid_notice}"
+                raise CogniteFileCreateError(message=msg, code=e.code, x_request_id=e.x_request_id) from e
+            raise e
```
haakonvt, line 574:
```suggestion
            raise
```

cognite/client/_api/functions.py
```diff
@@ -132,6 +133,7 @@ def create(
             index_url (str | None): Index URL for Python Package Manager to use. Be aware of the intrinsic security implications of using the `index_url` option. `More information can be found on official docs, <https://docs.cognite.com/cdf/functions/#additional-arguments>`_
             extra_index_urls (list[str] | None): Extra Index URLs for Python Package Manager to use. Be aware of the intrinsic security implications of using the `extra_index_urls` option. `More information can be found on official docs, <https://docs.cognite.com/cdf/functions/#additional-arguments>`_
             skip_folder_validation (bool): When creating a function using the 'folder' argument, pass True to skip the extra validation step that attempts to import the module. Skipping can be useful when your function requires several heavy packages to already be installed locally. Defaults to False.
+            data_set_id (int | None): Data set to upload the function code to.
```
haakonvt, line 136:
```suggestion
            data_set_id (int | None): Data set to upload the function code to. Note: Does not affect the function itself.
```

haakonvt (commented):
Just some nit + longer changelog entry 😉

haakonvt (approved):
Lgtm!

cognite/client/_api/files.py
haakonvt: I guess `CogniteAuthError` was not used as it doesn't inherit from `CogniteAPIError`, but `CogniteException`.

PR #1434, opened 2023-10-22
[v7] Fix asserts meant for the SDK user

haakonvt:
## Description
See changelog.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -1362,12 +1362,13 @@ class DatapointsAPI(APIClient):
                 >>> c = CogniteClient()
                 >>> c.time_series.data.delete_range(start="1w-ago", end="now", id=1)
         """
-        start = timestamp_to_ms(start)
-        end = timestamp_to_ms(end)
-        assert end > start, "end must be larger than start"
+        start_ms = timestamp_to_ms(start)
+        end_ms = timestamp_to_ms(end)
+        if end_ms <= start_ms:
+            raise ValueError(f"{end=} must be larger than {start=}")
 
         identifier = Identifier.of_either(id, external_id).as_dict()
-        delete_dps_object = {**identifier, "inclusiveBegin": start, "exclusiveEnd": end}
+        delete_dps_object = {**identifier, "inclusiveBegin": start_ms, "exclusiveEnd": end_ms}
         self._delete_datapoints_ranges([delete_dps_object])
 
     def delete_ranges(self, ranges: list[dict[str, Any]]) -> None:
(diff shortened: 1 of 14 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -1512,19 +1513,17 @@ def _validate_and_format_datapoints(
...
-                assert "timestamp" in dp, "A datapoint is missing the 'timestamp' key"
-                assert "value" in dp, "A datapoint is missing the 'value' key"
-                valid_datapoints.append((timestamp_to_ms(dp["timestamp"]), dp["value"]))
-        return valid_datapoints
+            return [(timestamp_to_ms(t), v) for t, v in datapoints]
+        datapoints = cast(List[Dict[str, Any]], datapoints)
+        try:
+            return [(timestamp_to_ms(dp["timestamp"]), dp["value"]) for dp in datapoints]
+        except KeyError:
+            raise KeyError("A datapoint is missing one or both keys ['value', 'timestamp'].")
```
doctrino, line 1526: I think the error message can be could be made more helpful by capturing all the failed datapoints, and display something like missing x of y timestamps and z of y values.

cognite/client/_api/functions.py
```diff
@@ -964,9 +964,9 @@ def list(
             try:
                 IdentifierSequence.load(ids=function_id, external_ids=function_external_id).assert_singleton()
             except ValueError:
-                raise AssertionError(
+                raise ValueError(
                     "Both 'function_id' and 'function_external_id' were supplied, pass exactly one or neither."
-                )
+                ) from None
```
doctrino, line 969: Why from None here and not the Value Error above? Is it a syntactic sugar?

cognite/client/data_classes/_base.py
```diff
@@ -393,7 +401,9 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
         Returns:
             dict[str, Any]: A dictionary representation of the instance.
         """
-        assert camel_case is True, "snake_case is currently unsupported"  # TODO maybe (when unifying classes)
+        if camel_case is False:
+            # TODO maybe (when unifying classes)
+            raise ValueError("snake_case is currently unsupported")
```
doctrino, line 406: Why even have the option fi it is not supported? This is not a `CogniteResource` so suggest just removing the `camel_case` parameter.

doctrino (commented):
Dumping a few suggestions.

cognite/client/_api/datapoints.py
haakonvt: I agree, but I don't think it belongs in this PR which seeks to remove wrong usage of `assert`s and `AssertionError`.

cognite/client/_api/functions.py
haakonvt: Because the wording in one of the two different errors from`assert_singleton`:
```
1) ValueError: Exactly one of id or external id must be specified
2) ValueError: No identifiers specified
```
...mentions id + external id instead of what the user expects, function_id and function_external_id.

We need to refactor the Identifier stuff to more easily work with different identifiers, like user_identifier, name, key to name a few...

cognite/client/data_classes/_base.py
haakonvt: Yeah, that's better

haakonvt: Unable to remove it, as several methods call dump with `camel_case=True` on list of objects that's `Union[CogniteUpdate, CogniteResource]`. Changed the type hint to `Literal[True]` instead and removed the if-test.

PR #1438, opened 2023-10-22
[v7] Change cog resource to pandas

haakonvt:
## TODO:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -141,7 +141,7 @@ class CogniteResource(_WithClientMixin):
 
     def to_pandas(
         self,
-        expand_metadata: bool = True,
+        expand_metadata: bool = False,
         metadata_prefix: str = "metadata.",
         ignore: list[str] | None = None,
         camel_case: bool = False,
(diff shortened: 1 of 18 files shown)


cognite/client/data_classes/_base.py
```diff
@@ -140,30 +140,36 @@ def _load(
...
 
         Args:
-            expand (Sequence[str]): List of row keys to expand, only works if the value is a Dict. Will expand metadata by default.
-            ignore (list[str] | None): List of row keys to not include when converting to a data frame.
-            camel_case (bool): Convert column names to camel case (e.g. `externalId` instead of `external_id`)
+            expand_metadata (bool): Expand the metadata into separate rows.
+            metadata_prefix (str): Prefix to use for the metadata rows, if expanded.
+            ignore (list[str] | None): List of row keys to skip when converting to a data frame. Is applied before expansions.
+            camel_case (bool): Convert column names to camel case (e.g. `externalId` instead of `external_id`). Does not affect custom data like metadata.
+            convert_timestamps (bool): Convert known attributes storing CDF timestamps (milliseconds since epoch) to datetime. Does not affect custom data like metadata.
```
doctrino, line 157: Could this be a sequence of strings? Then, have a default with the known attributes, such that the user could just also specify the metadata field they want converted?

cognite/client/utils/_time.py
```diff
@@ -164,44 +165,54 @@ def timestamp_to_ms(timestamp: int | float | str | datetime) -> int:
...
-    return new_item
+def _convert_time_attrs_in_dict(item: dict, to_isoformat: bool) -> dict:
+    for k in TIME_ATTRIBUTES.intersection(item):
+        try:
+            if to_isoformat:
+                item[k] = ms_to_datetime(item[k]).isoformat(sep=" ", timespec="milliseconds")
+            else:
+                item[k] = ms_to_datetime(item[k])
+        except ValueError:
+            pass
```
doctrino, line 197:
```suggestion
        with suppress(ValueError):
            if to_isoformat:
                item[k] = ms_to_datetime(item[k]).isoformat(sep=" ", timespec="milliseconds")
            else:
                item[k] = ms_to_datetime(item[k])
```

And remember to `from contextlib import suppress`

cognite/client/data_classes/_base.py
doctrino: Very uncertain about this suggestion as it increases complexity.

doctrino (commented):
Very nice fix.

cognite/client/data_classes/_base.py
haakonvt: Unsure! Could be added later without breaking, so I'll leave it for now.

cognite/client/utils/_time.py
haakonvt: I went with try-except as `suppress` has quite the larger overhead (with enter and exit methods) - and we call this in a loop. *But I prefer suppress* 😎

PR #1443, opened 2023-10-25
Add support for `through` in node result expressions

erlendvollset:
- Support 'through' on node result expressions
- Bump version and update changelog

## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/data_modeling/query.py
+++ b/cognite/client/data_classes/data_modeling/query.py
@@ -162,6 +162,12 @@ class ResultSetExpression(ABC):
                 "from_": query_node.get("from"),
                 "filter": Filter.load(query_node["filter"]) if "filter" in query_node else None,
             }
+            if (through := query_node.get("through")) is not None:
+                node["through"] = [
+                    through["view"]["space"],
+                    through["view"]["externalId"] + "/" + through["view"]["version"],
+                    through["identifier"],
+                ]
             return NodeResultSetExpression(sort=sort, limit=query.get("limit"), **node)
         elif "edges" in query:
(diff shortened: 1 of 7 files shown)


cognite/client/data_classes/data_modeling/query.py
```diff
@@ -201,6 +209,18 @@ def dump(self, camel_case: bool = False) -> dict[str, Any]:
...
         if self.filter:
             nodes["filter"] = self.filter.dump()
+        if self.through:
+            if len(self.through) != 3:
+                raise ValueError(f"`through` must be on the form (space, view/version, property), was {self.through}")
+            nodes["through"] = {
+                "view": {
+                    "type": "view",
+                    "space": self.through[0],
+                    "externalId": self.through[1].split("/")[0],
```
j-bjorne, line 219: Silly question, can external id or version have `/` in it?

erlendvollset: Version yes, externalId no.

erlendvollset:
> LGTM, I am a but curious as to why through became a 3 string tuple / len 3 string array, instead of an object itself with say:
> 
> ```
> class ViewPropertySomething:
>       view: ViewReference,
>       identifier: String
> ```
> 
> So I will be curious if you can amuse me and answer that just so I learn more about the python code. Also asked a question whether `/` can be part of either externalId or version as I assume that will break the code suggested.

Just to make it less verbose. Property references are not consistent in the DMS API, some places it's an object some places it's an array of strings. I prefer the array, but it could just as well have been what you propose.

PR #1448, opened 2023-10-27
[v7] Load YAML

doctrino:
## Description

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/data_modeling/instances.py
+++ b/cognite/client/_api/data_modeling/instances.py
@@ -96,7 +96,7 @@ class _NodeOrEdgeList(CogniteResourceList):
     _RESOURCE = (Node, Edge)  # type: ignore[assignment]
 
     @classmethod
-    def _load(
+    def load(
         cls, resource_list: Iterable[dict[str, Any]] | str, cognite_client: CogniteClient | None = None
     ) -> _NodeOrEdgeList:
         resource_list = json.loads(resource_list) if isinstance(resource_list, str) else resource_list
@@ -111,7 +111,7 @@ class _NodeOrEdgeList(CogniteResourceList):
 
 class _NodeOrEdgeResourceAdapter:
     @staticmethod
-    def _load(data: str | dict, cognite_client: CogniteClient | None = None) -> Node | Edge:
+    def load(data: str | dict, cognite_client: CogniteClient | None = None) -> Node | Edge:
         data = json.loads(data) if isinstance(data, str) else data
         if data["instanceType"] == "node":
             return Node.load(data)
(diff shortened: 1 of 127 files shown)


cognite/client/_api/data_modeling/instances.py
```diff
@@ -101,7 +101,7 @@ def load(
     ) -> _NodeOrEdgeList:
         resource_list = json.loads(resource_list) if isinstance(resource_list, str) else resource_list
         resources: list[Node | Edge] = [
-            Node.load(data) if data["instanceType"] == "node" else Edge.load(data) for data in resource_list
+            Node._load(data) if data["instanceType"] == "node" else Edge._load(data) for data in resource_list
```
erlendvollset, line 104: All these changes from `load` to `_load` shouldn't be necessary, right? We can still use the public method here?

cognite/client/data_classes/_base.py
```diff
@@ -130,14 +131,35 @@ def dump(self, camel_case: bool = False) -> dict[str, Any]:
...
-    def load(
-        cls: type[T_CogniteResource], resource: dict | str, cognite_client: CogniteClient | None = None
-    ) -> T_CogniteResource:
-        if isinstance(resource, dict):
-            return fast_dict_load(cls, resource, cognite_client=cognite_client)
-        elif isinstance(resource, str):
-            return cls.load(json.loads(resource), cognite_client=cognite_client)
-        raise TypeError(f"Resource must be json str or dict, not {type(resource)}")
+    @final
+    def load(cls: type[Self], resource: dict | str, cognite_client: CogniteClient | None = None) -> Self:
```
erlendvollset, line 135: I think this one is redundant
```suggestion
    def load(cls, resource: dict | str, cognite_client: CogniteClient | None = None) -> Self:
```

cognite/client/utils/_auxiliary.py
```diff
@@ -65,6 +68,14 @@ def fast_dict_load(
     return instance
 
 
+def load_yaml_or_json(resource: str) -> dict[str, Any]:
+    try:
+        yaml = local_import("yaml")
```
erlendvollset, line 73: This will potentially get invoked a lot - is the overhead from importing negligible? If not, we can do a conditional definition of this method. e.g.

```
try:
  import yaml
  def load_yaml_or_json(resource: str) -> dict[str, Any]:
    return yaml.safe_load(resource)
except CogniteImportError:
  import json
  def load_yaml_or_json(resource: str) -> dict[str, Any]:
    return json.loads(resource)
  
```

cognite/client/_api/data_modeling/instances.py
doctrino: Yes, we can, it is an optimization. This is an internal method and we now that we have a dict, then we save a function call in the callstack by calling _load directly.

cognite/client/utils/_auxiliary.py
doctrino: What about simply skipping the extra function call and do 
```
    try:
        import yaml
        return yaml.safe_load(resource)
    except ImportError:
        return json.loads(resource)
```

erlendvollset (approved):
Lgtm

PR #1452, opened 2023-10-27
[v7] Remove dict subclasses

doctrino:
## Description

## Checklist:
- [x] Tests added/updated.


cognite/client/data_classes/_base.py
```diff
@@ -110,8 +113,11 @@ def _get_cognite_client(self) -> CogniteClient | None:
         return self.__cognite_client
 
 
-class CogniteResource(_WithClientMixin):
-    __cognite_client: CogniteClient | None
+class _SerializationMixin:
```
erlendvollset, line 116: A mixin providing public methods needs to be public itself.  Also mixins typically shouldn't have constructors.

I think it makes more sense to just call this `CogniteObject`, and let it be the de facto base class.

doctrino: My thinking was that it is more descriptive, but I agree with your argument so have changed it back now.

PR #1454, opened 2023-10-28
[v7] Add timestamp conversion to `CogniteResourceList.to_pandas`

haakonvt:
## Description
- Add default conversion of CDF (milliseconds-integer) timestamps to `datetime` for all known columns.
- Add test for to_pandas on values + dtypes. Reason: _this PR did not cause a single test to fail_... which is bad.

Tested with lowest supported pandas version (1.4.0) and newest.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,9 @@ in sequence produces an equal object to the original, for example,
 methods are `json` and `yaml` serializable.
 
 ### Added
-- `CogniteResource.to_pandas` now converts known timestamps to `datetime` by default. Can be turned off with
-  the new parameter `convert_timestamps`.
+- `CogniteResource.to_pandas` and `CogniteResourceList.to_pandas` now converts known timestamps to `datetime` by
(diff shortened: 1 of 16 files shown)


cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -422,8 +422,8 @@ def dump(self, camel_case: bool = False) -> dict[str, Any]:
         return output
 
     @classmethod
-    def load(cls, data: dict | str) -> NodeApply:
-        data = json.loads(data) if isinstance(data, str) else data
+    def load(cls, resource: dict[Any, Any] | str, cognite_client: CogniteClient | None = None) -> NodeApply:
+        data = json.loads(resource) if isinstance(resource, str) else resource
```
haakonvt, line 426: Note to self: merge error caused this, move changes into previous (squashed) commit.

Note 2: fix any-any

erlendvollset (approved):
Lgtm 💯

PR #1455, opened 2023-10-30
[CDF-19483] Workflows skip tasks

VerstraeteBert:
## Description
Behaviour: 
   - `skipTask`: for both failures and timeouts it will retry until the retries are exhausted, then the Task is marked as `COMPLETED_WITH_ERRORS` and the next tasks are executed.
   - `abortWorkflow`:
       -  in case of Failures, retries will be done until exhausted, then the task is marked as `FAILED` and the Workflow is marked the same.
       - in case of a Timeout, no retries will be performed, the task is marked as `TIMED_OUT` and the Workflow is marked as `FAILED`

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/workflows.py
+++ b/cognite/client/data_classes/workflows.py
@@ -332,9 +332,12 @@ class DynamicTaskParameters(WorkflowTaskParameters):
 class TaskFailurePolicy(Enum):
     """
     Defines the policy to handle failures and timeouts.
-    - `SKIP_TASK`: For both failures and timeouts, it will retry until the retries are exhausted.
+
+    - `SKIP_TASK`:
+        For both failures and timeouts, it will retry until the retries are exhausted.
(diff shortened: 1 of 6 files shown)


cognite/client/data_classes/workflows.py
```diff
@@ -352,6 +372,7 @@ def __init__(
         description: str | None = None,
         retries: int = 3,
         timeout: int = 3600,
+        on_failure: TaskFailurePolicy = TaskFailurePolicy.ABORT_WORKFLOW,
```
doctrino, line 375: Lets just use a literal here as we do in the rest of the SDK implementation
```suggestion
        on_failure: Literal["abortWorkflow", "skipTask"] = "abortWorkflow",
```

doctrino (approved):
Great work and speed!

PR #1456, opened 2023-10-30
[v7] Fix exceptions not being raised together as `CogniteMultiException`

haakonvt:
## Description
Fixes one of the oldest issues: #578 

With this change, `client.time_series.data.retrieve_latest(id=123)`, assuming this id doesn't randomly exist, now raises `CogniteNotFoundError` instead of `CogniteAPIError`.

PR #1460, opened 2023-10-31
[v7] Refactor task executors. Make read ops. fast fail by default

haakonvt:
## Description

## Checklist:
- [ ] Tests added/updated.


cognite/client/utils/_concurrency.py
```diff
@@ -39,88 +50,75 @@ def joined_results(self, unwrap_fn: Callable | None = None) -> list:
...
-    missing_exc = None
-    dup_exc = None
-    unknown_exc: Exception | None = None
-    for exc in exceptions:
-        if isinstance(exc, CogniteAPIError):
-            if exc.code in [400, 422] and exc.missing is not None:
+        self.raise_compound_exception_if_failed_tasks()
+        # # TODO: Change to 'raise_compound_exception_if_failed_tasks' in next major version?
+        # if self.exceptions:
+        #     raise self.exceptions[0]
```
haakonvt, line 83: Awaiting #1456

cognite/client/utils/_identifier.py
```diff
@@ -191,6 +192,11 @@ def unwrap_identifier(identifier: str | int | dict) -> str | int:
             return identifier["space"]
         raise ValueError(f"{identifier} does not contain 'id' or 'externalId' or 'space'")
 
+    @staticmethod
+    def extract_identifiers(dct: dict[str, Any]) -> dict[str, str | int]:
+        """An API payload might look like {"id": 1, "before": "2w-ago", ...}. This function extracts the identifiers"""
```
doctrino, line 197: Why is the `before: "2w-ago"` part of this example? Isn't this for identifiers?

doctrino (commented):
Great functionality, which I guess the API will be happy with as it saves requests. 

I am wondering if we need to write in the documentation what skipped means, and maybe have a longer section on why the API has atomic operations, while the SDK does not. But that can wait until after v7.

doctrino (approved):

cognite/client/utils/_identifier.py
haakonvt: Because this function extracts identifiers only from whatever the API request (payload) may contain. So it was just the first thing to pop into my mind, could have been `limit=123` or whatever

haakonvt:
> I am wondering if we need to write in the documentation what skipped means, and maybe have a longer section on why the API has atomic operations, while the SDK does not. But that can wait until after v7.

I forgot the change log 😅  On it.

Btw, I don't see how this change affects atomicity at all?

PR #1461, opened 2023-10-31
[v7] Bug in new capabilities code

doctrino:
## Description

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/capabilities.py
+++ b/cognite/client/data_classes/capabilities.py
@@ -46,7 +46,7 @@ class Capability(ABC):
                         f"For '{scope_cls.__name__}', the following unknown fields were ignored: {not_supported}. "
                         "Try updating to the latest SDK version, or create an issue on Github!"
                     )
-                    return scope_cls(**rename_and_exclude_keys(data, exclude=not_supported))
+                    return cast(Self, scope_cls(**rename_and_exclude_keys(data, exclude=not_supported)))
             return cast(Self, UnknownScope(name=name, data=data))
 
         def dump(self, camel_case: bool = False) -> dict[str, Any]:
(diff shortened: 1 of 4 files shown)


haakonvt:
Won't this treat a valid `relationshipsAcl` as unknown? I.e. the bugged one I had:
```
'capabilities': [
    {'relationshipsAcl': {'scope': {'datasetScope': {'ids': ['2332579319524236']}}, 'actions': ['READ', 'WRITE']}},
]
```

erlendvollset:
This is gonna require some rationale. I don't understand this one at all.

haakonvt:
I think `datasetScope` needs to accept both `ids` (and the other), then use post-init to assert only one was given. Then for dump, only dump the one received.

cognite/client/data_classes/capabilities.py
```diff
@@ -49,21 +49,22 @@ class CurrentUserScope(ScopeBase):
...
-
 @dataclass(frozen=True)
 class IDScope(ScopeBase):
     _scope_name = "idScope"
     ids: list[int]
 
 
 @dataclass(frozen=True)
-class ExtractionPipelineScope(IDScope):
+class ExtractionPipelineScope(ScopeBase):
```
doctrino, line 59: I changed this one to avoid having a inheritance hierarchy deeper than one, as the `ScopeBase` should be the interface of al scope implementations. I also think that `ExtractionPipelineScope` is not a type of `IDScope`, thus it would be misleading to have it as a parent.

doctrino:
> I think `datasetScope` needs to accept both `ids` (and the other), then use post-init to assert only one was given. Then for dump, only dump the one received.

Turns out it was just a typo in the SDK code, no inconsistency in the scopes from the API.

haakonvt:
> > I think `datasetScope` needs to accept both `ids` (and the other), then use post-init to assert only one was given. Then for dump, only dump the one received.
> 
> Turns out it was just a typo in the SDK code, no inconsistency in the scopes from the API.

Nice! Nothing beats consistency ❤️

haakonvt (approved):
🚀

PR #1464, opened 2023-11-01
fix transformations backup auth flow

haakonvt:
## Description
When backup flow is used, scopes is not dumped as a comma-separated string.


erlendvollset (approved):
Lgtm

PR #1467, opened 2023-11-03
fix `Capability.load` for responses with multiple top-level keys

haakonvt:
## Description

1. fix Capability.load for responses with mulitple top-level keys
2. change a few json.loads to load_yaml_or_json
3. performance: loading from dict is the most common pattern
4. unify all dump methods to camel_case=False

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/data_sets.py
+++ b/cognite/client/data_classes/data_sets.py
@@ -83,7 +83,7 @@ class DataSetFilter(CogniteFilter):
         self.external_id_prefix = external_id_prefix
         self.write_protected = write_protected
 
-    def dump(self, camel_case: bool = True) -> dict[str, Any]:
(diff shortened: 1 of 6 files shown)


cognite/client/data_classes/data_sets.py
```diff
@@ -83,7 +83,7 @@ def __init__(
         self.external_id_prefix = external_id_prefix
         self.write_protected = write_protected
 
-    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+    def dump(self, camel_case: bool = False) -> dict[str, Any]:
```
doctrino, line 86: Nice catches.

doctrino (approved):
Great even more usability with better error message.

PR #1468, opened 2023-11-03
[v7] Project Capability

doctrino:
## Description

The output of `client.iam.token.inspect()` was not properly loaded. Added GroupCapability and `GroupCapabilites`. In addition, added helper method `has_capability` to demonstrate why we want the container class `GroupCapabilities`.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,7 @@ with no easy way to add a prefix. Also, it no longer expands metadata by default
 - Classes `Geometry`, `AssetAggregate`, `AggregateResultItem`, `EndTimeFilter`, `Label`, `LabelFilter`, `ExtractionPipelineContact`,
   `TimestampRange`, `AggregateResult`, `GeometryFilter`, `GeoLocation`, `RevisionCameraProperties`, `BoundingBox3D` are no longer
   `dict` but classes with attributes matching the API.
-- Calling `client.iam.token.inspect()` now gives an object `TokenInspection` with attribute `cababilities` of type `GroupCapabilities`
+- Calling `client.iam.token.inspect()` now gives an object `TokenInspection` with attribute `cababilities` of type `ProjectCapabilities`
   instead of `list[dict]`
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,7 +149,8 @@ with no easy way to add a prefix. Also, it no longer expands metadata by default
 
 ## [6.38.1] - 2023-10-31
 ### Fixed
-- `onFailure` property in Workflows was expected as mandatory and was raising KeyError if it was not returnd by the API. The sdk now assumes the field to be optional loads it as None instead of raising an error.
+- `onFailure` property in Workflows was expected as mandatory and was raising KeyError if it was not returned by the API.
+  The SDK now assumes the field to be optional and loads it as None instead of raising an error.
 
 ## [6.38.0] - 2023-10-30
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,7 @@ with no easy way to add a prefix. Also, it no longer expands metadata by default
 - Classes `Geometry`, `AssetAggregate`, `AggregateResultItem`, `EndTimeFilter`, `Label`, `LabelFilter`, `ExtractionPipelineContact`,
   `TimestampRange`, `AggregateResult`, `GeometryFilter`, `GeoLocation`, `RevisionCameraProperties`, `BoundingBox3D` are no longer
   `dict` but classes with attributes matching the API.
-- Calling `client.iam.token.inspect()` now gives an object `TokenInspection` with attribute `cababilities` of type `ProjectCapabilities`
+- Calling `client.iam.token.inspect()` now gives an object `TokenInspection` with attribute `cababilities` of type `ProjectCapabilitiesList`
   instead of `list[dict]`
 
 ### Added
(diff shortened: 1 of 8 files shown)


cognite/client/data_classes/capabilities.py
haakonvt, line 1: Seems too much still ends up as `UnknownAcl` or `UnknownScope` or `Action.Unknown`:

Screenshot 2023-11-06 at 12 09 11

haakonvt, line 1: Same, but for `iam.groups`:

Screenshot 2023-11-06 at 12 09 27

haakonvt (commented):
Could we iterate on the name `GroupCapability`? It seems to me that this could be a source of confusion when the "get groups"-method does not return "group capabilities", but a list of capabilites.

cognite/client/data_classes/capabilities.py
```diff
@@ -102,34 +106,165 @@ def dump(self, camel_case: bool = False) -> dict[str, Any]:
...
+        capability = next(key for key in resource if key != "projectScope")
+
+        return cls(
+            capability=Capability.load({capability: resource[capability]}),
+            project_scope=ProjectScope.load({ProjectScope.name: resource[ProjectScope.name]}),
+        )
+
+    def dump(self, camel_case: bool = False) -> dict[str, Any]:
+        dumped = self.capability.dump(camel_case=camel_case)
+        dumped[ProjectScope.name] = self.project_scope.dump(camel_case=camel_case)
```
haakonvt, line 175: This means `ProjectScope.name` is always camelCased in the return dict

cognite/client/data_classes/capabilities.py
```diff
@@ -102,34 +106,165 @@ def dump(self, camel_case: bool = False) -> dict[str, Any]:
...
+        return dumped
+
+
+class GroupCapabilities(UserList):
```
haakonvt, line 179: Why don't we use `CogniteResourceList`?

cognite/client/data_classes/capabilities.py
```diff
@@ -174,6 +335,11 @@ class UnknownScope(Capability.Scope):
     def __getitem__(self, item: str) -> Any:
         return self.data[item]
 
+    def is_within(self, other: Self) -> bool:
+        if isinstance(other, AllScope):
+            return True
+        raise NotImplementedError("Unknown scopes cannot be compared")
```
haakonvt, line 341: We don't know if the unknown scope has an AllScope, so I suggest we just always raise here

cognite/client/data_classes/capabilities.py
haakonvt: Moved fix to other PR, #1473

haakonvt: Moved fix to other PR, #1473

cognite/client/data_classes/capabilities.py
doctrino: Up for discussion, but isn't that a specific name of a scope and thus should not be changed casing of?

cognite/client/data_classes/capabilities.py
doctrino: I was about to write a reason why not, and realized I did not agree with it, so will switch it.

haakonvt: Haha 😄 Would be nice if we had a `CogniteResourceList` that is without all the nonsense (client reference, id/xid lookup etc).

CHANGELOG.md
```diff
@@ -75,6 +75,8 @@ with no easy way to add a prefix. Also, it no longer expands metadata by default
 - Classes `Geometry`, `AssetAggregate`, `AggregateResultItem`, `EndTimeFilter`, `Label`, `LabelFilter`, `ExtractionPipelineContact`,
   `TimestampRange`, `AggregateResult`, `GeometryFilter`, `GeoLocation`, `RevisionCameraProperties`, `BoundingBox3D` are no longer
   `dict` but classes with attributes matching the API.
+- Calling `client.iam.token.inspect()` now gives an object `TokenInspection` with attribute `cababilities` of type `GroupCapabilities`
```
haakonvt, line 78: Do a quick search for the old class name

haakonvt (approved):
Lgtm, just one merge conflict tbd

cognite/client/data_classes/capabilities.py
```diff
@@ -102,34 +109,126 @@ def dump(self, camel_case: bool = False) -> dict[str, Any]:
...
+    def dump(self, camel_case: bool = False) -> dict[str, Any]:
+        dumped = self.capability.dump(camel_case=camel_case)
+        dumped.update(self.project_scope.dump(camel_case=camel_case))
+        return dumped
+
+
+class ProjectCapabilities(CogniteResourceList[ProjectCapability]):
+    _RESOURCE = ProjectCapability
+
+    def includes_capability(self, capability: Capability) -> bool:
```
erlendvollset, line 186: This will also ignore project scope for the capabilities. It needs to accept a project parameter.

erlendvollset: All in all I'm hesitant about adding this feature, as there are many intricacies wrt. how capabilities are handled, so having to duplicate that logic in SDKs is both error prone and a maintenance hassle. 

e.g. in data modeling you may implicitly have READ when the WRITE action is present, which is not reflected in the `has_capability` implementation for those capabilities.

doctrino: My point of adding it is to demonstrate why I want the `ProjectCapabilityList` to be there and not just a `list[ProjectCapability`. But this can be postponed. 

When it comes to whether we should do this feature at all, I know this is implemented in multiple projects which means it is worth having here. But I think we should have it well documented and clear that this implementation is not covering edge cases as the one you mentioned above. A basic implementation will cover 95% of all use cases.

haakonvt: Moved this to a different PR, but kept the `has_capability` on `Capability`. It checks 1:1, so it is easy to keep as it is "always correct" as opposed to having capabilities spread over several groups.

PR #1469, opened 2023-11-04
Add callback to global_config to allow tracking of one's own SDK usage

haakonvt:
# DRAFT
Ref. edm-pipeline debug discussion: @spex66 @thorkildcognite 

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_http_client.py
+++ b/cognite/client/_http_client.py
@@ -5,7 +5,7 @@ import random
 import socket
 import time
 from http import cookiejar
-from typing import Any, Callable, Literal, MutableMapping
+from typing import Any, Callable, ClassVar, Literal, MutableMapping
 
 import requests
 import requests.adapters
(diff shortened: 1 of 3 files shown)


spex66:
Not run it yet, but expected code consolidating counters across threads/processes?

Or does it require handling concurrency on callback side?

Parallelism still confuses me every time :)

haakonvt:
> Not run it yet, but expected code consolidating counters across threads/processes?

@spex66 For API methods that use concurrency (the majority are "serial" by default), the different threads will call the callback function. Note also that the SDK only uses threading as a means of concurrency. I.e., this should be totally fine as the exact ordering of `RequestDetails` objects is not needed nor useful, and the container-operations) are thread-safe thanks to the GIL.

If the callback function for some reason should mutate

haakonvt:
Accidentally clicked close... 👀

haakonvt:
> Not run it yet, but expected code consolidating counters across threads/processes?

@spex66 The callback will be called by the different threads executing the requests (the SDK _only uses threading_ as a means of concurrency). In the two examples given in the docstring, they would all write to the same object, which is totally fine (GIL + ordering doesn't matter).

Do you have a usage pattern in mind that you're afraid might have a race condition?

haakonvt:
@spex66 do you need this feature or should we put it on hold?

spex66:
Ok hold pls, and I need to set a reminder to look into it .. later

haakonvt:
@spex66 I'm closing as I have not heard back from you

spex66:
My bad 😞

PR #1476, opened 2023-11-07
Give actions better error messages

haakonvt:
## Description
A feature of our Capability subclasses is that they all have the same `.Action` attribute. A downside with this approach is that for users creating these do not get a acl-specific error message:

```py
ValueError: 'DELETE' is not a valid Action
```
...even though `DELETE` for certain Acls is a valid one - and this might cause confusion.

This PR adds the Acls name into the enum, so that this changes to:

```py
ValueError: 'DELETE' is not a valid AnnotationsAcl Action
```

I opted to not reimplement `__missing__` for every single one in favor of some slightly magic code 👀 
## Checklist:
- [x] Tests added/updated.


doctrino (approved):
Great!

PR #1481, opened 2023-11-09
[v7] feature: Compare capabilities and add validation

haakonvt:
## Description
Adds the feature of easily comparing capabilities across groups.

Also

- add validation to all Acls on instantiation
- pass client ref to `ProjectCapabilitiesList`
- add missing acl: `UserProfilesAcl`
- fix `TableScope`, remove `DatabaseTableScope`
- Simplify `ProjectCapability`

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/capabilities.py
+++ b/cognite/client/data_classes/capabilities.py
@@ -7,6 +7,7 @@ from abc import ABC
 from dataclasses import asdict, dataclass, field
 from itertools import groupby, product
 from operator import itemgetter
+from types import MappingProxyType
 from typing import TYPE_CHECKING, Any, ClassVar, Sequence, cast
 
 from typing_extensions import Self
@@ -31,6 +32,22 @@ class Capability(ABC):
     _capability_name: ClassVar[str]
     actions: Sequence[Action]
     scope: Scope
+
+    def __post_init__(self) -> None:
+        if (capability_cls := type(self)) is UnknownAcl:
+            return
+        acl = capability_cls.__name__
+        if bad_actions := [a for a in self.actions if a not in capability_cls.Action]:
+            raise ValueError(
+                f"{acl} got an unknown action: {bad_actions}, expected one of: {list(capability_cls.Action)}. "
+                f"Example usage: AssetsAcl(actions=[AssetsAcl.Action.Read], scope=AssetsAcl.Scope.All())"
+            )
+        allowed_scopes = _VALID_SCOPES_BY_CAPABILITY[capability_cls]
+        if allowed_scopes and type(self.scope) not in allowed_scopes:
+            raise ValueError(
+                f"{acl} got an unknown scope: {self.scope}, expected one of: {[s.__name__ for s in allowed_scopes]}. "
+                f"Example usage: AssetsAcl(actions=[AssetsAcl.Action.Read], scope=AssetsAcl.Scope.All())"
+            )
 
     class Action(enum.Enum):
         ...
@@ -1150,3 +1167,8 @@ for acl in _CAPABILITY_CLASS_BY_NAME.values():
     if acl.Action.__members__:
         _cls = type(next(iter(acl.Action.__members__.values())))
         _cls.__name__ = f"{acl.__name__} {_cls.__name__}"
+
+# Add lookup that knows which acls support which scopes:
+_VALID_SCOPES_BY_CAPABILITY: MappingProxyType[type[Capability], frozenset[type[Capability.Scope]]] = MappingProxyType(
+    {acl: frozenset(filter(inspect.isclass, vars(acl.Scope).values())) for acl in _CAPABILITY_CLASS_BY_NAME.values()}
+)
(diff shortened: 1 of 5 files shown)


cognite/client/data_classes/capabilities.py
```diff
@@ -63,9 +65,31 @@ def dump(self, camel_case: bool = False) -> dict[str, Any]:
...
+    def from_tuple(cls, tpl: tuple) -> Self:
+        acl_name, action, scope_name, *scope_params = tpl
+        capability_cls = _CAPABILITY_CLASS_BY_NAME[acl_name]
+        scope_cls = _SCOPE_CLASS_BY_NAME[scope_name]
+
+        if not scope_params:
+            scope = scope_cls()
+        elif len(scope_params) == 1:
+            scope = scope_cls(scope_params)  # type: ignore [call-arg]
+        elif len(scope_params) == 2 and scope_name == TableScope._scope_name:
```
haakonvt, line 85: Note to self: can do `scope_cls is`

tests/tests_unit/test_data_classes/test_capabilities.py
```diff
@@ -215,122 +215,77 @@ def test_load__action_does_not_exist(self, acl_cls_name: str, bad_action: str, d
...
+        ],
+    )
+    def test_is_missing_capability(
+        self, proj_capabs_list: ProjectCapabilitiesList, project_name: str, capability: Capability
+    ) -> None:
+        missing_acls = proj_capabs_list.compare([capability], project=project_name)
+        assert missing_acls == [capability]
+
+        missing_acls = proj_capabs_list.compare([capability], project="do-es-nt-ex-is-ts")
+        assert missing_acls == [capability]
```
doctrino, line 291: Should we also have a test when you some of the capabilities? Input a set of capabilities, and some of them are missing?

cognite/client/data_classes/capabilities.py
```diff
@@ -255,6 +398,12 @@ class AssetRootIDScope(Capability.Scope):
     _scope_name = "assetRootIdScope"
     root_ids: list[int]
 
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "root_ids", [int(i) for i in self.root_ids])
```
doctrino, line 402: Not saying we should not do this, but it will cause some pain for the template repo as there  you store a configuration with the id as an external id, and then load it into groups, and then converts all the external ids to ids. Main problem being that you cannot store access configs as internal ids.

doctrino (approved):

tests/tests_unit/test_data_classes/test_capabilities.py
haakonvt: That's what this test does? Or am I misunderstanding your suggestion? 

Theres `test_has_capability` and `test_is_missing_capability`.

haakonvt: Oh, a group input, yes, I can add that

cognite/client/data_classes/capabilities.py
haakonvt: I don't understand this, could you elaborate? Do you pretend the root_ids are external ids, then convert later?

cognite/client/utils/_auxiliary.py
```diff
@@ -243,3 +243,7 @@ def load_resource(dct: dict[str, Any], cls: type[T_CogniteResource], key: str) -
 
 def unpack_items_in_payload(payload: dict[str, dict[str, Any]]) -> list:
     return payload["json"]["items"]
+
+
+def combine_sets(*sets: set[T] | frozenset[T]) -> set[T]:
+    return set().union(*sets)
```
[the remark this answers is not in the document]
doctrino: I had the same doubt. I guess it is clearer not to have it, as then sets and union are well known, and a custom function is not.

cognite/client/data_classes/capabilities.py
doctrino: Yes, although I hate that solution, but I can imagen a typical way to store configs is
```
capabilities:
  - threedAcl:
      actions:
        - READ
      scope:
        datasetScope: {
          ids: ['infield_fornebu_location_source_data_dataset']
        }
  - assetsAcl:
      actions:
        - READ
      scope:
        datasetScope: {
          ids: ['infield_fornebu_location_source_data_dataset']
        }
```

cognite/client/utils/_auxiliary.py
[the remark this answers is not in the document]
haakonvt: Agreed, I'll delete it. Only used 2 places anyhow

cognite/client/data_classes/capabilities.py
haakonvt: Oh I see. The problem now is that in order to properly compare capabilities, the ids returned by the API are still strings, even though the spec says integers, and so a user would likely input integers, and then he comparison would fail...

tests/tests_unit/test_data_classes/test_capabilities.py
haakonvt: Will implement in another PR that makes comparisons simpler for e.g. ordinary capabilities.

PR #1453, opened 2023-10-27
v7 release

haakonvt:
## Description
v7 release branch that is without branch protection rules... :hide: 

Ref. #1432

PR #1483, opened 2023-11-09
[v7] change all `dump` methods to `camel_case=True` by default

haakonvt:
## Description
Dump is inherently tied to creating a response meant for the API, so it makes sense to make the default the same as what the API expects.


doctrino (commented):
Does this pass the tests? 

I expect you have to change this one: https://github.com/cognitedata/cognite-sdk-python/pull/1453/files#diff-bf125bb2f0fb50c184ac0c0d6a76ac57879eb40aa34c5ed5de9c0b39496a2db0R192

PR #1502, opened 2023-11-13
[v7] bugfix star import from `data_classes/data_modeling`

haakonvt:
## Description
```
>>> from cognite.client.data_classes.data_modeling import *
AttributeError: module 'cognite.client.data_classes.data_modeling' has no attribute 'RequiresConstraintDefinition'
```

### DIFF
--- a/cognite/client/data_classes/data_modeling/__init__.py
+++ b/cognite/client/data_classes/data_modeling/__init__.py
@@ -12,6 +12,8 @@ from cognite.client.data_classes.data_modeling.containers import (
     ContainerList,
     ContainerProperty,
     Index,
+    RequiresConstraint,
+    UniquenessConstraint,
 )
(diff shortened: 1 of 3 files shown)


cognite/client/data_classes/data_modeling/__init__.py
```diff
@@ -131,8 +131,6 @@
     "ContainerProperty",
     "Primitive",
     "CDFExternalIdReference",
-    "RequiresConstraintDefinition",
```
erlendvollset, line 134: We still want to include them, they've just been renamed. I think they're just called RequiresConstraint and UniquenessConstraint.

haakonvt: Makes sense - they were missing in the import statement as well 👍

PR #1504, opened 2023-11-13
Add test to ensure CogniteObject subclasses handle unknown args when deserializing

erlendvollset:


doctrino (approved):
I do not know if it is worth updating the changelog for theses as they are not reported bugs, and they will not be triggered unless the API add new parameters.

Leaning towards not updating the changelog as you have not done.

erlendvollset:
Yeah, agree! I think it will just add noise to the changelog.

tests/utils.py
```diff
@@ -306,6 +307,13 @@ def create_instance(self, resource_cls: type[T_Object]) -> T_Object:
         if resource_cls is DataPointSubscriptionCreate:
             # DataPointSubscriptionCreate requires either timeseries_ids or filter
             keyword_arguments.pop("filter", None)
+        if resource_cls is Query:
```
doctrino, line 310: Great to see you managed to extend this one.

PR #1505, opened 2023-11-14
Add information about python version in CONTRIBUTING

per-tol:
Failed installing using `poetry install -E all` on my mac, as the default version of python is 3.12 on my machine.


erlendvollset (approved):

erlendvollset (changes requested):
adasdf

erlendvollset (approved):
asdfasdf

PR #1512, opened 2023-11-16
add `client_credentials` param to workflow execution trigger method

haakonvt:
#### Added
- The `trigger` method of the Workflow Execution API, now accepts a `client_credentials` to allow specifying specific
  credentials to run with. Previously, the current credentials set on the CogniteClient object doing the call would be used.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/tests/tests_unit/test_docstring_examples.py
+++ b/tests/tests_unit/test_docstring_examples.py
@@ -40,7 +40,7 @@ def run_docstring_tests(module):
 
 
 @patch("cognite.client.CogniteClient", CogniteClientMock)
-@patch("os.environ", defaultdict(lambda: "value"))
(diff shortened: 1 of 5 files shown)


tests/tests_unit/test_docstring_examples.py
```diff
@@ -39,6 +40,7 @@ def run_docstring_tests(module):
 
 
 @patch("cognite.client.CogniteClient", CogniteClientMock)
+@patch("os.environ", defaultdict(lambda: "value"))
```
erlendvollset, line 43: A small comment here would be nice.

haakonvt, line 43:
```suggestion
@patch("os.environ", defaultdict(lambda: "value"))  # ensure env.var. lookups does not fail in doctests
```

PR #1514, opened 2023-11-17
Fix `rawAcl` database scope

haakonvt:
#### Fixed
- The new compare methods for capabilities in major version 7, `IAMAPI.verify_capabilities` and `IAMAPI.compare_capabilities` now works correctly for rawAcl with database scope ("all tables").

#### Removed
- Capability scopes no longer have the `is_within` method, and capabilities no longer have `has_capability`. Use the more general `IAMAPI.compare_capabilities` instead.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/iam.py
+++ b/cognite/client/_api/iam.py
@@ -22,9 +22,11 @@ from cognite.client.data_classes import (
     SessionList,
 )
 from cognite.client.data_classes.capabilities import (
+    AllScope,
     Capability,
     ProjectCapability,
     ProjectCapabilityList,
+    RawAcl,
 )
 from cognite.client.data_classes.iam import TokenInspection
 from cognite.client.utils._identifier import IdentifierSequence
@@ -46,8 +48,6 @@ ComparableCapability: TypeAlias = Union[
 
 
 def _convert_capability_to_tuples(capabilities: ComparableCapability, project: str | None = None) -> set[tuple]:
-    from cognite.client.data_classes import Group, GroupList
-
     if isinstance(capabilities, ProjectCapability):
         return ProjectCapabilityList([capabilities]).as_tuples(project)
     if isinstance(capabilities, ProjectCapabilityList):
@@ -59,7 +59,6 @@ def _convert_capability_to_tuples(capabilities: ComparableCapability, project: s
         capabilities = capabilities.capabilities or []
     elif isinstance(capabilities, GroupList):
         capabilities = [cap for grp in capabilities for cap in grp.capabilities or []]
-
     if isinstance(capabilities, Sequence):
         tpls: set[tuple] = set()
         for cap in capabilities:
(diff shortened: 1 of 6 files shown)


cognite/client/_api/iam.py
```diff
@@ -100,8 +99,10 @@ def compare_capabilities(
             project (str | None): If a ProjectCapability or ProjectCapabilityList is passed, we need to know which CDF project
                 to pull capabilities from (existing might be from several). If project is not passed, and ProjectCapabilityList
                 is used, it will be inferred from the CogniteClient used to call retrieve it via token/inspect.
-            ignore_allscope_meaning (bool): Option on how to treat allScopes. When True, this function will return
-                e.g. an Acl scoped to a dataset even if the user have the same Acl scoped to all. Defaults to False.
+            ignore_allscope_meaning (bool): Option on how to treat scopes that encompass other scopes, like allScope. When True,
```
erlendvollset, line 102: When would you want this set to True?

cognite/client/data_classes/capabilities.py
```diff
@@ -98,9 +98,6 @@ def as_tuples(self) -> set[tuple]:
             # Basic implementation for all simple Scopes (e.g. all or currentuser)
             return {(self._scope_name,)}
 
-        def is_within(self, other: Self) -> bool:
```
doctrino, line 101: Surprised that these slipped through the reviews.

tests/tests_unit/test_data_classes/test_capabilities.py
```diff
@@ -347,6 +350,40 @@ def test_partly_missing_capabilities(
...
+        has_db_scope = RawAcl([RawAcl.Action.Read], TableScope(dbs_to_tables={"db1": []}))
+        assert not cognite_client.iam.compare_capabilities(has_all_scope, has_db_scope)
+
+    @pytest.mark.parametrize(
+        "extra_existing",
+        [
+            [],
+            [RawAcl(actions=[RawAcl.Action.Read], scope=AllScope)],
+            [RawAcl(actions=[RawAcl.Action.Read], scope=TableScope({"db1": []}))],
+            [RawAcl(actions=[RawAcl.Action.Read], scope=TableScope({"db1": {"tables": []}}))],
```
doctrino, line 365: These are confusing, but if I understand it correctly, it means all in `db1`? Both of them are equal?

doctrino (approved):
Brutal edge case. A bit surprised that we forgot to remove `is_within` before the v7 release.

cognite/client/_api/iam.py
doctrino: That is the most logical to me. You have access to all data sets when you have all, which would be the most common use case.

tests/tests_unit/test_data_classes/test_capabilities.py
haakonvt: This means access is granted to all current and all future tables in the database `"db1"`:
`TableScope({"db1": []})`

I don't like the syntax one bit.

haakonvt: However, that's how the API does it, and hence we shouldn't add our own `DataBaseScope` imo

cognite/client/data_classes/capabilities.py
haakonvt: I actually kept it as a "simple 1:1 check", but decided now with the raw-database changes to not duplicate the error-prone scope logic.

cognite/client/_api/iam.py
haakonvt: If you want two groups of capabilities to be equal, and they differ in scopes "not all", then this option will tell you exactly what those differences are. For Peter's bootstrap CLI (and maybe others...) it means the manual work of removing any possible all-scopes or raw-database-scopes before comparing is not needed.

PR #1517, opened 2023-11-21
Add example on how to update the nonce for transformations

muradsater:
## Description
Extend docs for transformations update

## Checklist:
- [X] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.

### DIFF
--- a/cognite/client/_api/transformations/__init__.py
+++ b/cognite/client/_api/transformations/__init__.py
@@ -299,34 +299,28 @@ class TransformationsAPI(APIClient):
 
             Perform a partial update on a transformation, updating the query and making it private::
 
-                >>> from cognite.client import CogniteClient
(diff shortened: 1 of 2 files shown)


cognite/client/_api/transformations/__init__.py
haakonvt, line 1: I wonder if this should accept a created session as well, would make the example much easier and more intuitive. Wdyt?

Also, please add some indentation to make it easier to read 🙏

haakonvt (approved):

cognite/client/_api/transformations/__init__.py
haakonvt: Can't be done because the CreatedSession does not store a reference to the `CogniteClient`, so we don't know what `project` to pass.

PR #1519, opened 2023-11-21
Fix: url encode workflows external id

ddonukis:
## Description

Workflows service allows arbitrary strings as workflow external id or version id. However, these are used as url path parameters and should be url-encoded. Fixing it now.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,13 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.3.2] - 2023-11-21
+### Fixed
+- `workflows.retrieve` and `workflows.versions.retrieve` returned None if the provided workflow external id contained special characters. This is now fixed.
+
 ## [7.3.1] - 2023-11-21
 ### Fixed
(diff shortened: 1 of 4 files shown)


doctrino (commented):
There are a few more strings used directly in the URL in the workflows.py, 103, 139, and 206.

In addition, is this worth a simple test?

ddonukis:
> There are a few more strings used directly in the URL in the workflows.py, 103, 139, and 206.
> 
> In addition, is this worth a simple test?

The other url path params are provided by the Workflows service and are uuids, that's why I didn't update those.

Do you think they should be quoted anyways to be safe?

doctrino:
> > There are a few more strings used directly in the URL in the workflows.py, 103, 139, and 206.
> > In addition, is this worth a simple test?
> 
> The other url path params are provided by the Workflows service and are uuids, that's why I didn't update those.
> 
> Do you think they should be quoted anyways to be safe?

Ah, no I don't think it is necessary.

doctrino (approved):

haakonvt:
> In addition, is this worth a simple test?

@ddonukis 👀

PR #1526, opened 2023-11-27
Revise partitions+generator logic and move max_workers to global_config

erlendvollset:
Current implementation of partitions+generator will keep `n_partitions*chunk_size` items in memory. With this change we will
1) Ignore partitions when using generator as it really only makes sense when using `list`
2) always fetch LIST_LIMIT items when using generator, and just yield `chunk_size` items, to avoid excessive requests being fired.

On max_workers, we have moved that to global_config, but will still accept it in ClientConfig (although with a deprecatio warning).

## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -416,11 +416,9 @@ class APIClient:
         resource_path = resource_path or self._RESOURCE_PATH
         total_items_retrieved = 0
         current_limit = self._LIST_LIMIT
-        if chunk_size and chunk_size <= self._LIST_LIMIT:
-            current_limit = chunk_size
         next_cursor = initial_cursor
         filter = filter or {}
-        current_items = []
+        unprocessed_items = []
         while True:
             if limit:
                 num_of_remaining_items = limit - total_items_retrieved
@@ -463,16 +461,20 @@ class APIClient:
                 for item in last_received_items:
                     yield resource_cls._load(item, cognite_client=self._cognite_client)
             else:
-                current_items.extend(last_received_items)
-                if len(current_items) >= chunk_size:
-                    items_to_yield = current_items[:chunk_size]
-                    current_items = current_items[chunk_size:]
-                    yield list_cls._load(items_to_yield, cognite_client=self._cognite_client)
+                unprocessed_items.extend(last_received_items)
+                if len(unprocessed_items) >= chunk_size:
+                    chunks = split_into_chunks(unprocessed_items, chunk_size)
+                    if chunks and len(chunks[-1]) < chunk_size:
+                        unprocessed_items = chunks.pop(-1)
+                    else:
+                        unprocessed_items = []
+                    for chunk in chunks:
+                        yield list_cls._load(chunk, cognite_client=self._cognite_client)
 
             next_cursor = res.json().get("nextCursor")
             if total_items_retrieved == limit or next_cursor is None:
-                if chunk_size and current_items:
-                    yield list_cls._load(current_items, cognite_client=self._cognite_client)
+                if chunk_size and unprocessed_items:
+                    yield list_cls._load(unprocessed_items, cognite_client=self._cognite_client)
                 break
 
     def _list(
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -5,6 +5,7 @@ import gzip
 import json as _json
 import logging
 import re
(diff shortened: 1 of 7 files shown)


cognite/client/_api_client.py
haakonvt, line 1: This is a breaking change (unintended I guess). Previously, passing chunk_size & partitions would yield resources one-by-one, e.g.:
```
>>> client.events.aggregate_count()
55897
>>> # On master
>>> events = list(client.events(chunk_size=200, partitions=5))
>>> len(events)
55897
>>> # This PR
>>> events = list(client.events(chunk_size=200, partitions=5))
>>> len(events)
280
>>> sum(map(len, events))
55897
```

haakonvt: Imo an improvement, but still :/

haakonvt, line 1: One more comment, this time regarding speed/number of requests. I think the slowdown from no longer using multiple workers is perfectly reasonable (how it should be), but the new implementation _emits a ton of requests_.

In my test project I have 2k events, and so if I run with `chunk_size=2` and `partitions=5`, the old version sends a total of 5 requests, while the new sends 1000 requests... I think we should accept that `_LIST_LIMIT` number of resources is still considered "tiny" enough (for the purpose of RAM usage) that we fetch `_LIST_LIMIT` at the time and yield from it (which will introduce slightly annoying logic).

erlendvollset: Great catch. Fixed now!

erlendvollset: I agree that's better, it's how the generator _without_ partitions always should have behaved. Will fix!

cognite/client/_api_client.py
```diff
@@ -628,7 +581,7 @@ def get_partition(partition: int) -> list[dict[str, Any]]:
             return retrieved_items
 
         tasks = [(f"{i + 1}/{partitions}",) for i in range(partitions)]
-        tasks_summary = execute_tasks(get_partition, tasks, max_workers=partitions, fail_fast=True)
+        tasks_summary = execute_tasks(get_partition, tasks, max_workers=self._config.max_workers, fail_fast=True)
```
haakonvt, line 584: We shouldn't accept `partitions` > `max_workers` imo., won't speed up fetching and maybe cause a higher backend load (just guessing, I'm clueless about databases). What about setting `partitions = min(partitions, max_workers)`, possibly with a warning?

cognite/client/config.py
```diff
@@ -26,6 +27,7 @@ class GlobalConfig:
             Defaults to 50.
         disable_ssl (bool): Whether or not to disable SSL. Defaults to False
         proxies (Dict[str, str]): Dictionary mapping from protocol to url. e.g. {"https": "http://10.10.1.10:1080"}
+        max_workers (int | None): Max number of workers to spawn when parallelizing API calls. Defaults to 10.
```
haakonvt, line 30: Should we take the opportunity to lower this number to 5? 😏

cognite/client/config.py
```diff
@@ -53,7 +56,8 @@ class ClientConfig:
         credentials (CredentialProvider): Credentials. e.g. Token, ClientCredentials.
         api_subversion (str | None): API subversion
         base_url (str | None): Base url to send requests to. Defaults to "https://api.cognitedata.com"
-        max_workers (int | None): Max number of workers to spawn when parallelizing data fetching. Defaults to 10. Can not be changed after your first API call.
+        max_workers (int | None): DEPRECATED. use global_config.max_workers instead.
+            Max number of workers to spawn when parallelizing data fetching. Defaults to 10.
```
haakonvt, line 60:
```suggestion
        max_workers (int | None): DEPRECATED. Use global_config.max_workers instead.
            Max number of workers to spawn when parallelizing data fetching. Defaults to 10.
```

haakonvt (approved):
Lgtm 🚀

cognite/client/_api_client.py
erlendvollset: Agree. fixed!

erlendvollset: That broke something. Couldn't immediately see what. Let's follow up in another PR instead.

cognite/client/config.py
haakonvt: 👀

PR #1535, opened 2023-12-01
Fix import in docs for filters.Not

sighol:
## Description
The `Not` filter had imports for `Or` and `Equals` in the docs, when it should have `Not` and `Equals.

Since this is just docs, I don't think we need a new version of the SDK. The public docs should be re-built when this PR is merged.

## Checklist:
- [ ] Tests added/updated.


haakonvt:
How was this not picked up by the doctests?!

PR #1538, opened 2023-12-04
make subscription data calls more efficient

matiascognite:
## Description
Use a more efficient method to query the subscription list data API

The timeout parameter will enable active wait:
Until there is more data available, we will keep the request waiting in the backend. Only when the timeout has expired, we will return. 

Not a visible change to the user; no need to update version or changelog?

## Checklist:
- [ ] ~Tests added/updated.~


cognite/client/_api/datapoints_subscriptions.py
```diff
@@ -263,7 +263,7 @@ def iterate_data(
                 "externalId": external_id,
                 "partitions": [p.dump(camel_case=True) for p in current_partitions],
                 "limit": limit,
-                "timeoutSeconds": 5, # keep the request open until data is available, up to 5 seconds
+                "timeoutSeconds": 5,  # keep the request open until data is available
```
haakonvt, line 266: This seems super useful! Accept it as an argument to the function with default being 5? 🚀

matiascognite:
Thanks for review, even if I didn't ask for it @haakonvt !
Now that we actually implemented it in the backend, the tests pass, but I have some questions

- Given that we change the method signature, we should probably do a (minor) version bump?
- Is poll_timeout a good name? We have pollTimeoutSeconds in the API, but I have the impression that python defaults to seconds (correct me if I'm wrong)
- The behavior changes. If I read it correctly, it will now block while waiting for new data. Before it would just respond with an empty change immediately. It should mostly be an improvement though, and also protect against writing very heavy applications with constant polling, by accident.

haakonvt (approved):
> In java and other languages, timestamps are often measured in milliseconds. So a variable named timeout could be thought to be given in milliseconds, unless it is named timeoutSeconds. Witch is also the name in the API itself.
> However, in python, I have the impression that time is measured in seconds, unless specified otherwise.
> Anyway, if you are happy, I am happy :)

Today I learned! At least epoch time is given in float sec at least (`time.time()`). Looks very good to me!

PR #1551, opened 2023-12-12
Make `AnnotationsAPI.list` support paging also for `limit=None/inf`

haakonvt:
## Description
Awaiting confirmation from @olacognite before bumping version & change log

Ref. slack thread: https://cognitedata.slack.com/archives/CG10VQPFX/p1702389070057459

### DIFF
--- a/cognite/client/_api/annotations.py
+++ b/cognite/client/_api/annotations.py
@@ -10,7 +10,7 @@ from cognite.client.data_classes._base import CogniteResource, PropertySpec
 from cognite.client.data_classes.annotations import AnnotationReverseLookupFilter
 from cognite.client.data_classes.contextualization import ResourceReference, ResourceReferenceList
 from cognite.client.utils._identifier import IdentifierSequence
-from cognite.client.utils._text import to_camel_case
+from cognite.client.utils._text import convert_all_keys_to_camel_case
 from cognite.client.utils._validation import assert_type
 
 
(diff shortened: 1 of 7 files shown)


cognite/client/_api/annotations.py
```diff
@@ -183,18 +183,13 @@ def list(self, filter: AnnotationFilter | dict, limit: int | None = DEFAULT_LIMI
         Returns:
             AnnotationList: list of annotations
         """
-        assert_type(limit, "limit", [int], allow_none=False)
```
olacognite, line 186: Any reason for not just changing to allow_none=True?

olacognite (approved):
LGTM, but maybe add a test too?

cognite/client/_api/annotations.py
haakonvt: Yes, it would fail for `math.inf` which is valid. Also, the `_list` method handles this for us automatically.

tests/tests_integration/test_api/test_annotations.py
```diff
@@ -321,13 +324,7 @@ def test_retrieve_multiple(self, cognite_client: CogniteClient, base_annotation:
         ids = [c.id for c in created_annotations]
         retrieved_annotations = cognite_client.annotations.retrieve_multiple(ids)
         assert isinstance(retrieved_annotations, AnnotationList)
-
-        # TODO assert the order and do without sorting
-        # as soon as the API is fixed
```
doctrino, line 326: Nice fix.

cognite/client/_api/annotations.py
doctrino: So this is the actual fix? Remove the custom handling of `limit` and just pass it to `_list`?

doctrino (approved):
Nice catch on `.to_pandas(...)` as that cannot have been working at any time. I am wondering if we should create a similar test as we have `.dump()` and `.load()` to ensure that `.to_pandas()` work.

cognite/client/_api/annotations.py
haakonvt: Yup!

tests/tests_integration/test_api/test_annotations.py
haakonvt: This was unrelated actually, I just checked if the api response had been fixed 😄

PR #1566, opened 2023-12-26
Remove dependency: `sortedcontainers`

haakonvt:
## Description
After #1425 was merged, the need for auto-sorted containers is gone (we now only need 1 sort per data container). This PR replaces `SortedList` and `SortedDict`, and completely removes`sortedcontainers` from our dependencies.

## ~~TBD~~
- tests ✅
- changelog / version ✅


erlendvollset (approved):
Nice!

PR #1570, opened 2024-01-03
Cancel workflows

silvavelosa:
## Description
Just an update in documentation

## Checklist:
- [ ] Tests added/updated.


cognite/client/_api/workflows.py
haakonvt, line 1: Methods should not return generics, but be specific on the type. In this case, it should be `list[WorkflowExecution]`. Did a quick check, and `WorkflowExecutionList` exists - probably even better? (Would also need to change the load logic below).

haakonvt, line 1: All the other Workflow classes are prefixed with Workflow, so I think `CancelExecution` should be `WorkflowCancelExecution`?

haakonvt:
Closing after #1742 was merged

PR #1572, opened 2024-01-04
Make `cdf_project_name` required in `OidcCredentials` (transformations)

haakonvt:
### Changed
- When using `OidcCredentials` to create a transformation, `cdf_project_name` is no longer optional as required
  by the API.


doctrino (approved):
Nice catch.

PR #1573, opened 2024-01-04
Handle the isAutoRetryable flag when retrying

erlendvollset:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_http_client.py
+++ b/cognite/client/_http_client.py
@@ -80,7 +80,7 @@ class _RetryTracker:
         backoff_time_adjusted = self._max_backoff_and_jitter(backoff_time)
         return backoff_time_adjusted
 
-    def should_retry(self, status_code: int | None) -> bool:
+    def should_retry(self, status_code: int | None, is_auto_retryable: bool = False) -> bool:
         if self.total >= self.config.max_retries_total:
             return False
         if self.status > 0 and self.status >= self.config.max_retries_status:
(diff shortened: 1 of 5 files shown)


cognite/client/_http_client.py
```diff
@@ -110,23 +110,28 @@ def __init__(
...
+                except Exception:
+                    # if the response is not JSON or it doesn't conform to the api design guide,
+                    # we assume it's not auto-retryable
+                    is_auto_retryable = False
+                if not retry_tracker.should_retry(status_code=res.status_code, is_auto_retryable=is_auto_retryable):
                     return res
             except CogniteReadTimeout as e:
                 retry_tracker.read += 1
-                if not retry_tracker.should_retry(status_code=last_status):
+                if not retry_tracker.should_retry(status_code=None, is_auto_retryable=True):
```
doctrino, line 130: Isn't this a change of behavior? It used to retry depending on the latest error, while now it will always retry?

doctrino (approved):
Just a question

cognite/client/_http_client.py
erlendvollset: Shouldn't be any change in behaviour, but the code used to be a little misleading. 

It would use the latest status code yes, but that status code will always be retryable, otherwise an error would have already been raised. So in the context of a read timeout/connection error, status_code may aswell always be set to null here.

tests/tests_integration/test_api/test_documents.py
```diff
@@ -219,6 +219,7 @@ def test_iterate_over_text_documents(self, cognite_client: CogniteClient):
             assert text_doc.mime_type == "text/plain"
             break
 
+    @pytest.mark.skip(reason="Currently broken in staging")
```
haakonvt, line 222: Works again now btw

PR #1604, opened 2024-01-31
add retry to annotations, make `annotations/list` auto-chunk on `annotatedResourceIds`

haakonvt:
## Description
Currently, when the number of `annotatedResourceIds` exceeds the API limit (1000), the user of `list` needs to chunk ids themselves and then call `list` in a loop, then recombine results. That is now taken care of automatically.

Additionally, it seems that the annotations requests are not being retried due to a missing regex for the `POST` endpoint for it. 

### Fixed
- Calls to `AnnotationsAPI.[list|retrieve|retrieve_multiple]` are now retried automatically.
### Improved
- Calls to `AnnotationsAPI.list` with more than 1000 `annotated_resource_ids` are now batched automatically for the user.
  Previously these would raise an API error.

## Checklist:
- [x] Tests added/updated.


haakonvt:
To be added, an entry `_RETRYABLE_POST_ENDPOINT_REGEX_PATTERNS` for `annotations`:
FAILED tests/tests_unit/test_api_client.py::TestRetryableEndpoints::test_is_retryable[POST-https://api.c.com/api/v1/projects/bla/annotations/list-True] - AssertionError: assert True == False

olacognite (approved):
LGTM!

PR #1611, opened 2024-02-04
fixed_labels_on_file_upload

CogniteJohan:
## Description
From v7.0.0 it has not been possible to call client.files.upload with Labels. This PR is a minimal change to solve this problem quickly, a more thorough investigation about other possible consequences of the dump-methods should be done in a separate PR.

The main problem comes from the dump method used in FilesAPI._upload_file_from_path where the methond upload_bytes is called, dumping FileMetadata with camel_case=False. This fails in Label._load_list where "ExternalId" is in CamelCase. The instance checks in _load_list can probably be removed. Other consequences of the dump method and camel-case vs snake-case can also potentially be an issue.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.17.2] - 2024-02-04
+### Fixed
+- Uploading files will now accept Labels as part of file Metadata.
+This has not been possible after v 7.0.0.
+
 ## [7.17.1] - 2024-02-02
 ### Fixed
 - An (extreme) edge case where an empty, unnecessary API request for datapoints would be sent leading to a `CogniteAPIError`.
--- a/tests/tests_unit/test_api/test_files.py
+++ b/tests/tests_unit/test_api/test_files.py
@@ -429,6 +429,10 @@ class TestFilesAPI:
         path = os.path.join(os.path.dirname(__file__), "files_for_test_upload", "file_for_test_upload_1.txt")
         cognite_client.files.upload(path, external_id="blabla", name="bla", data_set_id=42)
 
+    def test_upload_with_label(self, cognite_client, mock_file_upload_response):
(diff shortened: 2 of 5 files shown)


tests/tests_unit/test_api/test_files.py
```diff
@@ -429,6 +429,10 @@ def test_upload_with_external_id(self, cognite_client, mock_file_upload_response
         path = os.path.join(os.path.dirname(__file__), "files_for_test_upload", "file_for_test_upload_1.txt")
         cognite_client.files.upload(path, external_id="blabla", name="bla", data_set_id=42)
 
+    def test_upload_with_label(self, cognite_client, mock_file_upload_response):
+        path = os.path.join(os.path.dirname(__file__), "files_for_test_upload", "file_for_test_upload_1.txt")
+        cognite_client.files.upload(path, external_id="blabla", name="bla", data_set_id=42, labels=[Label("PUMP")])
```
haakonvt, line 434: Let's add a note here about why/when this broke (versions affected). (Could also consider not passing anything but `name` & `label`)

(`ValueError: Could not parse label: {'external_id': 'foo-bar'}`)

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.17.2] - 2024-02-04
+### Fixed
+- Uploading files will now accept Labels as part of file Metadata.
+This has not been possible after v 7.0.0.
```
haakonvt, line 23: Suggestion:
```suggestion
## [7.17.2] - 2024-02-04
### Fixed
- Uploading files now accepts Labels again as part of file metadata. This addresses a bug introduced in v7
  (that ironically made `dump` work correctly).
```

haakonvt (approved):
Good catch! Approved with a few suggestions

CHANGELOG.md
CogniteJohan: Something like this?

PR #1614, opened 2024-02-05
fix race conditions in tests that uses the same identifiers

haakonvt:
… for the  (win+linux test runners)

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/tests/tests_integration/test_api/test_data_modeling/test_containers.py
+++ b/tests/tests_integration/test_api/test_data_modeling/test_containers.py
@@ -40,15 +40,11 @@ class TestContainersAPI:
     def test_list(
         self, cognite_client: CogniteClient, movie_containers: ContainerList, integration_test_space: Space
     ) -> None:
-        # Arrange
         expected_containers = ContainerList([c for c in movie_containers if c.space == integration_test_space.space])
         expected_ids = set(expected_containers.as_ids())
-
-        # Act
-        actual_containers = cognite_client.data_modeling.containers.list(space=integration_test_space.space, limit=-1)
-
-        # Assert
         assert expected_ids, "The movie model is missing containers"
+
+        actual_containers = cognite_client.data_modeling.containers.list(space=integration_test_space.space, limit=-1)
         assert expected_ids <= set(actual_containers.as_ids())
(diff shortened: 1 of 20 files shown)


haakonvt:
These keep failing, e.g. a day ago: https://github.com/cognitedata/cognite-sdk-python/actions/runs/7782388421/job/21218699237

tests/tests_integration/test_api/test_data_modeling/test_containers.py
```diff
@@ -56,7 +57,7 @@ def test_apply_retrieve_and_delete(self, cognite_client: CogniteClient, integrat
         # Arrange
         new_container = ContainerApply(
             space=integration_test_space.space,
-            external_id="IntegrationTestContainer",
+            external_id="IntegrationTestContainer" + random_string(5),
```
erlendvollset, line 60: We don't want to create a new container for each test, this delete + recreate containers pattern that really only happens in tests has a bunch of undesireable effects on the system (e.g. elasticsearch mapping explosion). Let's keep it as a fixed externalId and remove the cleanup below.

erlendvollset (changes requested):

tests/tests_integration/test_api/test_data_modeling/test_containers.py
haakonvt: Updated now! Not much was left of the test after, so please have another look. There's also two other tests that seem to create+delete containers: `test_apply_failed_and_successful_task` and `test_load_and_create_only_required_fields`. I've left them as is - but let me know if you want them changed/removed as well 👌

haakonvt: ...I also... did some... comment deletion 👀 Don't tell Anders 🤭

erlendvollset: Yeah, let's update those aswell 👍

haakonvt: Done, without the create+delete, it was already covered by other tests, so I deleted both.

erlendvollset (approved):
Lgtm!

PR #1616, opened 2024-02-06
Fix geo_location handling for documents

tommythorsen:
## Description
The geo_location fields in the Documents API differ from those in the Files API because they contain the Geometry objects directly, without a Feature object to wrap them. The Documents API also supports GeometryCollection objects in addition to the standard set of Geometry types.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/documents.py
+++ b/cognite/client/data_classes/documents.py
@@ -27,7 +27,7 @@ class DocumentsGeoJsonGeometry(CogniteObject):
     """Represents the points, curves and surfaces in the coordinate space.
 
     Args:
-        type (Literal['Point', 'MultiPoint', 'LineString', 'MultiLineString', 'Polygon', 'MultiPolygon', 'GeometryCollection']): The geometry type.
+        type (Literal["Point", "MultiPoint", "LineString", "MultiLineString", "Polygon", "MultiPolygon", "GeometryCollection"]): The geometry type.
         coordinates (list | None): An array of the coordinates of the geometry. The structure of the elements in this array is determined by the type of geometry.
         geometries (Collection[Geometry] | None): No description.
 
--- a/cognite/client/data_classes/documents.py
+++ b/cognite/client/data_classes/documents.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, List, Literal, Union, cast
+from typing import TYPE_CHECKING, Any, Collection, List, Literal, Union, cast
 
 from typing_extensions import Self, TypeAlias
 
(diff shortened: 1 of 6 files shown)


tommythorsen:
Running lint locally does not give me any changes, so I'm a bit stuck on fixing that bit. If some kind soul with a working setup could check this branch out and run lint for me, that would be fantastic

cognite/client/data_classes/documents.py
```diff
@@ -36,7 +125,7 @@ class SourceFile(CogniteObject):
         size (int | None): The size of the file in bytes.
         asset_ids (list[int] | None): The ids of the assets related to this file.
         labels (list[Label | str | LabelDefinition] | None): A list of labels associated with this document's source file in CDF.
-        geo_location (GeoLocation | None): The geolocation of the source file.
+        geo_location (DocumentsGeoJsonGeometry | None): The geolocation of the source file.
```
haakonvt, line 128: I feel like we are missing a class here: The old usage of `GeoLocation` should be replaced (for Document/-List/SourceFile etc.) by a new `DocumentsGeoLocation` class. This in turn should accept a collection of `DocumentGeoJsonGeometry`  (instead of `Geometry`) for `geometries`.

tommythorsen: The `geo_location` field _is_ a geometry object. It can be one of the standard geometry types, such as Point or Polygon, or the special geometry type `GeometryCollection`, which is the only case the `geometries` field should be used.

haakonvt: Oh, does that mean the `GeoLocation` is redundant/should be removed?

tommythorsen: It looks like it is still used for other endpoints such as assets and files, so I don't think it can be removed.

In terms of the geojson spec, the `GeoLocation` object maps to the [Feature](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2) type, which is some kind of a wrapper object around Geometries. The Documents API uses the [Geometry](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2) objects directly.

An alternative thing I could have done instead of making the new `DocumentsGeoJsonGeometry` object, was to use the existing `Geometry` class directly. The existing `Geometry` class is a bit weird, because it explicitly does not support the `GeometryCollection` type, but it has a `geometries` field, which I don't think is actually used for anything at all. But I did not want to properly add `GeometryCollection` support to this class, since other services don't support it.

Another thing I probably could have done, was to make the `geo_location` field be a union type of `Geometry | GeometryCollection | None`, where `GeometryCollection` could be a new class that has the `geometries` field and type=`GeometryCollection`. Whether this will work or not, depends on the serialization/deserialization capabilities of the SDK, which I'm not so familiar with.

doctrino (commented):
I think I understand this one. I guess it could extracted a shared base class for `DocumentsGeoJsonGeometry` and `Geometry`. However, there are differences

* coordinates is optional in `DocumentsGeoJsonGeometry` and required in `Geometry`.
* The type `GeometryCollection` is allowed in `DocumentsGeoJsonGeometry` but not in `Geometry`.

Thus I agree to keep these separate.

cognite/client/data_classes/documents.py
```diff
@@ -11,18 +11,107 @@
...
 from cognite.client.data_classes.aggregations import UniqueResult
 from cognite.client.data_classes.labels import Label, LabelDefinition
-from cognite.client.data_classes.shared import GeoLocation
 from cognite.client.utils._text import convert_all_keys_to_snake_case
 
 if TYPE_CHECKING:
     from cognite.client import CogniteClient
 
 
+class DocumentsGeoJsonGeometry(CogniteObject):
```
doctrino, line 26: This seems like a class that could be use outside of the documents API. Could it be called `GeometryCollection` instead?
```suggestion
class GeometryCollection(CogniteObject):
```

cognite/client/data_classes/documents.py
```diff
@@ -11,18 +11,107 @@
...
+    """Represents the points, curves and surfaces in the coordinate space.
+
+    Args:
+        type (Literal["Point", "MultiPoint", "LineString", "MultiLineString", "Polygon", "MultiPolygon", "GeometryCollection"]): The geometry type.
+        coordinates (list | None): An array of the coordinates of the geometry. The structure of the elements in this array is determined by the type of geometry.
+        geometries (Collection[Geometry] | None): No description.
```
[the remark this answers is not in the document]
tommythorsen: It's a list of geometry objects (as defined here: https://datatracker.ietf.org/doc/html/rfc7946#section-3.1)

It is only relevant if the type is `GeometryCollection`

haakonvt (approved):
I think I finally understand the structure now after playing around attaching random geoLocation "types" to my files, then seeing what gets returned 😅 I rebased on master now btw

PR #1621, opened 2024-02-13
Fix default json encoder

erlendvollset:
default json encoder would just use `.__dict__()`, which won't work most of the time, so this change makes the encoder understand the CogniteObject and CogniteFilter interfaces, so that they will invoke `.dump()` instead.

### DIFF
--- a/cognite/client/data_classes/datapoints.py
+++ b/cognite/client/data_classes/datapoints.py
@@ -847,7 +847,7 @@ class DatapointsList(CogniteResourceList[Datapoints]):
         item = self.dump()
         for i in item:
             i["datapoints"] = convert_and_isoformat_time_attrs(i["datapoints"])
-        return _json.dumps(item, default=lambda x: x.__dict__, indent=4)
+        return _json.dumps(item, indent=4)
 
     def to_pandas(  # type: ignore [override]
         self,
--- a/cognite/client/data_classes/datapoints.py
+++ b/cognite/client/data_classes/datapoints.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import json
 import operator as op
 import typing
 import warnings
@@ -20,6 +19,7 @@ from typing import (
 )
 
 from cognite.client.data_classes._base import CogniteResource, CogniteResourceList
+from cognite.client.utils import _json
 from cognite.client.utils._auxiliary import find_duplicates, no_op
 from cognite.client.utils._identifier import Identifier
 from cognite.client.utils._importing import local_import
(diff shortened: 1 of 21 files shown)


cognite/client/data_classes/datapoints.py
```diff
@@ -847,7 +847,7 @@ def __str__(self) -> str:
         item = self.dump()
         for i in item:
             i["datapoints"] = convert_and_isoformat_time_attrs(i["datapoints"])
-        return json.dumps(item, default=lambda x: x.__dict__, indent=4)
+        return _json.dumps(item, default=lambda x: x.__dict__, indent=4)
```
haakonvt, line 850: I think we can remove this: `default=lambda x: x.__dict__`

tests/tests_unit/test_utils/test_json.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @pytest.mark.dsl
+    def test_json_serializable_numpy(self):
+        np = local_import("numpy")
+        arr = np.array([1.2, 3.4], dtype=np.float32)
+        with pytest.raises(TypeError):
+            json.dumps(arr)
+        with pytest.raises(TypeError):
+            json.dumps(arr[0])
+        with pytest.raises(TypeError):  # core sdk makes it hard to serialize np.ndarray
```
haakonvt, line 32: Not sure who wrote this comment originally ...but this can be trivially fixed with another instance check in `_default_json_encoder` (not that I see a use case for it)

tests/tests_unit/test_utils/test_json.py
```diff
@@ -0,0 +1,52 @@
...
+        with pytest.raises(TypeError):
+            json.dumps(Decimal(1))
+
+        assert _json.dumps(Decimal(1))
+
+    def test_json_not_serializable_sets(self):
+        with pytest.raises(TypeError):
+            json.dumps({1, 2})
+        with pytest.raises(TypeError):
+            json.dumps({1, 2})
```
haakonvt, line 22:
```suggestion
        with pytest.raises(TypeError):
            _json.dumps({1, 2})
```

tests/tests_unit/test_utils/test_json.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @pytest.mark.dsl
+    def test_json_serializable_numpy(self):
+        np = local_import("numpy")
+        arr = np.array([1.2, 3.4], dtype=np.float32)
+        with pytest.raises(TypeError):
+            json.dumps(arr)
+        with pytest.raises(TypeError):
+            json.dumps(arr[0])
```
haakonvt, line 31: `json` or `_json` here? 🤔

haakonvt (approved):
Very nice! A quick search for uses of dump reveals 400+ uses of `.dump(camel_case...`, e.g.:
```py
if isinstance(filter, CogniteFilter):
    dumped_filter = filter.dump(camel_case=True)
```
that no longer are required. Keep/prune?

cognite/client/data_classes/datapoints.py
erlendvollset: Nice. Will try that!

PR #1624, opened 2024-02-16
Make property access simpler on instances

haakonvt:
### Added
- Make properties on instances (`Node`, `Edge`) easier to work with, by implementing support for direct indexing (and a `.get` method).
  If the instances have properties from no source or multiple sources, an error is raised instead. Example usage: `instance["my_prop"]`
  (short-cut for: `instance.properties[ViewId("space", "ext.id", "version")]["my_prop"]`)

## TODO
- [x] implement item access `[]` (`__getitem__`++) on `Instance`
- [X] ~~implement attribute item access `__getattr__`++ on `Instance`~~ reverted
- [ ] ~~Add great examples to appropriate methods to showcase the simplified property access!~~ Did not find a good place to write such examples 😢 
- [x] Tests added/updated.
- [x] Changelog updated.
- [x] Version bumped.

### DIFF
--- a/tests/tests_unit/test_api/test_data_modeling/test_instances.py
+++ b/tests/tests_unit/test_api/test_data_modeling/test_instances.py
@@ -1,13 +1,9 @@
 from __future__ import annotations
 
-from typing import Any
-
 import pytest
 
 from cognite.client._api.data_modeling.instances import InstancesAPI
-from cognite.client.data_classes.data_modeling import Edge, Node
 from cognite.client.data_classes.data_modeling.ids import ViewId
-from cognite.client.data_classes.data_modeling.instances import Instance
 from tests.tests_unit.test_api.test_data_modeling.conftest import make_test_view
 
 SINGLE_SRC_DUMP = {"source": {"space": "a", "externalId": "b", "version": "c", "type": "view"}}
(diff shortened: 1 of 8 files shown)


tests/tests_unit/test_api/test_data_modeling/test_instances.py
```diff
@@ -33,3 +39,116 @@ def test_instances_api_dump_instance_source(sources, expected):
...
+        "version": 4,
+        "lastUpdatedTime": 5,
+        "createdTime": 6,
+        "type": {"space": "foo-space", "externalId": "bar-xid"},
+        "startNode": {"space": "the-space", "externalId": "the-xid"},
+        "endNode": {"space": "the-space", "externalId": "the-xid"},
+    }
+
+
+@pytest.mark.parametrize("instance_type", (Node, Edge))
```
doctrino, line 69: I would expect these test to be in the `tests/tests_unit/test_data_classes/test_data_models/test_instances.py`

tests/tests_unit/test_api/test_data_modeling/test_instances.py
```diff
@@ -33,3 +39,116 @@ def test_instances_api_dump_instance_source(sources, expected):
...
+
+    assert hasattr(inst, "id") is False
+    assert hasattr(inst, "space") is True
+    assert inst.space == "space"
+    inst.space = "more-space"
+    assert inst.space == "more-space"
+
+    # Property should work fine with both access/set/delete protocols:
+    assert inst.prop1 == 1
+    assert inst["prop1"] == 1
```
doctrino, line 94: This is more of a personal preference, but I am typically not a big fan of overwriting `__set_attr__`. I know this is done for example in pandas, but it tend to confuse what is an attribute with what is actually data.

doctrino (commented):
In addition, I suggest writing a `def _repr_html_(self) -> str:` method for the `Properties` class. As one of the challenges when working with instances in a notebook is just to see which properties are available.

tests/tests_unit/test_api/test_data_modeling/test_instances.py
haakonvt: Agreed, reverted to just "dict" syntax that exclusively targets properties 👍

haakonvt:
> In addition, I suggest writing a `def _repr_html_(self) -> str:` method for the `Properties` class. As one of the challenges when working with instances in a notebook is just to see which properties are available.

Agree here as well, but seems appropriate to move to a separate PR.

doctrino (approved):
Nice :)

PR #1630, opened 2024-02-19
Replaced c = CogniteClient() with client = CogniteClient

andeplane:
## Description
In the Python SDK documentation, we usually have `c = CogniteClient()` as the example to initiate the SDK client. However, shortening this from `client` to `c` has some consequences. One is that it's often challenging to read code with such variable names, in addition to AI being worse at code generation with non descriptive variable names.

In the AI team we work on teaching LLMs how to write code with Python SDK and rely heavily on the SDK documentation where we'd like to use `client = ` instead of `c = ` to get a stronger signal on that variable for the LLM. Human users will likely also benefit from this.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "cognite-sdk"
 
-version = "7.20.2"
+version = "7.20.1"
 description = "Cognite Python SDK"
 readme = "README.md"
 documentation = "https://cognite-sdk-python.readthedocs-hosted.com"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "cognite-sdk"
 
-version = "7.20.1"
+version = "7.20.2"
 description = "Cognite Python SDK"
 readme = "README.md"
 documentation = "https://cognite-sdk-python.readthedocs-hosted.com"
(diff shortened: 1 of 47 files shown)


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "cognite-sdk"
 
-version = "7.20.1"
+version = "7.20.2"
```
doctrino, line 4: Not necessary?

doctrino (approved):

pyproject.toml
andeplane: I'd like the docs to be published for us to pick it up for AI consumption (we use the deployed SDK as documentation source).

haakonvt:
Very nice, please update: `scripts/generate_code_snippets.py`

pyproject.toml
haakonvt: Will be updated without version bump

andeplane:
> Very nice, please update: `scripts/generate_code_snippets.py`

I don't understand what this does. Any hints?

andeplane:
> match code to the official documentation based on the links/urls in the docstrings. You can u

What do I need to update? It seems to already cover `client = CogniteClient()`

PR #1642, opened 2024-02-23
Data Modeling Units

doctrino:
## Description

Add unit support to data modeling. **Note** Taking the `includeTyping` parameter to a different PR in interest of keeping it small.

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/data_modeling/instances.py
```diff
@@ -163,7 +163,13 @@ def __call__(
...
         limit: int | None = None,
         include_typing: bool = False,
-        sources: list[ViewIdentifier] | ViewIdentifier | None = None,
+        sources: ViewIdentifier
+        | Sequence[ViewIdentifier]
+        | View
+        | Sequence[View]
+        | SourceDef
+        | Sequence[SourceDef]
+        | None = None,
```
haakonvt, line 172: Is it time to introduce a type alias?

cognite/client/_api/data_modeling/instances.py
```diff
@@ -707,6 +752,7 @@ def search(
             query (str): Query string that will be parsed and used for search.
             instance_type (Literal["node", "edge"]): Whether to search for nodes or edges.
             properties (list[str] | None): Optional array of properties you want to search through. If you do not specify one or more properties, the service will search all text fields within the view.
+            target_units (list[PropertyUnitReference] | None): Properties to convert to another unit. The API can only convert to another unit, if a unit has been defined as part of the type on the underlying container being queried.
```
haakonvt, line 755: The comma here does not belong (5 other occurrence):
```suggestion
            target_units (list[PropertyUnitReference] | None): Properties to convert to another unit. The API can only convert to another unit if a unit has been defined as part of the type on the underlying container being queried.
```

cognite/client/data_classes/data_modeling/data_types.py
```diff
@@ -125,23 +125,58 @@ class Json(ListablePropertyType):
...
+    source_unit: str | None = None
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        output = {"externalId" if camel_case else "external_id": self.external_id}
+        if self.source_unit:
+            output["sourceUnit" if camel_case else "source_unit"] = self.source_unit
+        return output
+
+    @classmethod
+    def load(cls, data: dict) -> UnitReference:
```
haakonvt, line 140: Should this be?
```suggestion
    def _load(cls, data: dict) -> UnitReference:
```

cognite/client/data_classes/data_modeling/views.py
```diff
@@ -863,3 +866,88 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+    def from_view_id(cls, view_id: ViewId, target_units: list[PropertyUnitReference] | None = None) -> SourceDef:
+        return cls(
+            space=view_id.space,
+            external_id=view_id.external_id,
+            version=view_id.version,
+            target_units=target_units or [],
+        )
+
+    @classmethod
+    def load(cls, data: dict | SourceDef | tuple[str, str] | tuple[str, str, str] | ViewId | View) -> SourceDef:
```
haakonvt, line 917: Same question here, I think I forgot when we use private vs public load 😅

tests/tests_integration/test_api/test_data_modeling/test_containers.py
```diff
@@ -121,3 +123,26 @@ def test_dump_json_serialize_load(self, movie_containers: ContainerList) -> None
...
+
+            retrieved = cognite_client.data_modeling.containers.retrieve(created.as_id())
+
+            assert retrieved is not None
+            pressure_type = retrieved.properties["pressure"].type
+            assert isinstance(pressure_type, Float64)
+            assert pressure_type.unit is not None
+            assert pressure_type.unit.external_id == "pressure:bar"
+        finally:
+            cognite_client.data_modeling.containers.delete(unit_container.as_id())
```
haakonvt, line 148: @erlendvollset made some good points about why we don't want to create+destroy containers in our tests. Maybe it would suffice to move the creation to a fixture with the "create if missing" pattern, then use it for testing?

tests/tests_integration/test_api/test_data_modeling/test_instances.py
```diff
@@ -530,6 +556,66 @@ def test_query_oscar_winning_actors_before_2000(
         assert all(actor.properties.get(actor_id, {}).get("wonOscar") for actor in result["actors"])
         assert result["actors"] == sorted(result["actors"], key=lambda x: x.external_id)
 
+    def test_apply_retrieve_list_search_aggregate_in_units(
+        self, cognite_client: CogniteClient, integration_test_space: Space, unit_view: View
```
haakonvt, line 560: That is one long test 😅  (Maybe consider splitting up, idk)

tests/tests_integration/test_api/test_data_modeling/test_instances.py
```diff
@@ -530,6 +556,66 @@ def test_query_oscar_winning_actors_before_2000(
...
+
+        source = SourceDef.from_view_id(
+            unit_view.as_id(), [PropertyUnitReference("pressure", UnitReference("pressure:pa"))]
+        )
+
+        created = cognite_client.data_modeling.instances.apply(node, replace=True)
+
+        retrieved = cognite_client.data_modeling.instances.retrieve(created.nodes.as_ids(), sources=[source])
+        assert retrieved.nodes
+        assert abs(retrieved.nodes[0]["pressure"] - 1.1 * 1e5) < 1e-5
```
haakonvt, line 581: I prefer not doing float compares manually 😜 
```suggestion
        assert math.isclose(retrieved.nodes[0]["pressure"], 1.1 * 1e5)
```

tests/tests_integration/test_api/test_data_modeling/test_instances.py
```diff
@@ -530,6 +556,66 @@ def test_query_oscar_winning_actors_before_2000(
...
+
+        is_node = filters.Prefix(["node", "externalId"], node.external_id)
```
haakonvt, line 583: Just in case someone (me) would add another node with `node.external_id + "-troll"`
```suggestion
        is_node = filters.Equals(["node", "externalId"], node.external_id)
```

haakonvt (commented):
Damn, this was implemented fast! Good stuff, mostly nitpicking from me

cognite/client/data_classes/data_modeling/data_types.py
doctrino: For example, `DirecRelationReference` which is a very similar class also has a public load, I don't see why it should be private

cognite/client/data_classes/data_modeling/views.py
doctrino: We typically use private load when inheriting from `CogniteObject` which then expose a public load that assumes a private load method existing. But if you don't inherit from a `CogniteObject` I think it should be a public load method.

haakonvt: Ah, right, makes sense 👌

cognite/client/data_classes/data_modeling/views.py
```diff
@@ -863,3 +866,72 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        return cls(
+            property=data["property"],
+            unit=UnitReference.load(data["unit"])
+            if "externalId" in data["unit"]
+            else UnitSystemReference.load(data["unit"]),
+        )
+
+
+@dataclass(frozen=True)
+class SourceDef:
```
erlendvollset, line 890: Looks like we already have `SourceSelector`. let's just use that instead.. Will need to introduce a new superclass for that though since `properties` shouldn't be supplied when used in context of `list`.

so smth like
```
@dataclass
class SourceSelectorWithoutProperties(CogniteObject):
   source: ViewId
   target_units: list[PropertyUnit]

@dataclass
class SourceSelector(SourceSelectorWithoutProperties):
   properties: list[str]
```

cognite/client/data_classes/data_modeling/views.py
```diff
@@ -863,3 +866,72 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
             output["connection_type"] = "multi_reverse_direct_relation"
 
         return output
+
+
+@dataclass
+class PropertyUnitReference:
```
erlendvollset, line 872: I would suggest this name instead.
```suggestion
class PropertyUnit:
```

cognite/client/_api/data_modeling/instances.py
```diff
@@ -92,6 +94,8 @@
 
 logger = logging.getLogger(__name__)
 
+Source: TypeAlias = Union[SourceDef, View, ViewId, Tuple[str, str], Tuple[str, str, str]]
```
erlendvollset, line 97: Won't this be sufficient?
```suggestion
Source: TypeAlias = Union[SourceDef, View, ViewIdentifier]
```

cognite/client/data_classes/data_modeling/data_types.py
```diff
@@ -125,23 +125,58 @@ class Json(ListablePropertyType):
...
+
+    @classmethod
+    def load(cls, data: dict) -> UnitSystemReference:
+        return cls(unit_system_name=data["unitSystemName"])
+
+
 @dataclass
 class ListablePropertyTypeWithUnit(ListablePropertyType, LoadablePropertyType, ABC):
-    unit: NodeId | None = None
+    unit: UnitReference | None = None
```
erlendvollset, line 161: Breaking change?

cognite/client/data_classes/data_modeling/views.py
```diff
@@ -863,3 +866,83 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
             output["connection_type"] = "multi_reverse_direct_relation"
 
         return output
+
+
+@dataclass
+class PropertyUnit(CogniteObject):
```
erlendvollset, line 872: Is TargetUnit maybe a better name? The parameter in the API is named targetUnits.

cognite/client/_api/data_modeling/instances.py
```diff
@@ -92,6 +94,8 @@
 
 logger = logging.getLogger(__name__)
 
+Source: TypeAlias = Union[SourceDef, View, ViewIdentifier]
```
erlendvollset, line 97: I think I was being ambiguous in my other comment. IMO this should accept a SourceSelector instead of a SourceDef. Do we really need SourceDef at all?

cognite/client/data_classes/data_modeling/views.py
```diff
@@ -863,3 +866,83 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+
+class SourceDef(SourceCore):
+    @classmethod
+    def _load(
+        cls,
+        data: dict | SourceDef | tuple[str, str] | tuple[str, str, str] | ViewId | View,
+        cognite_client: CogniteClient | None = None,
+    ) -> SourceDef:
+        if isinstance(data, dict):
+            if "source" in data:
```
erlendvollset, line 916: This could use a small comment explaining _why_.

cognite/client/data_classes/data_modeling/views.py
```diff
@@ -863,3 +866,83 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+    def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> PropertyUnit:
+        return cls(
+            property=resource["property"],
+            unit=UnitReference.load(resource["unit"])
+            if "externalId" in resource["unit"]
+            else UnitSystemReference.load(resource["unit"]),
+        )
+
+
+class SourceCore(CogniteObject):
```
erlendvollset, line 889: I don't think this serves much purpose as is. Why not make this the superclass of SourceSelector?

erlendvollset (changes requested):

cognite/client/data_classes/data_modeling/data_types.py
doctrino: Looked into this. This was from the last implementation of Units which was quickly removed. So I would say that this is technically a bug now which has been fixed by setting it correctly.

cognite/client/data_classes/data_modeling/views.py
```diff
@@ -863,3 +865,21 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
             output["connection_type"] = "multi_reverse_direct_relation"
 
         return output
+
+
+@dataclass
+class TargetUnit(CogniteObject):
```
erlendvollset, line 871: Seems kinda randomly placed in this module? It's only used in InstancesAPI, so I'd suggest putting it in data_classes/instances.py

cognite/client/data_classes/data_modeling/data_types.py
erlendvollset: sgtm.

erlendvollset (approved):
Looks good. I would move that TargetUnit class though.

PR #1657, opened 2024-03-04
Basic methods for multipart file upload

einarmo:
## Description
Add basic methods for multipart file upload, required to support files larger than 5 GB.

This can be augmented with more opinionated/specific methods in the future, but any such methods require more knowledge about what is being uploaded, our usecase operates on some very basic streams, so no matter what we will need access to the bare methods.

In the future it might be nice to add methods specifically for uploading files on disk, or more generally, a method taking a callable returning a seekable stream.

## Checklist:
- [x] Tests added/updated.


pyproject.toml
```diff
@@ -14,6 +14,9 @@ license = "Apache-2.0"
 
 packages = [{ include="cognite", from="." }]
 
+[tool.ruff]
+line-length = 120
```
einarmo, line 18: Objections to this, or any reason why ruff isn't configured here? My IDE explodes the files if it doesn't find any line length config in pyproject. I can work around it, but it's convenient to avoid that...

cognite/client/_api/files.py
```diff
@@ -598,6 +598,151 @@ def upload_bytes(
...
+        Examples:
+
+            Upload binary data in two chunks
+
+                >>> from cognite.client import CogniteClient
+                >>> client = CogniteClient()
+                >>> res = client.files.begin_multipart_upload("my_file.txt", parts=2)
+                >>> file_metadata = res[0]
+                >>> upload_urls = res[1]
+                >>> upload_id = res[2]
```
einarmo, line 658: Doctest doesn't understand tuple unpacking apparently? I'm open for suggestions.

pyproject.toml
haakonvt: It's defined in `.pre-commit-config.yaml`

cognite/client/_api/files.py
haakonvt: You could return an object, e.g. `MultiPartUpload`? Or add `  # doctest: +SKIP`

pyproject.toml
einarmo: I'm aware, but this is more general than defining it as an argument in pre-commit.

einarmo: Alright. I really think that's just wrong, but I'll remove it.

cognite/client/_api/files.py
einarmo: Do you prefer an object here? I can do that.

erlendvollset: object is extendable. definitely prefer that.

einarmo: I changed it to an object now.

cognite/client/data_classes/files.py
```diff
@@ -466,3 +466,18 @@ class FileMetadataList(WriteableCogniteResourceList[FileMetadataWrite, FileMetad
     def as_write(self) -> FileMetadataWriteList:
         """Returns this FileMetadataList in its writing format."""
         return FileMetadataWriteList([item.as_write() for item in self.data], cognite_client=self._get_cognite_client())
+
+
+class FileMultipartUploadInit:
```
erlendvollset, line 471: Would FileMultipartUploadSession be a better name?

cognite/client/_api/files.py
```diff
@@ -596,6 +597,144 @@ def upload_bytes(
...
+            timeout=self._config.file_transfer_timeout,
+            headers=headers,
+        )
+        if not upload_response.ok:
+            raise CogniteFileUploadError(
+                message=upload_response.text,
+                code=upload_response.status_code,
+            )
+
+    def complete_multipart_upload(self, file_id: int, upload_id: str) -> None:
```
erlendvollset, line 726: This could just accept the Init/Session object instead.

erlendvollset (commented):
How about making this a context manager which yields the Session/Init object instead?

einarmo:
> How about making this a context manager which yields the Session/Init object instead?

Like using `__enter__` and `__exit__`?

cognite/client/_api/files.py
```diff
@@ -596,6 +597,143 @@ def upload_bytes(
...
+            timeout=self._config.file_transfer_timeout,
+            headers=headers,
+        )
+        if not upload_response.ok:
+            raise CogniteFileUploadError(
+                message=upload_response.text,
+                code=upload_response.status_code,
+            )
+
+    def complete_multipart_upload(self, session: FileMultipartUploadSession) -> None:
```
msusag, line 726: After the complete_multipart_upload call is completed, the client normally would need to also check for if the file is marked as uploaded=true.

To simplify this checking for the client, the complete_multipart_upload could have a optional setting, for example: `wait_for_upload`, which will call the file `retrieve` operation on the fileID, and repeatedly check (with some backoff delay) until it sees the file is marked as uploaded = true. Normally this is done within 1 second after complete_multipart_upload has returned. Sometimes it can take several seconds though.

tests/tests_integration/test_api/test_files.py
```diff
@@ -209,8 +211,37 @@ def test_upload_bytes_with_nordic_characters(self, cognite_client: CogniteClient
...
         assert retrieved_content == content.encode("utf-8")
+
+    def test_upload_multipart(self, cognite_client: CogniteClient) -> None:
+        # Min file chunk size is 5MiB
+        content_1 = "abcde" * 1_200_000
+        content_2 = "fghij"
+
+        external_id = "test_upload_multipart"
+
+        cognite_client.config.file_transfer_timeout = 10
```
msusag, line 230: Transferring 5 * 1.2 = 6 MiB may perhaps take more than 10 seconds in some cases ?
Should the timeout limit be increased to avoid flaky tests?

tests/tests_integration/test_api/test_files.py
```diff
@@ -209,8 +211,37 @@ def test_upload_bytes_with_nordic_characters(self, cognite_client: CogniteClient
...
+
+        assert len(session.upload_urls) == 2
+
+        cognite_client.files.upload_multipart_part(session.upload_urls[0], content_1, session.file_metadata.mime_type)
+        cognite_client.files.upload_multipart_part(session.upload_urls[1], content_2, session.file_metadata.mime_type)
+
+        cognite_client.files.complete_multipart_upload(session)
+
+        retrieved_content = cognite_client.files.download_bytes(external_id=external_id)
+        assert len(retrieved_content) == 6000005
```
msusag, line 247: It should delete the file afterwards?
Or are the files cleaned up in some other way by the python SDK integration test tooling?

cognite/client/_api/files.py
msusag: Should this method return a boolean, to return status about that the uploaded flag is true or not?
Or just always true, as long as the response is a 200 or 201.

erlendvollset:
> > How about making this a context manager which yields the Session/Init object instead?
> 
> 
> 
> Like using `__enter__` and `__exit__`? It could maybe be done that way, though I'm not sure if it will really be that intuitive.

The user wouldn't need to explicitly invoke the begin/complete methods.

erlendvollset:
> > > How about making this a context manager which yields the Session/Init object instead?
> > 
> > 
> > 
> > Like using `__enter__` and `__exit__`? It could maybe be done that way, though I'm not sure if it will really be that intuitive.
> 
> The user wouldn't need to explicitly invoke the begin/complete methods. And you can use the @contextmanager decorator, that's simpler than implementing those dunder methods.

erlendvollset:
(accidentally deleted that comment)

tests/tests_integration/test_api/test_files.py
haakonvt: Also, need to use `monkeypatch.setattr` to avoid side effects on other tests

cognite/client/_api/files.py
einarmo: Would they need to check that? Or would that be something they _could_ check if they wanted to download the file immediately. Generally when you upload files you don't download them immediately.

einarmo: I'll check this in the test, but for now I don't think it makes sense to check it in `complete`

tests/tests_integration/test_api/test_files.py
einarmo: I just removed the change, it was a leftover from some weird issues we were having with this test.

cognite/client/_api/files.py
```diff
@@ -596,6 +597,146 @@ def upload_bytes(
 
         return FileMetadata._load(returned_file_metadata)
 
+    def begin_multipart_upload(
```
erlendvollset, line 600: I would suggest this name
```suggestion
    def multipart_upload_session(
```

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.27.0] - 2024-03-04
+### Added
+- Added support for multipart file uploads.
```
erlendvollset, line 22:
```suggestion
- Added support for multipart file uploads using the `client.files.multipart_upload_session` method.
```

cognite/client/_api/files.py
erlendvollset: This method needs to be added to the docs aswell. see `docs/source/files.rst`.

cognite/client/_api/files.py
```diff
@@ -596,6 +597,146 @@ def upload_bytes(
...
+        overwrite: bool = False,
+    ) -> FileMultipartUploadSession:
+        """Begin uploading a file in multiple parts. This allows uploading files larger than 5GiB.
+        Note that the size of each part may not exceed 4000MiB, and the size of each part except the last
+        must be greater than 5MiB.
+
+        The file chunks may be uploaded in any order, and in parallel, but the client must ensure that
+        the parts are stored in the correct order by uploading each chunk to the correct upload URL.
+
+        After calling this, call `upload_multipart_part` for each chunk, then `complete_multipart_upload`
```
erlendvollset, line 625: This is outdated now

docs/source/files.rst
```diff
@@ -32,6 +32,10 @@ Upload a string or bytes
 ^^^^^^^^^^^^^^^^^^^^^^^^
 .. automethod:: cognite.client._api.files.FilesAPI.upload_bytes
 
+Upload a file larger than 5GiB in multiple parts
```
erlendvollset, line 35: I wouldn't specify "larger than 5GiB" - that's a limitation on the single-part upload endpoint, not a limitation here. It can and will be used for files smaller than that.

erlendvollset (approved):
Great work! A few more doc nitpicks, but mostly good to go!

cognite/client/_api/files.py
msusag: It is perhaps only for certain use cases it would be important to know that the file is fully persisted, and can be downloaded.

The suggestion I had can be added later if we see a real need.

docs/source/files.rst
msusag: Yes, multipart upload can be used to upload even a 1-byte file.

If a client script wants to upload file sizes of any size between 1 byte and 1000 GiB, they should use the multipart upload function, and not bother switching between the old upload and the new upload_multipart

msusag (approved):
Looks good!

Just adjust those minor documentation changes suggested by Erlend.

PR #1666, opened 2024-03-06
Improve memory usage for raw rows iter

haakonvt:
#### Improved
- When iterating raw rows concurrently, a max queue size for pending results has been added to keep a stable low
  bounded memory usage profile (for when the caller's code isn't processing fast enough to keep up). Worth noting
  that this has no effect on the total retrieval time.

This has the added benefit that our request to the Raw Rows API gets more spread out in time 🚀 

#### Fetching 600k rows, in chunks of 10k, with 10 cursors.
Current SDK on `master`

Screenshot 2024-03-06 at 22 37 42

----
This PR:

Screenshot 2024-03-06 at 22 37 25

Note the remarkably similar fetching times

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/raw.py
+++ b/cognite/client/_api/raw.py
@@ -454,7 +454,6 @@ class RawRowsAPI(APIClient):
         while not all(f.done() for f in futures):
             while results:
                 yield results.popleft()
-            time.sleep(0.001)
         yield from results
 
     def _read_rows_limited(
@@ -473,7 +472,6 @@ class RawRowsAPI(APIClient):
                     quit_early.set()
                     yield part[: limit - n_total]
                     return
-            time.sleep(0.001)
             if all(f.done() for f in futures) and not results:
                 return
 
--- a/cognite/client/_api/raw.py
+++ b/cognite/client/_api/raw.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import math
+import random
 import threading
 import time
 from collections import deque
@@ -355,6 +356,10 @@ class RawRowsAPI(APIClient):
 
         Returns:
             Iterator[Row] | Iterator[RowList]: An iterator yielding the requested row or rows.
+
+        Note:
+            When iterating using partitions > 1, the memory usage is bounded at 2 x partitions x chunk_size. This is implemented
+            by halting retrival speed when the callers code can't keep up.
         """
         if partitions is None or _RUNNING_IN_BROWSER:
             return self._list_generator(
@@ -418,16 +423,24 @@ class RawRowsAPI(APIClient):
             for initial in cursors
         ]
 
-        def exhaust(iterator: Iterator, results: deque[RowList], quit_early: threading.Event) -> None:
+        def exhaust(iterator: Iterator) -> None:
             for res in iterator:
                 results.append(res)
                 if quit_early.is_set():
                     return
+                # User code might be processing slower than the rate of row-fetching, and we want this
+                # iteration-based method to have a upper-bounded memory impact, so we keep a max queue
+                # size of unprocessed row-chunks:
+                while len(results) >= partitions:
+                    # Sleep randomly per thread to avoid sending all new fetch requests at the same time
+                    time.sleep(random.uniform(0, partitions))
+                    if quit_early.is_set():
+                        return
 
         quit_early = threading.Event()
         results: deque[RowList] = deque()  # fifo, not that ordering matters anyway...
         pool = ConcurrencySettings.get_thread_pool_executor_or_raise(max_workers=self._config.max_workers)
-        futures = [pool.submit(exhaust, task, results, quit_early) for task in read_iterators]
+        futures = [pool.submit(exhaust, task) for task in read_iterators]
 
         if finite_limit:
             yield from self._read_rows_limited(futures, results, cast(int, limit), quit_early)
@@ -441,7 +454,7 @@ class RawRowsAPI(APIClient):
         while not all(f.done() for f in futures):
             while results:
                 yield results.popleft()
-            time.sleep(0.5)
+            time.sleep(0.001)
         yield from results
 
     def _read_rows_limited(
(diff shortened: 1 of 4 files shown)


cognite/client/_api/raw.py
```diff
@@ -418,16 +423,24 @@ def _list_generator_concurrent(
...
-        def exhaust(iterator: Iterator, results: deque[RowList], quit_early: threading.Event) -> None:
+        def exhaust(iterator: Iterator) -> None:
             for res in iterator:
                 results.append(res)
                 if quit_early.is_set():
                     return
+                # User code might be processing slower than the rate of row-fetching, and we want this
+                # iteration-based method to have a upper-bounded memory impact, so we keep a max queue
+                # size of unprocessed row-chunks:
+                while len(results) >= partitions:
```
pavelzubarev, line 434: I have a feature request here: Can we add `or self.user_throttler()` where user_throttler is a function user can provide and control the throttling behavior of their own will? 

My user story: As a user I want to ensure the memory footprint of my data-heavy app, so I want to write the following code:
```python
import os, psutil
process = psutil.Process()

def my_throttling_logic() -> bool:
  # returns true when throttling is needed
  return process.memory_info().rss) > 2 * math.pow(1024, 3)

....
# somewhere in my code:
for rows in client.raw.rows(
            **rows_args,
            chunk_size=10_000,
            limit=-1,
            min_last_updated_time=int(_min_time),
            max_last_updated_time=int(_max_time),
            partitions=3,
            throttle=my_throttling_logic,
        )
```

cognite/client/_api/raw.py
```diff
@@ -441,7 +454,7 @@ def _read_rows_unlimited(self, futures: list[Future], results: deque[RowList]) -
         while not all(f.done() for f in futures):
             while results:
                 yield results.popleft()
-            time.sleep(0.5)
+            time.sleep(0.001)
```
erlendvollset, line 457: This sleep seems redundant.

cognite/client/_api/raw.py
```diff
@@ -460,7 +473,7 @@ def _read_rows_limited(
                     quit_early.set()
                     yield part[: limit - n_total]
                     return
-            time.sleep(0.5)
+            time.sleep(0.001)
```
erlendvollset, line 476: This sleep seems redundant.

cognite/client/_api/raw.py
```diff
@@ -418,16 +423,24 @@ def _list_generator_concurrent(
...
         ]
 
-        def exhaust(iterator: Iterator, results: deque[RowList], quit_early: threading.Event) -> None:
+        def exhaust(iterator: Iterator) -> None:
             for res in iterator:
                 results.append(res)
                 if quit_early.is_set():
                     return
+                # User code might be processing slower than the rate of row-fetching, and we want this
+                # iteration-based method to have a upper-bounded memory impact, so we keep a max queue
```
erlendvollset, line 432: Is this really upper-bounded? What if the processing time far exceeds `time.sleep(random.uniform(0, partitions))`, then `results` will just keep growing?

What about something like:
```
while True:
    if (len(results) >= partitions):
        # Sleep randomly per thread to avoid sending all new fetch requests at the same time
        time.sleep(random.uniform(0, partitions))
    else:
        try:
            res = next(iterator)
        except StopIteration:
            return
        results.append(res)
    if quit_early.is_set():
        return
```

cognite/client/_api/raw.py
haakonvt: Agree

cognite/client/_api/raw.py
haakonvt: It is inside a while loop so it will just continuously wait until there's room in `results` (see the screenshot if you don't believe me 😜 ). Your code risks being stuck forever in `sleep` if `quit_early` signal is given.

erlendvollset: But currently it appends to `results` in every iteration of the loop, regardless of whether `len(results) >= partitions` or not.

erlendvollset: How does my code get stuck?

erlendvollset: In your screenshot, the chunk-processing time doesn't exceed 2 seconds, so it's fast enough to not cause `results` to grow beyond some threshold.

haakonvt: If we ignore the "limit-is-reached-quit-trigger", the code is this:
```py
for res in iterator:
    results.append(res)
    while len(results) >= partitions:
        time.sleep(random.uniform(0, partitions))
```
Imo, it should stay in the while loop until there's room

erlendvollset: I read it as:
```
for res in iterator: 
    results.append(res) 
    if len(results) >= partitions:  # <--- `if` not `while`
        time.sleep(random.uniform(0, partitions))
```
🤦

erlendvollset: Thank you for spoon feeding me 🙏

cognite/client/_api/raw.py
```diff
@@ -418,16 +423,24 @@ def _list_generator_concurrent(
...
+                # size of unprocessed row-chunks:
+                while len(results) >= partitions:
+                    # Sleep randomly per thread to avoid sending all new fetch requests at the same time
+                    time.sleep(random.uniform(0, partitions))
```
erlendvollset, line 436: Do we have an upper bound on number of partitions?
If not, this should probably be
```suggestion
time.sleep(random.uniform(0, min(partitions, 10)))
```

haakonvt: My thought process around this was: This sleep only triggers if they are "severely congested", so random imo sleep should scale with the number of workers, but happy to discuss/think more on it (I find it hard to reason about threading in general) 🤔

erlendvollset: I think that makes sense, but num_workers != partitions. So, should it be `time.sleep(random.uniform(0, self._config.max_workers))` instead then?

erlendvollset: ah, I see now that partitions is already capped at max_workers 👍

erlendvollset (approved):
LGTM! Nice work 💯

erlendvollset (approved):
LGTM! Nice work 💯

PR #1667, opened 2024-03-07
Add `advanced_filter` (and `sort`) param to `list()` method on `assets`, `events`, `sequences`, `time_series` APIs

ddonukis:
## Description
Add `advanced_filter` param to `list()` and `__call__()` method on assets, events, sequences, time_series APIs. Some of the api also were missing sort and/or parallelism params. Added those as well.

This PR is effectively making filter() method obsolete, so I added a deprecation warning.

It's now possible to combine simple and advanced filter in a single request making certain use cases possible, such as a filter on asset subtree ids in combination with an advanced property filter.

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/assets.py
```diff
@@ -1006,6 +1010,8 @@ def list(
         aggregated_properties: Sequence[AggregateAssetProperty] | None = None,
         partitions: int | None = None,
         limit: int | None = DEFAULT_LIMIT_READ,
+        advanced_filter: Filter | dict | None = None,
```
erlendvollset, line 1013: If we are gonna do this, it should be done across the board - for all resource types supporting advanced filters. I'd like your input on whether we should do this at all though @haakonvt @doctrino.

cognite/client/_api/assets.py
```diff
@@ -1081,12 +1125,22 @@ def list(
             root=root,
             external_id_prefix=external_id_prefix,
         ).dump(camel_case=True)
+
+        prep_sort = None
+        if advanced_filter is not None:
+            self._validate_filter(advanced_filter)
+            prep_sort = prepare_filter_sort(sort, AssetSort)
+
+        agg_props = self._process_aggregated_props(aggregated_properties)
```
erlendvollset, line 1134: This is already done above on l1109

tests/tests_integration/test_api/test_assets.py
```diff
@@ -279,24 +279,26 @@ def test_upsert_2_asset_one_preexisting(self, cognite_client: CogniteClient) ->
                 external_id=[new_asset.external_id, preexisting.external_id], ignore_unknown_ids=True
             )
 
-    def test_filter_on_metadata_key(
+    def test_advanced_filter_on_metadata_key(
```
erlendvollset, line 282: Let's not remove existing tests on `.filter`, we want to have those. Add new tests instead.

tests/tests_integration/test_api/test_assets.py
erlendvollset, line 1: Let's also add a test for behaviour when mixing standard filters and advanced filters.

erlendvollset (changes requested):

erlendvollset (commented):
Also needs changelog entry and version bump

tests/tests_integration/test_api/test_assets.py
ddonukis: Sure, I can do that. Was trying not to add to many integration tests as they take a while to run.

cognite/client/_api/assets.py
```diff
@@ -171,12 +175,21 @@ def __call__(
             external_id_prefix=external_id_prefix,
         ).dump(camel_case=True)
 
+        prep_sort = None
+        if sort is not None:
+            prep_sort = prepare_filter_sort(sort, AssetSort)
```
haakonvt, line 180: This check is not needed as it is handled by the function `prepare_filter_sort`

cognite/client/_api/assets.py
```diff
@@ -171,12 +175,21 @@ def __call__(
...
+
+        if advanced_filter is not None:
+            self._validate_filter(advanced_filter)
```
haakonvt, line 183: This check is not needed as it is handled by the function `_validate_filter`.

cognite/client/_api/assets.py
```diff
@@ -867,6 +880,10 @@ def filter(
                 ...     sort=(SortableAssetProperty.external_id, "asc"))
 
         """
+        warnings.warn(
+            "This method is deprecated. Use the list method with advanced_filter parameter instead.",
```
haakonvt, line 884: I would add which method this is + a note on removal in the next major version of the SDK

cognite/client/_api/assets.py
```diff
@@ -1026,8 +1045,10 @@ def list(
...
             aggregated_properties (Sequence[AggregateAssetProperty] | None): Set of aggregated properties to include. Options are childCount, path, depth.
-            partitions (int | None): Retrieve assets in parallel using this number of workers. Also requires `limit=None` to be passed. To prevent unexpected problems and maximize read throughput, API documentation recommends at most use 10 partitions. When using more than 10 partitions, actual throughout decreases. In future releases of the APIs, CDF may reject requests with more than 10 partitions.
+            partitions (int | None): Retrieve assets in parallel using this number of workers. Also requires `limit=None` to be passed. To prevent unexpected problems and maximize read throughput, API documentation recommends at most use 10 partitions. When using more than 10 partitions, actual throughout decreases. In future releases of the APIs, CDF may reject requests with more than 10 partitions. Note, when using partitions sort is not supported. Since partitions are done independently of sorting, there would be no guarantee of the sort order between elements from different partitions.
```
haakonvt, line 1048: Consider making a separate "note" section regarding this interplay between partitions and sort.

What I mean:
```
Args:
    foo (int): asdf

Returns:
    int: the computed asdf

Note:
    Bla bla
```

cognite/client/_api/assets.py
```diff
@@ -1026,8 +1045,10 @@ def list(
...
             limit (int | None): Maximum number of assets to return. Defaults to 25. Set to -1, float("inf") or None to return all items.
+            advanced_filter (Filter | dict | None): Advanced filter query using the filter DSL (Domain Specific Language). It allows defining complex filtering expressions that combine simple operations, such as equals, prefix, exists, etc., using boolean operators and, or, and not. It applies to basic fields as well as metadata.
+            sort (SortSpec | list[SortSpec] | None): The criteria to sort by. Can be up to two properties to sort by default to ascending order. Note, when using sort, partitions is not supported. See the note on partitions for more information.
```
haakonvt, line 1051: "Not supported" indicates to me that it won't work - but the API will return "as normal", just that there is no sorting outside each partition, so in effect, and what I think is a better wording, _sort is ignored when partitions is used_.

cognite/client/_api/assets.py
```diff
@@ -1061,6 +1082,42 @@ def list(
...
+                ...     sort=(SortableAssetProperty.external_id, "asc"))
+
+            Combine filter and advanced filter:
+
+                >>> from cognite.client import CogniteClient
+                >>> from cognite.client.data_classes import filters
+                >>> client = CogniteClient()
+                >>> not_instrument_lvl5 = filters.And(
+                ...    filters.ContainsAny(("labels"), ["Level5"]),
+                ...    filters.Not(filters.ContainsAny(("labels"), ["Instrument"]))
```
haakonvt, line 1117: Extra parens snuck in it seems:
```suggestion
                ...    filters.ContainsAny("labels", ["Level5"]),
                ...    filters.Not(filters.ContainsAny("labels", ["Instrument"]))
```

cognite/client/_api/sequences.py
```diff
@@ -127,13 +133,24 @@ def __call__(
...
         return self._list_generator(
             list_cls=SequenceList,
             resource_cls=Sequence,
             method="POST",
             chunk_size=chunk_size,
             filter=filter,
+            advanced_filter=advanced_filter,
             limit=limit,
+            sort=prep_sort,
+            partitions=partitions,
```
haakonvt, line 153: This was missing? 😂 Good catch!

tests/tests_unit/test_api/test_signatures.py
```diff
@@ -28,45 +30,83 @@ class TestListAndIterSignatures:
...
-            (sequences.SequencesAPI, sequences.SequenceFilter, ["asset_subtree_external_ids", "data_set_external_ids"]),
+            (
+                sequences.SequencesAPI,
+                sequences.SequenceFilter,
+                [
+                    "asset_subtree_external_ids",
+                    "data_set_external_ids",
+                    "partitions",
+                    "advanced_filter",
+                    "sort",
```
haakonvt, line 49: Huh, never seen these "meta tests" before

haakonvt (commented):
Good work! A few thoughts and comments from me

cognite/client/_api/assets.py
ddonukis: Updated

cognite/client/_api/assets.py
ddonukis: updated

cognite/client/_api/assets.py
ddonukis: I think I used them intentionally to be consistent with the api docs but parens in this case don't change the output:
```python
>>> f.ContainsAny(("labels"), ["blah", "blah2"]).dump()
{'containsAny': {'property': ['labels'], 'values': ['blah', 'blah2']}}
>>> f.ContainsAny("labels", ["blah", "blah2"]).dump()
{'containsAny': {'property': ['labels'], 'values': ['blah', 'blah2']}}
```

ddonukis: fixed in all 4 apis

cognite/client/_api/assets.py
ddonukis: agreed, updated the wording

cognite/client/_api/assets.py
ddonukis: Might be good, but I think it's getting out of scope for this PR. Actually, since this section is pretty much copy-paste across the different apis. Maybe better to add this to a shared docs section with notes and usage guides?

cognite/client/_api/assets.py
ddonukis: updated

ddonukis:
Addressed the last batch of the comments. Is it good to go now? 🙂

cognite/client/_api/time_series.py
```diff
@@ -791,6 +808,41 @@ def list(
...
+                >>> client = CogniteClient()
+                >>> in_timezone = flt.Prefix(TimeSeriesProperty.metadata_key("timezone"), "Europe")
+                >>> res = client.time_series.list(
+                ...     advanced_filter=in_timezone,
+                ...     sort=(SortableTimeSeriesProperty.external_id, "asc"))
+
+            Combine filter and advanced filter:
+
+                >>> from cognite.client import CogniteClient
+                >>> from cognite.client.data_classes import filters
```
haakonvt, line 839: You import filters two different ways here

cognite/client/_api/sequences.py
```diff
@@ -805,6 +825,9 @@ def list(
...
+            advanced_filter (Filter | dict | None): Advanced filter query using the filter DSL (Domain Specific Language). It allows defining complex filtering expressions that combine simple operations, such as equals, prefix, exists, etc., using boolean operators and, or, and not. It applies to basic fields as well as metadata.
+            sort (SortSpec | list[SortSpec] | None): The criteria to sort by. Can be up to two properties to sort by default to ascending order. Note, sort is ignored when partitions > 1 is used, since partitions are processed independently of each other. See the note on partitions for more information.
```
haakonvt, line 830: I think the description for `partitions`, `advanced_filter` and `sort` should be the same across the resources, and I think they are saying slightly different things now.

cognite/client/_api/events.py
```diff
@@ -748,6 +761,42 @@ def list(
...
+                ...     sort=(SortableEventProperty.external_id, "asc"))
+
+            Combine filter and advanced filter:
+
+                >>> from cognite.client import CogniteClient
+                >>> from cognite.client.data_classes import filters
+                >>> client = CogniteClient()
+                >>> not_instrument_lvl5 = filters.And(
+                ...    filters.ContainsAny("labels", ["Level5"]),
+                ...    filters.Not(filters.ContainsAny("labels", ["Instrument"]))
```
haakonvt, line 796: Same comment here, you've already imported filters as `flt`

cognite/client/_api/assets.py
haakonvt: You added functionality that has "special behavior" with certain existing parameters, that seems very much in-scope to me 😄

> Maybe better to add this to a shared docs section with notes and usage guides?

Sounds good 👍

ddonukis:
Reviewed the docstrings and standardized the across the APIs.

@haakonvt regarding sort, it's not allowed, because the sdk raises an exception when both partitions and sort are provided.

cognite/client/_api/events.py
```diff
@@ -100,6 +101,7 @@ def __call__(
             sort (SequenceNotStr[str] | None): Sort by array of selected fields. Ex: ["startTime:desc']. Default sort order is asc when omitted. Filter accepts following field names: startTime, endTime, createdTime, lastUpdatedTime. We only support 1 field for now.
             limit (int | None): Maximum number of events to return. Defaults to return all items.
             partitions (int | None): Retrieve assets in parallel using this number of workers. Also requires `limit=None` to be passed. To prevent unexpected problems and maximize read throughput, API documentation recommends at most use 10 partitions. When using more than 10 partitions, actual throughout decreases. In future releases of the APIs, CDF may reject requests with more than 10 partitions.
+            advanced_filter (Filter | dict | None): Advanced filter query using the filter DSL (Domain Specific Language). It allows defining complex filtering expressions that combine simple operations, such as equals, prefix, exists, etc., using boolean operators and, or, and not. It applies to basic fields as well as metadata.
```
haakonvt, line 104: Should we update the `sort` description? And type annotation?

cognite/client/_api/time_series.py
```diff
@@ -710,6 +719,10 @@ def filter(
                 >>> is_numeric = Equals(TimeSeriesProperty.is_string, False)
                 >>> res = client.time_series.filter(filter=is_numeric, sort=SortableTimeSeriesProperty.external_id)
         """
+        warnings.warn(
+            f"{self.__class__.__name__}.filter() method is deprecated and may be removed in the next major version of the SDK. Use the {self.__class__.__name__}.list() method with advanced_filter parameter instead.",
```
haakonvt, line 723:
```suggestion
            f"{self.__class__.__name__}.filter() method is deprecated and will be removed in the next major version of the SDK. Use the {self.__class__.__name__}.list() method with advanced_filter parameter instead.",
```

CHANGELOG.md
```diff
@@ -17,6 +17,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.28.0] - 2024-03-11
+### Added
+- Added support for advanced filter query in the `list()` (and `__call__()`) method of `assets`, `events`, `sequences`, and `time_series` APIs. Now you are able to use advanced filter (like in `filter()`) at the same time as the simple filter properties allowing for more complex requests. Additionally, added `sort` and `partitions` params to `list()` and `__call__()` methods of all the APIs that were missing either.
+### Deprecated
+- Added a deprecation warning on the `filter()` method of `assets`, `events`, `sequences`, and `time_series` APIs as its functionality is fully replicated in the `list()` method.
```
haakonvt, line 24:
```suggestion
## [7.28.0] - 2024-03-12
### Added
- Added support for advanced filter query in the `list()` (and `__call__()`) method of `assets`, `events`, `sequences`,
  and `time_series` APIs. Now you are able to use advanced filter (like in `filter()`) at the same time as the simple
  filter properties, allowing for more complex requests.
- Added missing `sort` parameter to `list()` and `__call__()` methods for `AssetsAPI`.
- Added missing `sort` parameter to `list()` and `__call__()` methods for `TimeSeriesAPI`.
- Added missing `sort` and `partitions` parameters to `list()` and `__call__()` methods for `SequencesAPI`.
### Deprecated
- Added a deprecation warning on the `filter()` method of `assets`, `events`, `sequences`, and `time_series` APIs as
  its functionality is fully covered by the `list()` method.
```

haakonvt (approved):
Great work @ddonukis ! 💯 (I have a few last few suggestions/comments btw)

cognite/client/_api/events.py
haakonvt: I ask because it seems only events already had the `sort` parameter

cognite/client/_api/time_series.py
```diff
@@ -764,12 +779,20 @@ def list(
...
-            partitions (int | None): Retrieve time series in parallel using this number of workers. Also requires `limit=None` to be passed.
+            partitions (int | None): Retrieve resources in parallel using this number of workers (values up to 10 allowed), limit must be set to `None` (or `-1`).
             limit (int | None): Maximum number of time series to return.  Defaults to 25. Set to -1, float("inf") or None to return all items.
+            advanced_filter (Filter | dict | None): Advanced filter query using the filter DSL (Domain Specific Language). It allows defining complex filtering expressions that combine simple operations, such as equals, prefix, exists, etc., using boolean operators and, or, and not. See examples below for usage.
+            sort (SortSpec | list[SortSpec] | None): The criteria to sort by. Defaults to desc for `_score_` and asc for all other properties. Sort is not allowed if `partitions` is used.
 
         Returns:
             TimeSeriesList: The requested time series.
 
+        .. note::
```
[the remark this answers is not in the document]
ddonukis: I see, good to know. I think for me it's a bit more transparent that it's a special block rather than just plan text when I use special syntax.

PR #1670, opened 2024-03-08
Robustify data model tests

doctrino:
## Description
Problem: The fixture `cdf_data_models` uses a `.list(limit=-1)` call. When the tests run in parallel the tests such as `test_apply_retrieve_and_delete` create temporarily data models which are retrieved by the `.list(limit=-1)` call. When this fixture is used in testing later, the temporary model is no longer there and thus the test fails.

https://github.com/cognitedata/cognite-sdk-python/actions/runs/8188053206/job/22389846913

Solution: Avoid the `.list(limit=-1)` and instead just use two permanent models for the `cdf_data_models` fixture.

## Checklist:
- [ ] Tests added/updated.


erlendvollset (approved):
Nice! 💯

PR #1674, opened 2024-03-12
Fix the problem for multipart uploads on Google by not setting Content-Type header in the individual upload_part calls

msusag:
Setting the content-type for the whole file is handled by initmultipartupload or completemultipartupload.

## Description

In the files API multipart upload support, the Content-Type is set by the files service in the initmultipartupload or completemultipartupload. It is not necessary, and actually causes a problem on at least Google clusters, to set it in the part upload requests for the upload URLs.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/files.py
+++ b/cognite/client/_api/files.py
@@ -699,8 +699,7 @@ class FilesAPI(APIClient):
     def _upload_multipart_part(
         self,
         upload_url: str,
-        content: str | bytes | TextIO | BinaryIO,
-        mime_type: str | None = None,
+        content: str | bytes | TextIO | BinaryIO
     ) -> None:
         """Upload part of a file to an upload URL returned from `multipart_upload_session`.
         Note that if `content` does not somehow expose its length, this method may not work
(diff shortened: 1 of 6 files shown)


cognite/client/_api/files.py
```diff
@@ -709,18 +708,16 @@ def _upload_multipart_part(
...
             content = content.encode("utf-8")
 
-        headers = {"Content-Type": mime_type}
         upload_response = self._http_client_with_retry.request(
             "PUT",
             upload_url,
             data=content,
             timeout=self._config.file_transfer_timeout,
-            headers=headers,
+            headers=None,
```
einarmo, line 720: I think `headers=None` is the default, you can probably remove this line completely.

msusag: I prefer to keep it as it is then expressing that this request MUST not have any headers.

einarmo (approved):

cognite/client/_api/files.py
haakonvt: ...then you should add a comment

PR #1675, opened 2024-03-12
limit warning

ks93:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/data_modeling/instances.py
+++ b/cognite/client/_api/data_modeling/instances.py
@@ -232,7 +232,7 @@ class InstancesAPI(APIClient):
 
         Args:
             chunk_size (int | None): Number of data_models to return in each chunk. Defaults to yielding one instance at a time.
-            instance_type (Literal['node', 'edge']): Whether to query for nodes or edges.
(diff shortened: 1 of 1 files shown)


cognite/client/_api/data_modeling/instances.py
```diff
@@ -864,6 +865,12 @@ def aggregate(
         if instance_type not in ("node", "edge"):
             raise ValueError(f"Invalid instance type: {instance_type}")
 
+        if limit is None:
```
erlendvollset, line 868: IMO `limit=max_limit` makes more sense in this case, with no warning.

ks93: My main hesitation is that as an SDK user I've come to expect `limit=None` gives me `n` results even if `n>max_limit_per_request`, which is not true for this specific endpoint. It kind-off leads me to believe `limit=None` is a think the SDK handles 🤷 

A secondary reason is that the SDK would need to know (and be up to date) on the limit. Not a big problem.

erlendvollset: Can you check the behaviour on other aggregation/search methods (which don't support paging) in the SDK?

PR #1680, opened 2024-03-19
Add `_repr_html_` to `Properties` class

haakonvt:
As requested here: https://github.com/cognitedata/cognite-sdk-python/pull/1624#pullrequestreview-1897448951

### Added
- `Properties` class, as used on e.g. `Node` and `Edge`, now renders in Jupyter Notebooks (`_repr_html_` added).

### DIFF
--- a/cognite/client/data_classes/data_modeling/instances.py
+++ b/cognite/client/data_classes/data_modeling/instances.py
@@ -258,6 +258,19 @@ class Properties(MutableMapping[ViewIdentifier, MutableMapping[PropertyIdentifie
         view_id = ViewId.load(view)
         self.data[view_id] = properties
 
+    def _repr_html_(self) -> str:
+        pd = local_import("pandas")
+        index_names = "space", "view", "version"
+        if not self:
+            df = pd.DataFrame(index=pd.MultiIndex(levels=([], [], []), codes=([], [], []), names=index_names))
+        else:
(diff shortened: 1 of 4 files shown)


haakonvt:
Example showing how Properties renders with a single source, multiple sources and no source:
----
Screenshot 2024-03-19 at 14 07 09

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -258,6 +258,19 @@ def __setitem__(self, view: ViewIdentifier, properties: MutableMapping[PropertyI
         view_id = ViewId.load(view)
         self.data[view_id] = properties
 
+    def _repr_html_(self) -> str:
+        pd = local_import("pandas")
+        index_names = "space", "view", "version"
```
doctrino, line 263: For nodes/edges, there will never be multiple versions existing simultaneously as version is just an integer that increments every time something in the node/edge changes. Thus, maybe not worth to include in an index?

haakonvt: This is the version of the view

doctrino: My mistake. Then I agree it makes sense. 

I guess the alternative would be to just make one index of `space:externa_id/version` but I do not mind the multiindex.

doctrino (approved):

PR #1681, opened 2024-03-19
Add quick column access to `Row`

haakonvt:
### Added
- Direct access to the columns/data stored on raw rows have been added (alongside a `.get` method). Example usage:
  `row["my_col"]` (short-cut for: `row.columns["my_col"]`).

Also added missing dunder method contains to `Instance`.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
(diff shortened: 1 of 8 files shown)


erlendvollset (approved):
Very nice! Maybe update an example or two to use this new syntax?

haakonvt:
Added the following:

Screenshot 2024-03-20 at 14 08 44

Screenshot 2024-03-20 at 14 08 55

PR #1685, opened 2024-03-21
Get rid of Optional on read-required CogniteResource properties

erlendvollset:
Currently we allow using subclasses of `CogniteResource` to both read _and_ write resources. e.g. `Asset` is returned from the `list`/`retrieve` methods, but is also accepted in the `create`/`update` methods.

The `id`, `last_updated_time`, and `created_time` attributes are all required when using `CogniteResource` classes to read. But when using them to write, the props are optional - they also don't make sense when using the classes to _write_. So this change just forces the type of those attributes to be non-optional. i.e. `int | None -> int`. That means we no longer need uneccesary asserts, casts, or type ignores in code using these classes. e.g.

```
asset = c.assets.retrieve("someId")

# these variants of dealing with the `Optional` are no longer necessary
assert c.id is not None
asset_id: int = cast(int, asset.id)
asset_id: int = asset.id  # type: ignore
```

With this change, we will break clients with a static type checker which rely on explicitly setting these properties to `None` or some optional variable. So, this will fail a static type checker now:

```
id: Optional[int] = 2

asset = Asset()
asset.id = id
# or
asset.id = None
```

But this doesn't match with any intended usage of the SDK that I'm aware of, so we deem this an acceptable tradeoff. We won't break anyone at runtime.

### DIFF
--- a/cognite/client/data_classes/functions.py
+++ b/cognite/client/data_classes/functions.py
@@ -133,9 +133,14 @@ class Function(FunctionCore):
             runtime=runtime,
             metadata=metadata,
         )
-        self.id = cast(int, id)
+        # id/created_time are required when using the class to read,
+        # but don't make sense passing in when creating a new object. So in order to make the typing
+        # correct here (i.e. int and not Optional[int]), we force the type to be int rather than
+        # Optional[int].
+        # TODO: In the next major version we can make these properties required in the constructor
+        self.id: int = id  # type: ignore
+        self.created_time: int = created_time  # type: ignore
         self.status = status
-        self.created_time = created_time
         self.runtime_version = runtime_version
         self.error = error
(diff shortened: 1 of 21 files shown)


cognite/client/data_classes/functions.py
```diff
@@ -133,9 +133,14 @@ def __init__(
             runtime=runtime,
             metadata=metadata,
         )
-        self.id = cast(int, id)
+        # id/created_time/last_updated_time are required when using the class to read,
```
haakonvt, line 136:
```suggestion
        # id/created_time are required when using the class to read,
```

haakonvt (approved):
Awesome! 🚀

doctrino (approved):
Looks good.

PR #1686, opened 2024-03-22
fix pyright issues with UNKNOWN

haakonvt:
## Description
Various list methods were addressed in a [previous PR](https://github.com/cognitedata/cognite-sdk-python/pull/1682), which led me to discover that more methods, like search, give similar `pyright` errors.

### DIFF
--- a/cognite/client/_api/documents.py
+++ b/cognite/client/_api/documents.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from collections.abc import Iterator
 from pathlib import Path
-from typing import IO, TYPE_CHECKING, BinaryIO, Literal, cast, overload
+from typing import IO, TYPE_CHECKING, Any, BinaryIO, Literal, cast, overload
 
 from cognite.client._api_client import APIClient
 from cognite.client._constants import DEFAULT_LIMIT_READ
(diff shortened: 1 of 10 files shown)


erlendvollset:
LGTM! We should estimate the effort it would take to also run pyright in automated checks - if they're even compatible with the settings we want.

erlendvollset (approved):

haakonvt:
> LGTM! We should estimate the effort it would take to also run pyright in automated checks - if they're even compatible with the settings we want.

Amen 👏

cognite/client/_api/documents.py
```diff
@@ -225,7 +225,7 @@ def __call__(
 
         Args:
             chunk_size (int | None): Number of documents to return in each chunk. Defaults to yielding one document at a time.
-            filter (Filter | dict | None): Filter | dict | None): The filter to narrow down the documents to return.
+            filter (Filter | dict[str, Any] | None): Filter | dict[str, Any] | None): The filter to narrow down the documents to return.
```
haakonvt, line 228:
```suggestion
            filter (Filter | dict[str, Any] | None): The filter to narrow down the documents to return.
```

PR #1688, opened 2024-03-22
Drop support for protobuf v3

haakonvt:
## Description
Other important DS libraries like plotly, streamlit and tensorflow no longer require `protobuf<4` (and have been for quite a long time now), so we drop the dual support in preparation for the status codes expansion of datapoints API.


erlendvollset (approved):
LGTM!

PR #1689, opened 2024-03-23
[Bug] Notebook repr of SequenceRowsList

doctrino:
## Description
**Problem**: `SequencesRowList` have a custom implementation of the `.to_pandas()` method with default arguments that 
returns a `dict` instead of `pd.Dataframe`. This causes the default `_repr_html` method to fail.

**Fix**: Add custom implementation of `_repr_html` for the `SequenceRowsList` class.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/sequences.py
+++ b/cognite/client/data_classes/sequences.py
@@ -897,6 +897,9 @@ class SequenceRowsList(CogniteResourceList[SequenceRows]):
 
         raise ValueError(f"Invalid key value '{key}', should be one of ['id', 'external_id']")
 
+    def _repr_html_(self) -> str:
+        return self.to_pandas(key="external_id", concat=True).to_html()
+
 
 SequenceDataList = SequenceRowsList
 
--- a/cognite/client/data_classes/sequences.py
+++ b/cognite/client/data_classes/sequences.py
@@ -898,7 +898,7 @@ class SequenceRowsList(CogniteResourceList[SequenceRows]):
         raise ValueError(f"Invalid key value '{key}', should be one of ['id', 'external_id']")
 
     def _repr_html_(self) -> str:
-        return self.to_pandas(key="external_id", concat=True).to_html()
+        return self.to_pandas(key="external_id", concat=True)._repr_html_()
 
 
 SequenceDataList = SequenceRowsList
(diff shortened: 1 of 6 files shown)


cognite/client/data_classes/sequences.py
```diff
@@ -897,6 +897,9 @@ def to_pandas(
 
         raise ValueError(f"Invalid key value '{key}', should be one of ['id', 'external_id']")
 
+    def _repr_html_(self) -> str:
+        return self.to_pandas(key="external_id", concat=True).to_html()
```
haakonvt, line 901: What's the difference between calling `.to_html()` vs `._repr_html_()`?

haakonvt (approved):
LGTM!

cognite/client/data_classes/sequences.py
haakonvt: @doctrino  Just tested, and I think you really should use `_repr_html_` (`to_html` seems to miss the length checks on the output)

doctrino: I was a bit quick, seems `_repr_html` is the correct one as it contains extra information for notebooks:
```python
    def _repr_html_(self) -> str | None:
        """
        Return a html representation for a particular DataFrame.

        Mainly for IPython notebook.
        """
        if self._info_repr():
            buf = StringIO()
            self.info(buf=buf)
            # need to escape the <class>, should be the first line.
            val = buf.getvalue().replace("<", r"&lt;", 1)
            val = val.replace(">", r"&gt;", 1)
            return f"<pre>{val}</pre>"

        if get_option("display.notebook_repr_html"):
            max_rows = get_option("display.max_rows")
            min_rows = get_option("display.min_rows")
            max_cols = get_option("display.max_columns")
            show_dimensions = get_option("display.show_dimensions")

            formatter = fmt.DataFrameFormatter(
                self,
                columns=None,
                col_space=None,
                na_rep="NaN",
                formatters=None,
                float_format=None,
                sparsify=None,
                justify=None,
                index_names=True,
                header=True,
                index=True,
                bold_rows=True,
                escape=True,
                max_rows=max_rows,
                min_rows=min_rows,
                max_cols=max_cols,
                show_dimensions=show_dimensions,
                decimal=".",
            )
            return fmt.DataFrameRenderer(formatter).to_html(notebook=True)
```

PR #1690, opened 2024-03-24
fix; SequenceNotStr

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/data_modeling/instances.py
+++ b/cognite/client/_api/data_modeling/instances.py
@@ -171,7 +171,7 @@ class InstancesAPI(APIClient):
         limit: int | None = None,
         include_typing: bool = False,
         sources: Source | Sequence[Source] | None = None,
-        space: str | Sequence[str] | None = None,
+        space: str | SequenceNotStr[str] | None = None,
         sort: list[InstanceSort | dict] | InstanceSort | dict | None = None,
         filter: Filter | dict[str, Any] | None = None,
     ) -> Iterator[Node]: ...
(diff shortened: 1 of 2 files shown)


cognite/client/_api/data_modeling/instances.py
```diff
@@ -1093,10 +1093,11 @@ def _query_or_sync(self, query: Query, endpoint: Literal["query", "sync"]) -> Qu
     @overload
     def list(
         self,
+        *,
         instance_type: Literal["node"] = "node",
-        include_typing: bool = False,
+        include_typing: Literal[False] = False,
```
erlendvollset, line 1098: Let's add a comment about why for this.

doctrino: That was a mistake. I did not intend to do this change yet, that is coming in the next PR.

haakonvt (commented):
Not sure I agree that we should change these; imo when both `str` and `Sequence[str]` are accepted, there is no reason to use `SequenceNotStr`. I think the latter can be a bit harder to understand for new users.

However, when writing overloads, there might not be a way around using it, but is that the case here?

doctrino:
> Not sure I agree that we should change these; imo when both `str` and `Sequence[str]` are accepted, there is no reason to use `SequenceNotStr`. I think the latter can be a bit harder to understand for new users.
> 
> However, when writing overloads, there might not be a way around using it, but is that the case here?

You are right it is not necessary for overload here. The reason I suggest it is for consistency, in the entire code base, this is the only place we have `Sequence[str]`

erlendvollset (approved):
Lgtm.

PR #1422, opened 2023-10-12
Fix filter dumps

haakonvt:
## Description
User given string would end up getting camelCased:
```
>>> f.Equals(EventProperty.metadata_key("some_key"), "some_value").dump(camel_case=True)
{'equals': {'property': ['metadata', 'someKey'], 'value': 'some_value'}}
                                     ^^^^^^^^^
```
https://cognitedata.slack.com/archives/CG10VQPFX/p1694426855201089

## Checklist:
- [x] Tests added/updated.
- [ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [x] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [x] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #1695, opened 2024-03-25
Only convert single-element prop refs to camelCase

erlendvollset:
- Only convert single-element property references to camel_case when dumping filters
- Remove NoCaseConversionPropertyList

## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/filters.py
+++ b/cognite/client/data_classes/filters.py
@@ -61,7 +61,10 @@ def _dump_property(property_: PropertyReference, camel_case: bool) -> list[str] 
     elif isinstance(property_, str):
         return [to_camel_case(property_) if camel_case else property_]
     elif isinstance(property_, (list, tuple)):
-        return type(property_)(map(to_camel_case, property_)) if camel_case else property_
+        if len(property_) == 1:
+            return [to_camel_case(property_[0])] if camel_case else property_
+        else:
+            return property_
     else:
         raise ValueError(f"Invalid property format {property_}")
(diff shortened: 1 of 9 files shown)


cognite/client/data_classes/filters.py
```diff
@@ -55,12 +55,15 @@ def _load_filter_value(value: Any) -> FilterValue | FilterValueList:
...
+    if isinstance(property_, EnumProperty):
         return property_.as_reference()
     elif isinstance(property_, str):
         return [to_camel_case(property_) if camel_case else property_]
     elif isinstance(property_, (list, tuple)):
-        return type(property_)(map(to_camel_case, property_)) if camel_case else property_
+        if len(property_) == 1:
+            return [to_camel_case(property_[0])] if camel_case else property_
+        else:
+            return property_
```
haakonvt, line 66: For an input `property_` of type `tuple`, with `len=1` , this no longer preserves the type.

haakonvt (commented):
I can't think of a single filter that allows a user-defined property to become the first element, so I think this fix is Gucci 🚀 

Back when I did #1422, I remember feeling strongly that this wasn't an option, but I struggle to remember my reasons 😅

haakonvt:
> ...but I struggle to remember my reasons

Ah, I suddenly remember - this will break anyone passing filters as `dict`s, which we allow.

haakonvt (approved):
LGTM!

PR #1699, opened 2024-03-28
Fix dtype conversion on `DatapointsArray`

haakonvt:
### Fixed
- Several methods for `DatapointsArray` that previously failed for string datapoints due to bad handling
  of numpy `dtype`-to-native conversion.


doctrino (approved):
Nice catch :)

PR #1702, opened 2024-04-03
Bugfix: dropna how-setting for `SequencesDataAPI.insert_dataframe`

haakonvt:
### Fixed
- Inserting sequence data using `insert_dataframe` would by default drop all rows that contained at least one missing value. This has now been fixed to only remove rows where all values are missing.


erlendvollset (commented):
Let's add a test!

haakonvt:
> Let's add a test!

@erlendvollset ✅

erlendvollset (approved):
Beautiful! LGTM 💯

PR #1707, opened 2024-04-05
Support inserting datapoints with status codes

haakonvt:
### Added
- Datapoints insert methods `insert` and `insert_multiple` now support ingesting (optional) status codes.


cognite/client/_api/datapoints.py
```diff
@@ -1335,91 +1338,128 @@ def insert(
...
 
+        Note:
+            All datapoints inserted without a status code (or symbol) is assumed to be good (code 0). To mark a value, pass
+            either the status code (int) or status symbol (str). Only one of code and symbol is required. If both are given,
+            they must match or an API error will be raised.
+
+            Datapoints marked bad can, in addition to the normal numeric range [-1e100, 1e100], take on any of the following
+            values: None (missing), NaN, and +/- Infinity.
+            Your datapoints can be a list of tuples where the first element is the timestamp and the second element is the value.
+            The third element is optional and may contain the status code for the datapoint. To pass by symbol, a dictionary must be used.
```
haakonvt, line 1422:
```suggestion
```

sighol (commented):
I've noticed that if I send in `"symbol": "good"`, the status will be omitted, and it will work as if it was `Good`, but if I send in `uncertain`, it will fail.

haakonvt:
> I've noticed that if I send in `"symbol": "good"`, the status will be omitted, and it will work as if it was `Good`, but if I send in `uncertain`, it will fail.

@sighol yeah, but that is the API raising an error, right? Do you think we should only ignore exactly "Good"?

sighol:
@haakonvt I think we should treat it as case-sensitive. Sending in `"symbol": "good"` should result in an API error to the user.

We are following the opc-ua standard for status codes, and they use `PascalCase` for their [symbols](https://github.com/OPCFoundation/UA-Nodeset/blob/UA-1.05.03-2023-12-15/Schema/StatusCode.csv).

haakonvt:
> @haakonvt I think we should treat it as case-sensitive. Sending in `"symbol": "good"` should result in an API error to the user.
> 
> We are following the opc-ua standard for status codes, and they use `PascalCase` for their [symbols](https://github.com/OPCFoundation/UA-Nodeset/blob/UA-1.05.03-2023-12-15/Schema/StatusCode.csv).

@sighol makes sense, changed now 👌

cognite/client/_api/datapoints.py
```diff
@@ -1590,14 +1590,14 @@ def dump(self) -> dict[str, Any]:
         dumped: dict[str, Any] = {"timestamp": timestamp_to_ms(self.ts), "value": self.value}
         if self.status_code:  # also skip if 0
             dumped["status"] = {"code": self.status_code}
-        if self.status_symbol and self.status_symbol.lower() != "good":
+        if self.status_symbol and self.status_symbol.lower() != "Good":
```
sighol, line 1593: Now you should no longer use `.lower()`, right?

cognite/client/_api/datapoints.py
```diff
@@ -1590,14 +1590,14 @@ def dump(self) -> dict[str, Any]:
...
             dumped.setdefault("status", {})["symbol"] = self.status_symbol
         # Out-of-range float values must be passed as strings:
         dumped["value"] = _json.convert_nonfinite_float_to_str(dumped["value"])
         return dumped
 
     def requires_api_subversion_beta(self) -> bool:
-        return bool(self.status_code or self.status_symbol and self.status_symbol.lower() != "good")
+        return bool(self.status_code or self.status_symbol and self.status_symbol.lower() != "Good")
```
sighol, line 1600: same here: `*.lower()` can never be == `"Good"`

sighol:
@haakonvt
Unfortunately, according to the standard, both `0` and `1024` has the same symbolic representation. They are both `"Good"`. It does look correct in this PR, though :100:

cognite/client/utils/_json.py
```diff
@@ -41,3 +43,31 @@ def dumps(
 
 
 loads = json.loads
+
+# As opposed to protobuf, datapoints in JSON returns out-of-range float values as strings. This means
+# we're forced to do a lookup for every single datapoint we try to insert (and read in retrieve_latest
+# when not ignoring bad)... so we allow some ugly optimizations in the translation code:
+_FLOAT_API_MAPPING = MappingProxyType({"Infinity": math.inf, "-Infinity": -math.inf, "NaN": math.nan})
+_FLOAT_API_MAPPING_REVERSE = MappingProxyType({math.inf: "Infinity", -math.inf: "-Infinity", math.nan: "NaN"})
```
sighol, line 51: Why do you need this? Is it just for the `.dump()` function? I thought you used protobuf both for ingestion and retrieval.

haakonvt: `retrieve_latest` and insertion is done using JSON

haakonvt: To avoid having to rewrite even more logic... 😬

sighol (approved):
LGTM!

cognite/client/_api/datapoints.py
```diff
@@ -1497,15 +1560,14 @@ def insert_dataframe(self, df: pd.DataFrame, external_id_headers: bool = True, d
 
         Warning:
             You can not insert datapoints with status codes using this method (``insert_dataframe``), you'll need
-            to use the :py:meth:`~DatapointsAPI.insert` method instead (once support is added in an upcoming release).
+            to use the :py:meth:`~DatapointsAPI.insert` method instead (or :py:meth:`~DatapointsAPI.insert_multiple`)!
```
doctrino, line 1563: Probably a personal preference, but feels more serious
```suggestion
            to use the `~DatapointsAPI.insert` method instead (or `~DatapointsAPI.insert_multiple`)!
```

doctrino (approved):
Not the biggest fan of the structure of the tests :p, but otherwise not much complain about :)

tests/tests_unit/test_api/test_datapoints.py
```diff
@@ -686,30 +692,157 @@ def test_insert_dataframe_with_strings(self, cognite_client, mock_post_datapoint
...
-        bin = DatapointsBin(1, 1)
-        dps_object = {"id": 100, "datapoints": [{"timestamp": 1, "value": 1}]}
-        bin.add(dps_object)
-        assert not bin.will_fit(1)
+# Increase readability in test data:
+d, t, v = "datapoints", "timestamp", "value"
+
+
+class TestDatapointsPoster:
+    @pytest.mark.parametrize(
```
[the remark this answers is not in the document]
haakonvt: Yes and no 😉  I decided to rewrite the datapoints poster from scratch given the many changes necessary to make status codes and symbols work. The old dps poster also used a request-filling logic that was O(N^2) _and_ it didn't fill up requests to the limit... so I could not restrain myself from rewriting/simplifying it to O(N) + make full requests 😅

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -2308,3 +2317,172 @@ def test_delete_range_before_epoch(self, cognite_client, new_ts):
...
+        empty_dps = cognite_client.time_series.data.retrieve(**ts_kwargs)
+        assert empty_dps.timestamp == empty_dps.value == empty_dps.status_code == []
+
+        # Test insert DatapointsArray object:
+        cognite_client.time_series.data.insert(id=new_ts.id, datapoints=dps_array1)
+        dps_array2 = cognite_client.time_series.data.retrieve_arrays(**ts_kwargs)
+        assert_correct_data(dps_array2)
+
+    def test_tuples_and_dps_objects_with_status_codes__string_ts(self, cognite_client, new_ts_string):
+        ts_kwargs = dict(id=new_ts_string.id, start=-123, limit=50, include_status=True, ignore_bad_datapoints=False)
```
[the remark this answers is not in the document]
haakonvt: I have a bad excuse, due to how we run tests, parametrized ones can run concurrently and I don't want to create N new time series just to be safe 😅 I'll see if I can add a timestamp offset to make sure it always works 👍

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -2308,3 +2317,172 @@ def test_delete_range_before_epoch(self, cognite_client, new_ts):
...
+    def test_bad_status_symbols_and_codes(self, cognite_client, new_ts):
+        with pytest.raises(CogniteAPIError, match="^Invalid status code"):
+            # code=1 is not allowed: When info type is 00, all info bits must be 0
+            cognite_client.time_series.data.insert(datapoints=[(1, 3.1, 1)], id=new_ts.id)
+
+        symbol = random.choice(("good", "uncertain", "bad"))  # should be PascalCased
+        with pytest.raises(CogniteAPIError, match="^Invalid status code symbol"):
+            cognite_client.time_series.data.insert(
+                datapoints=[{"timestamp": 0, "value": 2.3, "status": {"symbol": symbol}}], id=new_ts.id
+            )
```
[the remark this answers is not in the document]
haakonvt: Felt short enough for a single imo, but I'll split it into two but keeping the random part; no need to send 3 requests for this scenario.

cognite/client/utils/useful_types.py
```diff
@@ -21,7 +21,7 @@ def __len__(self) -> int: ...
 
     def __iter__(self) -> Iterator[_T_co]: ...
 
-    def index(self, value: Any, /, start: int = 0, stop: int = ...) -> int: ...
+    def index(self, value: Any, start: int = 0, stop: int = ..., /) -> int: ...
```
haakonvt, line 24: Piggybacked #1720 here

cognite/client/_api/datapoints.py
haakonvt: This removes the clickable links to the other SDK functions when I build the docs locally. How/what change does this cause on your machine?

Without ~~:py:meth:~~
image

With :py:meth:
Screenshot 2024-04-15 at 23 40 09

haakonvt: Or do you mean just without the `()`?

haakonvt: Oh, nvm, they are inserted automatically by sphinx

tests/tests_integration/test_api/test_datapoints.py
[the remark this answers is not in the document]
haakonvt: Decided to keep this; fixing the race conditions is quite the hassle

PR #1722, opened 2024-04-16
Handle Unknown and Invalid filters

erlendvollset:
- Handle unknown filter classes
- Support InvalidFilter type

## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/filters.py
+++ b/cognite/client/data_classes/filters.py
@@ -8,7 +8,7 @@ from typing_extensions import TypeAlias
 
 from cognite.client.data_classes._base import EnumProperty, Geometry
 from cognite.client.data_classes.labels import Label
-from cognite.client.utils._text import to_camel_case
+from cognite.client.utils._text import convert_all_keys_to_camel_case, to_camel_case
 from cognite.client.utils.useful_types import SequenceNotStr
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


tests/tests_unit/test_data_classes/test_data_models/test_filters.py
```diff
@@ -185,8 +185,8 @@ def test_dump_filter(user_filter: Filter, expected: dict) -> None:
 
 
 def test_unknown_filter_type() -> None:
-    with pytest.raises(ValueError, match="Unknown filter type: unknown"):
-        Filter.load({"unknown": {}})
+    unknown = Filter.load({"unknown": {}})
+    assert isinstance(unknown, f.UnknownFilter)
```
doctrino, line 189: Maybe extend with a dump to ensure that loading and dumping a filter returns the same.
```suggestion
    assert isinstance(unknown, f.UnknownFilter)
    assert unknown.dump() == {"unknown": {}}
```

cognite/client/data_classes/filters.py
```diff
@@ -185,8 +185,13 @@ def load(cls, filter_: dict[str, Any]) -> Filter:
                 property=filter_body["property"],
                 value=_load_filter_value(filter_body["value"]),
             )
+        elif filter_name == InvalidFilter._filter_name:
```
doctrino, line 188: Do not think I get this case? When will this happen? I do not see any test hitting it either.

erlendvollset: When a filter references resources which no longer exist there are cases when we don't have necessary information to map these filters to the necessary external representation, so we just need to use a "invalid" placeholder to indicate to the user that they need to fix their filter.

doctrino (approved):

PR #1726, opened 2024-04-17
Xid_for_function_schedule_creation

CogniteJohan:
## Description

## Checklist:
- [x] Tests added/updated.


CogniteJohan (commented):
Added support to use the function external id when creating new schedules

cognite/client/_api/functions.py
haakonvt, line 1: Do I understand correctly that this is purely a client-side helper? (As in a request is first made to get the internal ID, then that is used when creating the schedule)

If yes, then I think we should be even more explicit in the docstring, and maybe even create a `Note:` section to make this distinction from the API crystal clear!

CogniteJohan: Added note, also added extra test if user tries to use both internal and external id.

tests/tests_unit/test_api/test_functions.py
```diff
@@ -910,6 +907,17 @@ def test_list_schedules_with_function_id_and_function_external_id_raises(self, c
...
+    @pytest.mark.usefixtures("mock_function_schedules_response_with_xid")
+    def test_create_schedules_with_function_external_id(self, cognite_client):
+        res = cognite_client.functions.schedules.create(
+            name="my-schedule",
+            function_external_id="my-func",
+            cron_expression="*/5 * * * *",
+            description="Hi",
+        )
+
+        assert isinstance(res, FunctionSchedule)
```
haakonvt, line 919: I think you should also add a check here on the mock that `function_id` is what is passed to `_post` (and not external ID)

tests/tests_unit/test_api/test_functions.py
haakonvt, line 1: [suggestion] In file `tests/tests_integration/test_api/test_functions.py` you can also add an integration test that fails if the external ID does not exist

cognite/client/_api/functions.py
haakonvt: @CogniteJohan Where's this test?

HMEiding (commented):
I'm a little bit on the fence about this change, as it might fool the user. So I agree, let's try to make it as clear as possible that the link is done by internal id under the hood.

cognite/client/_api/functions.py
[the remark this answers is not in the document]
CogniteJohan: Added

CogniteJohan: test_create_schedules_with_function_id_and_function_external_id_raises

tests/tests_unit/test_api/test_functions.py
CogniteJohan: Done. Not sure if this is the right way to test variables passed but it seems to test the necessary functionality

tests/tests_unit/test_api/test_functions.py
CogniteJohan: Done

PR #1735, opened 2024-04-22
Add a flag to `to_pandas()` to convert np.nan to None in object columns

ddonukis:
## Description
It can be surprising to find NaN values in an object column. And it's not consistent when pandas creates NaN or None.

`to_pandas()` method is creating a dataframe from a list of dicts created by the `dump()` method on the CogniteResource. Depending on the specific resource implementations some field keys like `labels` are not included if there are no labels attached. In this case pandas assigns `np.nan` value in the label column for this row. But if the empty field was dumped with `{"labels": None}` then pandas would use `None` value. The behaviour is inconsistent and could lead to bugs, considering that lables is a list of dicts and the user would need to write python code of arbitrary complexity to process the values in the column. A surprising property of `np.nan` is that it's truthy, unlike `None`.

To address the problems of having `nan` I'm suggesting to convert nans to None in the `to_pandas()` method.

Proposed implementation provided.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -36,6 +36,7 @@ from cognite.client.utils._pandas_helpers import (
     convert_nullable_int_cols,
     convert_timestamp_columns_to_datetime,
     notebook_display_with_fallback,
+    convert_nan_to_none_for_object_cols,
 )
 from cognite.client.utils._text import convert_all_keys_to_camel_case, to_camel_case
 from cognite.client.utils._time import TIME_ATTRIBUTES, convert_and_isoformat_time_attrs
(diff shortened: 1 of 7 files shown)


cognite/client/data_classes/_base.py
```diff
@@ -351,6 +353,7 @@ def to_pandas(
             expand_metadata (bool): Expand the metadata column into separate columns.
             metadata_prefix (str): Prefix to use for metadata columns.
             convert_timestamps (bool): Convert known columns storing CDF timestamps (milliseconds since epoch) to datetime. Does not affect custom data like metadata.
+            convert_nan_to_none (bool): Convert any missing value type (such as NaN) to None for each column with object dtype (e.g. metadata).
```
ddonukis, line 356:
```suggestion
            convert_missing_to_none (bool): Convert any missing value type (such as NaN) to None for each column with object dtype (e.g. metadata).
```
better name?

ddonukis:
It was not meant to be

PR #1736, opened 2024-04-22
Forward Combability Fixes

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -191,9 +191,7 @@ class UnknownCogniteObject(CogniteObject):
         return cls(resource)
 
     def dump(self, camel_case: bool = True) -> dict[str, Any]:
-        if isinstance(self.__data, dict):
-            return convert_all_keys_recursive(self.__data, camel_case=camel_case)
-        return self.__data
+        return convert_all_keys_recursive(self.__data, camel_case=camel_case)
 
 
 T_CogniteObject = TypeVar("T_CogniteObject", bound=CogniteObject)
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -191,7 +191,9 @@ class UnknownCogniteObject(CogniteObject):
         return cls(resource)
 
     def dump(self, camel_case: bool = True) -> dict[str, Any]:
-        return convert_all_keys_recursive(self.__data, camel_case=camel_case)
+        if isinstance(self.__data, dict):
+            return convert_all_keys_recursive(self.__data, camel_case=camel_case)
+        return self.__data
 
 
 T_CogniteObject = TypeVar("T_CogniteObject", bound=CogniteObject)
--- a/cognite/client/data_classes/data_modeling/containers.py
+++ b/cognite/client/data_classes/data_modeling/containers.py
@@ -106,17 +106,14 @@ class ContainerApply(ContainerCore):
         return ContainerApply(
             space=resource["space"],
             external_id=resource["externalId"],
-            properties={k: ContainerProperty.load(v) for k, v in resource["properties"].items() if isinstance(v, dict)},
+            properties={k: ContainerProperty.load(v) for k, v in resource["properties"].items()},
             description=resource.get("description"),
             name=resource.get("name"),
             used_for=resource.get("usedFor"),
-            constraints={k: Constraint.load(v) for k, v in resource["constraints"].items() if isinstance(v, dict)}
-            or None
+            constraints={k: Constraint.load(v) for k, v in resource["constraints"].items()} or None
             if "constraints" in resource
             else None,
-            indexes={k: Index.load(v) for k, v in resource["indexes"].items() if isinstance(v, dict)} or None
-            if "indexes" in resource
-            else None,
+            indexes={k: Index.load(v) for k, v in resource["indexes"].items()} if "indexes" in resource else None,
         )
 
     def as_write(self) -> ContainerApply:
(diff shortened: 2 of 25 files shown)


cognite/client/data_classes/data_modeling/containers.py
```diff
@@ -285,10 +289,11 @@ def dump(self, camel_case: bool = True) -> dict[str, str | dict]:
 class Constraint(CogniteObject, ABC):
     @classmethod
     def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
-        if resource["constraintType"] == "requires":
-            return cast(Self, RequiresConstraint.load(resource))
-        elif resource["constraintType"] == "uniqueness":
-            return cast(Self, UniquenessConstraint.load(resource))
+        if isinstance(resource, dict):
```
erlendvollset, line 292: Why would this ever not be a dict?

cognite/client/data_classes/data_modeling/containers.py
```diff
@@ -57,11 +57,15 @@ def __init__(
...
+            output["constraints"] = {
+                k: v.dump(camel_case) for k, v in self.constraints.items() if isinstance(v, Constraint)
+            }
         if self.indexes:
-            output["indexes"] = {k: v.dump(camel_case) for k, v in self.indexes.items()}
+            output["indexes"] = {k: v.dump(camel_case) for k, v in self.indexes.items() if isinstance(v, Index)}
         if self.properties:
-            output["properties"] = {k: v.dump(camel_case) for k, v in self.properties.items()}
+            output["properties"] = {
+                k: v.dump(camel_case) for k, v in self.properties.items() if isinstance(v, ContainerProperty)
```
erlendvollset, line 67: Why would these objects ever not be instances of these classes?

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -195,10 +195,18 @@ def load(
     ) -> Properties:
         props: MutableMapping[ViewId, MutableMapping[PropertyIdentifier, PropertyValue]] = {}
         for space, view_properties in data.items():
+            if not isinstance(view_properties, dict):
```
erlendvollset, line 198: Why would this ever not be a dict?

cognite/client/data_classes/data_modeling/containers.py
```diff
@@ -334,10 +339,11 @@ def dump(self, camel_case: bool = True) -> dict[str, str | dict]:
 class Index(CogniteObject, ABC):
     @classmethod
     def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
-        if resource["indexType"] == "btree":
-            return cast(Self, BTreeIndex.load(resource))
-        if resource["indexType"] == "inverted":
-            return cast(Self, InvertedIndex.load(resource))
+        if isinstance(resource, dict):
```
erlendvollset, line 342: Same comment. This should always be a dict

cognite/client/data_classes/filters.py
```diff
@@ -89,7 +90,10 @@ def dump(self, camel_case_property: bool = False) -> dict[str, Any]:
 
     @classmethod
     def load(cls, filter_: dict[str, Any]) -> Filter:
-        (filter_name,) = filter_
+        if len(filter_) > 1:
+            warnings.warn(f"Filter {filter_} has more than one key, only the first key will be used", stacklevel=2)
+
+        filter_name = next(iter(filter_))
```
erlendvollset, line 96: This won't necessarily pick the right key. See how it was fixed for `Aggregation` here https://github.com/cognitedata/cognite-sdk-python/commit/43e4d2883e0b90aa5f3f304086a68d815f1528e8

cognite/client/data_classes/data_modeling/instances.py
doctrino: I realize that this might be overkill, but if any thing else than an object is returned here then the current code would break without this extra check.
image

cognite/client/data_classes/filters.py
doctrino: Good point, fixed.

cognite/client/data_classes/data_modeling/instances.py
erlendvollset: That won't happen, don't need to take that into consideration.

cognite/client/data_classes/_base.py
```diff
@@ -191,7 +191,9 @@ def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None =
         return cls(resource)
 
     def dump(self, camel_case: bool = True) -> dict[str, Any]:
-        return convert_all_keys_recursive(self.__data, camel_case=camel_case)
+        if isinstance(self.__data, dict):
```
erlendvollset, line 194: Not necessary, this should always be a dict. If it can be anything else, then the `_init_` and `load` signatures should be updated.

erlendvollset (commented):
Let's get rid of all the redundant `isinstance(resource, dict)` checks

cognite/client/data_classes/filters.py
```diff
@@ -89,20 +90,20 @@ def dump(self, camel_case_property: bool = False) -> dict[str, Any]:
 
     @classmethod
     def load(cls, filter_: dict[str, Any]) -> Filter:
-        (filter_name,) = filter_
-        filter_body = filter_[filter_name]
+        if len(filter_) > 1:
+            warnings.warn(f"Filter {filter_} has more than one key, only the first key will be used", stacklevel=2)
```
erlendvollset, line 94: Don't really see the value of this warning - can be removed?

cognite/client/data_classes/filters.py
```diff
@@ -89,20 +89,17 @@ def dump(self, camel_case_property: bool = False) -> dict[str, Any]:
 
     @classmethod
     def load(cls, filter_: dict[str, Any]) -> Filter:
-        (filter_name,) = filter_
-        filter_body = filter_[filter_name]
-
-        if filter_name == And._filter_name:
+        if filter_body := filter_.get(And._filter_name):
```
erlendvollset, line 92: Hmm. I think 'and' and 'or' filters can return empty arrays, or we might allow it at some point in the future. So this needs to be more precise with its truthiness check.

tests/tests_unit/test_base.py
```diff
@@ -250,6 +267,44 @@ def test_load_has_no_side_effects(
         # Ensure no part of load mutates the input dict:
         assert dumped == original_dumped
 
+    # These properties are guaranteed to not change from object in the API,
```
erlendvollset, line 270: Not quite following this. Any object in the spec will remain an object, why won't we inject extra keys into these particular ones?

doctrino: The tests below dumps all CogniteObject instances and injects new parameters in any dictionary it finds in the dumped object, including in nested structures. If you do this in a container object, you will find the properties dictionary and inject into than, which is the reason for all the previous `isinstance(, dict)` checks. 

These are the exceptions where we are not injecting an arbitrary parameter to check that it does not break `_load(...)`.

PR #1739, opened 2024-04-23
Add `DiagramDetectConfig` to `DiagramsAPI.detect()`

ddonukis:
## Description
The diagram detect endpoint has a lot of configuration options that can be passed as a dict. But for that the user needs to know precisely which parameters are available and how they are spelled. To simplify configuring the detect job I've added a config data class. Of course the endpoint still supports raw dicts for backwards compatibility.

```python
from congite.client.data_classes.contextualization import DiagramDetectConfig, CustomizeFuzziness, ConnectionFlags

config = DiagramDetectConfig(
    read_embedded_text=True,
    customize_fuzziness=CustomizeFuzziness(
        min_chars=6,
    ),
    connection_flags=ConnectionFlags(natural_reading_order=True)
)

client.diagrams.detect(entities, file_ids, configuration=config)
```

## Checklist:
- [x] Tests added/updated.


olacognite (approved):
LGTM, mind that there is already a new parameter in town. Substitutions.

ddonukis:
> LGTM, mind that there is already a new parameter in town. Substitutions.

Added it. Also updated the param descriptions in line with the API spec PR.

cognite/client/data_classes/contextualization.py
```diff
@@ -1036,3 +1038,176 @@ def __init__(
 
 class ResourceReferenceList(CogniteResourceList[ResourceReference]):
     _RESOURCE = ResourceReference
+
+
+class NestableDiagramDetectConfig:
```
erlendvollset, line 1043: Let's make these classes extend `CogniteObject`.

tests/tests_unit/test_data_classes/test_contextualization.py
```diff
@@ -181,3 +188,62 @@ def test_DetectJobBundle_div(self) -> None:
...
+        expected = {
+            "annotationExtract": True,
+            "caseSensitive": False,
+            "directionWeights": {"right": 0.1, "down": 0.1},
+            "minFuzzyScore": 0.95,
+            "removeLeadingZeros": True,
+            "newParameterStr": "something",
+            "newParamaterDict": {"my_key": "my_value"},
+        }
+        assert config.dump() == expected
```
erlendvollset, line 236: You'll get these tests generated when extending `CogniteObject`.

ddonukis: Ohh! That's convenient.

cognite/client/data_classes/contextualization.py
ddonukis: I answered in more detail in a different thread, but it doesn't seem like a good fit.

Although I just remembered the API returns this config as part of the detect job result 🤔 
Maybe we could load the results into this object then. I'll investigate further.

ddonukis:
@erlendvollset made the config classes inherit from `CogniteObject` as you suggested 👌

cognite/client/_api/diagrams.py
```diff
@@ -230,14 +237,31 @@ def detect(
...
         entities = [
             entity.dump(camel_case=True) if isinstance(entity, CogniteResource) else entity for entity in entities
         ]
         beta_parameters = {}
         if pattern_mode is not None or configuration is not None:
-            beta_parameters = dict(pattern_mode=pattern_mode, configuration=configuration)
+            config = configuration.dump() if isinstance(configuration, DiagramDetectConfig) else configuration
+            beta_parameters = dict(pattern_mode=pattern_mode, configuration=config)
+            self._detect_beta_params_warning.warn()
+            self._api_subversion = "beta"
```
ddonukis, line 264: @haakonvt does this look correct? I saw the datapoints endpoint appends `"-beta"` to subversion but setting subversion to `"beta"` seems to work for `digrams.detect`.

haakonvt: The API versioning guide is very clear on what format it should be:
https://cognitedata.atlassian.net/wiki/spaces/PD/pages/3842900086/CDF+API+versioning

olacognite: I think beta is equivalent to <latest>-beta, but better safe than sorry maybe

ddonukis: Oh, thanks for sharing! I wasn't aware of that doc.

Will change to append '-beta'.

cognite/client/_api/diagrams.py
```diff
@@ -230,14 +240,31 @@ def detect(
...
```
ddonukis, line 267:
```suggestion
            if self._api_subversion and not self._api_subversion.endswith("-beta"):
                self._api_subversion += "-beta"
```

ddonukis:
Updated API subversion logic. Any other comments or can this be merged?

cognite/client/_api/diagrams.py
ddonukis: Hmm, I'm not sure why the integration test is saying the api subversion is bad now 🤔

haakonvt: @olacognite `Invalid version in cdf-version header`... 😉

haakonvt: (Just double-check that it is in fact sending `"20230101-beta"`)

cognite/client/data_classes/contextualization.py
```diff
@@ -1036,3 +1038,213 @@ def __init__(
...
+    @classmethod
+    def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
+        resouce_copy = resource.copy()
+
+        if con_flg := resouce_copy.pop("connectionFlags", None):
+            con_flg = ConnectionFlags.load(con_flg)
+        if cus_fuz := resouce_copy.pop("customizeFuzziness", None):
+            cus_fuz = CustomizeFuzziness.load(cus_fuz)
+        if dir_wgt := resouce_copy.pop("directionWeights", None):
+            dir_wgt = DirectionWeights.load(dir_wgt)
```
haakonvt, line 1241: Perhaps we can change `pop` (mutates) to `get` (and skip initial copy)?
```suggestion
        if con_flg := resouce_copy.get("connectionFlags"):
            con_flg = ConnectionFlags.load(con_flg)
        if cus_fuz := resouce_copy.pop("customizeFuzziness"):
            cus_fuz = CustomizeFuzziness.load(cus_fuz)
        if dir_wgt := resouce_copy.pop("directionWeights"):
            dir_wgt = DirectionWeights.load(dir_wgt)
```

cognite/client/data_classes/contextualization.py
```diff
@@ -1036,3 +1038,213 @@ def __init__(
...
+
+        snake_cased = convert_all_keys_to_snake_case(resouce_copy)
```
haakonvt, line 1243: Perhaps just pass these on without any conversion?

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "cognite-sdk"
 
-version = "7.39.0"
+version = "7.39.1"
```
haakonvt, line 4: I think this deserves a minor bump 👌

cognite/client/data_classes/contextualization.py
```diff
@@ -1036,3 +1038,213 @@ def __init__(
...
+class DirectionWeights(CogniteObject):
+    """Direction weights for the text graph that control how far text boxes can be one from another
+    in a particular direction before they are no longer connected in the same graph. Lower value means
+    larger distance is allowed.
+
+    Args:
+        left (float | None): explanation
+        right (float | None): explanation
+        up (float | None): explanation
+        down (float | None): explanation
```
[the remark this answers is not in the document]
ddonukis: not self explanatory enough? 😜

yeah, I forgot about it

cognite/client/data_classes/contextualization.py
```diff
@@ -1036,3 +1038,213 @@ def __init__(
...
+    def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
+        return cls(
+            left=resource.get("left", None),
+            right=resource.get("right", None),
+            up=resource.get("up", None),
+            down=resource.get("down", None),
+        )
+
+
+class CustomizeFuzziness(CogniteObject):
```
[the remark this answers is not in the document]
ddonukis: It's a bit of an unfortunate name, straight from the context api.

ddonukis:
@haakonvt  thanks for the review, I'll fix next week 👍

cognite/client/data_classes/contextualization.py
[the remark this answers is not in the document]
ddonukis: I thought about changing the name and I believe it's better to keep it consistent with the API for the sake of user experience. Now the name matches the API field exactly so it's easy to look it up in the docs and code.

The name really should changed on the API side which talked to Ola about. He doesn't like the name either.

cognite/client/data_classes/contextualization.py
```diff
@@ -1036,3 +1038,213 @@ def __init__(
...
+                    f"Provided parameter name `{param_name}` collides with a known parameter `{known}`. Please use it insead."
+                )
+            else:
+                self._extra_params[param_name] = value
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        dumped = super().dump(camel_case=camel_case)
+
+        for param_name, v in self._extra_params.items():
+            dumped[to_camel_case(param_name) if camel_case else param_name] = v
```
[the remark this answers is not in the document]
ddonukis: I don't understand what you mean. I'm just respecting the `camel_case` param. It's true by default anyways. If for some reason the user wants to dump the object in snake case I think it's more weird that some of the config options are converted and some are not.

I'm thinking that the user can stick to the pythonic snake_case when using the config class but could also use camelCase, so we don't really know which case the use will provide when initializing the class.

When loading we expect the values to come from the api, in camelCase, since it's not arbitrary metadata but part of the api schema, albeit not documented.

cognite/client/data_classes/contextualization.py
```diff
@@ -1036,3 +1038,212 @@ def __init__(
...
+                    remove_keys.append(param_name)
+
+        for param_name in remove_keys:
+            del dumped[param_name]
+
+        return dumped
+
+    @classmethod
+    def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
+        resource_copy = resource.copy()
```
ddonukis, line 1233: @haakonvt I do the copy and pop because I expect undocumented parameters to be in this resource dict.

If I use get and then pass the original dict into kwargs, the parsed and loaded params like `ConnectionFlags` will be passed to `cls()` twice, causing a TypeError.

I need some way to separate known loadable params from unknown ones, so I decided to use `pop()` to remove the items we need to load separately.

ddonukis:
@haakonvt I fixed the subversion handling and the failing test (it was modifying subversion separately, so api received subversion `"beta-beta"`). Also addressed you other comments.

I'm pretty happy with how the PR looks now. What do you think?

cognite/client/data_classes/contextualization.py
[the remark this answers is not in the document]
haakonvt: Ah, I see. This still has to be passed as `customize_fuzziness=CustomizeFuzziness(...)` so I think you are free to choose a better name for it 😄 

Maybe @erlendvollset thinks otherwise

cognite/client/data_classes/contextualization.py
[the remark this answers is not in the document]
haakonvt: All I'm trying to say is that a roundtrip using `to_camel_case` and `to_snake_case` can only be considered a no-op for well-behaved strings, that's why I think you should not pass them through either of these functions, not in load and not in dump.

If a user is passing these super-custom settings, they need to account for CorrectCasing themselves. Just my 2 cents (I'll leave the decision up to you!)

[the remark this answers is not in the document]
haakonvt: (A bit like how we in data modeling aren't case converting properties)

cognite/client/_api/diagrams.py
```diff
@@ -230,14 +243,34 @@ def detect(
...
+        api_subversion = None
         beta_parameters = {}
         if pattern_mode is not None or configuration is not None:
-            beta_parameters = dict(pattern_mode=pattern_mode, configuration=configuration)
+            config = configuration.dump() if isinstance(configuration, DiagramDetectConfig) else configuration
+            beta_parameters = dict(pattern_mode=pattern_mode, configuration=config)
+
+            self._detect_beta_params_warning.warn()
+            if self._api_subversion and not self._api_subversion.endswith("beta"):
+                api_subversion = f"{self._api_subversion}-beta"
```
haakonvt, line 273: Is this calling `beta` even when no beta features are used?

haakonvt: ...or is passing `configuration` always beta perhaps

haakonvt (approved):
LGTM!

PR #1740, opened 2024-04-23
fix test pollution

haakonvt:


doctrino (approved):
Is it the `@chached_property` that was causing the issue?

haakonvt:
> Is it the `@chached_property` that was causing the issue?

The mappingproxy object it stored

PR #1750, opened 2024-05-05
Bugfix: limited dps queries yielding more than limit

haakonvt:
Oneliner fix of a bug introduced in #1425 .

### Fixed
- An edge case when a request for datapoints from several hundred time series (with specific finite limits) would return
  more datapoints than the user-specified limit.

## Checklist:
- [x] Tests added/updated.


doctrino (approved):

erlendvollset (approved):
LGTM

PR #1763, opened 2024-05-10
Fix load text property type

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.


cognite/client/data_classes/data_modeling/data_types.py
```diff
@@ -73,7 +73,7 @@ def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None =
         type_ = resource["type"]
         obj: Any
         if type_ == "text":
-            obj = Text(is_list=resource["list"], collation=resource["collation"])
+            obj = Text(is_list=resource["list"], collation=resource.get("collation", "ucs_basic"))
```
nodegard, line 76: **issue:** `is_list` should also be optional, and the same applies to all the other types. Also wouldn't it be best to not set the default value in the SDK and just rely on the API to set the default?

nodegard: Nevermind "ucs_basic" is already defined as the default type for collation in `Text` so disregard the original comment but don't need to use that here

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.43.2] - 2024-05-10
+### Fixed
+- In containers, `PropertyType` `Text` required parameter `collation` is now optional, matching the API spec.
```
erlendvollset, line 22:
```suggestion
- In containers, `PropertyType` `Text` required parameter `collation` is now optional when `load()`ing, matching the API spec.
```

erlendvollset (approved):
LGTM

PR #1766, opened 2024-05-15
bugfix: always convert 'expires_in' to float

haakonvt:
### Fixed
- Identity providers that return `expires_in` as a string no longer cause `TypeError` when authenticating.


erlendvollset (approved):
LGTM

PR #1769, opened 2024-05-16
Improve Capability.load

haakonvt:
## Description
1. Greatly improve error messages when loading a `Capability` fails (https://cognitedata.atlassian.net/browse/CDF-21155)
2. Change `Scope.load` to also use the `allow_unknown` pattern and future-proof with no expectation of the presence of a single key.

## Example
Before
```
from cognite.client.data_classes.capabilities import Capability

Capability.load(
    {'EventsAcl': {'actions': ['READ', 'WRITE'], 'scope': {'datasetScope': {'ids': [12378]}}}},
    allow_unknown=False,
)
```
Raises -> `ValueError: 'READ' is not a valid Capability.Action`

Now, you get quite the mouthful: (list has been shortened here)
```
ValueError: Unable to parse Capability, none of the top-level keys in the input, ['EventsAcl'],
    matched known ACLs, - or - multiple was found. Pass `allow_unknown=True` to force
    loading it as an unknown capability. List of known ACLs:
    ['analyticsAcl', 'annotationsAcl', ..., 'workflowOrchestrationAcl'].
```

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/capabilities.py
+++ b/cognite/client/data_classes/capabilities.py
@@ -190,7 +190,7 @@ class Capability(ABC):
             try:
                 scopes = Capability.Scope.load(resource[name]["scope"], allow_unknown)
             except Exception as err:
-                raise ValueError(f"Could not instantiate {capability_cls.__name__} due to: {err}") from None
+                raise ValueError(f"Could not instantiate {capability_cls.__name__} due to: {err}")
 
             return cast(
                 Self,
--- a/cognite/client/data_classes/capabilities.py
+++ b/cognite/client/data_classes/capabilities.py
@@ -116,12 +116,13 @@ class Capability(ABC):
         _scope_name: ClassVar[str]
 
         @classmethod
-        def load(cls, resource: dict | str) -> Self:
+        def load(cls, resource: dict | str, allow_unknown: bool = False) -> Self:
             resource = resource if isinstance(resource, dict) else load_yaml_or_json(resource)
-            for cls_name, scope_cls in _SCOPE_CLASS_BY_NAME.items():
-                if cls_name not in resource:
-                    continue
-                data = convert_all_keys_to_snake_case(resource[cls_name])
+            known_scopes = set(resource).intersection(_SCOPE_CLASS_BY_NAME)
+            if len(known_scopes) == 1:
+                (name,) = known_scopes
+                scope_cls = _SCOPE_CLASS_BY_NAME[name]
+                data = convert_all_keys_to_snake_case(resource[name])
                 try:
                     return cast(Self, scope_cls(**data))
                 except TypeError:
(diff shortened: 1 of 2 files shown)


cognite/client/data_classes/capabilities.py
```diff
@@ -176,31 +187,39 @@ def load(cls, resource: dict | str, allow_unknown: bool = False) -> Self:
             (name,) = known_acls
             capability_cls = _CAPABILITY_CLASS_BY_NAME[name]
             # TODO: We ignore extra keys that might be present like 'projectUrlNames' - are these needed?
+            try:
+                scopes = Capability.Scope.load(resource[name]["scope"], allow_unknown)
+            except Exception as err:
+                raise ValueError(f"Could not instantiate {capability_cls.__name__} due to: {err}") from None
```
erlendvollset, line 193: why not like this? In my experience, getting the traceback for the wrapped exception is more useful than just getting the `exception.message` field.
```suggestion
                raise ValueError(f"Could not instantiate {capability_cls.__name__}") from err
```

erlendvollset (approved):
Good stuff! Just one nitpick

cognite/client/data_classes/capabilities.py
```diff
@@ -176,31 +187,39 @@ def load(cls, resource: dict | str, allow_unknown: bool = False) -> Self:
...
+                    allow_unknown=True,
+                )
+            return cast(Self, unknown_acl)
+
         raise ValueError(
-            "Unable to parse capability from API-response, please create an issue on Github: "
-            "https://github.com/cognitedata/cognite-sdk-python"
+            f"Unable to parse Capability, none of the top-level keys in the input, {sorted(resource)}, "
+            f"matched known ACLs, - or - multiple was found. Pass `allow_unknown=True` to force loading it "
+            f"as an unknown capability. List of known ACLs: {sorted(_CAPABILITY_CLASS_BY_NAME)}."
```
doctrino, line 222: Could we use the `builin` `difflib` to give a shorter l list the most relevant?

cognite/client/data_classes/capabilities.py
haakonvt: Yeah, I agree. But I think I should have just not used the from-syntax at all (the first error is not the direct cause of the one I'm raising, right). Will fix

cognite/client/data_classes/capabilities.py
doctrino: The reason for this is that I think a typical issue here are typos, thus it would be useful to have a curated list of what could be the intintion.

haakonvt: Something like the top 3 most similar ACLs by name?

cognite/client/data_classes/capabilities.py
haakonvt, line 193:
```suggestion
            except Exception as err:
                raise ValueError(f"Could not instantiate {capability_cls.__name__} due to: {err}")
```

cognite/client/data_classes/capabilities.py
haakonvt: Makes sense to me, could also introduce and reference a constant that lists all?

doctrino: That is a great idea, then you avoid spam and allows the user to check it.

PR #1772, opened 2024-05-20
add timestamp format utility function intended for GraphQL

CogniteJohan:
## Description

## Checklist:
- [x] Tests added/updated.


CogniteJohan:
@haakonvt , can you have a look at the failed tests? Seems to be a problem with access tokens for the integration tests, so I am innocent!

haakonvt:
> @haakonvt , can you have a look at the failed tests? Seems to be a problem with access tokens for the integration tests, so I am innocent!

Outdated credentials I see. Erlend will update later today 👍

tests/tests_unit/test_utils/test_time.py
```diff
@@ -35,6 +36,15 @@
     import pandas
 
 
+class TestDatetimeToGqlTimestamp:
+    def test_datetime_to_gql_timestamp_correct_type(self):
+        assert datetime_to_gql_timestamp(datetime(2021, 1, 1, 0, 0, 0, 0)) == "2021-01-01T00:00:00.000"
+
+    def test_datetime_to_gql_timestamp_incorrect_type(self):
+        with pytest.raises(TypeError):
+            datetime_to_gql_timestamp("2021-01-01T00:00:00.000")
```
haakonvt, line 45: Let’s add a test with a tz aware date time as well!

cognite/client/utils/__init__.py
```diff
@@ -18,4 +19,12 @@
...
 
-__all__ = ["ZoneInfo", "MAX_TIMESTAMP_MS", "MIN_TIMESTAMP_MS", "datetime_to_ms", "ms_to_datetime", "timestamp_to_ms"]
+__all__ = [
+    "ZoneInfo",
+    "MAX_TIMESTAMP_MS",
+    "MIN_TIMESTAMP_MS",
+    "datetime_to_ms",
+    "ms_to_datetime",
+    "timestamp_to_ms",
+    "datetime_to_gql_timestamp",
```
haakonvt, line 29: Let's also add this to the docs! A bonus would be to incorporate it in a docstring code example 🍾

tests/tests_unit/test_utils/test_time.py
CogniteJohan: Good catch!

CogniteJohan:
@haakonvt in the Checklist it is written "If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring. "  But there is no such rst file anymore. Should we update the "pull_request_template.md" to specify the necessary steps to get documentation generated?

cognite/client/utils/__init__.py
CogniteJohan: Done for the new function. Examples for the other functions is out of scope for this PR

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+
+## [7.44.0] - 2024-05-24
```
haakonvt, line 21:
```suggestion
## [7.44.0] - 2024-05-24
```

cognite/client/_api/data_modeling/instances.py
```diff
@@ -623,6 +623,24 @@ def apply(
...
+                >>> from datetime import datetime
+                >>> my_date = datetime(2020, 3, 14, 15, 9, 26, 535000)
+                >>> client = CogniteClient()
+                >>> node = NodeApply(
+                ...     space="actors",
+                ...     external_id="arnold_schwarzenegger",
+                ...     sources=[
+                ...         NodeOrEdgeData(
+                ...             ViewId("mySpace", "PersonView", "v1"),
+                ...             {"name": "Piday", "date": datetime_to_ms_iso_timestamp(my_date)}
```
haakonvt, line 640: Nice!

cognite/client/utils/_time.py
```diff
@@ -60,6 +60,10 @@ def _import_zoneinfo_not_found_error() -> type[ZoneInfoNotFoundError]:
     return ZoneInfoNotFoundError
 
 
+def _datetime_is_tz_unaware(dt: datetime) -> bool:
+    return dt.tzinfo is None or dt.tzinfo.utcoffset(dt) is None
```
haakonvt, line 64: Why the last part? It is still aware if it is set to UTC imo
```suggestion
def _datetime_is_tz_unaware(dt: datetime) -> bool:
    return dt.tzinfo is None
```

cognite/client/utils/_time.py
```diff
@@ -102,6 +106,26 @@ def ms_to_datetime(ms: int | float) -> datetime:
...
+        dt (datetime): Naive or aware datetime object. Naive datetimes are interpreted as local time.
+
+    Raises:
+        TypeError: If dt is not a datetime object
+
+    Returns:
+        str: Timestamp string in ISO 8601 format with milliseconds
+    """
+    if isinstance(dt, datetime):
+        if _datetime_is_tz_unaware(dt):
```
haakonvt, line 122: Imo you dont need this function

haakonvt (approved):
A few questions/suggestions, but looks good!

tests/tests_unit/test_utils/test_time.py
```diff
@@ -35,6 +37,30 @@
     import pandas
 
 
+class TestDatetimeToIsoTimestamp:
```
haakonvt, line 40:
```suggestion
class TestDatetimeToMsIsoTimestamp:
```

tests/tests_unit/test_utils/test_time.py
```diff
@@ -35,6 +37,30 @@
...
+    @pytest.mark.skipif(platform.system() == "Windows", reason="Overriding timezone is too much hassle on Windows")
+    def test_datetime_to_iso_timestamp_timezone_unaware(self):
```
haakonvt, line 42: With the class approach, you then dont need to repeat it here:
```suggestion
    def test_timezone_unaware(self):
```

cognite/client/utils/_time.py
CogniteJohan: I would agree if it was one simple check, but as we discussed earlier we need two which is not obvious :) And I think this could be useful for other functions as well. But I should have called it "naive" instead of "unaware". Agree?

cognite/client/utils/_time.py
```diff
@@ -60,6 +60,10 @@ def _import_zoneinfo_not_found_error() -> type[ZoneInfoNotFoundError]:
     return ZoneInfoNotFoundError
 
 
+def _datetime_is_tz_naive(dt: datetime) -> bool:
+    return dt.tzinfo is None or dt.tzinfo.utcoffset(dt) is None
```
haakonvt, line 64: Feel free to remove the leading underscore, this lives in `_time.py` so everything is implied internal/private already (those public we explicitly expose)

cognite/client/utils/_time.py
```diff
@@ -102,6 +106,26 @@ def ms_to_datetime(ms: int | float) -> datetime:
...
+
+    Returns:
+        str: Timestamp string in ISO 8601 format with milliseconds
+    """
+    if isinstance(dt, datetime):
+        if _datetime_is_tz_naive(dt):
+            return dt.astimezone().isoformat(timespec="milliseconds")
+        return dt.isoformat(timespec="milliseconds")
+    else:
+        raise TypeError(f"Expected datetime object, got {type(dt)}")
```
haakonvt, line 126: Suggested code golfing 😅 
```suggestion
        if _datetime_is_tz_naive(dt):
            dt = dt.astimezone()
        return dt.isoformat(timespec="milliseconds")
    raise TypeError(f"Expected datetime object, got {type(dt)}")
```

tests/tests_unit/test_utils/test_time.py
```diff
@@ -35,6 +37,30 @@
...
+class TestDatetimeToMsIsoTimestamp:
+    @pytest.mark.skipif(platform.system() == "Windows", reason="Overriding timezone is too much hassle on Windows")
+    def test_timezone_unaware(self):
+        input_datetime = datetime(2021, 1, 1, 0, 0, 0, 0)
+        with tmp_set_envvar("TZ", "CET"):
+            time.tzset()
+            assert datetime_to_ms_iso_timestamp(input_datetime) == "2021-01-01T00:00:00.000+01:00"
+
+    def test_timezone_cet(self):
+        input_datetime = datetime(2021, 1, 1, 0, 0, 0, 0, tzinfo=pytz.timezone("CET"))
```
haakonvt, line 49: This example - and if the pytz example has been cloned from somewhere else - should be changed, reason being:

> The main pitfall of using the `pytz` package is that it requires you to use the `localize` method to correctly attach timezone information to naive `datetime` objects. Simply setting the `tzinfo` attribute directly often leads to incorrect timezone behavior, such as wrong offset calculations due to daylight saving time transitions.

Example:
```
import datetime
import pytz

eastern = pytz.timezone("US/Eastern")

user_date = eastern.localize(datetime.datetime(2020, 4, 15, 6, 0, 0))
utc_date = user_date.astimezone(pytz.utc)
print(utc_date)  # 2020-04-15 10:00:00+00:00
```

cognite/client/_api/data_modeling/instances.py
haakonvt, line 640: I just saw that the number of examples here is quite lengthy already, should we do something like just showcase the `datetime_to_ms_iso_timestamp` function?

tests/tests_unit/test_utils/test_time.py
```diff
@@ -35,6 +37,30 @@
...
+
+    @pytest.mark.skipif(platform.system() == "Windows", reason="Overriding timezone is too much hassle on Windows")
+    def test_timezone_cet_in_local_tz(self):
+        input_datetime = datetime(2021, 1, 1, 0, 0, 0, 0, tzinfo=pytz.timezone("CET"))
+        with tmp_set_envvar("TZ", "UTC"):
+            time.tzset()
+            assert datetime_to_ms_iso_timestamp(input_datetime) == "2021-01-01T00:00:00.000+01:00"
+
+    def test_incorrect_type(self):
+        with pytest.raises(TypeError):
```
haakonvt, line 60: This should get a `match` since TypeError is "generic" and might come from somewhere else than your raise statement

tests/tests_unit/test_utils/test_time.py
```diff
@@ -35,6 +37,30 @@
     import pandas
 
 
+class TestDatetimeToMsIsoTimestamp:
+    @pytest.mark.skipif(platform.system() == "Windows", reason="Overriding timezone is too much hassle on Windows")
```
haakonvt, line 41: If this `skipif` is used for all tests in this class, I think you can change it to a class decorator instead to avoid repetition!

cognite/client/utils/_time.py
haakonvt: I agree!

cognite/client/utils/_time.py
CogniteJohan: Agree. The original reason was testing the tz awareness more easily :)

tests/tests_unit/test_utils/test_time.py
CogniteJohan: Absolutely. Fixed

tests/tests_unit/test_utils/test_time.py
CogniteJohan: Something like this?

tests/tests_unit/test_utils/test_time.py
CogniteJohan: Some tests do not use this, and those can be run on Windows (with fingers crossed...)

tests/tests_unit/test_utils/test_time.py
CogniteJohan: Yes!

cognite/client/utils/_time.py
CogniteJohan: Agree

cognite/client/_api/data_modeling/instances.py
CogniteJohan: Something like this?

cognite/client/_api/data_modeling/instances.py
```diff
@@ -623,6 +623,13 @@ def apply(
                 ...     auto_create_end_nodes=True
                 ... )
 
+            Using helper function to create valid graphql timestamp for a datetime object:
+
+                >>> from cognite.client.utils import datetime_to_ms_iso_timestamp
+                >>> from datetime import datetime
+                >>> import pytz
+                >>> my_date = datetime(2020, 3, 14, 15, 9, 26, 535000, tzinfo=pytz.utc)
```
haakonvt, line 631: This felt much better!

I just think we should change `pytz` to built-in:
```suggestion```suggestion
                >>> from datetime import datetime, timezone
                >>> my_date = datetime(2020, 3, 14, 15, 9, 26, 535000, tzinfo=timezone.utc)
```

haakonvt, line 631: This felt much better! I just think we should change `pytz` to built-in:
```suggestion
                >>> from datetime import datetime, timezone
                >>> my_date = datetime(2020, 3, 14, 15, 9, 26, 535000, tzinfo=timezone.utc)
```

tests/tests_unit/test_utils/test_time.py
```diff
@@ -8,6 +8,7 @@
 from unittest import mock
 
 import pytest
+import pytz
```
haakonvt, line 11: I thought you had to move this inside something marked with "dsl", but if it works it works 😅

CogniteJohan: I have absolutely no idea... But it seems to test properly, so 🙄

PR #1774, opened 2024-05-21
prepare DatapointsAPI for tz and calendar queries

haakonvt:


doctrino (approved):
Great simplification.

PR #1775, opened 2024-05-22
add `environment:CD` back into release wf

haakonvt:
Attempt at fixing PyPI releases 403ing

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -41,7 +41,7 @@ jobs:
         env:
           TWINE_USERNAME: __token__
           TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
-        run: twine upload --verbose dist/* || echo 'Version probably exists'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -40,7 +41,7 @@ jobs:
         env:
           TWINE_USERNAME: __token__
           TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
-        run: twine upload --verbose dist/* || echo 'Version exists'
+        run: twine upload --verbose dist/* || echo 'Version probably exists'
```
doctrino, line 44: Lets actually get a failed workflow if this fails for any other reason than the version exist.
```suggestion
        run: twine upload --skip-existing --verbose dist/*
```

doctrino (approved):

.github/workflows/release.yaml
haakonvt: Oooooh like it 🤩

PR #1781, opened 2024-05-24
skip json decoding attempt when expecting protobuf (or other)

haakonvt:
## Description
When expecting e.g. `protobuf`, it causes a significant slowdown to attempt json decoding on every request as was introduced in #1573. This PR changes the default behavior to only do so when `accept=application/json`.

Some timings on current `master`, around ~400 ms:

Screenshot 2024-05-24 at 15 09 33

Some timings on this branch, around ~260 ms:

Screenshot 2024-05-24 at 15 09 21

----

# Note: Draft, tests not updated!


cognite/client/_http_client.py
```diff
@@ -107,24 +107,28 @@ def __init__(
         self.refresh_auth_header = refresh_auth_header
         self.retry_tracker_factory = retry_tracker_factory  # needed for tests
 
-    def request(self, method: str, url: str, **kwargs: Any) -> requests.Response:
+    def request(self, method: str, url: str, accept: str, **kwargs: Any) -> requests.Response:
         retry_tracker = self.retry_tracker_factory(self.config)
         headers = kwargs.get("headers")
+        accepts_json = accept == "application/json"
+        is_auto_retryable = False
```
haakonvt, line 114: Should this be inside the `while` loop?

erlendvollset: It can be outside, `is_auto_retryable` is always static for a given endpoint.

erlendvollset (approved):
LGTM!

cognite/client/_http_client.py
```diff
@@ -107,24 +107,28 @@ def __init__(
         self.refresh_auth_header = refresh_auth_header
         self.retry_tracker_factory = retry_tracker_factory  # needed for tests
 
-    def request(self, method: str, url: str, **kwargs: Any) -> requests.Response:
+    def request(self, method: str, url: str, accept: str, **kwargs: Any) -> requests.Response:
```
erlendvollset, line 110: Would be nice to get away from the **kwargs here and in `_do_request`. Would be slightly more verbose, but wouldn't need to deal with this untyped dict, which would make more approachable IMO

PR #1784, opened 2024-05-27
Fix type hint to transformations.preview

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/transformations/__init__.py
+++ b/cognite/client/_api/transformations/__init__.py
@@ -502,7 +502,7 @@ class TransformationsAPI(APIClient):
         self,
         query: str | None = None,
         convert_to_string: bool = False,
-        limit: int = 100,
+        limit: int | None = 100,
         source_limit: int | None = 100,
         infer_schema_limit: int | None = 1000,
     ) -> TransformationPreviewResult:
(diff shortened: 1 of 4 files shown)


cognite/client/_api/transformations/__init__.py
```diff
@@ -502,7 +502,7 @@ def preview(
         self,
         query: str | None = None,
         convert_to_string: bool = False,
-        limit: int = 100,
+        limit: int | None = 100,
```
haakonvt, line 505: - Does it work to pass None for all?

- Also, I see that the API spec uses 10k for inferSchemaLimit instead of the 1k in the SDK. Perhaps update as well? 

- Looks like `timeout` i also missing from the params

doctrino: Yes it work for `None`, will fix the other issues.

haakonvt: Cool, I wonder why we call it preview if you can page through unlimited results? 🤔

haakonvt (approved):

cognite/client/_api/transformations/__init__.py
doctrino: I think because you are actually not writing the results to the destination.

PR #1798, opened 2024-06-05
Fix typehint transformation notification

doctrino:
## Description

Just want to become best friends with mypy 😜

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/transformations/notifications.py
+++ b/cognite/client/data_classes/transformations/notifications.py
@@ -55,11 +55,11 @@ class TransformationNotification(TransformationNotificationCore):
         cognite_client: CogniteClient | None = None,
     ) -> None:
         super().__init__(destination)
-        self.id = id
-        self.transformation_id = transformation_id
-        self.transformation_external_id = transformation_external_id
-        self.created_time = created_time
-        self.last_updated_time = last_updated_time
+        self.id = cast(int, id)
+        self.transformation_id = cast(int, transformation_id)
+        self.transformation_external_id = cast(str, transformation_external_id)
+        self.created_time = cast(int, created_time)
+        self.last_updated_time = cast(int, last_updated_time)
         self._cognite_client = cast("CogniteClient", cognite_client)
 
     def __hash__(self) -> int:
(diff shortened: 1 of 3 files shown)


cognite/client/data_classes/transformations/notifications.py
```diff
@@ -55,11 +55,11 @@ def __init__(
...
-        self.id = id
-        self.transformation_id = transformation_id
-        self.transformation_external_id = transformation_external_id
-        self.created_time = created_time
-        self.last_updated_time = last_updated_time
+        self.id = cast(int, id)
+        self.transformation_id = cast(int, transformation_id)
+        self.transformation_external_id = cast(str, transformation_external_id)
+        self.created_time = cast(int, created_time)
+        self.last_updated_time = cast(int, last_updated_time)
```
haakonvt, line 62: Is this the same as Erlend did for many of the read classes some time ago? E.g. for `class Asset`:

```py
# id/created_time/last_updated_time are required when using the class to read,
        # but don't make sense passing in when creating a new object. So in order to make the typing
        # correct here (i.e. int and not Optional[int]), we force the type to be int rather than
        # Optional[int].
        # TODO: In the next major version we can make these properties required in the constructor
        self.id: int = id  # type: ignore
        self.created_time: int = created_time  # type: ignore
        self.last_updated_time: int = last_updated_time  # type: ignore
```

doctrino: Yes, updated.

cognite/client/data_classes/transformations/notifications.py
```diff
@@ -55,11 +55,16 @@ def __init__(
         cognite_client: CogniteClient | None = None,
     ) -> None:
         super().__init__(destination)
-        self.id = id
-        self.transformation_id = transformation_id
-        self.transformation_external_id = transformation_external_id
-        self.created_time = created_time
-        self.last_updated_time = last_updated_time
+        # id/created_time/last_updated_time are required when using the class to read,
```
haakonvt, line 58: Nit: list all or write «e.g.»

haakonvt (approved):
LGTM

PR #1804, opened 2024-06-10
Fix reset dataset for 3d model update

msusag:
## Description

This fixes resetting the dataSetId for 3d model updates.

Previously , trying to use ThreeDModelUpdate with a `data_set_id.set(None)` would lead to a invalid request sent to 3d api.

The dataSetId property is a "primitive" field, not an object, so the fix is to use the correct return type for it, CognitePrimitiveUpdate.

## Checklist:
- [x] Tests added/updated.


erlendvollset (approved):
LGTM

PR #1806, opened 2024-06-12
Fix raw rows to pandas (keep all rows)

haakonvt:
### Fixed
- Converting rows (`RowList` and `RowListWrite`) to a pandas DataFrame no longer silently drops rows that do not have
  any columnar data.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/raw.py
+++ b/cognite/client/data_classes/raw.py
@@ -105,6 +105,13 @@ class RowWrite(RowCore):
         columns (dict[str, Any]): Row data stored as a JSON object.
     """
 
+    def __init__(
+        self,
+        key: str,
+        columns: dict[str, Any],
+    ) -> None:
+        super().__init__(key, columns)
+
     @classmethod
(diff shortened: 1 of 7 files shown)


cognite/client/data_classes/raw.py
```diff
@@ -130,7 +130,10 @@ def to_pandas(self) -> pandas.DataFrame:  # type: ignore[override]
             pandas.DataFrame: The pandas DataFrame representing this instance.
         """
         pd = local_import("pandas")
-        return pd.DataFrame.from_dict(OrderedDict((d.key, d.columns) for d in self.data), orient="index")
+        if self:
+            index, data = zip(*((row.key, row.columns) for row in self))
+            return pd.DataFrame.from_records(data, index=index)
+        return pd.DataFrame(columns=[], index=[])
```
erlendvollset, line 136: Nitpick: flip em'!
```suggestion
        if not self:
            return pd.DataFrame(columns=[], index=[])
        index, data = zip(*((row.key, row.columns) for row in self))
        return pd.DataFrame.from_records(data, index=index)

```

erlendvollset (approved):

PR #1809, opened 2024-06-14
timezone and calendar aggregates -> GA

haakonvt:
### Changed
- DatapointsAPI support for timezones and calendar-based aggregates reaches general availability (GA).
### Deprecated
- The function `DatapointsAPI.retrieve_dataframe_in_tz` is deprecated. Use the other retrieve methods instead
  and pass in `timezone`.

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -6,6 +6,7 @@ import heapq
 import itertools
 import math
 import time
+import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from itertools import chain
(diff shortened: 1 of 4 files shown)


cognite/client/_api/datapoints.py
```diff
@@ -1177,6 +1172,15 @@ def retrieve_dataframe_in_tz(
                 ...     start=datetime(2020, 1, 1, tzinfo=ZoneInfo("America/New_York")),
                 ...     end=datetime(2022, 12, 31, tzinfo=ZoneInfo("America/New_York")))
         """
+        warnings.warn(
+            (
+                "This SDK function is deprecated and will be removed in the next major release. Reason: Cognite Data "
```
haakonvt, line 1177: Note to self: Should prob mention name specifically

matiascognite (approved):
I haven't actually tested any of this, so take my review with a grain of salt, but the code looks good to me! 🚀

PR #1820, opened 2024-06-18
Typed nodes and edges

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/data_modeling/instances.py
+++ b/cognite/client/_api/data_modeling/instances.py
@@ -58,6 +58,8 @@ from cognite.client.data_classes.data_modeling.instances import (
     NodeApplyResultList,
     NodeList,
     SubscriptionContext,
+    T_Edge,
+    T_Node,
     TargetUnit,
 )
 from cognite.client.data_classes.data_modeling.query import (
@@ -66,16 +68,9 @@ from cognite.client.data_classes.data_modeling.query import (
     SourceSelector,
 )
 from cognite.client.data_classes.data_modeling.typed_instances import (
-    T_Edge,
-    T_Node,
     TypedEdge,
-    TypedEdgeList,
-    TypedEdgeWrite,
     TypedInstance,
-    TypedInstancesResult,
     TypedNode,
-    TypedNodeList,
-    TypedNodeWrite,
 )
 from cognite.client.data_classes.data_modeling.views import View
 from cognite.client.data_classes.filters import _BASIC_FILTERS, Filter, _validate_filter
@@ -100,7 +95,7 @@ Source: TypeAlias = Union[SourceSelector, View, ViewId, Tuple[str, str], Tuple[s
 
 
 class _NodeOrEdgeResourceAdapter(Generic[T_Node, T_Edge]):
-    def __init__(self, node_cls: type[T_Node] | type[Node], edge_cls: type[T_Edge] | type[Edge]):
+    def __init__(self, node_cls: type[T_Node], edge_cls: type[T_Edge]):
         self._node_cls = node_cls
         self._edge_cls = edge_cls
 
@@ -112,9 +107,8 @@ class _NodeOrEdgeResourceAdapter(Generic[T_Node, T_Edge]):
 
 
 class _TypedNodeOrEdgeListAdapter:
-    def __init__(self, instance_cls: type, typed_list: type[TypedNodeList] | type[TypedEdgeList]) -> None:
+    def __init__(self, instance_cls: type) -> None:
         self._instance_cls = instance_cls
-        self._typed_list = typed_list
 
     def __call__(self, items: Any, cognite_client: CogniteClient | None = None) -> Any:
         return self._typed_list(items, cognite_client)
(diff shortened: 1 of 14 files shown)


cognite/client/data_classes/data_modeling/typed_instances.py
```diff
@@ -0,0 +1,172 @@
...
+)
+from cognite.client.data_classes.data_modeling.ids import ContainerId, EdgeId, NodeId, ViewId
+from cognite.client.data_classes.data_modeling.instances import PropertyValueWrite
+from cognite.client.utils._text import to_camel_case
+
+if TYPE_CHECKING:
+    from cognite.client import CogniteClient
+
+
+class InstanceProperty:
```
erlendvollset, line 22: I'd suggest this name instead
```suggestion
class PropertyOptions:
```

cognite/client/data_classes/data_modeling/typed_instances.py
```diff
@@ -0,0 +1,172 @@
...
+    """This is a descriptor class for instance properties in a typed class.
+
+    It is used when you have a property that has a different name in the Data Model
+    compared to the name in the Python class.
+    """
+
+    def __init__(self, alias: str | None = None) -> None:
```
erlendvollset, line 29: It's not really an alias, the attribute name is really the alias and this is the actual identifier. Maybe `identifer` or `identifier_override`?
```suggestion
    def __init__(self, identifier: str | None = None) -> None:
```

cognite/client/data_classes/data_modeling/typed_instances.py
```diff
@@ -0,0 +1,172 @@
...
+            if key in self._instance_properties:
+                continue
+            if isinstance(value, Iterable) and not isinstance(value, (str, dict)):
+                properties[key] = [self._serialize_value(v, camel_case) for v in value]
+            else:
+                properties[key] = self._serialize_value(value, camel_case)
+        return properties
+
+    @staticmethod
+    def _serialize_value(value: PropertyValueWrite, camel_case: bool) -> str | int | float | bool | dict | list:
```
erlendvollset, line 101: Let's extract and reuse the function in data_classes/data_modeling/instances.py

cognite/client/data_classes/data_modeling/typed_instances.py
```diff
@@ -0,0 +1,172 @@
...
+            if camel_key in resource:
+                args[key] = resource[camel_key]
+        return args
+
+    @classmethod
+    def _load_properties(cls, resource: dict[str, Any]) -> dict[str, Any]:
+        raise NotImplementedError()
+
+
+class TypedNodeWrite(TypedInstanceWrite, ABC):
```
erlendvollset, line 135: Would be good with some unit tests for a subclass of this, since we won't get any generated when these are abstract

cognite/client/_api/data_modeling/instances.py
```diff
@@ -638,13 +639,13 @@ def apply(
...
-        nodes = nodes if isinstance(nodes, Sequence) else [nodes]
+        nodes = nodes if isinstance(nodes, Sequence) else [nodes]  # type: ignore[assignment]
 
         edges = edges or []
-        edges = edges if isinstance(edges, Sequence) else [edges]
+        edges = edges if isinstance(edges, Sequence) else [edges]  # type: ignore[assignment]
 
         res = self._create_multiple(
-            items=cast(Sequence[WriteableCogniteResource], (*nodes, *edges)),
+            items=cast(Sequence[WriteableCogniteResource], (*nodes, *edges)),  # type: ignore[misc]
```
erlendvollset, line 648: Hmm. Should TypedInstanceWrite actually inherit from `WriteableCogniteResource` instead of `CogniteResource`?

cognite/client/_api/data_modeling/instances.py
```diff
@@ -638,13 +639,13 @@ def apply(
             "replace": replace,
         }
         nodes = nodes or []
-        nodes = nodes if isinstance(nodes, Sequence) else [nodes]
+        nodes = nodes if isinstance(nodes, Sequence) else [nodes]  # type: ignore[assignment]
 
         edges = edges or []
-        edges = edges if isinstance(edges, Sequence) else [edges]
+        edges = edges if isinstance(edges, Sequence) else [edges]  # type: ignore[assignment]
```
erlendvollset, line 645: Why did these suddenly start acting up? 🤔

erlendvollset (approved):
Very nice! I really like this!

erlendvollset:
Would be great with an example in the docs aswell!

cognite/client/_api/data_modeling/instances.py
doctrino: No, I don't think so. Writable is for the read version of the resource.

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -389,7 +401,16 @@ def __contains__(self, attr: str) -> bool:
             self.__raise_if_non_singular_source(attr)
 
     def dump(self, camel_case: bool = True) -> dict[str, Any]:
-        dumped = super().dump(camel_case)
```
doctrino, line 392: We cannot call the default dump method, as the TypedNode/Edge, will have properties directly on the class, which the default method just serialize in place.

cognite/client/data_classes/data_modeling/core.py
```diff
@@ -77,54 +70,6 @@ def __init__(
         self.description = description
 
 
-class DataModelingInstancesList(WriteableCogniteResourceList, Generic[T_WriteClass, T_WritableCogniteResource], ABC):
```
doctrino, line 80: Probably need to add an alias here.

PR #1824, opened 2024-06-20
Typed Core Model

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.


cognite/client/_api_client.py
```diff
@@ -1109,7 +1109,7 @@ def _prepare_item_chunks(
         return [
             {"items": chunk, **(extra_body_fields or {})}
             for chunk in split_into_chunks(
-                [it.dump(camel_case=True) if isinstance(it, CogniteResource) else it for it in items],
+                [it.dump(camel_case=True) if isinstance(it, CogniteObject) else it for it in items],
```
doctrino, line 1112:
```suggestion
                [it.dump(camel_case=True) if isinstance(it, CogniteResource) else it for it in items],
```

CHANGELOG.md
```diff
@@ -18,6 +18,15 @@ Changes are grouped as follows
 - `Security` in case of vulnerabilities.
 
 
+## [7.53.2] - 2024-07-03
```
doctrino, line 21: Patch bump because this is an alpha feature.

cognite/client/data_classes/cdm/__init__.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import annotations
+
+from cognite.client.config import global_config
+from cognite.client.utils._experimental import FeaturePreviewWarning
+
+FeaturePreviewWarning("alpha", "alpha", "Core Data Model").warn()
+
+if global_config.cdm_model_version == "v1":
```
erlendvollset, line 8: I don't think we should do this. In migration scenarios it will be useful to import multiple versions at the same time. And I don't really think it's worth introducing complexity just to avoid being explicit about which version to import.

CHANGELOG.md
```diff
@@ -18,6 +18,15 @@ Changes are grouped as follows
...
+### Fixed
+- If you derived from `TypedNode` or `TypedEdge`, and then derived the `load` method would not include the parent
+  class properties. Same if you used multiple inheritance. This is now fixed.
+
+### Added
+
+- [Feature Preview - alpha] Core Model, available from `cognite.client.data_classes import cdm`.
```
erlendvollset, line 28:
```suggestion
- [Feature Preview - alpha] Core Model, available `from cognite.client.data_classes import cdm`. 
```

cognite/client/data_classes/data_modeling/typed_instances.py
```diff
@@ -230,17 +230,33 @@ def _load_properties(
...
+        for name, value in current_cls.__dict__.items():
+            if isinstance(value, PropertyOptions):
+                property_by_name[name] = value
+        to_search.extend(
+            [
+                b
+                for b in current_cls.__bases__
+                if {b not in {object, TypedNode, TypedEdge, Edge, Node, TypedNodeApply, TypedEdgeApply}}
+            ]
+        )
```
erlendvollset, line 246: Let's add unit tests to `tests_unit/test_data_classes/test_data_models/test_typed_instances.py` for these cases.

cognite/client/data_classes/cdm/__init__.py
erlendvollset: (explicit is good!)

tests/utils.py
```diff
@@ -83,7 +84,9 @@ def all_concrete_subclasses(base: T_Type) -> list[T_Type]:
     return [
         sub
         for sub in all_subclasses(base)
-        if all(base is not abc.ABC for base in sub.__bases__) and not inspect.isabstract(sub)
+        if all(base is not abc.ABC for base in sub.__bases__)
+        and not inspect.isabstract(sub)
+        and all(parent not in {TypedNodeApply, TypedEdgeApply, TypedEdge, TypedNode} for parent in sub.__mro__)
```
erlendvollset, line 89: Why shouldn't these be tested?

doctrino: They should, they actually showed a bug. A was thinking these should be treated differently.

doctrino: Added a comment on it now, but it is that the `FakeCogniteResourceGenerator` does not support the descriptors.

cognite/client/data_classes/cdm/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+    "Object3DApply",
+    "Schedulable",
+    "SchedulableApply",
+    "Sourceable",
+    "SourceableApply",
+    "TimeSeriesBase",
+    "TimesSeriesBaseApply",
+]
+
+FeaturePreviewWarning("alpha", "alpha", "Core Data Model").warn()
```
erlendvollset, line 54: Let's remove these imports and move the FeaturePreviewWarning into `cdm/v1.py`. When v2 comes along, it will have many naming collisions with these, so it's best to just keep their namespaces separate.

erlendvollset (approved):
LGTM

PR #1828, opened 2024-06-24
Extending NodeListWithCursor

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/data_modeling/instances.py
+++ b/cognite/client/data_classes/data_modeling/instances.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import threading
 import warnings
 from abc import ABC, abstractmethod
-from collections import defaultdict
+from collections import UserList, defaultdict
 from collections.abc import Iterable
 from dataclasses import dataclass
 from datetime import date, datetime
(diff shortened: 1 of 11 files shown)


tests/tests_unit/test_data_classes/test_data_models/test_instances.py
```diff
@@ -161,6 +162,32 @@ def test_dump_and_load(self) -> None:
         }
 
 
+class TestNodeListWithCursor:
```
erlendvollset, line 165: Add test for EdgeListWithCursor too?

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -995,6 +1008,19 @@ def __init__(
...
 
+    def extend(self, other: EdgeListWithCursor) -> None:  # type: ignore[override]
+        if not isinstance(other, type(self)):
+            raise ValueError("Unable to extend as the types do not match")
+        other_res_list = type(self)(other, other.cursor)  # See if we can accept the types
+        if set(self._id_to_item).isdisjoint(other_res_list._id_to_item):
+            # Skip super() as we need the original type.
+            UserList.extend(self, other)
+            self._external_id_to_item.update(other_res_list._external_id_to_item)
+            self._id_to_item.update(other_res_list._id_to_item)
```
erlendvollset, line 1019: Edges and nodes don't have internal ids, so this is redundant.

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -995,6 +1008,19 @@ def __init__(
         super().__init__(resources, cognite_client)
         self.cursor = cursor
 
+    def extend(self, other: EdgeListWithCursor) -> None:  # type: ignore[override]
+        if not isinstance(other, type(self)):
+            raise ValueError("Unable to extend as the types do not match")
+        other_res_list = type(self)(other, other.cursor)  # See if we can accept the types
+        if set(self._id_to_item).isdisjoint(other_res_list._id_to_item):
```
erlendvollset, line 1015: `_id_to_item` will always be empty, so this will always return `True`, need to use `_external_id_to_item`. Let's add unit tests for duplicates in extends.

Also 
```
set(self._external_id_to_item).isdisjoint(other_res_list._external_id_to_item)
```
is about 10 times slower than
```
self._external_id_to_item.keys().isdisjoint(other_res_list._external_id_to_item)
```

erlendvollset (changes requested):

cognite/client/data_classes/data_modeling/instances.py
doctrino: Nice did not know that one, learned something new.

PR #1834, opened 2024-07-03
Make it safe to run test_auto_apply_create_nodes in parallel

erlendvollset:
## Description

## Checklist:
- [ ] Tests added/updated.


tests/tests_integration/test_api_client.py
```diff
@@ -392,7 +392,7 @@ def test_retrieve_multiple__ordering_matches_input(
     ) -> None:
         # Between SDK version 7.0.0 and 7.33.1, ordering of results was broken when >> 1k elements
         # was requested (meaning multiple requests were used):
-        event_ids = cognite_client.events.list(limit=1000).as_ids()
+        event_ids = cognite_client.events.list(limit=10).as_ids()
```
haakonvt, line 395: I think this value should be 2-3x (at least) of the retrieve limit to not break what the test is for. Hey from the rainy cabin 😅

erlendvollset: Ack. Problem is that it keeps listing out stuff that is getting deleted by the time the byids call is made, but we can sort by descending createdTime to only get events which are not likely to be part of a current test run.

tests/tests_unit/test_api/test_data_modeling/test_containers.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    rsps.add(rsps.POST, url_pattern, status=200, json=response_body)
+    yield rsps
+
+
+class TestContainersApi:
+    def test_apply_retrieve_and_delete_index(
+        self, cognite_client: CogniteClient, mock_containers_response: Any, mock_delete_index_response: Any
+    ) -> None:
+        new_container = ContainerApply(
```
doctrino, line 74: I am almost in favor of removing this test. I think by making it an unit test we only testing that the serialization is correct, which should be tested on the data class, and that we call the correct endpoint in the API method. If there are any changes to the endpoint, it would give a false sense of security.

doctrino (commented):
Wonder if we should put some effort into smoke testing for tests like these, to ensure that the SDK is working.

tests/tests_unit/test_api/test_data_modeling/test_containers.py
erlendvollset: It's not the python sdk tests' responsibility to ensure that the service is working correctly. We have smoke tests for all services. The pysdk tests should ensure that we correctly integrate with the API. Any changes to the endpoint should be backwards compatible.

doctrino (approved):

tests/tests_integration/test_api/test_assets.py
```diff
@@ -114,7 +114,7 @@ def is_integration_test() -> Filter:
 class TestAssetsAPI:
     def test_get(self, cognite_client):
         res = cognite_client.assets.list(limit=1)
-        assert res[0] == cognite_client.assets.retrieve(res[0].id)
+        assert res[0].id == cognite_client.assets.retrieve(res[0].id).id
```
haakonvt, line 117: I still don't understand what/how this test fails; is it perhaps xdist related? 🤔

PR #1835, opened 2024-07-03
[CDF-21996] Make sdk data workflows test more robust

VerstraeteBert:
## Description
Initially, only a single execution was created for the test suite, which always was successful before the fixture dependant tests ran. Now, however, there's multiple executions and some are not terminal yet. To fix the test, we just select the first completed (i.e., now immutable) execution.


erlendvollset (approved):
Lgtm

PR #1837, opened 2024-07-05
Better error message filter

doctrino:
## Description
Some user struggled to understand this error message:
image

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/filters.py
+++ b/cognite/client/data_classes/filters.py
@@ -252,6 +252,8 @@ class CompoundFilter(Filter, ABC):
     _filter_name = "compound"
 
     def __init__(self, *filters: Filter) -> None:
+        if not_flt := [flt for flt in filters if not isinstance(flt, Filter)]:
+            raise TypeError(f"One or more invalid filters, expected Filter, got: {not_flt}")
         self._filters = filters
 
     def _filter_body(self, camel_case_property: bool) -> list | dict:
--- a/cognite/client/data_classes/filters.py
+++ b/cognite/client/data_classes/filters.py
@@ -193,10 +193,11 @@ class Filter(ABC):
         output = {type(self)}
         if isinstance(self, CompoundFilter):
             for filter_ in self._filters:
-                output.update(filter_._involved_filter_types())
-        elif isinstance(self, Filter):
-            return output
-        raise ValueError(f"Unknown filter type: {type(self)}. Expected Filter got {type(self)}")
+                if isinstance(filter_, Filter):
+                    output.update(filter_._involved_filter_types())
+                else:
+                    raise ValueError(f"Unknown filter type: {type(filter_)}. Expected Filter got {type(filter_)}")
+        return output
 
 
 class UnknownFilter(Filter):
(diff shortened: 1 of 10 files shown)


haakonvt:
So they passed in a set when the type says `Filter` or `dict`? Hmm......

haakonvt:
Had a look again at your screenshot. This won't fix the user issue as the validate is called before "merge-space-into-filter" (which you have modified)

cognite/client/data_classes/filters.py
```diff
@@ -238,9 +238,12 @@ def _validate_filter(
...
         return
-    if not_supported := (filter._involved_filter_types() - supported_filters):
-        names = [f.__name__ for f in not_supported]
-        raise ValueError(f"The filters {names} are not supported for {api_name}")
+    elif isinstance(filter, Filter):
+        if not_supported := (filter._involved_filter_types() - supported_filters):
+            names = [f.__name__ for f in not_supported]
+            raise ValueError(f"The filters {names} are not supported for {api_name}")
+    else:
+        raise ValueError(f"Invalid filter type: {type(filter)}. Must be Filter or dict")
```
haakonvt, line 246: This only raises if the top level filter is incorrect. Using e.g. `filter=And(MatchAll(), set())`, still gives the same error:
`AttributeError: 'set' object has no attribute '_involved_filter_types'`

doctrino: I think that is less likely user error to do, but added a check for that as well.

haakonvt: Nice

cognite/client/data_classes/filters.py
```diff
@@ -193,7 +193,10 @@ def _involved_filter_types(self) -> set[type[Filter]]:
         output = {type(self)}
         if isinstance(self, CompoundFilter):
             for filter_ in self._filters:
-                output.update(filter_._involved_filter_types())
+                if isinstance(filter_, Filter):
+                    output.update(filter_._involved_filter_types())
+                else:
+                    raise ValueError(f"Unknown filter type: {type(filter_)}. Expected Filter got {type(filter_)}")
```
haakonvt, line 199: Should this rather be checked in `CompoundFilter.__init__`?
```suggestion
                    raise TypeError(f"Unknown filter type: {type(filter_)}")
```

cognite/client/data_classes/filters.py
```diff
@@ -238,9 +241,12 @@ def _validate_filter(
...
         return
-    if not_supported := (filter._involved_filter_types() - supported_filters):
-        names = [f.__name__ for f in not_supported]
-        raise ValueError(f"The filters {names} are not supported for {api_name}")
+    elif isinstance(filter, Filter):
+        if not_supported := (filter._involved_filter_types() - supported_filters):
+            names = [f.__name__ for f in not_supported]
+            raise ValueError(f"The filters {names} are not supported for {api_name}")
+    else:
+        raise ValueError(f"Invalid filter type: {type(filter)}. Must be Filter or dict")
```
haakonvt, line 249:
```suggestion
    else:
        raise TypeError(f"Invalid filter type: {type(filter)}. Must be Filter or dict")
```

haakonvt (approved):
Pushed the changes we talked about 😄

PR #1691, opened 2024-03-24
Instances: Include typing

doctrino:
## Description
Introduce `TypeInformation` class to both match the spec, and also provide useful helper methods such that the data structure becomes easier to work with (compared to a triple nested dict)

## Checklist:
- [ ] Tests added/updated.
- [ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [ ] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ ] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #1839, opened 2024-07-06
Retrieve/list responses with more than items

doctrino:
## Description
This is a preparation for #1691 

### Problem
The `_list` and `_retrieve_multiple` methods does account for any other than `items` in the response object:
image

This PR gives the option to get the `typing` key from the response.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -549,7 +549,7 @@ class APIClient:
             )
 
         resource_path = resource_path or self._RESOURCE_PATH
-        items: list[T_CogniteResource] = []
+        items: T_CogniteResourceList | None = None
         for resource_list in self._list_generator(
             list_cls=list_cls,
             resource_cls=resource_cls,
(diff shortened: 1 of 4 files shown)


cognite/client/_api_client.py
```diff
@@ -483,12 +492,24 @@ def _list_generator(
                     else:
                         unprocessed_items = []
                     for chunk in chunks:
-                        yield list_cls._load(chunk, cognite_client=self._cognite_client)
+                        if list_cls._support_dict_load:
+                            yield list_cls._load(
+                                {"items": chunk, **first_response},  # type: ignore[dict-item]
```
haakonvt, line 497: I think that assuming the other top-level keys always are static across responses - and thus only the first is needed - is not an assumption we can make here. (For type info it obviously works perfectly 😄 )

cognite/client/_api_client.py
```diff
@@ -545,8 +566,11 @@ def _list(
             advanced_filter=advanced_filter,
             api_subversion=api_subversion,
         ):
-            items.extend(cast(T_CogniteResourceList, resource_list).data)
-        return list_cls(items, cognite_client=self._cognite_client)
+            if items is None:
+                items = cast(T_CogniteResourceList, resource_list)
+            else:
+                items.data.extend(cast(T_CogniteResourceList, resource_list).data)
+        return items if items is not None else list_cls([], cognite_client=self._cognite_client)
```
haakonvt, line 573: I don't understand what's going on here? You change `items` initial value from empty list to `None` in order to...?

doctrino: Originally items was a regular list which was then passed into the list_cls. 

This means we lose extra information stored in the list, namely, typing.

doctrino:
Solved by alternative approach in #1843 instead

PR #1841, opened 2024-07-09
InstanceID for TimeSeries

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.


cognite/client/data_classes/_base.py
```diff
@@ -503,7 +503,7 @@ def dump(self, camel_case: Literal[True] = True) -> dict[str, Any]:
 
     @classmethod
     @abstractmethod
-    def _get_update_properties(cls) -> list[PropertySpec]:
+    def _get_update_properties(cls, item: CogniteResource | None = None) -> list[PropertySpec]:
```
doctrino, line 506: This change is to allow the updated properties to be different depending on whether InstanceId is passed or not.

cognite/client/data_classes/time_series.py
```diff
@@ -39,6 +41,7 @@ class TimeSeriesCore(WriteableCogniteResource["TimeSeriesWrite"], ABC):
 
     Args:
         external_id (str | None): The externally supplied ID for the time series.
+        instance_id (NodeId | None): No description.
```
doctrino, line 44:
```suggestion
        instance_id (NodeId | None): The ID of the instance this time series belongs to.
```

cognite/client/data_classes/time_series.py
```diff
@@ -86,6 +91,54 @@ def __init__(
...
 
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        """Dump the object to a dictionary"""
+        output: dict[str, Any] = {}
+        if self.external_id is not None:
+            output["externalId" if camel_case else "external_id"] = self.external_id
+        if self.instance_id is not None:
+            output["instanceId" if camel_case else "instance_id"] = self.instance_id.dump(
+                camel_case=camel_case, include_instance_type=False
+            )
```
doctrino, line 102: Need manual implementation of load/dump methods now as instance id introduces a nested structure.

cognite/client/data_classes/time_series.py
```diff
@@ -390,21 +493,31 @@ def data_set_id(self) -> _PrimitiveTimeSeriesUpdate:
...
-            PropertySpec("unit"),
-            PropertySpec("unit_external_id"),
-            PropertySpec("asset_id"),
-            PropertySpec("description"),
-            PropertySpec("is_step", is_nullable=False),
-            PropertySpec("security_categories", is_container=True),
-            PropertySpec("data_set_id"),
-        ]
+    def _get_update_properties(cls, item: CogniteResource | None = None) -> list[PropertySpec]:
+        if isinstance(item, (TimeSeries, TimeSeriesWrite)) and item.instance_id:
```
haakonvt, line 497: Could we instead have a method that says `_get_update_properties_dms` (or something)? I think the caller of the function should decide which update properties to get rather than this method deciding for itself.

cognite/client/utils/_identifier.py
```diff
@@ -314,6 +317,16 @@ def load(
                 raise TypeError(
                     f"{id_name}external_ids must be of type str or SequenceNotStr[str]. Found {type(external_ids)}"
                 )
+        if instance_ids is not None:
+            if isinstance(instance_ids, InstanceId):
+                value_passed_as_primitive = True
+                all_identifiers.append(instance_ids)
+            elif isinstance(instance_ids, Sequence):
+                all_identifiers.extend(instance_ids)
```
haakonvt, line 325: The other steps before this one enforce int/str before extending, so I think we should enforce `InstanceId` here

cognite/client/data_classes/time_series.py
doctrino: The default implementation only accepts basic basic types not nested. The moment you need nested you need to implement load/dump. It is possible to write this

```
        """Dump the object to a dictionary"""
        output = super().dump(camel_case=camel_case)
        if self.instance_id is not None:
            output["instanceId" if camel_case else "instance_id"] = self.instance_id.dump(
                camel_case=camel_case, include_instance_type=False
            )
        return output
```

but I have a preference for the explicit method here.

cognite/client/_api/files.py
```diff
@@ -919,7 +919,7 @@ def download(
     @staticmethod
     def _get_ids_filepaths_directories(
         directory: Path,
-        id_to_metadata: dict[str | int, FileMetadata],
+        id_to_metadata: dict[str | int | InstanceId, FileMetadata],
```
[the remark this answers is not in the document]
doctrino: Motivation for doing it now is `mypy` complains, but though it was ok as the next PR will add `InstanceId` for files

cognite/client/_api/time_series.py
```diff
@@ -224,25 +229,32 @@ def retrieve(self, id: int | None = None, external_id: str | None = None) -> Tim
                 >>> client = CogniteClient()
                 >>> res = client.time_series.retrieve(external_id="1")
         """
-        identifiers = IdentifierSequence.load(ids=id, external_ids=external_id).as_singleton()
+        headers: dict | None = None
+        if instance_id is not None:
+            self._use_alpha()
+            headers = {"cdf-version": "alpha"}
```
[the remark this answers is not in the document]
doctrino: `api_subversion` is another part of the header. They are combined 
image

cognite/client/_api/time_series.py
```diff
@@ -612,12 +637,19 @@ def update(
                 >>> my_update = TimeSeriesUpdate(id=1).description.set("New description").metadata.add({"key": "value"})
                 >>> res = client.time_series.update(my_update)
         """
+        headers: dict | None = None
+        if (isinstance(item, Sequence) and any(ts.instance_id for ts in item)) or (
+            isinstance(item, (TimeSeries, TimeSeriesWrite, TimeSeriesUpdate)) and item.instance_id
+        ):
```
haakonvt, line 643: I don't see you do
`isinstance(item, (TimeSeries, TimeSeriesWrite, TimeSeriesUpdate))`
for the elements in the sequence - does this mean we can remove the check, or should it be added both places?

haakonvt (approved):
2 final comments, first I leave up to you (api guidelines on alpha header), but the second I think should be adressed 🤙 

Pre-approving though

cognite/client/_api/time_series.py
doctrino: Good catch the instance check was added before Instance ID was supported in TimeSeriesUpdate.

PR #1843, opened 2024-07-10
Support loading non-items in API responses

haakonvt:
## Description
Adds the possibility of loading the full response(s) - and hence any other data than items - to `_retrieve_multiple` and `_list` via `_list_generator_raw_responses`.

It also simplifies and abstracts out a lot of details from `_list_generator` to ease future dev work.

### DIFF
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -453,7 +453,7 @@ class APIClient:
             next_cursor = response.get("nextCursor")
             total_retrieved += len(response["items"])
             if total_retrieved == limit or next_cursor is None:
-                if unprocessed_items:  # can only happen when using chunk_size
+                if unprocessed_items:  # may only happen when -not- yielding one-by-one
                     yield list_cls._load(unprocessed_items, cognite_client=self._cognite_client)
                 break
 
--- a/cognite/client/_api_client.py
+++ b/cognite/client/_api_client.py
@@ -427,87 +427,35 @@ class APIClient:
         advanced_filter: dict | Filter | None = None,
         api_subversion: str | None = None,
     ) -> Iterator[T_CogniteResourceList] | Iterator[T_CogniteResource]:
-        verify_limit(limit)
-        if is_unlimited(limit):
-            limit = None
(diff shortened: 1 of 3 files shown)


cognite/client/_api_client.py
```diff
@@ -418,79 +427,147 @@ def _list_generator(
...
-            next_cursor = res.json().get("nextCursor")
-            if total_items_retrieved == limit or next_cursor is None:
-                if chunk_size and unprocessed_items:
+            response = res.json()
+            yield from self._process_into_chunks(response, chunk_size, resource_cls, list_cls, unprocessed_items)
+
+            next_cursor = response.get("nextCursor")
+            total_retrieved += len(response["items"])
+            if total_retrieved == limit or next_cursor is None:
+                if unprocessed_items:  # can only happen when using chunk_size
```
doctrino, line 456: Maybe an extra sentence why this can only happen when using chunk_size. 

Is it become the `_process_into_chunck` will take care of it?

doctrino (commented):
I think this looks good and is in line with discussion we have had. Suggest waiting with merging it until the include typing PR is written on top of this.

cognite/client/_api_client.py
haakonvt: I'll change to:
# only happens when not yielding one-by-one

doctrino: That helps

doctrino (approved):

cognite/client/_api_client.py
```diff
@@ -355,6 +356,7 @@ def _retrieve_multiple(
         params: dict[str, Any] | None = None,
         executor: ThreadPoolExecutor | None = None,
         api_subversion: str | None = None,
+        settings_forcing_raw_response_loading: list[str] | None = None,
```
doctrino, line 359: Remember to update the overload methods with the new parameter :)

haakonvt: I don’t think that’s necessary when it doesn’t affect the return type?

PR #1844, opened 2024-07-11
files-api: join base-url and upload-url

HMEiding:
## Description
On Private Link projects, the `"uploadUrl"` does not contain the netloc portion of the URL. Updating `FilesAPI.upload_from_bytes()` so that the `"uploadUrl"` is joined with  `self._config.base_url` whenever the netloc portion is missing.

## Checklist:
- [x] Tests added/updated.


rbtcollins (approved):

rbtcollins:
I would prefer a unit test to demonstrate this behaviour, but I think the code change is right

doctrino (approved):

qTipTip:
> I would prefer a unit test to demonstrate this behaviour, but I think the code change is right

I added two unit tests.

PR #1846, opened 2024-07-11
add `last_updated_time_in_index` to `retrieve_dataframe`

KristianVangsnes:
## Description
Adding last_updated_time_in_index to retrieve_dataframe function, which makes index into multi-index. Also updated docstring.

## Checklist:
- [x] Tests added/updated.


haakonvt (commented):
LGTM! Let's add a test making sure it works when `len(rows)` is 0, 1, and many 👌

KristianVangsnes:
Where would be best to add the test? tests_unit/test_api/test_raw.py?

haakonvt:
> Where would be best to add the test? tests_unit/test_api/test_raw.py?

Yup!

cognite/client/_api/raw.py
```diff
@@ -756,7 +756,9 @@ def retrieve_dataframe(
         pd = local_import("pandas")
         rows = self.list(db_name, table_name, min_last_updated_time, max_last_updated_time, columns, limit, partitions)
         if last_updated_time_in_index:
-            idx = pd.MultiIndex.from_tuples([(r.key, r.last_updated_time) for r in rows], names=["key", "last_updated_time"])
+            idx = pd.MultiIndex.from_tuples(
+                [(r.key, r.last_updated_time) for r in rows], names=["key", "last_updated_time"]
+            )
```
haakonvt, line 761: Should we convert epoch-ms to datetime?

tests/tests_unit/test_api/test_raw.py
```diff
@@ -64,6 +64,47 @@ def mock_retrieve_raw_row_response(rsps, cognite_client):
...
+        rsps.GET,
+        cognite_client.raw._get_base_url_with_base_path() + "/raw/dbs/db1/tables/table1/rows",
+        status=200,
+        json=response_body,
+    )
+    yield rsps
+
+
+@pytest.fixture
+def mock_retrieve_raw_rows_response_one_rows(rsps, cognite_client):
```
haakonvt, line 85: 😅 
```suggestion
def mock_retrieve_raw_rows_response_one_row(rsps, cognite_client):
```

cognite/client/_api/raw.py
KristianVangsnes: Good questions, I don't see why not.

haakonvt (approved):
LGTM! Might need to update one or more tests after making it a timestamp? Also, please try to revert the unrelated docstring changes (unless you actually fixed something in them 👌 )

Issue #1831, opened 2024-06-26
ThreadPoolExecutor locks/hangs when getting all items from client.relationships.list


korvalds:
**System information:**
 - OS: Linux/5.15.153.1-microsoft-standard-WSL2-x86_64-64bit
 - Python Version: CPython/3.10.12
 - SDK Version: CognitePythonSDK/7.52.0 

**Describe the bug**
`ThreadPoolExecutor` locks/hangs when getting all items (`limit=-1`) from `client.relationships.list` endpoint using `partitions=10` and `source_external_ids=[...]`  when `len(source_external_id_list) > self._LIST_SUBQUERY_LIMIT`

**To Reproduce**
Runnable code reproducing the error.
```
from cognite.client import CogniteClient

client = CogniteClient()

workorders = client.events.list(
    data_set_external_ids=["some_id"],
    type="work_order",
    metadata=({"woMainAsset": "asset_name"}),
    limit=-1,
    partitions=10
).to_pandas(camel_case=True)

ops = client.relationships.list(
    data_set_external_ids=["some_id"],
    source_types=["event"],
    target_types=["event"],
    source_external_ids=workorders["externalId"].tolist(), # this is maybe 50000 ids
    limit=-1,
    partitions=10,
).to_pandas(camel_case=True)

 # Then hangs indefinitely, until killing the process:

  File "/home/dev/test_cognite_sdk_client.py", line 111, in <module>
    ops = client.relationships.list(
  File "/home/dev/.virtualenvs/core/lib/python3.10/site-packages/cognite/client/_api/relationships.py", line 353, in list
    tasks_summary = execute_tasks(
  File "/home/dev/.virtualenvs/core/lib/python3.10/site-packages/cognite/client/utils/_concurrency.py", line 300, in execute_tasks
    for fut in as_completed(futures_dct):
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 245, in as_completed
    waiter.event.wait(wait_timeout)
  File "/usr/lib/python3.10/threading.py", line 607, in wait
    signaled = self._cond.wait(timeout)
  File "/usr/lib/python3.10/threading.py", line 320, in wait
    waiter.acquire()
KeyboardInterrupt

Exception ignored in: <module 'threading' from '/usr/lib/python3.10/threading.py'>
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1537, in _shutdown
    atexit_call()
[...]

PR #1849, opened 2024-07-16
Fix deadlock in RelationshipsAPI.list

haakonvt:
From issue: #1831 

DM-1896: https://cognitedata.atlassian.net/browse/DM-1896

### Fixed
- A bug in the list method of the RelationshipsAPI that could cause a thread deadlock.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/relationships.py
+++ b/cognite/client/_api/relationships.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import itertools
+import warnings
 from functools import partial
 from typing import TYPE_CHECKING, Iterator, Literal, Sequence, overload
 
@@ -303,6 +304,7 @@ class RelationshipsAPI(APIClient):
                 method="POST",
                 limit=limit,
                 filter=filter,
+                partitions=partitions,
                 other_params={"fetchResources": fetch_resources},
             )
         if not is_unlimited(limit):
@@ -314,6 +316,8 @@ class RelationshipsAPI(APIClient):
         target_chunks = split_into_chunks(target_external_ids, self._LIST_SUBQUERY_LIMIT) or [[]]
         source_chunks = split_into_chunks(source_external_ids, self._LIST_SUBQUERY_LIMIT) or [[]]
 
+        # All sources (if any) must be checked against all targets (if any). When either is not
+        # given, we must exhaustively list all matching just the source or the target:
         for target_xids, source_xids in itertools.product(target_chunks, source_chunks):
             task_filter = filter.copy()
             if target_external_ids:  # keep null if it was
@@ -322,6 +326,12 @@ class RelationshipsAPI(APIClient):
                 task_filter["sourceExternalIds"] = source_xids
             tasks.append({"filter": task_filter})
 
+        if partitions is not None:
+            warnings.warn(
+                f"When one or both of source/target external IDs have more than {self._LIST_SUBQUERY_LIMIT} "
+                "elements, `partitions` is ignored",
+                UserWarning,
+            )
         tasks_summary = execute_tasks(
             partial(
                 self._list,
@@ -329,7 +339,7 @@ class RelationshipsAPI(APIClient):
                 resource_cls=Relationship,
                 method="POST",
                 limit=None,
-                partitions=None,
+                partitions=None,  # Otherwise, workers will spawn workers -> deadlock (singleton threadpool)
                 other_params={"fetchResources": fetch_resources},
             ),
             tasks,
--- a/cognite/client/_api/relationships.py
+++ b/cognite/client/_api/relationships.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-import copy
+import itertools
+from functools import partial
 from typing import TYPE_CHECKING, Iterator, Literal, Sequence, overload
 
 from cognite.client._api_client import APIClient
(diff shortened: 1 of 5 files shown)


cognite/client/utils/_auxiliary.py
```diff
@@ -176,19 +178,25 @@ def split_into_n_parts(seq: Sequence[T], *, n: int) -> Iterator[Sequence[T]]:
 
 
 @overload
-def split_into_chunks(collection: set | list, chunk_size: int) -> list[list]: ...
+def split_into_chunks(collection: SequenceNotStr[T], chunk_size: int) -> list[list[T]]: ...
 
 
 @overload
-def split_into_chunks(collection: dict, chunk_size: int) -> list[dict]: ...
+def split_into_chunks(collection: set[T], chunk_size: int) -> list[list[T]]: ...
```
doctrino, line 185: Can't be combined as they now return the same value?
```python
@overload
def split_into_chunks(collection: set[T] | SequenceNotStr[T], chunk_size: int) -> list[list[T]]: ...
```

cognite/client/utils/_auxiliary.py
```diff
@@ -176,19 +178,25 @@ def split_into_n_parts(seq: Sequence[T], *, n: int) -> Iterator[Sequence[T]]:
...
 
 
-def split_into_chunks(collection: set | list | dict, chunk_size: int) -> list[list] | list[dict]:
+@overload
+def split_into_chunks(collection: dict[K, T], chunk_size: int) -> list[dict[K, T]]: ...
+
+
+def split_into_chunks(
+    collection: SequenceNotStr[T] | set[T] | dict[K, T], chunk_size: int
```
doctrino, line 193: Nice type support for `SequenceNotStr`

cognite/client/_api/relationships.py
```diff
@@ -32,36 +33,6 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
         super().__init__(config, api_version, cognite_client)
         self._LIST_SUBQUERY_LIMIT = 1000
 
-    def _create_filter(
```
doctrino, line 35: Removed because it is stupid? Just do a big call to do another big call

cognite/client/_api/relationships.py
```diff
@@ -327,53 +293,50 @@ def list(
...
-            other_params={"fetchResources": fetch_resources},
+        target_external_ids, source_external_ids = target_external_ids or [], source_external_ids or []
+        if all(len(xids) <= self._LIST_SUBQUERY_LIMIT for xids in (target_external_ids, source_external_ids)):
+            return self._list(
+                list_cls=RelationshipList,
+                resource_cls=Relationship,
+                method="POST",
+                limit=limit,
+                filter=filter,
+                other_params={"fetchResources": fetch_resources},
```
doctrino, line 306: Missing `partitions`. Or is that one of the problems?

cognite/client/_api/relationships.py
```diff
@@ -327,53 +293,50 @@ def list(
...
+        target_chunks = split_into_chunks(target_external_ids, self._LIST_SUBQUERY_LIMIT) or [[]]
+        source_chunks = split_into_chunks(source_external_ids, self._LIST_SUBQUERY_LIMIT) or [[]]
+
+        for target_xids, source_xids in itertools.product(target_chunks, source_chunks):
+            task_filter = filter.copy()
+            if target_external_ids:  # keep null if it was
+                task_filter["targetExternalIds"] = target_xids
+            if source_external_ids:
+                task_filter["sourceExternalIds"] = source_xids
+            tasks.append({"filter": task_filter})
```
doctrino, line 323: Maybe a quick comment that for each target chunk you need to find match for each source chunk.

cognite/client/_api/relationships.py
```diff
@@ -327,53 +293,50 @@ def list(
...
+
+        tasks_summary = execute_tasks(
+            partial(
+                self._list,
+                list_cls=RelationshipList,
+                resource_cls=Relationship,
+                method="POST",
+                limit=None,
+                partitions=None,
```
doctrino, line 332: What is the reason partitions does not work for unlimited?

doctrino (commented):
Main issue is the unused `partition` parameter.

cognite/client/utils/_auxiliary.py
haakonvt: Yes, nice 👌

cognite/client/_api/relationships.py
haakonvt: Yeah, was completely unneccessary 😅

cognite/client/_api/relationships.py
haakonvt: I didn't change that code, but you are absolutely right about it being missing - will add!

cognite/client/_api/relationships.py
haakonvt: That is the whole reason for this PR; each of these parallel workers would spawn new parallel workers on the same thread pool, leading to a deadlock.

haakonvt: I could add a warning that partitions is ignored?

haakonvt: Decided to add a warning

doctrino (approved):

PR #1853, opened 2024-07-19
auto-expand properties in notebooks

haakonvt:
### Changed
- Instance classes like `Node` and `NodeList` now expand properties by default in notebook-like environments.

https://cognitedata.atlassian.net/browse/DM-1911


doctrino (approved):
Nice

PR #1859, opened 2024-07-23
feat: adding from_yaml to CogniteClient

nodegard:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_cognite_client.py
+++ b/cognite/client/_cognite_client.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
+import os
+from pathlib import Path
+from string import Template
 from typing import Any
 
+import yaml
 from requests import Response
 
 from cognite.client._api.annotations import AnnotationsAPI
@@ -29,7 +33,7 @@ from cognite.client._api.units import UnitAPI
 from cognite.client._api.vision import VisionAPI
 from cognite.client._api.workflows import WorkflowAPI
 from cognite.client._api_client import APIClient
-from cognite.client.config import ClientConfig, global_config
+from cognite.client.config import ClientConfig, GlobalConfig, global_config
 from cognite.client.credentials import CredentialProvider, OAuthClientCredentials, OAuthInteractive
 from cognite.client.utils._auxiliary import get_current_sdk_version
 
@@ -213,3 +217,64 @@ class CogniteClient:
         """
         credentials = OAuthInteractive.default_for_azure_ad(tenant_id, client_id, cdf_cluster)
         return cls.default(project, cdf_cluster, credentials, client_name)
+
+    @classmethod
+    def from_dict(cls, config: dict[str, Any]) -> CogniteClient:
+        cognite_sdk_config_input = config.get("cognite")
+        if cognite_sdk_config_input is None:
+            raise ValueError("cognite section is missing in the configuration file")
+
+        global_config_input = cognite_sdk_config_input.get("global_config")
+        if global_config_input:
+            # TODO: set global config based on input
+            # GlobalConfig.from_dictionary(**global_config_input)
+            global_config = GlobalConfig()  # noqa: F841
+
+        client_config_input = cognite_sdk_config_input.get("client_config")
+        if client_config_input:
+            credentials_config_input = client_config_input.get("credentials")
+            if credentials_config_input is None:
+                raise ValueError("credentials section is missing in the configuration file")
+            elif isinstance(credentials_config_input, dict) and len(credentials_config_input) == 1:
+                credential_type, credential_provider_input = next(iter(credentials_config_input.items()))
+                credentials = CredentialProvider.from_config(
+                    credential_type=credential_type, config=credential_provider_input
+                )
+                client_config_input["credentials"] = credentials
+                client_config = ClientConfig(**client_config_input)
+            else:
+                raise ValueError("credentials section should contain exactly one key-value pair")
+        else:
+            raise ValueError("client_config section is missing in the configuration file")
+
+        return cls(client_config)
+
+    @classmethod
+    def from_yaml(cls, file_path: str | Path) -> CogniteClient:
+        # TODO: docstring, type hints, and error handling
+
+        file_path = Path(file_path)
+        if not file_path.is_file():
+            raise ValueError(f"File {file_path} is not a file")
+
+        try:
+            with file_path.open("r") as file_raw:
+                sub_template = Template(file_raw.read())  # FIXME: use string.Template or expand yaml.SafeLoader class
+
+                if not sub_template.is_valid():  # type: ignore[attr-defined]
+                    raise ValueError("Invalid template")
+
+                all_identifiers = sub_template.get_identifiers()  # type: ignore[attr-defined]
+                env_dict = dict(os.environ)  # FIXME: is load_dotenv() needed?
+
+                missing_env_vars = set(all_identifiers) - set(env_dict.keys())
+                if missing_env_vars:
+                    raise ValueError(f"Missing environment variables: {missing_env_vars}")
+
+                file_env_parsed = sub_template.safe_substitute(env_dict)
+
+                config_input = yaml.safe_load(file_env_parsed)
+        except yaml.YAMLError as e:
+            raise ValueError(f"Error parsing YAML file {file_path}: {e}")
+
+        return cls.from_dict(config_input)
--- a/cognite/client/credentials.py
+++ b/cognite/client/credentials.py
@@ -23,6 +23,39 @@ class CredentialProvider(Protocol):
     def authorization_header(self) -> tuple[str, str]:
         raise NotImplementedError
 
+    @classmethod
+    def from_config(cls, credential_type: str, config: Any) -> CredentialProvider:
+        """Create a CredentialProvider from a configuration dictionary.
+
+        Args:
+            credential_type (str): The type of credential provider.
+            config (Any): The input data for the credential provider.
+
+        Returns:
+            CredentialProvider: Initialized credential provider of the specified type.
+
+        Examples:
+
+                >>> from cognite.client.credentials import CredentialProvider
+                >>> credential_provider = CredentialProvider.from_config("token",  "my secret token")
+        """
+        if credential_type == "token":
+            return Token(config)
+        elif credential_type == "o_auth_client_credentials":
+            return OAuthClientCredentials(**config)
+        elif credential_type == "o_auth_client_certificate_default_for_azure_ad":  # FIXME: do we want this?
+            return OAuthClientCredentials.default_for_azure_ad(**config)
+        elif credential_type == "o_auth_interactive":
+            return OAuthInteractive(**config)
+        elif credential_type == "o_auth_interactive_default_for_azure_ad":  # FIXME: do we want this?
+            return OAuthInteractive.default_for_azure_ad(**config)
+        elif credential_type == "o_auth_device_code":
+            return OAuthDeviceCode(**config)
+        elif credential_type == "o_auth_client_certificate":
+            return OAuthClientCertificate(**config)
+        else:
+            raise ValueError(f"The provided credential type {credential_type} is not supported.")
+
 
 class Token(CredentialProvider):
     """Token credential provider
--- a/tests/tests_unit/test_cognite_client.py
+++ b/tests/tests_unit/test_cognite_client.py
@@ -1,4 +1,6 @@
 import logging
+import os
+from unittest.mock import patch
 
 import pytest
 
@@ -54,6 +56,22 @@ def mock_token_inspect(rsps) -> None:
     yield
 
 
+@pytest.fixture(scope="function")
+def set_env_vars(monkeypatch):
+    with patch.dict(os.environ, clear=True):
+        env_vars = {
+            "COGNITE_PROJECT": "test-project",
+            "COGNITE_CLIENT_NAME": "test-project",
+            "credential_type": "o_auth_client_credentials",
+            "URL": "test",
+            "COGNITE_CLIENT_SECRET": "test-client-secret",
+            "COGNITE_DEBUG": "true",
+        }
+        for key, value in env_vars.items():
+            monkeypatch.setenv(key, value)
+        yield
+
+
 class TestCogniteClient:
     def test_project_is_empty(self):
         with pytest.raises(ValueError, match="Invalid value for ClientConfig.project: <>"):
@@ -101,6 +119,31 @@ class TestCogniteClient:
         assert rsps.calls[0][0].req_kwargs["verify"] is True
         assert client._api_client._http_client_with_retry.session.verify is True
         assert client._api_client._http_client.session.verify is True
+
+    def test_client_from_yaml(self):
+        path = os.path.join(os.path.dirname(__file__), "test_config.yaml")
+        client = CogniteClient.from_yaml(path)
+        assert client.config.project == "test-project"
+        assert client.config.credentials.client_id == "test-client-id"
+        assert client.config.credentials.client_secret == "test-client-secret"
+        assert client.config.credentials.token_url == TOKEN_URL
+        assert client.config.credentials.scopes == ["https://test.com/.default", "https://test.com/.admin"]
+
+    def test_client_from_yaml_with_envs(self, set_env_vars):
+        path = os.path.join(os.path.dirname(__file__), "test_config_envs.yaml")
+
+        client = CogniteClient.from_yaml(path)
+        assert client.config.project == "test-project"
+        assert client.config.credentials.client_id == "test-client-id-with-escaped-$"
+        assert client.config.credentials.client_secret == "test-client-secret"
+        assert client.config.credentials.token_url == TOKEN_URL
+        assert client.config.credentials.scopes == ["https://test.com/.default", "https://test.com/.admin"]
+        assert client.config.debug is True
+
+    def test_client_from_yaml_missing_envs(self):
+        path = os.path.join(os.path.dirname(__file__), "test_config_envs.yaml")
+        with pytest.raises(ValueError, match="Missing environment variables: .*"):
+            CogniteClient.from_yaml(path)
 
 
 class TestInstantiateWithClient:
--- a/tests/tests_unit/test_credential_providers.py
+++ b/tests/tests_unit/test_credential_providers.py
@@ -5,6 +5,7 @@ import pytest
 from oauthlib.oauth2 import InvalidClientIdError
 
 from cognite.client.credentials import (
+    CredentialProvider,
     OAuthClientCertificate,
     OAuthClientCredentials,
     Token,
@@ -21,6 +22,17 @@ class TestToken:
         creds = Token(lambda: "abc")
         assert "Authorization", "Bearer abc" == creds.authorization_header()
 
+    def test_create_from_credential_provider(self):
+        creds = CredentialProvider.from_config(credential_type="token", config="abc")
+        assert isinstance(creds, Token)
+        assert "Authorization", "Bearer abc" == creds.authorization_header()
+
+    def test_create_from_credential_provider_invalid(self):
+        with pytest.raises(
+            TypeError, match=r"'token' must be a string or a no-argument-callable returning a string, not .*"
+        ):
+            CredentialProvider.from_config(credential_type="token", config={"foo": "bar"})
+
 
 class TestOauthClientCredentials:
     DEFAULT_PROVIDER_ARGS: ClassVar = {
@@ -28,6 +40,7 @@ class TestOauthClientCredentials:
         "client_secret": "azure-client-secret",
         "token_url": "https://login.microsoftonline.com/testingabc123/oauth2/v2.0/token",
         "scopes": ["https://greenfield.cognitedata.com/.default"],
+        "other_custom_arg": "some_value",
     }
 
     @patch("cognite.client.credentials.BackendApplicationClient")
@@ -64,6 +77,17 @@ class TestOauthClientCredentials:
         assert "Authorization", "Bearer azure_token_expired" == creds.authorization_header()
         assert "Authorization", "Bearer azure_token_refreshed" == creds.authorization_header()
 
+    def test_create_from_credential_provider(self):
+        creds = CredentialProvider.from_config(
+            credential_type="o_auth_client_credentials", config=self.DEFAULT_PROVIDER_ARGS
+        )
+        assert isinstance(creds, OAuthClientCredentials)
+        assert creds.client_id == "azure-client-id"
+        assert creds.client_secret == "azure-client-secret"
+        assert creds.token_url == "https://login.microsoftonline.com/testingabc123/oauth2/v2.0/token"
+        assert creds.scopes == ["https://greenfield.cognitedata.com/.default"]
+        assert creds.token_custom_args == {"other_custom_arg": "some_value"}
+
 
 class TestOAuthClientCertificate:
     DEFAULT_PROVIDER_ARGS: ClassVar = {
@@ -82,3 +106,19 @@ class TestOAuthClientCertificate:
         }
         creds = OAuthClientCertificate(**self.DEFAULT_PROVIDER_ARGS)
         assert "Authorization", "Bearer azure_token" == creds.authorization_header()
+
+    @patch("cognite.client.credentials.ConfidentialClientApplication")
+    def test_create_from_credential_provider(self, mock_msal_app):
+        mock_msal_app().acquire_token_for_client.return_value = {
+            "access_token": "azure_token",
+            "expires_in": 1000,
+        }
+        creds = CredentialProvider.from_config(
+            credential_type="o_auth_client_certificate", config=self.DEFAULT_PROVIDER_ARGS
+        )
+        assert isinstance(creds, OAuthClientCertificate)
+        assert creds.authority_url == "https://login.microsoftonline.com/xyz"
+        assert creds.client_id == "azure-client-id"
+        assert creds.cert_thumbprint == "XYZ123"
+        assert creds.certificate == "certificatecontents123"
+        assert creds.scopes == ["https://greenfield.cognitedata.com/.default"]
--- a/tests/tests_unit/test_config.yaml
+++ b/tests/tests_unit/test_config.yaml
@@ -1,0 +1,13 @@
+cognite:
+  client_config:
+    project: "test-project"
+    client_name: "cognite-sdk-python"
+    credentials:
+      o_auth_client_credentials:
+        token_url: "https://test.com/token"
+        client_id: "test-client-id"
+        client_secret: "test-client-secret"
+        scopes: ["https://test.com/.default", "https://test.com/.admin"]
+    debug: true
+  global_config:
+    max_retries: 5
--- a/tests/tests_unit/test_config_envs.yaml
+++ b/tests/tests_unit/test_config_envs.yaml
@@ -1,0 +1,13 @@
+cognite:
+  client_config:
+    project: ${COGNITE_PROJECT}
+    client_name: $COGNITE_CLIENT_NAME
+    credentials:
+      ${credential_type}: # o_auth_client_credentials
+        token_url: "https://${URL}.com/token"
+        client_id: "test-client-id-with-escaped-$$"
+        client_secret: ${COGNITE_CLIENT_SECRET}
+        scopes: ["https://${URL}.com/.default", "https://${URL}.com/.admin"]
+    debug: ${COGNITE_DEBUG}
+  # global_config:
+  #   max_retries: 5


cognite/client/_cognite_client.py
```diff
@@ -213,3 +217,64 @@ def default_oauth_interactive(
         """
         credentials = OAuthInteractive.default_for_azure_ad(tenant_id, client_id, cdf_cluster)
         return cls.default(project, cdf_cluster, credentials, client_name)
+
+    @classmethod
+    def from_dict(cls, config: dict[str, Any]) -> CogniteClient:
```
haakonvt, line 222: I did some digging, and to keep this in line with the rest of the SDK, let's use load for both `dict` and `yaml`. I also suggest we don't support loading from a path-like object.

```py
@classmethod
    def load(cls, resource: dict | str) -> Self:
        ...
```

Taking a look at CogniteObject (in `_base.py`), consider adding: `dump` and `dump_yaml` methods 😄

cognite/client/credentials.py
```diff
@@ -23,6 +23,39 @@ class CredentialProvider(Protocol):
     def authorization_header(self) -> tuple[str, str]:
         raise NotImplementedError
 
+    @classmethod
+    def from_config(cls, credential_type: str, config: Any) -> CredentialProvider:
```
haakonvt, line 27: I suggest we do the same here, use `load`, then expect the following:

```py
dct = {"token": {...}}
credz = CredentialProvider.load(dct)
assert type(credz) is Token

dct = {"client_credentials": {...}}
credz = CredentialProvider.load(dct)
assert type(credz) is OAuthClientCredentials
```

if more than one top-level key exists, probably raise

cognite/client/credentials.py
```diff
@@ -23,6 +23,39 @@ class CredentialProvider(Protocol):
...
+        if credential_type == "token":
+            return Token(config)
+        elif credential_type == "o_auth_client_credentials":
+            return OAuthClientCredentials(**config)
+        elif credential_type == "o_auth_client_certificate_default_for_azure_ad":  # FIXME: do we want this?
+            return OAuthClientCredentials.default_for_azure_ad(**config)
+        elif credential_type == "o_auth_interactive":
+            return OAuthInteractive(**config)
+        elif credential_type == "o_auth_interactive_default_for_azure_ad":  # FIXME: do we want this?
+            return OAuthInteractive.default_for_azure_ad(**config)
```
haakonvt, line 51: I think we should remove these alternate class constructors here, yea, they serve a different purpose imo

tests/tests_unit/test_cognite_client.py
```diff
@@ -54,6 +56,22 @@ def mock_token_inspect(rsps) -> None:
...
+            "COGNITE_PROJECT": "test-project",
+            "COGNITE_CLIENT_NAME": "test-project",
+            "credential_type": "o_auth_client_credentials",
+            "URL": "test",
+            "COGNITE_CLIENT_SECRET": "test-client-secret",
+            "COGNITE_DEBUG": "true",
+        }
+        for key, value in env_vars.items():
+            monkeypatch.setenv(key, value)
+        yield
```
haakonvt, line 72: I think you can simplify this a bit:
```suggestion
@pytest.fixture
def set_env_vars(monkeypatch):
    env_vars = {
        "COGNITE_PROJECT": "test-project",
        "COGNITE_CLIENT_NAME": "test-project",
        "credential_type": "o_auth_client_credentials",
        "URL": "test",
        "COGNITE_CLIENT_SECRET": "test-client-secret",
        "COGNITE_DEBUG": "true",
    }
    monkeypatch.setattr(os, 'environ', env_vars)
```

cognite/client/_cognite_client.py
```diff
@@ -213,3 +217,64 @@ def default_oauth_interactive(
...
+        file_path = Path(file_path)
+        if not file_path.is_file():
+            raise ValueError(f"File {file_path} is not a file")
+
+        try:
+            with file_path.open("r") as file_raw:
+                sub_template = Template(file_raw.read())  # FIXME: use string.Template or expand yaml.SafeLoader class
+
+                if not sub_template.is_valid():  # type: ignore[attr-defined]
+                    raise ValueError("Invalid template")
```
haakonvt, line 265: This was added in version 3.11

nodegard: Yea 🥲 saw it after I pushed but got stuck doing some other things. Will update to only use the `substitute/safe_ substitute` that's available in 3.8>=

cognite/client/credentials.py
nodegard: Agreed, that's what I was leaning too but just added them to be discussed 👍🏻

cognite/client/credentials.py
nodegard: I had that originally and was going to bring it up to discuss 👌🏻

cognite/client/_cognite_client.py
```diff
@@ -213,3 +217,93 @@ def default_oauth_interactive(
...
+            CogniteClient: A cognite client object.
+
+        Examples:
+                Create a cognite client object from a YAML file, using envs from the current environment:
+
+                >>> config.yaml
+                >>> project: $MY_CDF_PROJECT
+                >>> base_url: https://${MY_CDF_CLUSTER}.cognitedata.com/
+                >>> client_credentials:
+                >>>     token: ${MY_CDF_TOKEN}
```
doctrino, line 276: All cognite resources assume camelCase in the load method. Suggest following the same practice here

nodegard: These are all instances of user input, not responses from the CogniteAPI, so I would think the snake case is the standard to use @haakonvt?

haakonvt: I discussed this with @nodegard and camelCase makes a lot of sense when loading resources from the API (which is 99.95% of what we do in the SDK), but for these configuration/setup classes, just matching parameters 1:1 (i.e. snake_case), makes the most sense - to us at least.

cognite/client/config.py
```diff
@@ -163,3 +200,45 @@ def default(
...
+        """
+        loaded = load_dict_or_str(config)
+
+        try:
+            credentials_config_input = loaded.pop("credentials")
+        except KeyError:
+            raise ValueError("'credentials' is a required field and must be included in the input dictionary.")
+
+        credentials = CredentialProvider.load(credentials_config_input)
+        return ClientConfig(credentials=credentials, **loaded)
```
haakonvt, line 244: (TL;DR) Let's write out all the arguments explicitly

cognite/client/credentials.py
```diff
@@ -23,6 +24,69 @@ class CredentialProvider(Protocol):
...
+        Examples:
+                Get a token credential provider:
+
+                >>> from cognite.client.credentials import CredentialProvider
+                >>> config = {"token": "my secret token"}
+                >>> credential_provider = CredentialProvider.load(config)
+
+                Get a client credential provider:
+
+                >>> from cognite.client.credentials import CredentialProvider
```
haakonvt, line 49:
```suggestion
```

cognite/client/credentials.py
```diff
@@ -23,6 +24,69 @@ class CredentialProvider(Protocol):
...
+                ...     }
+                ... }
+                >>> credential_provider = CredentialProvider.load(config)
+        """
+
+        loaded = load_dict_or_str(config)
+
+        if len(loaded) != 1:
+            raise ValueError(
+                "Credential provider configuration must be a dictionary containing exactly one top level key."
```
haakonvt, line 68: Let's repeat the options here as well

cognite/client/credentials.py
```diff
@@ -23,6 +24,69 @@ class CredentialProvider(Protocol):
...
+            )
+
+        credential_type, credential_config = next(iter(loaded.items()))
```
haakonvt, line 71: If you want this to be less readable (🥇 ):
```suggestion
        (credential_type, credential_config), = loaded.items()
```

cognite/client/credentials.py
```diff
@@ -23,6 +24,69 @@ class CredentialProvider(Protocol):
...
+        }
+
+        if credential_type not in supported_credential_types.keys():
+            raise ValueError(
+                f"Invalid credential provider type provided, the valid options are: {list(supported_credential_types.keys())}."
+            )
+        elif credential_type == "token":
+            return supported_credential_types[credential_type](credential_config)
+        else:
+            return supported_credential_types[credential_type](**credential_config)
```
haakonvt, line 88: Consider implementing and calling load (sorry :hide:)

cognite/client/credentials.py
```diff
@@ -23,6 +24,69 @@ class CredentialProvider(Protocol):
...
+
+        credential_type, credential_config = next(iter(loaded.items()))
+
+        supported_credential_types = {
+            "token": Token,
+            "client_credentials": OAuthClientCredentials,
+            "interactive": OAuthInteractive,
+            "device_code": OAuthDeviceCode,
+            "client_certificate": OAuthClientCertificate,
+        }
```
haakonvt, line 79: Lets define this either as a class variable or as a module-level constant. I suggest wrapping in `MappingProxyType` to avoid mutation

cognite/client/data_classes/_base.py
```diff
@@ -407,7 +401,7 @@ def _load(
 
     @classmethod
     def _load_raw_api_response(cls, responses: list[dict[str, Any]], cognite_client: CogniteClient) -> Self:
-        # Certain classes may need more than just 'items' from the raw repsonse. These need to provide
+        # Certain classes may need more than just 'items' from the raw response. These need to provide
```
haakonvt, line 404: 🫶

cognite/client/data_classes/_base.py
```diff
@@ -764,7 +758,7 @@ def load(
                 nulls=data[2],
             )
         elif isinstance(data, str) and (prop_order := data.split(":", 1))[-1] in ("asc", "desc"):
-            # Syntax "<fieldname>:asc|desc" is depreacted but handled for compatibility
+            # Syntax "<fieldname>:asc|desc" is deprecated but handled for compatibility
```
haakonvt, line 761: Clearly none of us core maintainers can read

cognite/client/credentials.py
```diff
@@ -23,6 +24,69 @@ class CredentialProvider(Protocol):
...
+
+        if credential_type not in supported_credential_types.keys():
```
haakonvt, line 81: You probably know this, but `.keys()` is implied for both contain checks and iteration so you can remove (we have a slight preference, but your call)

cognite/client/utils/_auxiliary.py
```diff
@@ -55,6 +55,18 @@ def get_accepted_params(cls: type[T_CogniteResource]) -> dict[str, str]:
     return {to_camel_case(k): k for k in vars(cls()) if not k.startswith("_")}
 
 
+def load_dict_or_str(resource: dict | str) -> dict:
```
haakonvt, line 58: Since this method enforces dict, I'd love for the name to reflect that. Maybe load_resource_to_dict? 🤔 

Also, keys must be str, right?
```suggestion
def load_dict_or_str(resource: dict[str, Any] | str) -> dict[str, Any]:
```

docs/source/cognite_client.rst
```diff
@@ -1,6 +1,7 @@
 CogniteClient
 =============
 
+.. _class_client_CogniteClient:
```
haakonvt, line 4: What does this do?

docs/source/quickstart.rst
```diff
@@ -1,7 +1,68 @@
...
+    from cognite.client import CogniteClient
+    from cognite.client.config import GlobalConfig
+
+    file_path = Path("cognite-sdk-config.yaml")
+
+    # Read in yaml file and substitute environment variables in the file string
+    with file_path.open("r") as file_raw:
+        env_sub_template = Template(file_raw.read())
+    try:
+        file_env_parsed = env_sub_template.substitute(dict(os.environ))
```
haakonvt, line 50: Can we skip this?
```suggestion
        file_env_parsed = env_sub_template.substitute(os.environ)
```

docs/source/quickstart.rst
```diff
@@ -1,7 +1,68 @@
 Quickstart
 ==========
+Instantiate a new client from a configuration file
+--------------------------------------------------
+Use this code to instantiate a client in order to execute API calls to Cognite Data Fusion (CDF) using a configuration file.
+
+.. note::
+    How you read in the configuration file is up to you as the :ref:`CogniteClient <class_client_CogniteClient>` load method only
+    accepts a dictionary. So for the purposes of this example, we will use the yaml library to read in a yaml file and
```
haakonvt, line 9: I think needs updating:
_...load method only accepts a dictionary._

docs/source/quickstart.rst
```diff
@@ -1,7 +1,68 @@
...
+    # Read in yaml file and substitute environment variables in the file string
+    with file_path.open("r") as file_raw:
+        env_sub_template = Template(file_raw.read())
+    try:
+        file_env_parsed = env_sub_template.substitute(dict(os.environ))
+    except (KeyError, ValueError) as e:
+        raise ValueError(f"Error substituting environment variable: {e}")
+
+    # Load yaml file string into a dictionary
+    cognite_config = yaml.safe_load(file_env_parsed)
```
haakonvt, line 55: I'd suggest we remove this step

docs/source/quickstart.rst
```diff
@@ -1,7 +1,68 @@
...
+
+    # If you want to set a global configuration it must be done before creating the client
+    global_config = GlobalConfig.load(cognite_config["global"])
```
haakonvt, line 58: As mentioned above, this will not take effect

tests/tests_unit/test_cognite_client.py
```diff
@@ -102,6 +114,71 @@ def test_verify_ssl_enabled_by_default(self, rsps, client_config_w_token_factory
...
+        assert client.config.credentials.client_secret == "test-client-secret"
+        assert client.config.credentials.token_url == TOKEN_URL
+        assert client.config.credentials.scopes == ["https://test.com/.default", "https://test.com/.admin"]
+        assert client.config.debug is True
+        log = logging.getLogger("cognite.client")
+        log.handlers = []
+        log.propagate = False
+
+    def test_client_load_docs_example(self, set_env_vars):
+        """Test the example in the Quickstart docs, docs/source/quickstart.rst"""
```
haakonvt, line 143: You may be able to use the doctests for this. Check out:
`tests/tests_unit/test_docstring_examples.py`

tests/tests_unit/test_utils/test_auxiliary.py
```diff
@@ -228,3 +229,27 @@ class TestFastDictLoad:
...
+    def test_load_dict_or_str(self, input, expected):
+        assert expected == load_dict_or_str(input)
+
+    @pytest.mark.parametrize(
+        "input",
+        (
+            "foo",
+            100,
+        ),
+    )
```
haakonvt, line 252: The longest oneliner I've seen 😜

haakonvt (commented):
This will be great step towards unified deliveries!

cognite/client/_cognite_client.py
```diff
@@ -213,3 +213,36 @@ def default_oauth_interactive(
...
+        """Loads a dictionary of configuration fields into a cognite client object.
+
+        Args:
+            config (dict | str): A dictionary or dictionary parsable string containing configuration values needed to create a CogniteClient.
+
+        Returns:
+            CogniteClient: A cognite client object.
+
+        Examples:
+                Create a cognite client object from a dictionary input:
```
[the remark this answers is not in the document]
nodegard: 95% sure it's 4 😅 I'd copied the docstring formatting from an existing function but will triple-check

cognite/client/_cognite_client.py
```diff
@@ -213,3 +213,36 @@ def default_oauth_interactive(
...
+                ...     "project": "cdf-project",
+                ...     "base_url": "https://api.cognitedata.com/",
+                ...     "client_credentials": {
+                ...         "client_id": "abcd",
+                ...         "client_secret": os.environ["OAUTH_CLIENT_SECRET"],
+                ...         "token_url": os.environ["TOKEN_URL"],
+                ...         "scopes": ["https://greenfield.cognitedata.com/.default"],
+                ...         # Any additional IDP-specific token args. e.g.
+                ...         "audience": "some-audience",
+                ...     }
```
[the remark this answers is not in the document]
nodegard: Oh yeah, thanks for catching that! The docstring tests had seen it in the ClientConfig docs, but running those tests for the CogniteClient & CredentialProviders wasn't as easy 🫠

cognite/client/data_classes/_base.py
nodegard: I'll add a spellchecker to your linters for you 🤣

tests/tests_unit/test_utils/test_auxiliary.py
nodegard: I always expect it to just get auto-reformated by some formatting tool 🤣

tests/tests_unit/test_cognite_client.py
nodegard: That's what I was thinking, but it doesn't look like those tests cover anything in the `docs` folder?

cognite/client/credentials.py
nodegard: the linters configured in my other projects would get so mad at that 🤣

cognite/client/credentials.py
nodegard: fair point 🙏🏻 but 🫠

docs/source/quickstart.rst
nodegard: yea was just from mypy originally but not needed for example

cognite/client/config.py
nodegard: I'm not manipulating any of the other args though, so don't we just want to leave it to the `__init__` to use the unpacked dict?

docs/source/quickstart.rst
nodegard: It would though, since line 55 cognite_config is just a dict? I should probably rename line 55 since that could be confusing

docs/source/quickstart.rst
nodegard: if I want to make a GlobalConfig and CogniteClient in the example then I'd have to do this so I can get the key for each. but maybe easiest to take an offline chat about this whole example since there were other comments too

docs/source/cognite_client.rst
nodegard: I wasn't able to get a link directly to [this](https://cognite-sdk-python.readthedocs-hosted.com/en/latest/cognite_client.html#cognite.client.CogniteClient) part of the docs before, it only let me get to [here](https://cognite-sdk-python.readthedocs-hosted.com/en/latest/cognite_client.html#) by using `:ref:'cognite_client:CogniteClient'` (which is almost the same but not exactly in my opinion but I can switch it to using the second one if preferred)

cognite/client/config.py
```diff
@@ -46,6 +47,42 @@ def __init__(self) -> None:
...
+                >>> config = {
+                ...     "max_retries": 5,
+                ...     "disable_ssl": True,
+                ... }
+                >>> global_config = GlobalConfig.load(config)
+        """
+
+        loaded = load_dict_or_str(config)
+
+        global_config = cls()
```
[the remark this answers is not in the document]
nodegard: raise what? 😅

cognite/client/config.py
```diff
@@ -46,6 +47,42 @@ def __init__(self) -> None:
         self.max_workers: int = 5
         self.silence_feature_preview_warnings: bool = False
 
+    @classmethod
+    def load(cls, config: dict | str) -> GlobalConfig:
+        """Loads a dictionary of configuration fields into a client config object.
```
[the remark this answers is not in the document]
nodegard: `set_global_config` maybe?

cognite/client/config.py
haakonvt: Yeah, we've had enough issues with old SDK versions and dict unpacking in the past... 😅 
```py
return ClientConfig(
    client_name=config["client_name"],
    project=config["project"],
    credentials=config["credentials"],
    api_subversion=config.get("api_subversion"),
    base_url=config.get("base_url"),
    max_workers=config.get("max_workers"),
    headers=config.get("headers"),
    timeout=config.get("timeout"),
    file_transfer_timeout=config.get("file_transfer_timeout"),
    debug=config.get("debug"),
)

File: cognite/client/data_classes/_base.py (hunk 14 above)
Comment (haakonvt) [reply cognite/client/data_classes/_base.py:761]: Yes, please 🚀

File: docs/source/quickstart.rst (hunk 21 above)
Comment (haakonvt) [reply docs/source/quickstart.rst:58]: No, what I mean is that `GlobalConfig.load` returns a new instance and no part of the SDK knows about this one, so any settings on it does not get picked up.

File: docs/source/quickstart.rst (hunk 20 above)
Comment (haakonvt) [reply docs/source/quickstart.rst:55]: Oh, right 👍

File: tests/tests_unit/test_cognite_client.py (hunk 22 above)
Comment (haakonvt) [reply tests/tests_unit/test_cognite_client.py:143]: It finds the docstrings and runs the code examples (it finds), so it should be possible if you point it to the cognite client file.

File: docs/source/quickstart.rst (hunk 21 above)
Comment (nodegard) [reply docs/source/quickstart.rst:58]: AHHH I see what you mean now

File: cognite/client/config.py (hunk 7 above)
Comment (nodegard) [reply cognite/client/config.py:245]: Ah ok, then I'll do that for all of them no problem 👌🏻

File: tests/tests_unit/test_config.py [hunk 28]
```diff
@@ -1,25 +1,33 @@
...
+        ):
+            GlobalConfig()
+
+    def test_apply_settings(self):
+        settings = {
             "max_workers": 5,
             "max_retries": 3,
         }
-        global_config = GlobalConfig.load(config)
+        global_config.apply_settings(settings)
```
haakonvt, line 21: This one is quite tricky... I think we should just mock `global_config` in this test to avoid all side effects. I'd also suggest making `apply_settings` a `staticmethod`, then adding `global global_config` to it. Wdyt?

haakonvt (commented):
Starting to look real good!

cognite/client/config.py
```diff
@@ -47,41 +57,39 @@ def __init__(self) -> None:
...
-        global_config = cls()
         for key, value in loaded.items():
             if not hasattr(global_config, key):
                 raise ValueError(f"Invalid key in global config: {key}")
             if key == "default_client_config":
-                global_config.default_client_config = ClientConfig.load(value)
+                if isinstance(value, ClientConfig):
+                    self.default_client_config = value
+                else:
+                    self.default_client_config = ClientConfig.load(value)
```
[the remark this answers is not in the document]
nodegard: I'm glad I had the epiphany because my brain was thinking so much in just a dict[str, str] mode (ie the loading from yaml scenario) 🤣 I'll add a test for it too

cognite/client/credentials.py
```diff
@@ -124,6 +118,27 @@ def thread_safe_get_token() -> str:
...
+        """Load a token credential provider object from a YAML/JSON string or dict.
+
+        Args:
+            config (dict[str, Any] | str): A dictionary or YAML/JSON string containing configuration values defined in the Token class.
+
+        Returns:
+            Token: Initialized token credential provider.
+
+        Warning:
+            A callable token is not supported if passing in a yaml string.
```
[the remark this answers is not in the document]
nodegard: Gotta make sure everyone knows 😜 but maybe it's actually just a Note not Warning 🤣

cognite/client/credentials.py
```diff
@@ -124,6 +118,27 @@ def thread_safe_get_token() -> str:
     def authorization_header(self) -> tuple[str, str]:
         return "Authorization", f"Bearer {self.__token_factory()}"
 
+    @classmethod
+    def load(cls, config: dict[str, Any] | str) -> Token:
```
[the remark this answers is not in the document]
nodegard: But a callable can be passed in if the dict is not coming from a yaml string, so the Any could be str or Callable so I'll update it to be explicit

cognite/client/utils/_auxiliary.py
```diff
@@ -55,14 +55,24 @@ def get_accepted_params(cls: type[T_CogniteResource]) -> dict[str, str]:
...
-def load_dict_or_str(resource: dict | str) -> dict:
+def load_resource_to_dict(resource: dict[str, Any] | str) -> dict[str, Any]:
     if isinstance(resource, dict):
-        return resource
+        if all(isinstance(key, str) for key in resource):
+            return resource
+        else:
+            raise TypeError(
+                "Resource must be a dictionary with string keys"
+            )  # FIXME: might not want to raise this error if it could cause breaking changes in the old load methods? (fast_dict_load expects dict[str, Any] so should be fine?)
```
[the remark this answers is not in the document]
nodegard: sounds good👌🏻 felt a bit overkill hence the fixme question

cognite/client/credentials.py
```diff
@@ -25,67 +26,60 @@ def authorization_header(self) -> tuple[str, str]:
...
-            return supported_credential_types[credential_type](credential_config)
+            if isinstance(credential_config, dict):
+                return _SUPPORTED_CREDENTIAL_TYPES[credential_type].load(credential_config)  # type: ignore [attr-defined]
+            elif isinstance(credential_config, str):
+                return _SUPPORTED_CREDENTIAL_TYPES[credential_type].load({"token": credential_config})  # type: ignore [attr-defined]
+            else:
+                raise TypeError(f"Token resource must be json or yaml str, or dict, not {type(credential_config)}")
         else:
-            return supported_credential_types[credential_type](**credential_config)
+            return _SUPPORTED_CREDENTIAL_TYPES[credential_type].load(credential_config)  # type: ignore [attr-defined]
```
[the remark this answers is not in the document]
nodegard: nice simplification 🙏🏻

tests/tests_unit/test_credential_providers.py
```diff
@@ -52,9 +54,61 @@ def test_token_non_string(self):
...
+            '{"token": "abc"}',
+            '{"token": {"token": "abc"}}',
+            {"token": (lambda: "abc")},
+            {"token": {"token": (lambda: "abc")}},
+        ],
+    )
+    def test_create_from_credential_provider(self, config):
+        creds = CredentialProvider.load(config)
         assert isinstance(creds, Token)
+        assert "Authorization", "Bearer abc" == creds.authorization_header()
```
haakonvt, line 71: Add another `Token.load` (maybe in the same test?) Could add a flag like "accepted"..

cognite/client/config.py
```diff
@@ -82,13 +82,16 @@ def apply_settings(self, settings: dict[str, Any] | str) -> None:
 
         loaded = load_resource_to_dict(settings)
 
+        if not set(loaded).issubset(set(self.__dict__)):
```
haakonvt, line 85: We can avoid converting both of the key views to sets:
```suggestion
        if not loaded.keys() <= self.__dict__.keys():
```

cognite/client/config.py
```diff
@@ -82,13 +82,16 @@ def apply_settings(self, settings: dict[str, Any] | str) -> None:
...
+            raise ValueError(
+                f"Invalid keys provided for global_config, no settings applied: {set(loaded) - set(self.__dict__)}"
```
haakonvt, line 87: What about: "One or more invalid keys..."?

cognite/client/config.py
```diff
@@ -36,7 +36,7 @@ class GlobalConfig:
 
     def __new__(cls) -> GlobalConfig:
         if hasattr(cls, "_instance"):
-            raise ValueError(
+            raise ValueError(  # TODO: raise a different error type?
```
haakonvt, line 39: ChatGPT suggested `RuntimeError`, Python uses `TypeError` for `typing.Any`. I leave this up to you 😜

cognite/client/config.py
```diff
@@ -82,13 +82,16 @@ def apply_settings(self, settings: dict[str, Any] | str) -> None:
...
+
         for key, value in loaded.items():
-            if not hasattr(global_config, key):
-                raise ValueError(f"Invalid key in global config: {key}")
             if key == "default_client_config":
-                if isinstance(value, ClientConfig):
+                if isinstance(value, ClientConfig) or value is None:
                     self.default_client_config = value
-                else:
+                elif value:
```
haakonvt, line 94: If someone passes in a value in disagreement with the type annotation, which are handled in the `if`, I'd say it's better to let `ClientConfig.load` decide if it can be accepted or not:
```suggestion
                else:
```

tests/tests_unit/test_config.py
```diff
@@ -13,23 +13,60 @@ def test_global_config_singleton(self):
...
+            "invalid_2": "foo",
         }
-        with pytest.raises(ValueError, match=r"Invalid key in global config: .*"):
+
+        with pytest.raises(ValueError, match=r"Invalid keys provided for global_config, no settings applied:"):
             global_config.apply_settings(settings)
 
+        assert (
+            global_config.max_workers != 0
+        )  # confirm that the valid keys were not applied since we don't want a partial application
```
haakonvt, line 68:
```suggestion
        # confirm that the valid keys were not applied since we don't want a partial application
        assert global_config.max_workers != 0
```

cognite/client/config.py
```diff
@@ -244,7 +247,7 @@ def load(cls, config: dict[str, Any] | str) -> ClientConfig:
         """
         loaded = load_resource_to_dict(config)
 
-        if isinstance(loaded["credentials"], CredentialProvider):
+        if not loaded["credentials"] or isinstance(loaded["credentials"], CredentialProvider):
```
haakonvt, line 250: Same comment here, if Credentials are required, let's always pass the content to the load method (except if already correct (sub)type)

cognite/client/config.py
haakonvt: Lol, I guess you can save a `not` by flipping the crocodile 😅

cognite/client/config.py
nodegard: should be fine that it will raise `TypeError(f"Resource must be json or yaml str, or dict, not {type(resource)}")` so reverted to not doing additional checks

tests/tests_unit/test_credential_providers.py
```diff
@@ -21,13 +48,119 @@ def test_token_factory_auth_header(self):
...
+    DEFAULT_PROVIDER_ARGS: ClassVar = {
+        "authority_url": "https://login.microsoftonline.com/xyz",
+        "client_id": "azure-client-id",
+        "scopes": ["https://greenfield.cognitedata.com/.default"],
+    }
+
+    @patch("cognite.client.credentials.PublicClientApplication")
+    @pytest.mark.parametrize("expires_in", (1000, "1001"))  # some IDPs return as string
+    def test_access_token_generated(self, mock_public_client, expires_in):
+        mock_public_client().return_value = Mock()
```
haakonvt, line 97: All tests run fine without this line (on my machine at least)
```suggestion
```

cognite/client/config.py
```diff
@@ -46,6 +59,43 @@ def __init__(self) -> None:
...
+            raise ValueError(
+                f"One or more invalid keys provided for global_config, no settings applied: {set(loaded) - set(self.__dict__)}"
+            )
+
+        for key, value in loaded.items():
+            if key == "default_client_config":
+                if isinstance(value, ClientConfig):
+                    self.default_client_config = value
+                else:
+                    self.default_client_config = ClientConfig.load(value)
```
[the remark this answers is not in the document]
nodegard: Will fix 👍🏻 but that raises another question because current implementation could then partially apply settings if you provide an invalid value to any of the other types like passing a dict to max_retries or something. So should probably validate all the fields then before applying them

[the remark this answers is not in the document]
haakonvt: No, all the others are just set as attributes and this won't raise

cognite/client/credentials.py
```diff
@@ -7,22 +7,75 @@
...
     @abstractmethod
     def authorization_header(self) -> tuple[str, str]:
         raise NotImplementedError
 
+    @classmethod
+    def load(cls, config: dict[str, Any] | str) -> CredentialProvider:
+        """Load a credential provider object from a YAML/JSON string or dict.
+
+        Args:
+            config (dict[str, Any] | str): A dictionary or YAML/JSON string containing the configuration for the credential provider. Note: The dictionary must contain exactly one top level key, which is the type of the credential provider and must be one of the following strings: "token", "client_credentials", "interactive", "device_code", "client_certificate". The value of the key is a dictionary containing the configuration for the credential provider.
```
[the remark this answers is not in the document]
nodegard: Were you thinking about using a Literal as an additional input argument?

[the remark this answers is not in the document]
haakonvt: Was thinking a TypedDict, but nevermind that! I would like us to move/duplicate the note about allowed keys and do something like:

```py
Note:
    The dictionary must contain exactly one top level key, which is the type of the credential provider and must be one of the following strings: "token", "client_credentials", "interactive", "device_code", "client_certificate". The value of the key is a dictionary containing the configuration for the credential provider.
```

cognite/client/config.py
```diff
@@ -46,6 +59,43 @@ def __init__(self) -> None:
...
+        if not loaded.keys() <= self.__dict__.keys():
+            raise ValueError(
+                f"One or more invalid keys provided for global_config, no settings applied: {set(loaded) - set(self.__dict__)}"
+            )
+
+        if "default_client_config" in loaded:
+            if not isinstance(loaded["default_client_config"], ClientConfig):
+                loaded["default_client_config"] = ClientConfig.load(loaded["default_client_config"])
+
+        vars(self).update(loaded)
```
haakonvt, line 97: You use `__dict__` directly two times, then go for `vars` at the end; I would suggest we either use dunder dict everywhere _or_ move `current_settings = vars(self)` then reference it?

(also added in a suggestion using `dict_keys.__sub__`)

```suggestion
        current_settings = vars(self)
        if not loaded.keys() <= current_settings:
            raise ValueError(
                f"One or more invalid keys provided for global_config, no settings applied: {loaded.keys() - current_settings}"
            )

        if "default_client_config" in loaded:
            if not isinstance(loaded["default_client_config"], ClientConfig):
                loaded["default_client_config"] = ClientConfig.load(loaded["default_client_config"])

        current_settings.update(loaded)
```

cognite/client/credentials.py
```diff
@@ -222,6 +296,37 @@ def _refresh_access_token(self) -> tuple[str, float]:
...
+            ...     "scopes": ["https://greenfield.cognitedata.com/.default"],
+            ... }
+            >>> credentials = OAuthDeviceCode.load(config)
+        """
+        loaded = load_resource_to_dict(config)
+        return cls(
+            authority_url=loaded["authority_url"],
+            client_id=loaded["client_id"],
+            scopes=loaded["scopes"],
+            token_cache_path=loaded.get("token_cache_path"),
```
haakonvt, line 324: `Path`?

pytest.ini
```diff
@@ -11,4 +11,8 @@ filterwarnings =
     ignore::DeprecationWarning:test
     ignore::cognite.client.utils._experimental.FeaturePreviewWarning
 
-addopts = --color=yes
+testpaths =
+    tests
+    docs
```
haakonvt, line 16: Nice 👌

haakonvt (commented):
Final round of comments 🤓 🚀

.github/workflows/build.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           COGNITE_PROJECT: python-sdk-test
           COGNITE_BASE_URL: https://greenfield.cognitedata.com
           COGNITE_CLIENT_NAME: python-sdk-integration-tests
-        run: pytest tests --durations=10 --cov --cov-report term --cov-report xml:coverage.xml -n8 --dist loadscope --reruns 2 --maxfail 20
+        run: pytest --durations=10 --cov --cov-report term --cov-report xml:coverage.xml -n8 --dist loadscope --reruns 2 --maxfail 20
```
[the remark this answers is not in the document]
haakonvt: Potentially add a comment here that points to the ini file for the multiple locations

haakonvt:
Heads'up @gregertw #1892 , you may need to update load methods after this PR was merged 😄

PR #1865, opened 2024-07-26
Allow unknown in read format of groups

doctrino:
## Checklist:
- [ ] Tests added/updated.


cognite/client/data_classes/iam.py
```diff
@@ -140,7 +141,7 @@ def is_managed_in_cdf(self) -> bool:
         return self.members is not None
 
     @classmethod
-    def _load(cls, resource: dict, cognite_client: CogniteClient | None = None, allow_unknown: bool = False) -> Group:
+    def _load(cls, resource: dict, cognite_client: CogniteClient | None = None, allow_unknown: bool = True) -> Group:
```
haakonvt, line 144: I really don't think we should change this default to true

doctrino: If CDF has a group with a capability which is not implemented in the ACL, when doing retrieve/list then we will fail to load the response from the API. 

My thinking is that in the read format (Group/GroupList) we should accept whatever capabilities the API returns.

doctrino: This is hardcoded into the `group.list` and `token.inspet` methods today, but is not there for `.create`

haakonvt: Right, yeah, agree that create should be updated to match 👌 Good catch!

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.54.10] - 2024-07-26,
+### Fixed
+- Creating a Group with an `UnknownAcl` supported by the API no longer raises a `ValueError`.
```
haakonvt, line 22: One typo + suggest we add in info that the create call still went through previously
```suggestion
## [7.54.10] - 2024-07-26
### Fixed
- Creating a Group with an `UnknownAcl` supported by the API no longer raises a client-side `ValueError` after
  successfully creating the group.
```

cognite/client/data_classes/iam.py
haakonvt, line 144:
```suggestion
    def _load(cls, resource: dict, cognite_client: CogniteClient | None = None, allow_unknown: bool = False) -> Group:
```

cognite/client/data_classes/iam.py
```diff
@@ -218,7 +231,7 @@ def _load(
         cls,
         resource_list: Iterable[dict[str, Any]],
         cognite_client: CogniteClient | None = None,
-        allow_unknown: bool = False,
+        allow_unknown: bool = True,
```
haakonvt, line 234:
```suggestion
        allow_unknown: bool = False,
```

cognite/client/data_classes/iam.py
doctrino: This is the fix, why do you want to revert that?

cognite/client/data_classes/iam.py
doctrino: Same here?

cognite/client/_api/iam.py
```diff
@@ -363,7 +363,43 @@ def create(self, group: Group | GroupWrite | Sequence[Group] | Sequence[GroupWri
...
+                cls,
+                resource: dict[str, Any],
+                cognite_client: CogniteClient | None = None,
+                allow_unknown: bool = False,
+            ) -> GroupWrite:
+                return GroupWrite._load(resource, cognite_client=cognite_client, allow_unknown=True)
+
+        return self._create_multiple(
+            list_cls=GroupListAdapter, resource_cls=GroupAdapter, items=group, input_resource_cls=GroupWriteAdapter
+        )
```
haakonvt, line 402: Nice, let's also use these for `list`?

haakonvt (approved):
## -> Let's move the class definitions out of the create method!

_Suggestion_: Also use them for `list`

PR #1886, opened 2024-08-26
Use Sequence instead of List for DataModelApply.view

erlendvollset:
- **Bump mypy**
- **Use Sequence instead of list in DataModelApply.views**

## Description

## Checklist:
- [ ] Tests added/updated.


pyproject.toml
```diff
@@ -51,6 +51,7 @@ shapely = { version = ">=1.7.0", optional = true }
 pyodide-http = { version = "^0.2.1", optional = true }
 graphlib-backport = { version = "^1.0.0", python = "<3.9" }
 PyYAML = { version = "^6.0", optional = true }
+mypy = "^1"
```
haakonvt, line 54: Should only be a dev deps, I think

erlendvollset: indeed

erlendvollset: fixed

haakonvt (approved):
LGTM

PR #1887, opened 2024-08-26
enable diagram detect based on file instance id

olacognite:
## Description

Make diagram detect available by file_instance_id. 

I have done it similarly to id and external id, but could consider to only make it available with file references, since file references with page ranges is the recommended way to run (at least when documents have variable numbers of pages). 

PR in context-api is [here](https://github.com/cognitedata/infrastructure/pull/13632)

## Checklist:
- [x] Tests added/updated.


cognite/client/data_classes/contextualization.py
```diff
@@ -356,24 +357,28 @@ def __init__(
...
 
-    def to_api_item(self) -> dict[str, str | int | dict[str, int]]:
+    def to_api_item(self) -> dict[str, str | int | dict[str, int] | dict[str, str]]:
         if self.file_id is None and self.file_external_id is not None:
-            item: dict[str, str | int | dict[str, int]] = {"fileExternalId": self.file_external_id}
+            item: dict[str, str | int | dict[str, int] | dict[str, str]] = {"fileExternalId": self.file_external_id}
         if self.file_id is not None and self.file_external_id is None:
             item = {"fileId": self.file_id}
+        if self.file_instance_id is not None:
+            item = {"fileInstanceId": self.file_instance_id.dump(include_instance_type=False)}
```
doctrino, line 381: A situation that should never happen, but this is more robust: 
```python
        item: dict[str, str | int | dict[str, int] | dict[str, str]] = {}
        if self.file_id is None and self.file_external_id is not None and self.file_instance_id is None:
            item["fileExternalId"] = self.file_external_id
        elif self.file_id is not None and self.file_external_id is None and self.file_instance_id is None:
            item["fileId"] = self.file_id
        elif self.file_instance_id is not None and self.file_id is None and self.file_external_id is None:
            item["fileInstanceId"] = self.file_instance_id.dump(include_instance_type=False)
        else:
            raise ValueError("Exactly one of file_id, file_external_id and file_instance_id must be set")
```

tests/tests_integration/test_api/test_diagrams.py
```diff
@@ -9,8 +9,11 @@
     DiagramDetectResults,
     FileReference,
 )
+from cognite.client.data_classes.data_modeling import NodeId
 
 PNID_FILE_ID = 3261066797848581
+DIAGRAM_SPACE = "diagram_space"
```
doctrino, line 15: Can you make a fixture to guarantee that this space will exist?

Something like
```python
def diagram_space(cognite_client: CogniteClient) -> Space:
    return cognite_client.data_modeling.spaces.apply(SpaceApply(space=DIAGRAM_SPACE))
```

tests/tests_integration/test_api/test_diagrams.py
```diff
@@ -103,3 +106,20 @@ def test_run_diagram_detect_in_pattern_mode(self, cognite_client):
 
         assert len(detected_by_resource_type["file_reference"]) >= 10  # 14 seen when making the test
         assert len(detected_by_resource_type["instrument"]) >= 60  # 72 seen when making the test
+
+    def test_run_diagram_detect_with_file_instance_id(self, cognite_client):
```
doctrino, line 110:
```suggestion
    @pytest.mark.usefixtures("diagram_space")
    def test_run_diagram_detect_with_file_instance_id(self, cognite_client):
```

doctrino (commented):
Just some nit picking :)

tests/tests_integration/test_api/test_diagrams.py
olacognite: Hm, could do that, but should also make sure that the file exists in that case. If the space does not exist, then the file cant exist either, so maybe add a file to the repo and upload it if it is not there or what do you think?

doctrino: Yes, I missed that. I would also make another fixture for the file then. If the file is not too big or sensitive then commit it to the repo.

doctrino (approved):

PR #1895, opened 2024-08-29
workflow_orchestration is now data_workflows rst and add trigger doc

vincent-cognite:
## Description
- workflow_orchestration.rst is now data_workflows.rst
- add missing doc methods
- fix property names according to latest api spec

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/workflows.py
+++ b/cognite/client/data_classes/workflows.py
@@ -1381,26 +1381,33 @@ class WorkflowTriggerRun(CogniteResource):
 
     def __init__(
         self,
-        trigger_external_id: str,
-        trigger_fire_time: int,
+        external_id: str,
+        fire_time: int,
         workflow_external_id: str,
         workflow_version: str,
         workflow_execution_id: str,
+        status: Literal["success", "failed"],
+        reason_for_failure: str | None = None,
     ) -> None:
-        self.trigger_external_id = trigger_external_id
-        self.trigger_fire_time = trigger_fire_time
+        self.external_id = external_id
+        self.fire_time = fire_time
         self.workflow_external_id = workflow_external_id
         self.workflow_version = workflow_version
         self.workflow_execution_id = workflow_execution_id
+        self.status = status
+        self.reason_for_failure = reason_for_failure
 
     def dump(self, camel_case: bool = True) -> dict[str, Any]:
         item = {
-            "external_id": self.trigger_external_id,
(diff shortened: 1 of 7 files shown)


cognite/client/data_classes/workflows.py
```diff
@@ -1385,18 +1385,21 @@ def __init__(
         trigger_fire_time: int,
```
doctrino, line 1385: Why not also update the data class names? It is better if this matches the spec, and as this feature is in alpha we can do it without a major bump.
```suggestion
        fire_time: int,
```

vincent-cognite: Not sure I follow: why you want to change the data class names?
(the sdk is tagged as beta for now, only a selection of projects have access)

doctrino: If I understand it correctly, this class represent the response object of the getTriggerHistory endpoint. 

Then I expect the attributes to be snake case of the parameters returned in the response. In addition, it seem the `status`and `reasonForFailure` is missing.
image

cognite/client/data_classes/workflows.py
```diff
@@ -1381,34 +1381,47 @@ class WorkflowTriggerRun(CogniteResource):
...
         self,
-        trigger_external_id: str,
-        trigger_fire_time: int,
+        external_id: str,
+        fire_time: int,
         workflow_external_id: str,
         workflow_version: str,
+        workflow_execution_id: str,
+        status: Literal["success", "failed"],
+        reason_for_failure: str | None = None,
```
doctrino, line 1390: Ideally also a literal.

doctrino (approved):
One comment, up to you if you do it :)

cognite/client/data_classes/workflows.py
vincent-cognite: There can be many reasons, and new reasons added at any time in the server. Why a literal?

doctrino: Just help for the user. Literals are just hints/docs, so will not break if you suddenly send something else. You know the intentions the best and it might make sense to just keep it as a string.

vincent-cognite: The reason is a sentence, it can come from other APIs, which have no control of (transformations, functions, simulators, ...)
So I think we keep it as a string.

PR #1902, opened 2024-09-02
[CDF-22368] Allow Subworkflow references

VerstraeteBert:
## Description
Adds support for subworkflow task referencing existing workflow versions.


doctrino (commented):
What about just introducing an new class?

```python
class SubWorkflowReference(WorkflowTaskParameters):
    """
    The subworkflow reference is used to specifying a reference to an other workflow.

    When a workflow is made of stages with dependencies between them, we can use subworkflow tasks for convenience.

    Args:
        workflow_external_id (str | None): The external ID of the referenced workflow.
        version (str | None): The version of the referenced workflow.
    """

    task_type = "subworkflow"

    def __init__(
        self,
        workflow_external_id: str,
        version: str
    ) -> None:
        self.workflow_external_id = workflow_external_id
        self.version = version

    @classmethod
    def _load(cls: type[Self], resource: dict, cognite_client: CogniteClient | None = None) -> Self:
        return cls(
            workflow_external_id=resource["workflowExternalId"],
            version=resource["version"]
        )

        return {
            self.task_type: {
                "workflowExternalId": self.workflow_external_id,
                "version": self.version,
            }
        }
```

And then change the `WorkflowTaskParameters` load to
```python
        elif type_ == "dynamic":
            return DynamicTaskParameters._load(parameters)
        elif type_ == "subworkflow" and "tasks" in parameters:
            return SubworkflowTaskParameters._load(parameters)
        elif type_ == "subworkflow" and "workflowExternalId" in parameters:
            return SubWorkflowReference._load(parameters)
```

vincent-cognite (approved):
I like the simple approach for handling the polymorphism of subworkflow 🚀

PR #1914, opened 2024-09-05
[fix] Zipping files older than 1980

qTipTip:
## Description

From the docs: https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile

_The strict_timestamps argument, when set to False, allows to zip files older than 1980-01-01 at the cost of setting the timestamp to 1980-01-01. Similar behavior occurs with files newer than 2107-12-31, the timestamp is also set to the limit._

We have been seeing some of these errors when creating functions lately. Not sure why they sporadically show up: 
```
ValueError: ZIP does not support timestamps before 1980
```
## Checklist:
- [ ] Tests added/updated.


cognite/client/_api/functions.py
```diff
@@ -58,7 +58,8 @@
 
 MAX_RETRIES = 5
 REQUIREMENTS_FILE_NAME = "requirements.txt"
-REQUIREMENTS_REG = re.compile(r"(\[\/?requirements\]){1}$", flags=re.M)  # Matches [requirements] and [/requirements]
```
qTipTip, line 61: Most of the changes here are linting fmt.

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.58.7] - 2024-09-05
+### Fixed
+- Allow zipping files older than 1980 when creating a function, this is done by
+setting the timestamp to 1980 if older than that.
```
haakonvt, line 23:
```suggestion
## [7.58.7] - 2024-09-06
### Fixed
- Creating a function using files dated before 1980 no longer raises ValueError,
  by overriding the timestamps to 1980-01-01.
```

haakonvt (approved):
LGTM

CHANGELOG.md
qTipTip: Thanks!

PR #1919, opened 2024-09-09
WIP: take CDM/v1 using typed instances to GA

haakonvt:
## Description
This PR takes CDM/v1 using typed instances to GA.

### DIFF
--- a/cognite/client/_api/data_modeling/instances.py
+++ b/cognite/client/_api/data_modeling/instances.py
@@ -349,7 +349,9 @@ class InstancesAPI(APIClient):
         Returns:
             EdgeList[T_Edge] | T_Edge | Edge | None: The requested edges.
 
-        Retrieve nodes using a custom Edge class Flow
+        Retrieve edges using a custom typed class "Flow". Any property that you want to look up by a different attribute name,
+        e.g. you want `my_edge.flow_rate` to return the data for property `flowRate`, must use the PropertyOptions as shown below.
+        We strongly suggest you use snake_cased attribute names, as is done here:
 
                 >>> from cognite.client import CogniteClient
                 >>> from cognite.client.data_classes.data_modeling import EdgeId, TypedEdge, PropertyOptions, DirectRelationReference, ViewId
@@ -367,7 +369,7 @@ class InstancesAPI(APIClient):
                 ...        flow_rate: float,
                 ...        start_node: DirectRelationReference,
                 ...        end_node: DirectRelationReference,
-                ...        deleted_time: Union[int, None] = None,
+                ...        deleted_time: int | None = None,
                 ...    ) -> None:
                 ...        super().__init__(
                 ...            space, external_id, version, type, last_updated_time, created_time, start_node, end_node, deleted_time
@@ -379,7 +381,9 @@ class InstancesAPI(APIClient):
                 ...        return ViewId("sp_model_space", "flow", "1")
                 ...
                 >>> client = CogniteClient()
-                >>> res = client.data_modeling.instances.retrieve_edges(NodeId("mySpace", "theFlow"), edge_cls=Flow)
+                >>> res = client.data_modeling.instances.retrieve_edges(
+                ...     EdgeId("mySpace", "theFlow"), edge_cls=Flow
+                ... )
                 >>> isinstance(res, Flow)
         """
         res = self._retrieve_typed(
@@ -437,7 +441,6 @@ class InstancesAPI(APIClient):
             without providing a custom node class, but in that case, the retrieved nodes will be of the
             built-in Node class.
 
-
         Args:
             nodes (NodeId | Sequence[NodeId] | tuple[str, str] | Sequence[tuple[str, str]]): Node id(s) to retrieve.
             node_cls (type[T_Node]): The custom node class to use, the retrieved nodes will automatically be serialized to this class.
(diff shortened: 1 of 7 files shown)


cognite/client/_api/data_modeling/instances.py
```diff
@@ -925,24 +933,23 @@ def apply(
...
+            Create a typed node. Any property that you want to look up by a different attribute name, e.g. you want
+            `my_node.birth_year` to return the data for property `birthYear`, must use the PropertyOptions as shown below.
+            We strongly suggest you use snake_cased attribute names, as is done here:
 
                 >>> from cognite.client import CogniteClient
-                >>> from datetime import date
                 >>> from cognite.client.data_classes.data_modeling import TypedNodeApply, PropertyOptions
                 >>> class Person(TypedNodeApply):
-                ...     birth_date = PropertyOptions(identifier="birthDate")
+                ...     birth_year = PropertyOptions(identifier="birthYear")
```
haakonvt, line 943: Perhaps `Person -> PersonApply`

cognite/client/_api/data_modeling/instances.py
```diff
@@ -369,7 +371,7 @@ def retrieve_edges(
                 ...        flow_rate: float,
                 ...        start_node: DirectRelationReference,
                 ...        end_node: DirectRelationReference,
-                ...        deleted_time: Union[int, None] = None,
+                ...        deleted_time: int | None = None,
                 ...    ) -> None:
                 ...        super().__init__(
                 ...            space, external_id, version, type, last_updated_time, created_time, start_node, end_node, deleted_time, None
```
haakonvt, line 377: Note to self, update call to remove `None`

erlendvollset:
What's the deal with the `use_attribute_name` parameter on `TypedInstance.dump`? It dumps the custom properties using the attribute names in the defined class? That doesn't seem valuable to me and it's adding a bunch of complexity. I say we remove that.

cognite/client/data_classes/_base.py
```diff
@@ -220,9 +220,9 @@ def to_pandas(
             pandas.DataFrame: The dataframe.
         """
         pd = local_import("pandas")
-        from cognite.client.data_classes.data_modeling.typed_instances import _TypedInstance
+        from cognite.client.data_classes.data_modeling.typed_instances import TypedInstance
 
-        if isinstance(self, _TypedInstance):  # TODO: This is hacky, but it works for now
+        if isinstance(self, TypedInstance):  # TODO: This is hacky, but it works for now
             dumped = self.dump(camel_case=camel_case, use_attribute_name=not camel_case)  # type: ignore [call-arg]
```
erlendvollset, line 226: This doesn't make any sense to me. What does `use_attribute_name` have to do with camel casing?

cognite/client/data_classes/data_modeling/typed_instances.py
```diff
@@ -65,68 +77,252 @@ def __delete__(self, instance: Any) -> None:
...
+        return output
+
+    @classmethod
+    def _load(cls, resource: dict, cognite_client: CogniteClient | None = None) -> Self:
+        sources = resource.pop("sources", [])
+        properties = sources[0].get("properties", {}) if sources else {}
+        return cast(Self, _load_typed_instance(cls, resource, properties, cls._instance_properties))
+
+
+class TypedNodeApply(TypedInstanceApply):
```
erlendvollset, line 225: This doesn't work. TypedNodeApply needs to subclass NodeApply, and same fore TypedNode and Node. Otherwise the InstanceAPI methods don't accept them. If this was just for the datetime/int thing, then we shouldn't do that.

cognite/client/_api/data_modeling/instances.py
doctrino: Yes, I think it is good to be consistent on that.

cognite/client/data_classes/_base.py
doctrino: In the Typed classes, we use snake_case version of the view properties. But when these are dumped, they are typically camelCase such that they can be written to the API.

When we do `.to_pandas` a Data Scientist would expect the columns to match the attributes of the Typed Instance and thus this workaround.

cognite/client/data_classes/data_modeling/cdm/v1.py
```diff
@@ -70,7 +70,7 @@ def __init__(
         existing_version: int | None = None,
         type: DirectRelationReference | tuple[str, str] | None = None,
     ) -> None:
-        TypedNodeApply.__init__(self, space, external_id, existing_version, None, type)
+        TypedNodeApply.__init__(self, space, external_id, existing_version, type)
```
doctrino, line 73: If we edit this file manually, we cannot regenerate it any more without losing the manual edits. 

Suggest we instead include this changes into the generation and then apply it both to the core and extractor extension models.

doctrino (commented):
Main comment is to avoid manually editing the generated files, and instead update the source (pygen) [template ](https://github.com/cognitedata/pygen/blob/main/cognite/pygen/_core/templates/typed_classes.py.jinja)

erlendvollset:
So per now TypedNode(..).properties returns an empty dict, and TypedNodeApply().sources returns None. They’re just being resolved correctly when dumping, which is not how it should be.

cognite/client/data_classes/_base.py
erlendvollset: "typically" camelCase

erlendvollset: what about cases where they're not?

haakonvt:
@erlendvollset / @doctrino PR is reverted to not include any changes to datetime and TypedNode inheritance changes.

cognite/client/data_classes/data_modeling/cdm/__init__.py
```diff
@@ -1,5 +1 @@
 from __future__ import annotations
-
-from cognite.client.utils._experimental import FeaturePreviewWarning
-
-FeaturePreviewWarning("alpha", "alpha", "Core Data Model").warn()
```
doctrino, line 5: Lets hold off on this one, I will make one more PR that does some changes.

doctrino (commented):
Just add back the warning and then good to go.

cognite/client/data_classes/data_modeling/cdm/__init__.py
haakonvt: Yeah, I agree. Will also update PR title

doctrino (approved):

PR #1934, opened 2024-09-17
Remove warning instance id ts

haakonvt:
Incorporate instance_id as an equal to id and external_id in the TimeSeries/Datapoints API.

https://cognitedata.atlassian.net/browse/DM-2128

## TODO
Some more tests, particularly on combining identifiers.

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -58,7 +58,6 @@ from cognite.client.utils._auxiliary import (
     unpack_items_in_payload,
 )
 from cognite.client.utils._concurrency import ConcurrencySettings, execute_tasks
-from cognite.client.utils._experimental import FeaturePreviewWarning
 from cognite.client.utils._identifier import Identifier, IdentifierSequence, IdentifierSequenceCore
 from cognite.client.utils._importing import import_as_completed, local_import
 from cognite.client.utils._time import (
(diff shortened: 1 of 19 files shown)


tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -184,7 +184,7 @@ def all_retrieve_endpoints(cognite_client, retrieve_endpoints):
 
 
 @pytest.fixture(scope="session")
-def instance_ts_id(cognite_client_alpha: CogniteClient, alpha_test_space: Space) -> InstanceId:
+def instance_ts_id(cognite_client: CogniteClient, alpha_test_space: Space) -> InstanceId:
```
doctrino, line 187: is the alpha client no longer used? Should we remove the fixture?

cognite/client/_api/datapoints.py
```diff
@@ -536,13 +526,9 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
         self,
         *,
-        id: None | int | DatapointsQuery | dict[str, Any] | Sequence[int | DatapointsQuery | dict[str, Any]] = None,
-        external_id: None
-        | str
-        | DatapointsQuery
-        | dict[str, Any]
-        | SequenceNotStr[str | DatapointsQuery | dict[str, Any]] = None,
-        instance_id: None | NodeId | Sequence[NodeId] = None,
+        id: None | int | DatapointsQuery | Sequence[int | DatapointsQuery] = None,
```
doctrino, line 529: isn't this breaking? not to allow `dict[str, Any]` any more?

doctrino (commented):
Just some clarification.

tests/tests_integration/test_api/test_datapoints.py
haakonvt: Still in use by files tests. Maybe these can also be changed. I decided not to remove it as we will certainly have alpha features in the futures

cognite/client/_api/datapoints.py
haakonvt: Just the type annotations being updated to push people a little bit 😉

doctrino (approved):

PR #1941, opened 2024-09-24
Update Core Model

doctrino:
## Description
There seems to be some updated description and order of properties in the CoreModel, which is also reflected in this PR change. 

The main change is to make the `isUploaded` and `uploadedTime` read-only properties in `CogniteFile`.

## Checklist:
- [ ] Tests added/updated.


cognite/client/data_classes/data_modeling/cdm/v1.py
```diff
@@ -2648,8 +2646,6 @@ def __init__(
         assets: list[DirectRelationReference | tuple[str, str]] | None = None,
         mime_type: str | None = None,
         directory: str | None = None,
-        is_uploaded: bool | None = None,
-        uploaded_time: datetime | None = None,
```
doctrino, line 2652: This is the main change, these are no considered read-only properties.

erlendvollset (approved):

PR #1950, opened 2024-09-30
Require >=Python3.10

erlendvollset:
- **Require python 3.10 or higher**
- **Fix mypy issues**

## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,11 +8,11 @@ repos:
           - --fix
           - --exit-non-zero-on-fix
           - --line-length=120
-          - --ignore=E731,E501,W605
+          - --ignore=E731,E501,W605,UP038
             # See https://beta.ruff.rs/docs/rules for an overview of ruff rules
           - --select=E,W,F,I,T,RUF,TID,UP
           - --fixable=E,W,F,I,T,RUF,TID,UP
-          - --target-version=py38
+          - --target-version=py310
           - --exclude=cognite/client/_proto
       - id: ruff-format
         args:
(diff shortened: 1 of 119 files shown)


.pre-commit-config.yaml
```diff
@@ -8,11 +8,11 @@ repos:
           - --fix
           - --exit-non-zero-on-fix
           - --line-length=120
-          - --ignore=E731,E501,W605
+          - --ignore=E731,E501,W605,UP038
```
doctrino, line 11: Personal preference, or a special issue with the code base?

doctrino (approved):

.pre-commit-config.yaml
haakonvt: https://docs.astral.sh/ruff/rules/non-pep604-isinstance/

doctrino (commented):
I guess a new minor release with changelog is in order.

.pre-commit-config.yaml
doctrino: I was wondering why ignore it, not what the it is.

haakonvt: I do think the link answers that 😉 

[Why is this bad?](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#why-is-this-bad)

Since Python 3.10, isinstance and issubclass can be passed a |-separated union of types, which is consistent with the union operator introduced in [PEP 604](https://peps.python.org/pep-0604/).

Note that this results in slower code. Ignore this rule if the performance of an isinstance or issubclass check is a concern, e.g., in a hot loop

PR #1960, opened 2024-10-07
make `allow_unknown` kw only to remove from subclasses

haakonvt:
Part of the changes to move to 3.10


doctrino (approved):
Agreed, no brainer

PR #1970, opened 2024-10-10
fix: instance type not possible to reset to null

haakonvt:


erlendvollset (changes requested):
edge.type cannot be set to null - it is non-nullable. So if omitted it should not be passed as null. I'm also not certain we'll want omitted meaning explicit null for node base props either. This might require changes in DMS too to be handled properly. i.e. we need to differentiate between omitted and null for base properties, like we do for custom properties.

haakonvt:
> edge.type cannot be set to null - it is non-nullable. So if omitted it should not be passed as null. I'm also not certain we'll want omitted meaning explicit null for node base props either. This might require changes in DMS too to be handled properly. i.e. we need to differentiate between omitted and null for base properties, like we do for custom properties.

oh right - I'll see what I can in the SDK as a workaround for now 👍

cognite/client/data_classes/data_modeling/cdm/v1.py
```diff
@@ -87,7 +88,7 @@ def __init__(
         station_360: DirectRelationReference | tuple[str, str] | None = None,
         taken_at: datetime | None = None,
         existing_version: int | None = None,
-        type: DirectRelationReference | tuple[str, str] | None = None,
+        type: DirectRelationReference | tuple[str, str] | None = NOT_SET,  # type: ignore [assignment]
```
erlendvollset, line 91: IMO, NOT_SET should be an instance of a `NotSet` class (or I prefer OMITTED/Omitted), and then we do this instead:
```suggestion
        type: DirectRelationReference | tuple[str, str] | Omitted | None = OMITTED,
```

haakonvt: While `class Omitted` with a (singleton instance) `OMITTED` is nicer for sure! ...I don't really like to expose this to the users (more than necessary, _the default value_).

It exposes internal details; users will see `Omitted` in the type annotation and perhaps wonder what it does, when in reality it is an implementation detail. This may lead to unnecessary confusion (do I need to know what this is/does?), which in turn may lead to unintended usage.

It also goes against the principle of what the type annotations mean; these are the allowed types you may **pass**. The point with the default value is when it is _not passed_.

haakonvt:
### Blocked until DMS starts to distinguish between "not given" and "given as null"

haakonvt:
Closing as this won't be addressed in DMS in probably a veeeeery long time.

PR #1976, opened 2024-10-13
fix 3.13 test runners

haakonvt:
Python 3.13 ships with some safety improvements to exec/eval that caused some tests to fail:

> This change to the semantics of locals() in optimized scopes also affects the default behavior of code execution functions that implicitly target locals() if no explicit namespace is provided (such as exec() and eval()).


doctrino (approved):

haakonvt:
How ironic, both 3.13 failed for unrelated reasons 😂 

Screenshot 2024-10-13 at 18 31 02

PR #1978, opened 2024-10-15
[CDF-22879] Update extraction pipeline with NotificationConfig

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/extractionpipelines.py
+++ b/cognite/client/data_classes/extractionpipelines.py
@@ -60,6 +60,7 @@ class ExtractionPipelineNotificationConfiguration(CogniteObject):
         allowed_not_seen_range_in_minutes (int | None): Time in minutes to pass without any Run. Null if extraction pipeline is not checked.
 
     """
+
     allowed_not_seen_range_in_minutes: int | None = None
 
 
--- a/cognite/client/data_classes/extractionpipelines.py
+++ b/cognite/client/data_classes/extractionpipelines.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from abc import ABC
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal, cast
 
 from cognite.client.data_classes._base import (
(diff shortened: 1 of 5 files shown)


cognite/client/data_classes/extractionpipelines.py
```diff
@@ -224,6 +243,7 @@ class ExtractionPipelineWrite(ExtractionPipelineCore):
         metadata (dict[str, str] | None): Custom, application specific metadata. String key -> String value. Limits: Maximum length of key is 128 bytes, value 10240 bytes, up to 256 key-value pairs, of total size at most 10240.
         source (str | None): Source text value for extraction pipeline.
         documentation (str | None): Documentation text value for extraction pipeline.
+        notification_config (ExtractionPipelineNotificationConfiguration | None): Notification configuration for the extraction pipeline.
```
haakonvt, line 246: Let's add in this bit of info from the API docs:

> Time in minutes to pass without any Run. Null if extraction pipeline is not checked.

cognite/client/data_classes/extractionpipelines.py
```diff
@@ -203,6 +218,10 @@ def _load(cls, resource: dict, cognite_client: CogniteClient | None = None) -> E
                 ExtractionPipelineContact._load(contact) if isinstance(contact, dict) else contact
                 for contact in instance.contacts
             ]
+        if instance.notification_config and isinstance(instance.notification_config, dict):
```
haakonvt, line 221: Can we skip the `isinstance` check on `dict`?
```suggestion
        if instance.notification_config:
```

haakonvt (approved):
I'd say just mark those tests as skip for now. You could add a condition to only skip for the next 7 or 14 days so that we don't forget about them 😄

cognite/client/data_classes/extractionpipelines.py
doctrino: No, then mypy complains. I don't like this pattern for loading, but did not what to do more changes than necessary.

PR #1980, opened 2024-10-15
Tests: TimeSeries Subscriptions

doctrino:
## Issue - TimeSeries can only be used in 10 subscriptions
image

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/tests/tests_integration/test_api/test_datapoint_subscriptions.py
+++ b/tests/tests_integration/test_api/test_datapoint_subscriptions.py
@@ -261,10 +261,7 @@ class TestDatapointSubscriptions:
             assert batch.subscription_changes.added[0].external_id == time_series_external_ids[0]
 
             existing_data = cognite_client.time_series.data.retrieve_dataframe(external_id=time_series_external_ids[0])
-            if existing_data.empty:
-                start = pd.Timestamp("now")
-            else:
-                start = existing_data.index[-1] + pd.Timedelta("1d")
+            start = pd.Timestamp("now") if existing_data.empty else existing_data.index[-1] + pd.Timedelta("1d")
             new_values = [42, 43]
             new_data = pd.DataFrame(
                 {time_series_external_ids[0]: new_values},
--- a/tests/tests_integration/test_api/test_datapoint_subscriptions.py
+++ b/tests/tests_integration/test_api/test_datapoint_subscriptions.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import math
-import platform
 import random
 import time
 from contextlib import contextmanager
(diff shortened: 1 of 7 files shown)


tests/tests_integration/test_api/test_datapoint_subscriptions.py
```diff
@@ -20,7 +21,9 @@
 )
 from cognite.client.utils._text import random_string
 
-TIMESERIES_EXTERNAL_IDS = [f"PYSDK DataPoint Subscription Test {no}" for no in range(20)]
+TIMESERIES_EXTERNAL_IDS = [
+    f"PYSDK DataPoint Subscription {platform.system()} {platform.python_version()} Test {no}" for no in range(20)
+]
```
haakonvt, line 26: May I suggest we make this a fixture?
```suggestion
def time_series_external_ids(os_and_py_version: str) -> list[str]:
    return [
        f"PYSDK DataPoint Subscription {os_and_py_version} Test {no}" for no in range(20)
    ]
```

tests/tests_integration/test_api/test_datapoint_subscriptions.py
```diff
@@ -261,10 +264,14 @@ def test_iterate_data_subscription_datapoints_added(
             assert batch.subscription_changes.added[0].external_id == time_series_external_ids[0]
 
             existing_data = cognite_client.time_series.data.retrieve_dataframe(external_id=time_series_external_ids[0])
+            if existing_data.empty:
+                start = pd.Timestamp("now")
+            else:
+                start = existing_data.index[-1] + pd.Timedelta("1d")
```
haakonvt, line 270: This seems like bootstrapping the time series with at least one datapoint? I think that belongs in a fixture!

haakonvt (commented):
I really hope this addresses the quite flaky dps sub tests 😄

tests/tests_integration/test_api/test_datapoint_subscriptions.py
doctrino: Sure can do that. What is the main motivation as it is just constants anyways?

haakonvt: I agree it is not necessary, but you get `os_an_py_version` for free which is nice 🤓

tests/tests_integration/test_api/test_datapoint_subscriptions.py
doctrino: This I am more skeptical about. If it is only used in one test, moving it to a fixture seems to obfuscate.

haakonvt (approved):

tests/tests_integration/test_api/test_datapoint_subscriptions.py
haakonvt: The test is using a random xid (see fixture `time_series_external_ids`), meaning all these xids should have an existing datapoint. Anyway, approving 👍

PR #1991, opened 2024-10-22
In filter with node ids

MortGron:
## Description
If you pass a list of NodeIds as values to the In filter and use this to filter instances in a data model, you get the following error:
`Invalid value for direct relation property 'property'. The value must be a json object with space and externalId fields. |  code: 400`
The same happens if you try to use the `dump` method on the NodeIds. You have to relalize that `include_instance_type` has to be set to False, which is not obvious by reading the error message.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Changes are grouped as follows
 
 ## [7.63.10] - 2024-10-22
 ### Added
-- Data modeling: When using the `In` filter to filter on direct relations, `NodeId`s can now be used directly as values.
+- Data modeling filters now support the use of `NodeId` (and `EdgeId`) directly.
 
 ## [7.63.9] - 2024-10-21
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.63.10] - 2024-10-22
+### Added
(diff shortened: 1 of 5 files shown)


cognite/client/data_classes/filters.py
```diff
@@ -30,7 +31,7 @@ class ParameterValue:
 
 
 FilterValue: TypeAlias = RawValue | PropertyReferenceValue | ParameterValue
-FilterValueList: TypeAlias = Sequence[RawValue] | PropertyReferenceValue | ParameterValue
+FilterValueList: TypeAlias = Sequence[RawValue | InstanceId] | PropertyReferenceValue | ParameterValue
```
haakonvt, line 34: I believe we should add it to `RawValue` instead, then `FilterValue` and `FilterValueList` will include it automatically.
```py
RawValue: TypeAlias = str | float | bool | Sequence | Mapping[str, Any] | Label
```

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.63.10] - 2024-10-22
+### Added
+- Data modeling: When using the `In` filter to filter on direct relations, `NodeId`s can now be used directly as values.
```
```suggestion
- Data modeling filters now support the use of `NodeId` (and `EdgeId`) directly.
```

haakonvt (approved):
LGTM! Another great addition! 🔥 A few comments left for you before you merge:

PR #1999, opened 2024-10-31
feat(workflows): Add support for simulation integration

lpereiracgn:
## Description

## Checklist:
- [x] Tests added/updated.


cognite/client/data_classes/workflows.py
```diff
@@ -219,6 +221,47 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        inputs (list[dict[str, Any]] | None): List of input overrides
+    """
+
+    task_type = "simulator"
+
+    def __init__(
+        self,
+        routine_external_id: str,
+        run_time: int | None = None,
+        inputs: list[dict[str, Any]] | None = None,
```
doctrino, line 239: Shouldn't this be `list[SimulationInput]`? Does it have to be a generic dict?

tests/tests_integration/test_api/test_data_workflows.py
```diff
@@ -540,3 +541,33 @@ def test_trigger_run_history(
         )
         # it would take too long to wait for the trigger to run, so we just check that the history is not None
         assert history is not None
+
+
+class TestSimIntInWorkflows:
+    def test_create_and_run_simint_workflow(
```
doctrino, line 547: Lets move this test into `TestWorkflowVersions`. In addition, use the try-finally pattern and a random string in the workflow id. See the `test_upsert_delete` test in that class. 

Note that the test name is misleading as you are not running the workflow. Suggest renaming to 
`test_upsert_delete_with_simulation_task`

cognite/client/data_classes/workflows.py
```diff
@@ -560,6 +605,49 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+    """
+
+    task_type: ClassVar[str] = "simulator"
+
+    def __init__(
+        self,
+        run_id: int | None,
+        logs: list[dict[str, Any]] | None,
+        status_message: str | None,
+        outputs: list[dict[str, Any]] | None,
```
doctrino, line 625: Both of these are not arbitrary objects. You need to add SimulationLog and SimulationRunResultData
image

doctrino (changes requested):
Great to have simulation support in Workflows.

Does this not need the `beta` flag set in the API? Can you verify with the workflow team that this is not necessary?
image

In addition, lets avoid `dict` when we can. For the users of the SDK it is difficult to work with.

lpereiracgn:
Already confirmed the beta header is not necessary on the workflows' side. Also, the review comments you left were already addressed @doctrino

cognite/client/data_classes/workflows.py
```diff
@@ -232,6 +239,47 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
         return output
 
 
+class SimulationTaskParameters(WorkflowTaskParameters):
+    """
+    The simulation parameters are used to specify the simulation routine to be executed.
+    Args:
+        routine_external_id (str): The external ID of the simulation routine to be executed.
+        run_time (int | None): Reference timestamp used for data pre-processing and data sampling.
+        inputs (list[SimulatorRoutineInputConstant | SimulatorRoutineInputTimeseries] | None): List of input overrides
```
evertoncolling, line 248: @lpereiracgn here inputs should be just (`referenceId`, `value` and `unit`) according to the API spec, but I'm seeing that `SimulatorRoutineInputConstant` and `SimulatorRoutineInputTimeseries` are the complete input definitions that we expect when defining a routine.

image

evertoncolling: This input type is the same one as the one required by the simulation run method
image

cognite/client/data_classes/workflows.py
```diff
@@ -573,6 +623,49 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+    """
+
+    task_type: ClassVar[str] = "simulation"
+
+    def __init__(
+        self,
+        run_id: int | None,
+        logs: list[dict[str, Any]] | None,
+        status_message: str | None,
+        outputs: list[SimulatorRoutineOutput] | None,
```
evertoncolling, line 643: Here as well, the type that comes out of workflow executions does not match the type that is being declared by `SimulatorRoutineOutput`:
image

polomani: removed the whole thing

cognite/client/data_classes/workflows.py
polomani: fixed this

cognite/client/data_classes/simulators.py
```diff
@@ -0,0 +1,49 @@
...
+        return cls(
+            name=resource["name"],
+        )
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        return super().dump(camel_case=camel_case)
+
+
+@dataclass
+class SimulationInputOverride(CogniteObject):
```
doctrino, line 31: What is the motivation to move this to a separate .py file?

Is it expected to be used with the simulator endpoints?

cognite/client/data_classes/simulators.py
```diff
@@ -0,0 +1,49 @@
...
+    reference_id: str
+    value: str | int | float
```
doctrino, line 33: It looks like the spec also accepts arrays.
```suggestion
    value: str | int | float | list[str] | list[int] | list[float]
```

cognite/client/data_classes/simulators.py
```diff
@@ -0,0 +1,49 @@
...
+    unit: SimulationValueUnitName | None = None
```
doctrino, line 34: Do you know how this relates to the Unit Endpoints? Is it something specific for simulations.

cognite/client/data_classes/workflows.py
```diff
@@ -238,6 +243,49 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        run_time (int | None): Reference timestamp used for data pre-processing and data sampling.
+        inputs (list[SimulationInputOverride] | None): List of input overrides.
+    """
+
+    task_type = "simulation"
+
+    def __init__(
+        self,
+        routine_external_id: str,
+        run_time: int | None = None,
```
doctrino, line 260: These are missing from the spec. Can you please update the API docs?

cognite/client/data_classes/workflows.py
```diff
@@ -238,6 +243,49 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+            if simulation.get("inputs")
+            else None,
+        )
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        simulation = {
+            "routineExternalId" if camel_case else "routine_external_id": self.routine_external_id,
+            "runTime" if camel_case else "run_time": self.run_time,
+            "inputs": [item.dump(camel_case) for item in self.inputs] if self.inputs else None,
+        }
```
doctrino, line 284: Skip optional 
```suggestion
        simulation = {
            "routineExternalId" if camel_case else "routine_external_id": self.routine_external_id,
        }
        if self.run_time:
            simulation["runTime" if camel_case else "run_time"] = self.run_time
        if self.inputs:
            simulation["inputs" if camel_case else "inputs"] = [item.dump(camel_case) for item in self.inputs]
```

cognite/client/data_classes/workflows.py
```diff
@@ -579,6 +629,44 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        status_message (str | None): Status message of the simulation execution.
+    """
+
+    task_type: ClassVar[str] = "simulation"
+
+    def __init__(
+        self,
+        run_id: int | None,
+        log_id: int | None,
+        status_message: str | None,
```
doctrino, line 647: Looking at the spec
```suggestion
        self,
        log_id: int,
        run_id: int | None = None,
        status_message: str | None = None,
```

doctrino (changes requested):
It will be good to get this one in, but two things.

1. Can you update the API docs to match what is actual there? This will cause a lot of confusion if it deviates.
2. A few smaller things, see the comments.

cognite/client/data_classes/workflows.py
polomani: https://api-docs.cognite.com/20230101-beta/tag/Workflow-versions/operation/CreateOrUpdateWorkflowVersion
Screenshot 2024-12-01 at 10 49 23
They're here, not sure what you mean?

cognite/client/data_classes/simulators.py
polomani: exactly. there will be many more simint classes added, @lpereiracgn is working on another PR for that

cognite/client/data_classes/simulators.py
polomani: Specific to simulators, this one doesn't have direct relation to units API yet.

cognite/client/data_classes/workflows.py
polomani: fixed

cognite/client/data_classes/simulators.py
doctrino: Does it make sense to move this into `/data_classes/simulators/<something>.py`? See for example what we did for `hosted_extractors/`. The thinking is that with more resources we start to get overlapping names, so we want to separate them into "namespace" which in python would be different packages.

polomani: moved

cognite/client/data_classes/simulators/runs.py
```diff
@@ -0,0 +1,58 @@
...
+
+from cognite.client.data_classes._base import (
+    CogniteObject,
+)
+from cognite.client.utils._experimental import FeaturePreviewWarning
+
+if TYPE_CHECKING:
+    from cognite.client import CogniteClient
+
+_WARNING = FeaturePreviewWarning(api_maturity="General Availability", sdk_maturity="alpha", feature_name="Simulators")
```
doctrino, line 16: Nice

doctrino (approved):

PR #2010, opened 2024-11-06
feature: iterate datapoints using `DatapointsAPI.__call__`

haakonvt:
https://cognitedata.atlassian.net/browse/DM-2319

```
`Iterate through datapoints in chunks, for one or more time series. <https://developer.cognite.com/api#tag/Time-series/operation/getMultiTimeSeriesDatapoints>`_

Note:
    Control memory usage by specifying ``chunk_size_time_series``, how many time series to iterate simultaneously and ``chunk_size_datapoints``,
    how many datapoints to yield per iteration. Note that in order to make efficient use of the API request limits, this method will never hold
    less than 100k datapoints in memory at a time, per time series.

    If you run with memory constraints, use ``return_arrays=True`` (the default).

    No empty chunk will ever be returned.

Args:
    queries (DatapointsQuery | Sequence[DatapointsQuery]): Query, or queries, using id, external_id or instance_id for time series to fetch data for. Individual settings in the DatapointsQuery take precedence over top-level settings. The options 'limit' and 'include_outside_points' are not supported.
    start (int | str | datetime.datetime | None): Inclusive start. Default: 1970-01-01 UTC.
    end (int | str | datetime.datetime | None): Exclusive end. Default: "now"
    aggregates (Aggregate | str | list[Aggregate | str] | None): Single aggregate or list of aggregates to retrieve. Available options: ``average``, ``continuous_variance``, ``count``, ``count_bad``, ``count_good``, ``count_uncertain``, ``discrete_variance``, ``duration_bad``, ``duration_good``, ``duration_uncertain``, ``interpolation``, ``max``, ``min``, ``step_interpolation``, ``sum`` and ``total_variation``. Default: None (raw datapoints returned)
    granularity (str | None): The granularity to fetch aggregates at. Can be given as an abbreviation or spelled out for clarity: ``s/second(s)``, ``m/minute(s)``, ``h/hour(s)``, ``d/day(s)``, ``w/week(s)``, ``mo/month(s)``, ``q/quarter(s)``, or ``y/year(s)``. Examples: ``30s``, ``5m``, ``1day``, ``2weeks``. Default: None.
    timezone (str | datetime.timezone | ZoneInfo | None): For raw datapoints, which timezone to use when displaying (will not affect what is retrieved). For aggregates, which timezone to align to for granularity 'hour' and longer. Align to the start of the hour, day or month. For timezones of type Region/Location, like 'Europe/Oslo', pass a string or ``ZoneInfo`` instance. The aggregate duration will then vary, typically due to daylight saving time. You can also use a fixed offset from UTC by passing a string like '+04:00', 'UTC-7' or 'UTC-02:30' or an instance of ``datetime.timezone``. Note: Historical timezones with second offset are not supported, and timezones with minute offsets (e.g. UTC+05:30 or Asia/Kolkata) may take longer to execute.
    target_unit (str | None): The unit_external_id of the datapoints returned. If the time series does not have a unit_external_id that can be converted to the target_unit, an error will be returned. Cannot be used with target_unit_system.
    target_unit_system (str | None): The unit system of the datapoints returned. Cannot be used with target_unit.
    ignore_unknown_ids (bool): Whether to ignore missing time series rather than raising an exception. Default: False
    include_status (bool): Also return the status code, an integer, for each datapoint in the response. Only relevant for raw datapoint queries, not aggregates.
    ignore_bad_datapoints (bool): Treat datapoints with a bad status code as if they do not exist. If set to false, raw queries will include bad datapoints in the response, and aggregates will in general omit the time period between a bad datapoint and the next good datapoint. Also, the period between a bad datapoint and the previous good datapoint will be considered constant. Default: True.
    treat_uncertain_as_bad (bool): Treat datapoints with uncertain status codes as bad. If false, treat datapoints with uncertain status codes as good. Used for both raw queries and aggregates. Default: True.
    chunk_size_datapoints (int): The number of datapoints per time series to yield per iteration. Must evenly divide 100k OR be an integer multiple of 100k. Default: 100_000.
    chunk_size_time_series (int | None): The max number of time series to yield per iteration (varies as time series get exhausted, but is always at least 1). Default: None (all given queries are returned).
    return_arrays (bool): Whether to return the datapoints as numpy arrays. Default: True.

Yields:
    DatapointsArray | DatapointsArrayList | Datapoints | DatapointsList: If return_arrays=True, a ``DatapointsArray`` object containing the datapoints chunk, or a ``DatapointsArrayList`` if multiple time series were asked for. When False, a ``Datapoints`` object containing the datapoints chunk, or a ``DatapointsList`` if multiple time series were asked for. If `ignore_unknown_ids` is `True`, a single time series is requested and it is not found, iteration will quit.

Examples:

    Iterate through the datapoints of a single time series with external_id="foo", in chunks of 25k:

        >>> from cognite.client.data_classes import DatapointsQuery
        >>> client = CogniteClient()
        >>> query = DatapointsQuery(external_id="foo", start="2w-ago")
        >>> for chunk in client.time_series.data(query, chunk_size=25_000):
        ...     pass  # do something with the datapoints chunk

    Iterate through datapoints from multiple time series, and do not return them as numpy arrays. As one or more time
    series get exhausted (no more data), they are no longer part of the returned "chunk list", but the order is
    still preserved (for the remaining).

    If you run with ``chunk_size_time_series=None``, an easy way to check is to use the ``.get`` method, as illustrated below:

        >>> from cognite.client.data_classes.data_modeling import NodeId
        >>> queries = [
        ...     DatapointsQuery(id=123),
        ...     DatapointsQuery(external_id="foo"),
        ...     DatapointsQuery(instance_id=NodeId("my-space", "my-ts-xid"))
        ... ]
        >>> for chunk_lst in client.time_series.data(query, return_arrays=False):
        ...     if chunk_lst.get(id=2) is None:
        ...         print("Time series with id=2 has no more datapoints!")

    A likely use case for iterating datapoints is to clone data from one project to another, while keeping a low memory
    footprint and without having to write very custom logic involving count aggregates (which won't work for string data)
    and time domain splitting.

    Here's an example of how to do so efficiently, while not ignoring bad and uncertain data (``ignore_bad_datapoints=False``)
    and copying status codes (automatically taken care of when requested, ``include_status=True``). The only assumption is that
    the time series are already created in the target project.

        >>> from cognite.client.utils import MIN_TIMESTAMP_MS, MAX_TIMESTAMP_MS
        >>> target_client = CogniteClient(...)
        >>> to_copy = client.time_series.list(data_set_external_ids="my-use-case")
        >>> queries = [DatapointsQuery(external_id=ts.external_id) for ts in to_copy]
        >>> for dps_chunk in client.time_series.data(
        ...     queries,  # may be several thousand time series...
        ...     chunk_size_time_series=20,  # control memory usage by specifying how many to iterate at a time
        ...     chunk_size_datapoints=100_000,
        ...     include_status=True,
        ...     ignore_bad_datapoints=False,
        ...     start=MIN_TIMESTAMP_MS,
        ...     end=MAX_TIMESTAMP_MS + 1,  # end is exclusive
        ... ):
        ...     target_client.time_series.data.insert_multiple(
        ...         [{"external_id": dps.external_id, "datapoints": dps} for dps in dps_chunk]
        ...     )
```

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -50,6 +50,7 @@ from cognite.client.utils._auxiliary import (
     exactly_one_is_not_none,
     find_duplicates,
     is_finite,
+    is_positive,
     split_into_chunks,
     split_into_n_parts,
     unpack_items_in_payload,
@@ -554,8 +555,8 @@ class DatapointsAPI(APIClient):
 
         Note:
             Control memory usage by specifying ``chunk_size_time_series``, how many time series to iterate simultaneously and ``chunk_size_datapoints``,
-            how many datapoints to yield per iteration. Note that in order to make efficient use of the API request limits, this method will never hold
-            less than 100k datapoints in memory at a time, per time series.
+            how many datapoints to yield per iteration. See full example in examples. Note that in order to make efficient use of the API request limits,
+            this method will never hold less than 100k datapoints in memory at a time, per time series.
 
             If you run with memory constraints, use ``return_arrays=True`` (the default).
 
@@ -649,7 +650,7 @@ class DatapointsAPI(APIClient):
             )
         chunk_fn = functools.partial(split_into_chunks, chunk_size=chunk_size_datapoints)
 
-        if not (chunk_size_time_series is None or is_finite(chunk_size_time_series)):
+        if not (chunk_size_time_series is None or is_positive(chunk_size_time_series)):
             raise ValueError(
                 f"'chunk_size_time_series' must be a positive integer or None, not {chunk_size_time_series}"
             )
@@ -709,6 +710,11 @@ class DatapointsAPI(APIClient):
             for query in to_fetch_queries:
                 ident = query.identifier
                 dps = dps_lst.get(**ident.as_dict(camel_case=False))
+                if isinstance(dps, list):
+                    raise RuntimeError(
+                        "When iterating datapoints, identifiers must be unique! You can not get around this by passing "
+                        "several of [id, external_id, instance_id] for the same underlying time series."
+                    )
                 # Update query.start for next iteration if ts is not yet exhausted:
                 if dps and len(dps) == request_limit:
                     new_start = cast(int, dps[-1].timestamp) + 1
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -707,21 +707,7 @@ class DatapointsAPI(APIClient):
             dps_lst: DatapointsArrayList | DatapointsList = (
                 fetcher.fetch_all_datapoints_numpy() if return_arrays else fetcher.fetch_all_datapoints()
             )
-            for query in to_fetch_queries:
-                ident = query.identifier
-                dps = dps_lst.get(**ident.as_dict(camel_case=False))
-                if isinstance(dps, list):
-                    raise RuntimeError(
-                        "When iterating datapoints, identifiers must be unique! You cannot get around this by passing "
-                        "several of [id, external_id, instance_id] for the same underlying time series."
-                    )
-                # Update query.start for next iteration if ts is not yet exhausted:
-                if dps and len(dps) == request_limit:
-                    new_start = cast(int, dps[-1].timestamp) + 1
-                    if query.end_ms > new_start:
-                        query.start = new_start  # manual cursoring ftw
-                        continue
-                alive_queries.pop(ident)
+            self._update_alive_queries_and_do_manual_cursoring(alive_queries, dps_lst, to_fetch_queries, request_limit)
 
             # We should never yield an empty chunk, so we filter out empty or exhausted time series from result
             # (need to rebuild to not keep references to those empty in various private "id lookups")
@@ -739,6 +725,29 @@ class DatapointsAPI(APIClient):
                 for all_chunks in itertools.zip_longest(*map(chunk_fn, dps_lst)):
                     # Filter out dps as ts get exhausted, then rebuild the Dps(Array)List container and yield chunk:
                     yield dps_lst_cls(list(filter(None, all_chunks)), cognite_client=self._cognite_client)
+
+    @staticmethod
+    def _update_alive_queries_and_do_manual_cursoring(
+        alive_queries: dict[Identifier, DatapointsQuery],
+        dps_lst: DatapointsArrayList | DatapointsList,
+        to_fetch_queries: list[DatapointsQuery],
+        request_limit: int,
+    ) -> None:
+        for query in to_fetch_queries:
+            ident = query.identifier
+            dps = dps_lst.get(**ident.as_dict(camel_case=False))
+            if isinstance(dps, list):
+                raise RuntimeError(
+                    "When iterating datapoints, identifiers must be unique! You cannot get around this by passing "
+                    "several of [id, external_id, instance_id] for the same underlying time series."
+                )
+            # Update query.start for next iteration if ts is not yet exhausted:
+            if dps and len(dps) == request_limit:
+                new_start = dps[-1].timestamp + 1
+                if query.end_ms > new_start:
+                    query.start = new_start  # manual cursoring ftw
+                    continue
+            alive_queries.pop(ident)
 
     def retrieve(
         self,
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -727,13 +727,13 @@ class DatapointsAPI(APIClient):
                 break
 
             if chunk_size_datapoints == request_limit:
-                yield dps_lst[0] if is_single else dps_lst  # type: ignore [return-value]
+                yield dps_lst[0] if is_single else dps_lst
             elif is_single:
                 yield from chunk_fn(dps_lst[0])  # type: ignore [misc]
             else:
                 for all_chunks in itertools.zip_longest(*map(chunk_fn, dps_lst)):
                     # Filter out dps as ts get exhausted, then rebuild the Dps(Array)List container and yield chunk:
-                    yield dps_lst_cls(list(filter(None, all_chunks)), cognite_client=self._cognite_client)  # type: ignore [return-value]
+                    yield dps_lst_cls(list(filter(None, all_chunks)), cognite_client=self._cognite_client)
 
     def retrieve(
         self,
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -536,17 +536,6 @@ class DatapointsAPI(APIClient):
         self,
         queries: DatapointsQuery | Sequence[DatapointsQuery],
         *,
-        start: int | str | datetime.datetime | None = None,
-        end: int | str | datetime.datetime | None = None,
-        aggregates: Aggregate | str | list[Aggregate | str] | None = None,
-        granularity: str | None = None,
-        timezone: str | datetime.timezone | ZoneInfo | None = None,
-        target_unit: str | None = None,
-        target_unit_system: str | None = None,
-        ignore_unknown_ids: bool = False,
-        include_status: bool = False,
-        ignore_bad_datapoints: bool = True,
-        treat_uncertain_as_bad: bool = True,
         chunk_size_datapoints: int = 100_000,
         chunk_size_time_series: int | None = None,
         return_arrays: bool = True,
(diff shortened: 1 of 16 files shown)


cognite/client/utils/_identifier.py
```diff
@@ -134,7 +146,7 @@ def is_instance_id(self) -> bool:
 
     def as_dict(self, camel_case: bool = True) -> dict[str, T_ID]:
         if isinstance(self.__value, InstanceId):
-            return {"instanceId": self.__value.dump(camel_case=camel_case, include_instance_type=False)}  # type: ignore[dict-item]
+            return {self.name(camel_case): self.__value.dump(camel_case=camel_case, include_instance_type=False)}  # type: ignore[dict-item]
```
haakonvt, line 149: Bugfix here for `camel_case=False`

docs/source/time_series.rst
```diff
@@ -98,6 +98,10 @@ Retrieve datapoints in time zone in pandas dataframe
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 .. automethod:: cognite.client._api.datapoints.DatapointsAPI.retrieve_dataframe_in_tz
 
+Iterate through datapoints in chunks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. automethod:: cognite.client._api.datapoints.DatapointsAPI.__call__
+
```
haakonvt, line 104: Any ideas on how to make sphinx make this pretty... hmm

doctrino: Tried a quick experiment in; https://github.com/cognitedata/cognite-sdk-python/pull/2015

My interpretation as a developer you should know that `__call__` and `__iter__` are dunder methods and how to call them. The moment we are talking about iterating in chunks, we are talking about a more advanced case and we expects more from the user.

cognite/client/_api/datapoints.py
```diff
@@ -517,6 +498,248 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
+
+        Examples:
+
+            Iterate through the datapoints of a single time series with external_id="foo", in chunks of 25k:
+
+                >>> from cognite.client import CogniteClient
+                >>> from cognite.client.data_classes import DatapointsQuery
+                >>> client = CogniteClient()
+                >>> query = DatapointsQuery(external_id="foo", start="2w-ago")
+                >>> for chunk in client.time_series.data(query, chunk_size=25_000):
```
doctrino, line 593:
```suggestion
                >>> for chunk in client.time_series.data(query, chunk_size_datapoints=25_000):
```

cognite/client/_api/datapoints.py
```diff
@@ -517,6 +498,248 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
+            If you run with ``chunk_size_time_series=None``, an easy way to check is to use the ``.get`` method, as illustrated below:
+
+                >>> from cognite.client.data_classes.data_modeling import NodeId
+                >>> queries = [
+                ...     DatapointsQuery(id=123),
+                ...     DatapointsQuery(external_id="foo"),
+                ...     DatapointsQuery(instance_id=NodeId("my-space", "my-ts-xid"))
+                ... ]
+                >>> for chunk_lst in client.time_series.data(query, return_arrays=False):
+                ...     if chunk_lst.get(id=2) is None:
```
doctrino, line 609:
```suggestion
                ...     if chunk_lst.get(id=123) is None:
```

cognite/client/_api/datapoints.py
```diff
@@ -517,6 +498,248 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
+                ...         print("Time series with id=2 has no more datapoints!")
```
doctrino, line 610:
```suggestion
                ...         print("Time series with id=123 has no more datapoints!")
```

cognite/client/_api/datapoints.py
```diff
@@ -517,6 +498,248 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
+                f"'chunk_size_time_series' must be a positive integer or None, not {chunk_size_time_series}"
+            )
+
+        user_queries = [queries] if (is_single := isinstance(queries, DatapointsQuery)) else queries
+        dps_lst_cls: type[DatapointsArrayList | DatapointsList] = (
+            DatapointsArrayList if return_arrays else DatapointsList
+        )
+
+        # Note: Next major we should remove id/xid/instance_id from the other retrieve methods and just accept queries:
+        qs_per_id_type = defaultdict(list)
```
doctrino, line 664: Just to help readability
```suggestion
        qs_per_id_type: dict[str, list[DatapointsQuery]] = defaultdict(list)
```

Should this loop be a private classmethod of `DatapointsQuery`?

cognite/client/_api/datapoints.py
```diff
@@ -517,6 +498,248 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
+        target_unit: str | None = None,
+        target_unit_system: str | None = None,
+        ignore_unknown_ids: bool = False,
+        include_status: bool = False,
+        ignore_bad_datapoints: bool = True,
+        treat_uncertain_as_bad: bool = True,
+        chunk_size_datapoints: int = 100_000,
+        chunk_size_time_series: int | None = None,
+        return_arrays: bool = True,
+    ) -> Iterator[DatapointsArray | DatapointsArrayList | Datapoints | DatapointsList]:
```
doctrino, line 553: These are all essentially default values as I understand it. Not sure about this suggestion, but what about having a factory method on  `DatapointQuery` that would achieve the same?

Then, you can simplify the signature of this function, which makes is more readable:
```suggestion
    def __call__(
        self,
        queries: DatapointsQuery | Sequence[DatapointsQuery],
        *,
        chunk_size_datapoints: int = 100_000,
        chunk_size_time_series: int | None = None,
        return_arrays: bool = True,
    ) -> Iterator[DatapointsArray | DatapointsArrayList | Datapoints | DatapointsList]:
```
If you do this reduction in parameter, we can use it as an experiment in training the user to use the `DatapointQuery` class instead as we are forcing them to do so. In case it turns out to be too complicated for endusers and we get a feedback, we can always introduce the default arguments later. Much easier to add then remove :)

cognite/client/_api/datapoints.py
```diff
@@ -517,6 +498,248 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
+                        "When iterating datapoints, identifiers must be unique! You can not get around this by passing "
+                        "several of [id, external_id, instance_id] for the same underlying time series."
+                    )
+                # Update query.start for next iteration if ts is not yet exhausted:
+                if dps and len(dps) == request_limit:
+                    new_start = cast(int, dps[-1].timestamp) + 1
+                    if query.end_ms > new_start:
+                        query.start = new_start  # manual cursoring ftw
+                        continue
+                alive_queries.pop(ident)
```
doctrino, line 724: Just for readability, could this be pulled out to a helper. Makes the overall loop easier to read (at least for me :P) 
```suggestion
self._update_query(alive_queries, dps_lst, request_limit, to_fetch_queries)
```

cognite/client/_api/datapoints.py
```diff
@@ -517,6 +498,248 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
+            )
+            dps_lst: DatapointsArrayList | DatapointsList = (
+                fetcher.fetch_all_datapoints_numpy() if return_arrays else fetcher.fetch_all_datapoints()
+            )
+            for query in to_fetch_queries:
+                ident = query.identifier
+                dps = dps_lst.get(**ident.as_dict(camel_case=False))
+                if isinstance(dps, list):
+                    raise RuntimeError(
+                        "When iterating datapoints, identifiers must be unique! You can not get around this by passing "
```
doctrino, line 715:
```suggestion
                        "When iterating datapoints, identifiers must be unique! You cannot get around this by passing "
```

doctrino (commented):
This is fantastic work! It is such a needed feature, great that you pick this one up.

One suggestion to simplify the signature and give maximum flexibility in the future. Otherwise just minor suggestions to make the code more readable (for me).

docs/source/time_series.rst
haakonvt: Interesting! As a developer, I really liked just having the functions listed directly (like in your experiment)! I do wonder if normal users prefer the "old way" with more descriptions though... But I take it you're okay with `__call__` just being that in the docs?

cognite/client/_api/datapoints.py
haakonvt: I'm always surprised on how little the doctests actually fail on 🫤 Great catch btw!

cognite/client/_api/datapoints.py
haakonvt: The problem with this approach is that you have to give the same parameters to all queries, instead of passing once at top-level. I am uncertain, but I think I favor keeping this method similar to the existing ones to make it feel as familiar as possible

cognite/client/_api/datapoints.py
haakonvt: Sure 👍 Like a manual cursoring method 😅

cognite/client/_api/datapoints.py
doctrino: My thinking is that you would have a factory method on the `DatapointsQuery` that allowed instansiate multiple queries with the same parameters. Something like shown below. The thinking is to make it clearer what is query parameters, and what are parameters about how to query. 

```python
DatapointsQuery.create_multiple(
        id: int | list[int],
        external_id: str | list[str],
        instance_id: NodeId | list[NodeId],
        start: int | str | datetime.datetime | None = None,
        end: int | str | datetime.datetime | None = None,
        aggregates: Aggregate | str | list[Aggregate | str] | None = None,
        granularity: str | None = None,
        timezone: str | datetime.timezone | ZoneInfo | None = None,
        target_unit: str | None = None,
        target_unit_system: str | None = None,
        ignore_unknown_ids: bool = False,
        include_status: bool = False,
        ignore_bad_datapoints: bool = True,
        treat_uncertain_as_bad: bool = True,
)
```

docs/source/time_series.rst
doctrino: Yes, I am absolutely okay with `__call__`. I agree with you as well that I like having the functions listed directly. It has the advantage that I can lookup the name of the function directly instead of having to know what it is called in the API.

cognite/client/_api/datapoints.py
haakonvt: Sure, I'll do it! I really like the short signature:
```py
        self,
        queries: DatapointsQuery | Sequence[DatapointsQuery],
        *,
        chunk_size_datapoints: int = 100_000,
        chunk_size_time_series: int | None = None,
        return_arrays: bool = True,
```

cognite/client/_api/datapoints.py
```diff
@@ -517,6 +498,216 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
...
+    ) -> Iterator[DatapointsList]: ...
+
+    def __call__(
+        self,
+        queries: DatapointsQuery | Sequence[DatapointsQuery],
+        *,
+        chunk_size_datapoints: int = 100_000,
+        chunk_size_time_series: int | None = None,
+        return_arrays: bool = True,
+    ) -> Iterator[DatapointsArray | DatapointsArrayList | Datapoints | DatapointsList]:
```
doctrino, line 542: This is very easy to read and understand.

doctrino (approved):
Great work, all good from me

PR #2016, opened 2024-11-10
Fix FunctionSchedulesAPI.__call__()

HMEiding:
## Description
`FunctionSchedulesAPI.__call__()` currently calls `APIClient._list_generator()`, which relies on `/schedules/list` supporting pagination. However, `/schedules/list` does not support pagination. Fix so that `FunctionSchedulesAPI.__call__()` instead calls `FunctionSchedulesAPI.list()`.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [7.64.12] - 2024-11-10
+## [7.64.12] - 2024-11-11
 ### Fixed
 - `FunctionSchedulesAPI.__call__()` calls `FunctionSchedulesAPI.list()` instead of `APIClient._list_generator()`.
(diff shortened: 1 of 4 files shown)


haakonvt (approved):
LGTM!

Can all 100 functions (typically the limit) have up to 100 schedules? If yes, you are setting max-limit to 10k inside `list`, and we do have customers with 200 as the limit I believe? (200 * 100 > 10k). Probably not a problem, but just wanted to check 👍

HMEiding:
> LGTM!
> 
> Can all 100 functions (typically the limit) have up to 100 schedules? If yes, you are setting max-limit to 10k inside `list`, and we do have customers with 200 as the limit I believe? (200 * 100 > 10k). Probably not a problem, but just wanted to check 👍

Yes, but we also have a limit of 1000 schedules per project, as documented [here](https://docs.cognite.com/cdf/functions/technical), so I think we're good. Let me know if I'm missing something.

PR #2046, opened 2024-11-24
anders/documents status

andersfylling:
- **feat(documents status): use new POST variant**
- **feat(document status): support external id and instance id**


haakonvt (dismissed):
LGTM

PR #2055, opened 2024-12-04
fix(workflows): support None value for simulation input unit override

polomani:
## Description
introduced a small bug my recent PR when adding simulation workflows.
basically, a null pointer exception, fixed now


doctrino (approved):
Nice catch

PR #2056, opened 2024-12-04
KeyError on retrieving an ExtractionPipeline with partial contact information

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/extractionpipelines.py
+++ b/cognite/client/data_classes/extractionpipelines.py
@@ -30,26 +30,23 @@ class ExtractionPipelineContact(CogniteObject):
     """A contact for an extraction pipeline
 
     Args:
-        name (str): Name of contact
-        email (str): Email address of contact
-        role (str): Role of contact, such as Owner, Maintainer, etc.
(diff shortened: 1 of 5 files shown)


cognite/client/data_classes/extractionpipelines.py
```diff
@@ -30,27 +30,24 @@ class ExtractionPipelineContact(CogniteObject):
...
+        send_notification (bool | None): Whether to send notifications to this contact or not
     """
 
-    def __init__(self, name: str, email: str, role: str, send_notification: bool) -> None:
+    def __init__(
+        self,
+        name: str | None = None,
+        email: str | None = None,
+        role: str | None = None,
+        send_notification: bool | None = None,
```
erlendvollset, line 44: Is this class used as both a read and write class? If only read, we don't really want the defaults.

erlendvollset (approved):

cognite/client/data_classes/extractionpipelines.py
doctrino: It is used for both.

PR #2057, opened 2024-12-04
[DOGE-15] Update Sequence Issue

doctrino:
## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/tests/tests_integration/test_api/test_sequences.py
+++ b/tests/tests_integration/test_api/test_sequences.py
@@ -1,3 +1,4 @@
+import unittest
 from unittest import mock
 
 import pytest
@@ -14,7 +15,12 @@ from cognite.client.data_classes import (
     SequenceUpdate,
     filters,
 )
-from cognite.client.data_classes.sequences import SequenceProperty, SortableSequenceProperty
+from cognite.client.data_classes.sequences import (
+    SequenceColumnWrite,
+    SequenceProperty,
+    SequenceWrite,
+    SortableSequenceProperty,
+)
 from cognite.client.exceptions import CogniteNotFoundError
 from cognite.client.utils._text import random_string
 from tests.utils import set_request_limit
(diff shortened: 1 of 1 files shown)


cognite/client/_api/sequences.py
```diff
@@ -692,15 +695,73 @@ def upsert(
...
+    def _convert_resource_to_patch_object(
+        cls,
+        resource: CogniteResource,
+        update_attributes: list[PropertySpec],
+        mode: Literal["replace_ignore_null", "patch", "replace"] = "replace_ignore_null",
+        cdf_item_by_id: dict[str, Sequence] | None = None,  # type: ignore[override]
+    ) -> dict[str, dict[str, dict]]:
+        update_obj = super()._convert_resource_to_patch_object(resource, update_attributes, mode)
+        if not isinstance(resource, SequenceWrite):
+            return update_obj
```
haakonvt, line 734: Add check to also return if `resource.columns=None and replace != "replace"`?

cognite/client/_api/sequences.py
```diff
@@ -692,15 +695,73 @@ def upsert(
...
+        else:
+            cdf_columns_by_external_id = {
+                column.external_id: column for column in cdf_item.columns or [] if column.external_id
+            }
+            columns_by_external_id = {column.external_id: column for column in resource.columns if column.external_id}
+            if to_remove := (set(cdf_columns_by_external_id.keys()) - set(columns_by_external_id.keys())):
+                update_obj["update"]["columns"]["remove"] = list(to_remove)
+            if to_add := (set(columns_by_external_id.keys()) - set(cdf_columns_by_external_id.keys())):
+                update_obj["update"]["columns"]["add"] = [columns_by_external_id[ext_id].dump() for ext_id in to_add]
+            if to_modify := (set(columns_by_external_id.keys()) & set(cdf_columns_by_external_id.keys())):
```
haakonvt, line 751: This doesn't seem to distinguish based on `mode` (`"replace_ignore_null", "patch", "replace"`)?

tests/tests_integration/test_api/test_sequences.py
```diff
@@ -329,3 +334,58 @@ def test_aggregate_unique_metadata_keys(self, cognite_client: CogniteClient, seq
...
+        created: Sequence | None = None
+        try:
+            created = cognite_client.sequences.create(original_sequence)
+
+            upserted = cognite_client.sequences.upsert(upsert)
+
+            retrieved = cognite_client.sequences.retrieve(external_id=upserted.external_id)
+
+            assert retrieved is not None
+            assert retrieved.as_write().columns.dump() == upsert.columns.dump()
```
haakonvt, line 388: This test implicitly uses `mode="replace_ignore_null"`, we should ensure `"replace"` and `"patch"` also works as expected

haakonvt (commented):
Great that you are addressing this one! We should decide on how we want to solve the general problem of not-updateable-fields for write-classes in general. Have you done some thinking around this in light of this PR?

cognite/client/_api/sequences.py
```diff
@@ -692,15 +695,73 @@ def upsert(
...
+            cdf_item_by_id = {}
         return self._upsert_multiple(
             item,
             list_cls=SequenceList,
             resource_cls=Sequence,
             update_cls=SequenceUpdate,
             input_resource_cls=Sequence,
             mode=mode,
+            # String is not recognized as hashable by mypy
+            cdf_item_by_id=cdf_item_by_id,  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
doctrino: Not all identifiers are string. Workflow and data modeling are two examples that uses frozen dataclasses.

cognite/client/_api/sequences.py
doctrino: You are right, good catch. Since columns are not nullable, I think `replace` and 'replace_ignore_null' should be the same behavior. While patch should not do the remove.

cognite/client/_api/sequences.py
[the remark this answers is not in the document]
haakonvt: Oh right, makes sense. Then you must use Mapping/MutableMapping which is covariant (`dict` is invariant)

cognite/client/_api/sequences.py
```diff
@@ -730,7 +730,7 @@ def _convert_resource_to_patch_object(
         cdf_item_by_id: dict[str, Sequence] | None = None,  # type: ignore[override]
     ) -> dict[str, dict[str, dict]]:
         update_obj = super()._convert_resource_to_patch_object(resource, update_attributes, mode)
-        if not isinstance(resource, SequenceWrite):
+        if not isinstance(resource, SequenceWrite) or not resource.columns:
```
haakonvt, line 733: ...but if `mode=="replace"`, empty columns should mean remove, right?

haakonvt:
There's still this question left from initial review:

> We should decide on how we want to solve the general problem of not-updateable-fields for write-classes in general. Have you done some thinking around this in light of this PR?

cognite/client/_api/sequences.py
doctrino: Good point

doctrino:
> There's still this question left from initial review:
> 
> > We should decide on how we want to solve the general problem of not-updateable-fields for write-classes in general. Have you done some thinking around this in light of this PR?

I consider that a separate topic apart from this PR. It is on my list of todos, but cannot be solved in this one.

haakonvt:
This PR breaks the example given in the docstring for `client.sequences.upsert`:
```
from cognite.client.data_classes import Sequence

client.sequences.upsert(
    Sequence(external_id="new_sequence", description="New sequence")
)
```

haakonvt:
Maybe, when I think about it, upsert should actually enforce that `columns` is set because if it doesn't exist - it is required from the API. So I suggest we just change the docstring example instead. Wdyt?

doctrino:
> Maybe, when I think about it, upsert should actually enforce that `columns` is set because if it doesn't exist (update failed), it will be created and then _it is required from the API_. So I suggest we just change the docstring example instead. Wdyt?

Agreed.

haakonvt (approved):
LGTM!

PR #2101, opened 2025-01-23
[CDFNEW-2189] Hotfix: mark `useTransformationCredentials` as optional

VerstraeteBert:


cognite/client/data_classes/workflows.py
```diff
@@ -324,7 +324,7 @@ def _load(cls, resource: dict, cognite_client: CogniteClient | None = None) -> T
         return cls(
             resource[cls.task_type]["externalId"],
             resource[cls.task_type]["concurrencyPolicy"],
-            resource[cls.task_type]["useTransformationCredentials"],
+            resource[cls.task_type].get("useTransformationCredentials"),
```
doctrino, line 327: I think we need to set the default here as this is used in the Write object for WorkflwoVersions.
```suggestion
            resource[cls.task_type].get("useTransformationCredentials", False),
```

cognite/client/data_classes/workflows.py
```diff
@@ -324,7 +324,7 @@ def _load(cls, resource: dict, cognite_client: CogniteClient | None = None) -> T
         return cls(
             resource[cls.task_type]["externalId"],
             resource[cls.task_type]["concurrencyPolicy"],
```
doctrino, line 326: This is also optional. with default 'fail'.

doctrino (changes requested):

doctrino (approved):

PR #2102, opened 2025-01-25
[DOGE-99] Setup new test project (add missing test data)

doctrino:
## Description
Missing test data from simulation team. They have been notified, in addition, the new project must enable simulations. But this PR is ready for review.

## Checklist:
- [ ] Tests added/updated.
- [ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [ ] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ ] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #2105, opened 2025-02-03
Set lower limits for create/update/delete of postgres gateway users

einarmo:
## Description
Set the limit for number of postgres gateway users to create/update/delete to 1.

These operations have a tendency to be extremely heavy, so the limits are low for now.

## Checklist:
- [ ] Tests added/updated.


haakonvt:
Unrelated tests fail currently due to #2102 , I'll see if I can get that fixed by EOD. Set it to auto-merge if you want, and I'll see it through.

PR #2120, opened 2025-02-22
Fix datapoints `retrieve_latest` with missing `instance_id` + ignore=True

haakonvt:
From slack: https://cognitedata.slack.com/archives/CM71GE7CN/p1740238206942469

### DIFF
--- a/cognite/client/utils/_identifier.py
+++ b/cognite/client/utils/_identifier.py
@@ -32,6 +32,22 @@ class InstanceId:
     def dump(self, camel_case: bool = True, include_instance_type: bool = True) -> dict[str, str]:
         return {"space": self.space, "externalId" if camel_case else "external_id": self.external_id}
 
+    @overload
+    @classmethod
+    def load_if(cls, data: None) -> None: ...
+
+    @overload
+    @classmethod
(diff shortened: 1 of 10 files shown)


doctrino (dismissed):

cognite/client/utils/_identifier.py
```diff
@@ -32,6 +32,22 @@ class InstanceId:
...
+    @overload
+    @classmethod
+    def load_if(cls, data: None) -> None: ...
+
+    @overload
+    @classmethod
+    def load_if(cls, data: dict[str, str] | tuple[str, str] | Self) -> Self: ...
+
+    @classmethod
+    def load_if(cls, data: dict[str, str] | tuple[str, str] | Self | None) -> Self | None:
```
erlendvollset, line 44: When do we not want this behaviour? Can we just add it to `load`?

haakonvt: Hmm, couldn’t this risk not raising on required-but-missing parts of objects? I’ll keep it in mind for v8, I don’t like the thought of extending the possible return types to include None in a minor version (although that may be okay to do?)

doctrino (approved):

PR #2121, opened 2025-02-25
[DOGE-106] Instance aggregate/search limit=None means maximum

doctrino:
### Fixed
- The `client.data_modeling.instances.aggregate()` method now correctly returns maximum, 1000, results when setting
  the `limit` parameter to `None`, `-1`, or `math.inf`.

## Checklist:
- [x] Tests added/updated.


erlendvollset (dismissed):

erlendvollset (dismissed):

cognite/client/_api/data_modeling/instances.py
```diff
@@ -1282,7 +1287,11 @@ def aggregate(
 
         self._validate_filter(filter)
         filter = self._merge_space_into_filter(instance_type, space, filter)
-        body: dict[str, Any] = {"view": view.dump(camel_case=True), "instanceType": instance_type, "limit": limit}
+        body: dict[str, Any] = {
+            "view": view.dump(camel_case=True),
+            "instanceType": instance_type,
+            "limit": self._AGGREGATE_LIMIT if is_unlimited(limit) or limit is None else limit,
```
haakonvt, line 1293:
```suggestion
            "limit": self._AGGREGATE_LIMIT if is_unlimited(limit) else limit,
```

ref its implementation:

```py
def is_unlimited(limit: float | int | None) -> bool:
    return limit in {None, -1, math.inf}
```

doctrino: The `limit is None` is just for `mypy`.

haakonvt: My `mypy` is happy without it

haakonvt (approved):
LGTM!

cognite/client/_api/data_modeling/instances.py
doctrino: Ok then it is me that did not trust mypy

cognite/client/_api/data_modeling/instances.py
```diff
@@ -1166,7 +1172,11 @@ def search(
         else:
             raise ValueError(f"Invalid instance type: {instance_type}")
 
-        body: dict[str, Any] = {"view": view.dump(camel_case=True), "instanceType": instance_type_str, "limit": limit}
+        body: dict[str, Any] = {
+            "view": view.dump(camel_case=True),
+            "instanceType": instance_type_str,
+            "limit": self._SEARCH_LIMIT if is_unlimited(limit) or limit is None else limit,
```
doctrino, line 1178:
```suggestion
            "limit": self._SEARCH_LIMIT if is_unlimited(limit) else limit,
```

haakonvt (dismissed):

finnag (changes requested):
It looks like you pretend that limit is an int (or None).  Should not be

doctrino:
> Previous comment was sent prematurely. Trying again:
> 
> How do you know the behavior is as you write? There should be tests. What happens if you specify -2, 0, or 1040 as limit?

The the API will give a 400 and tell you that it is an invalid limit. The philosophy of the PySDK is that we do as little validation of the input to the API as possible, this is such that we do not have to keep up with the validation of the API side. 

> Have we publicly documented 3 different ways of getting the max limit (-1, None, inf) ? If we have, we'll have to support that more or less forever. If we haven't, do we want to support it forever?

We support that for all other limit parameters in the SDK, so we have already committed to supporting it. I think we should continue to do so here, so we get consistency.

> Do we really support floats as limit (math.inf is a float..)

Yes, again for consistency with all other limit parameters.

cognite/client/_api/data_modeling/instances.py
```diff
@@ -164,6 +164,11 @@ def _load(data: dict, cognite_client: CogniteClient | None = None) -> NodeApply
 class InstancesAPI(APIClient):
     _RESOURCE_PATH = "/models/instances"
 
+    def __init__(self, config: ClientConfig, api_version: str | None, cognite_client: CogniteClient) -> None:
+        super().__init__(config, api_version, cognite_client)
+        self._AGGREGATE_LIMIT = 1000
```
dbrattli, line 169: Shouldn't these be class variables instead?

dbrattli:
Even if you don't do validation in the SDK, you still have to test that feature you make here actually works, that it's possible to return maximum by using e.g `None`.

finnag (changes requested):
This PR still lacks a test. 

There should be a test that that the values `None`, `-1`, or `math.inf` actually set the limit to 1000 instead, and that other values are passed on unmodified.

haakonvt (approved):
LGTM. Ideally the test would also include some random numbers/objects to ensure they are passed along untouched, but that is probably already covered by existing tests.

finnag (approved):

finnag:
:unicorn:

PR #2068, opened 2024-12-18
Add support for simulator models

abdullah-cognite:
## Description
Add support for simulator models to the SDK.

## Checklist:
- [x] Tests added/updated.
- [x] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [x] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [x] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #2127, opened 2025-03-05
fix(simulators) : delete model after update test

abdullah-cognite:
This part was missed out in the PR that added support for models here: https://github.com/cognitedata/cognite-sdk-python/pull/2068


haakonvt (approved):

finnag (approved):
:unicorn:

PR #2128, opened 2025-03-06
Use dedicated cdf test project for CD workflow

erlendvollset:
Using a dedicated cdf project for CD workflow for 2 reasons:
- No longer want a test project in staging since it's less stable
- Don't want devs to have write access to this project, so that we don't risk breaking master builds and we force devs to create test fixtures for setting up their data instead of ingesting it manually.

Variables and secrets have been updated accordingly through the github UI.

### DIFF
--- a/.github/workflows/cdf_auth.yaml
+++ b/.github/workflows/cdf_auth.yaml
@@ -16,11 +16,11 @@ jobs:
     container:
       image: cognite/toolkit:0.4.9
       env:
-        CDF_CLUSTER: ${{ vars.CDF_CLUSTER }}
-        CDF_PROJECT: ${{ vars.CDF_PROJECT }}
-        IDP_CLIENT_ID: ${{ vars.IDP_CLIENT_ID }}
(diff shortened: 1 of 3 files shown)


.github/workflows/cdf_auth.yaml
```diff
@@ -16,11 +16,11 @@ jobs:
     container:
       image: cognite/toolkit:0.4.9
       env:
-        CDF_CLUSTER: ${{ vars.CDF_CLUSTER }}
```
doctrino, line 19: I think you need to remove the `environment: CD` above.

erlendvollset: why?

doctrino: It looks like you are not using GitHub environment?

erlendvollset: Ah right, no just using repo variables.

doctrino (approved):

finnag (approved):
:unicorn:

PR #2139, opened 2025-03-27
Add support for simulator routine revisions

polomani:
## Description
Support for the /simulators/routines/revisions API endpoints. (list, retrieve, and create)
api docs: https://api-docs.cognite.com/20230101/tag/Simulator-Routines/operation/create_simulator_routine_revision_simulators_routines_revisions_post

## Note for the risk-review team
This is bigger than 500 lines of code but each commit is made the way that it's less than 500. It's currently not feasible to split this without a major rewrite.

## Checklist:
- [x] Tests added/updated.
- [ ] Documentation - no, we would like to do it once all the parts of the API are available.
- [ ] Version not bumped, we don't have to release it just yet

To avoid any misunderstanding here, the Simulators API is bigger, so this PR for instance doesn't include docs, to make it smaller. We will make a separate PR for docs once we have all the pieces.

### DIFF
--- a/cognite/client/_api/simulators/routine_revisions.py
+++ b/cognite/client/_api/simulators/routine_revisions.py
@@ -1,0 +1,358 @@
+from __future__ import annotations
+
+from collections.abc import Iterator, Sequence
+from typing import TYPE_CHECKING, NoReturn, overload
+
+from cognite.client._api_client import APIClient
+from cognite.client.data_classes.shared import TimestampRange
+from cognite.client.data_classes.simulators import PropertySort
+from cognite.client.data_classes.simulators.filters import SimulatorRoutineRevisionsFilter
+from cognite.client.data_classes.simulators.routine_revisions import (
+    SimulatorRoutineRevision,
+    SimulatorRoutineRevisionList,
+    SimulatorRoutineRevisionWrite,
+)
+from cognite.client.utils._experimental import FeaturePreviewWarning
+from cognite.client.utils._identifier import IdentifierSequence
+from cognite.client.utils._validation import assert_type
+from cognite.client.utils.useful_types import SequenceNotStr
+
+if TYPE_CHECKING:
+    from cognite.client import ClientConfig, CogniteClient
+
+
+class SimulatorRoutineRevisionsAPI(APIClient):
+    _RESOURCE_PATH = "/simulators/routines/revisions"
+
+    def __init__(self, config: ClientConfig, api_version: str | None, cognite_client: CogniteClient) -> None:
+        super().__init__(config, api_version, cognite_client)
+        self._warning = FeaturePreviewWarning(
+            api_maturity="General Availability", sdk_maturity="alpha", feature_name="Simulators"
+        )
+        self._LIST_LIMIT = 20
+        self._CREATE_LIMIT = 1
+        self._RETRIEVE_LIMIT = 20
+
+    def __iter__(self) -> Iterator[SimulatorRoutineRevision]:
+        """Iterate over simulator routine revisions
+
+        Fetches simulator routine revisions as they are iterated over, so you keep a limited number of simulator routine revisions in memory.
+
+        Returns:
+            Iterator[SimulatorRoutineRevision]: yields Simulator routine revisions one by one.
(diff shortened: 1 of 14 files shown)


cognite/client/_api/simulators/routine_revisions.py
```diff
@@ -0,0 +1,358 @@
...
+        identifiers = IdentifierSequence.load(ids=ids, external_ids=external_ids)
+        return self._retrieve_multiple(
+            resource_cls=SimulatorRoutineRevision,
+            list_cls=SimulatorRoutineRevisionList,
+            identifiers=identifiers,
+            resource_path="/simulators/routines/revisions",
+        )
+
+    @overload
+    def create(self, routine_revision: Sequence[SimulatorRoutineRevisionWrite]) -> SimulatorRoutineRevisionList: ...
```
thorkildcognite, line 185: Should these call the create() on line 190? Do we want placeholders like this in the ready code? (also see the one below)

tests/tests_integration/test_api/test_simulators/seed/data.py
```diff
@@ -232,3 +235,88 @@
     "name": "Simulator Routine - Test",
     "description": "Simulator Routine - Description Test",
 }
+
+
+simulator_routine_revision = {
+    "externalId": None,
+    "routineExternalId": resource_names["simulator_routine_external_id"],
+    "configuration": {
+        "schedule": {"enabled": True, "cronExpression": "*/10 * * * *"},
```
thorkildcognite, line 244: Again, the service should ensure some kind of jitter/random delay for these cron jobs

cognite/client/_api/simulators/routines.py
```diff
@@ -95,7 +99,7 @@ def __call__(
     def create(self, routine: Sequence[SimulatorRoutineWrite]) -> SimulatorRoutineList: ...
 
     @overload
-    def create(self, routine: SimulatorRoutineWrite) -> SimulatorRoutineList: ...
+    def create(self, routine: SimulatorRoutineWrite) -> SimulatorRoutine: ...
```
thorkildcognite, line 102: Save as previously. Do we want them to work for future compatibility?

tests/tests_integration/test_api/test_simulators/conftest.py
```diff
@@ -3,20 +3,26 @@
 import time
 from collections.abc import Iterator
 from pathlib import Path
+from typing import Any
```
thorkildcognite, line 6: AS before, Dag does not fully approve!

thorkildcognite:
@polomani I've added some comments, but they are not blocking. But, you must get a review by the team first before I can mark it as risk reviewed.

Also note that you use Any, which Dag claims is a bad smell https://cognitedata.slack.com/archives/C04U08HH4A3/p1742780947133449 . Although you may choose to not fix it here, you should fix it later on.

cognite/client/_api/simulators/routine_revisions.py
polomani: just following the sdk pattern here, maybe @cognitedata/python-sdk-maintainers have thoughs?

tests/tests_integration/test_api/test_simulators/conftest.py
polomani: it's a common approach in the SDK, Dag is looking for better ways, most likely it will require some rewrite of the internals

cognite/client/data_classes/simulators/routine_revisions.py
```diff
@@ -0,0 +1,437 @@
...
+            output["unit"] = self.unit.dump(camel_case=camel_case)
+
+        return output
+
+
+@dataclass
+class SimulatorRoutineSchedule(CogniteObject):
+    """
+    The schedule configuration of the simulator routine revision.
+
```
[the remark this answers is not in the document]
polomani: Totally makes sense, that's exactly what they do (spikes every 00 minute).
This cron actually runs on the connector (client side) and we are evaluating if it should be this way.

thorkildcognite:
risk review ok

PR #2144, opened 2025-04-01
Fix object agg bug when iterating

haakonvt:
# Description
### Fixed
- When iterating datapoints, object aggregates `min_datapoint` and `max_datapoint` no longer raise
  `ValueError: Unsupported aggregate` (used to require `include_status` to be explicitly passed.)

## Risk review
My opinion on the change: **Low risk**.

_Big picture_: Object aggregates were released yesterday and most probably don't know they exist yet. Also, the use of the iterative datapoints method is pretty new - and I've only seen internal usage, and those using it are only using it for raw copy/clone operations (which is why it was added).

_Why, but less handwavy_: Due to how we handle "parameters VS defaults VS not-given" in `DatapointsAPI.__call__`, requesting the brand new object aggregates (`min`- or `max_datapoint`) that was merged yesterday would fail with a ValueError because `include_status` did not use the correct default (which all existing retrieve methods use already as they have a concept of "overall defaults" built right into the function signature).

To future-proof `__call__`, the API defaults are now used when the user didn't specify them, similar to existing retrieve methods.

## Checklist:
- [x] Tests added/updated.


doctrino (dismissed):

dmivankov:
There's a merge conflict in changelog, resolving it will discard reviews so will need to be done before final round of reviews, but sooner is better :slightly_smiling_face:

cognite/client/data_classes/datapoints.py
```diff
@@ -266,22 +267,21 @@ class DatapointsQuery:
...
+            "granularity": None,
+            "timezone": None,
+            "target_unit": None,
+            "target_unit_system": None,
+            "limit": None,
+            "include_outside_points": False,
+            "ignore_unknown_ids": False,
+            "include_status": False,
+            "ignore_bad_datapoints": True,
+            "treat_uncertain_as_bad": True,
```
dmivankov, line 284: I'm not quite sure here, it seems to apply defaults to a range of parameters, not all of them were added recently

- do defaults here match defaults on API? Don't see all in https://api-docs.cognite.com/20230101/tag/Time-series/operation/getMultiTimeSeriesDatapoints some are handled in SDK?
- do existing tests cover this change? (see that no tests are updated, it could be that behavior stayed the same, but could be lack of coverage)

dmivankov: for 1st point I needed to expand "items" in body, there there's more parameters :+1:

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -754,6 +754,34 @@ def test_iterate_multiple_requests(self, cognite_client, all_test_time_series, r
         else:
             pytest.fail("Too few iterations/datapoints returned")
 
+    @pytest.mark.parametrize("retrieve_arrays", (False, True))
+    def test_iterating_object_aggregates(self, cognite_client, all_test_time_series, retrieve_arrays):
+        # Bug prior to 7.74.2, when iterating object aggregate datapoints like `min_datapoint` or `max_datapoint`,
+        # a 'ValueError: Unsupported aggregate' would be raised because `include_status` did not have a default value.
```
dmivankov, line 760: a quick question, if user had supplied value for include_status, iteration would work?

dmivankov: actually one more, on API side parameter looks optional, so what was failing?

cognite/client/data_classes/datapoints.py
haakonvt: These defaults should have been used from "day one". It now exactly matches the existing `retrieve` methods (`retrieve`, `retrieve_arrays`, `retrieve_dataframe` and `*_in_tz`). 

> ...not all of them were added recently

True, the problem was SDK code relating to the new object aggregates. The API happily returned datapoints, but unpacking failed.

tests/tests_integration/test_api/test_datapoints.py
haakonvt: > a quick question, if user had supplied value for include_status, iteration would work?

Yes.

> actually one more, on API side parameter looks optional, so what was failing?

The problem was SDK code relating to the new object aggregates. The API happily returned datapoints, but unpacking failed.

cognite/client/data_classes/datapoints.py
haakonvt: So, yes, these match the API, _although we are a tiny bit more lenient_. The API doesn't allow `ignoreUnknownIds` to be customized _per time series_ that you request datapoints from, but the SDK allows this to keep all options to be set individually, not "every option except ignoreUnknownIds".

The SDK also never passes any top-level values in the payload, and instead explicitly specifies all settings (that are not using the default value) for all time series.

doctrino (dismissed):

psalaberria002 (dismissed):
:unicorn:

psalaberria002 (approved):
🦄

PR #2152, opened 2025-05-05
feat(simint-alpha): consistency and dev experience of simulation run

evertoncolling:
## Description
Creating a new simulation run is a bit more complex than we would like it to be, for the basic use cases we would like to provide a way of running a single simulation in a simple way. It should be easy to wait syncronously for the simulation to finish and get logs when it fails.

- added a syntactic sugar to run the simulation via `routines.run()` that doesn't require creating s SimulationRunWrite object
- added retrieve for simulation runs
- added `SimulationRun.wait` so user can wait for the simulation to finish (similar to functions)
- syntactic sugar to read the logs of the run
- renamed `SimulatorRunsList` to `SimulatorRunList` for consistency, and other minor refactoring
- fixed a flaky test

NOTE: we have not officially released this part of the SDK yet, we wait for all resources to be in place before bumping the version/updating changelog / docs.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/simulators/routines.py
+++ b/cognite/client/_api/simulators/routines.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Iterator, Sequence
-from typing import TYPE_CHECKING, overload
+from typing import TYPE_CHECKING, Literal, overload
 
 from cognite.client._api.simulators.routine_revisions import SimulatorRoutineRevisionsAPI
 from cognite.client._api_client import APIClient
@@ -11,6 +11,12 @@ from cognite.client.data_classes.simulators.routines import (
     SimulatorRoutine,
     SimulatorRoutineList,
     SimulatorRoutineWrite,
+)
+from cognite.client.data_classes.simulators.runs import (
+    SimulationInputOverride,
+    SimulationRun,
+    SimulationRunWrite,
+    SimulatorRunsList,
 )
 from cognite.client.utils._experimental import FeaturePreviewWarning
 from cognite.client.utils._identifier import IdentifierSequence
@@ -23,6 +29,7 @@ if TYPE_CHECKING:
 
 class SimulatorRoutinesAPI(APIClient):
     _RESOURCE_PATH = "/simulators/routines"
+    _RESOURCE_PATH_RUN = "/simulators/run"
 
     def __init__(self, config: ClientConfig, api_version: str | None, cognite_client: CogniteClient) -> None:
         super().__init__(config, api_version, cognite_client)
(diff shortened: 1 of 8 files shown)


tests/tests_integration/test_api/test_simulators/test_runs.py
```diff
@@ -10,21 +9,14 @@ def test_list_filtering(self, cognite_client: CogniteClient, seed_resource_names
...
         for current_status in ["running", "success", "failure"]:
-            created_runs = cognite_client.simulators.runs.create(
-                [
-                    SimulationRunWrite(
-                        run_type="external",
-                        routine_external_id=routine_external_id,
-                    )
-                ]
-            )
+            created_run = cognite_client.simulators.routines.run(routine_external_id=routine_external_id)
```
polomani, line 12: `cognite_client.simulators.runs.create` supports multiple items (by firing multiple request), and this one is only one per call.
do we want to support multiple here?

evertoncolling: Just as Functions do not allow to trigger multiple calls in a single method, I think we could follow the same pattern here: https://cognite-sdk-python.readthedocs-hosted.com/en/latest/functions.html#call-function

If someone wants to create multiple, they need to be mindful of that and handle the logic.

It's a tradeoff of simplicity for newcomers, versus advanced feature for advanced usage. The previous pattern which required users to import `SimulationRunWrite` to trigger a simple simulation is too much verbose in my opinion.

evertoncolling: I gave it another thought after sleeping over it and it should be fine to have both options available in the SDK. One that is easy and intuitive for a newcomer and one that allows creating multiple runs in one go, since the SDK has support for this out of the box.

lpereiracgn (approved):

cognite/client/_api/simulators/routines.py
```diff
@@ -217,3 +222,45 @@ def list(
...
+            inputs=list(inputs) if inputs is not None else None,
+            run_time=run_time,
+            queue=queue,
+            log_severity=log_severity,
+        )
+
+        simulation_run = self._cognite_client.simulators.runs.create(run_object)
+
+        if wait:
+            simulation_run.wait()
```
abdullah-cognite, line 265: did we forget to pass the `timeout` to the wait function?

polomani: you are totally right, thanks

tests/tests_integration/test_api/test_simulators/test_routines.py
```diff
@@ -43,13 +43,6 @@ def test_sort(self, cognite_client: CogniteClient, seed_resource_names, seed_sim
...
-        )
-
-        assert len(routines_asc) > 0
-        assert len(routines_desc) > 0
-        assert len(routines_asc) == len(routines_desc)
-
-        assert routines_asc[0].external_id == routines_desc[-1].external_id
+        assert len(routines_asc) > 1
+        for i in range(1, len(routines_asc)):
+            assert routines_asc[i].created_time >= routines_asc[i - 1].created_time
```
polomani, line 48: this was flaky because we share the same `simulator_integration_external_id` between runs, so the list endpoint gives different items when multiple tests run at the same time

lpereiracgn (approved):

doctrino (commented):
One minor thing.

cognite/client/data_classes/simulators/runs.py
```diff
@@ -209,6 +215,44 @@ def as_write(self) -> SimulationRunWrite:
...
+        """
+        return self._cognite_client.simulators.logs.retrieve(id=self.log_id)
+
+    def update(self) -> None:
+        """Update the simulation run object to the latest state. Useful if the run was created with wait=False."""
+        # same logic as Cognite Functions
+        latest = self._cognite_client.simulators.runs.retrieve(ids=self.id)
+        if latest is None:
+            raise RuntimeError("Unable to update the simulation run object (it was not found)")
+        assert isinstance(latest, SimulationRun)
```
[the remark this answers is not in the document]
polomani: this was type narrowing, fixed by having "else"

doctrino (dismissed):
All good from me

cognite/client/data_classes/simulators/runs.py
```diff
@@ -201,6 +206,7 @@ def __init__(
         self.data_set_id = data_set_id
         self.user_id = user_id
         self.log_id = log_id
+        self._cognite_client = cast("CogniteClient", cognite_client)
```
psalaberria002, line 209: Is it guaranteed that we will get a client and not None?

polomani: this is a common pattern throughout the SDK, I assume it's not given it is available exactly on this line, but it will be injected by some internal magic.
@doctrino do you know why this is correct way of doing it, or we should remove the cast?

doctrino: Yes, it is. All `CogniteResource` (base class) uses a `_WithClientMixin`, https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/data_classes/_base.py#L90, that raises an appropriate error if the client is not there.

doctrino (approved):

PR #2156, opened 2025-05-19
Add an include_unit argument to datapoints retrieve_dataframe and to_pandas methods

MortGron:
…

## Description

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -1102,6 +1102,7 @@ class DatapointsAPI(APIClient):
         include_aggregate_name: bool = True,
         include_granularity_name: bool = False,
         column_names: Literal["id", "external_id", "instance_id"] = "instance_id",
+        include_unit: bool = False,
     ) -> pd.DataFrame:
         """Get datapoints directly in a pandas dataframe.
 
@@ -1132,6 +1133,7 @@ class DatapointsAPI(APIClient):
             include_aggregate_name (bool): Include 'aggregate' in the column name, e.g. `my-ts|average`. Ignored for raw time series. Default: True
             include_granularity_name (bool): Include 'granularity' in the column name, e.g. `my-ts|12h`. Added after 'aggregate' when present. Ignored for raw time series. Default: False
             column_names (Literal['id', 'external_id', 'instance_id']): Use either instance IDs, external IDs or IDs as column names. Time series missing instance ID will use external ID if it exists then ID as backup. Default: "instance_id"
+            include_unit (bool): Include units as a second level in the columns. If availible unit external IDs will be used, if not the basic unit property of the time series will be used. If none of the time series have units, the columns will not have a second level.
 
         Returns:
             pd.DataFrame: A pandas DataFrame containing the requested time series. The ordering of columns is ids first, then external_ids. For time series with multiple aggregates, they will be sorted in alphabetical order ("average" before "max").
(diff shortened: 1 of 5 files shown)


haakonvt:
@MortGron Is this still alive? Is it "still" ready for review?

MortGron:
> @MortGron Is this still alive? Is it "still" ready for review?

If you could have a look at it that would be great!

It is probably not ready in the sense that it has merge conflicts with the master branch.

The questions that probably needs to be discussed first are:

1.  Do we want unit informaion in the dataframe output? (I would say yes)
2. If yes, how should unit information be included? (I would say better with multiindex than to include information in the string columns that already exist)

cognite/client/_api/datapoints.py
```diff
@@ -1132,6 +1133,7 @@ def retrieve_dataframe(
             include_aggregate_name (bool): Include 'aggregate' in the column name, e.g. `my-ts|average`. Ignored for raw time series. Default: True
             include_granularity_name (bool): Include 'granularity' in the column name, e.g. `my-ts|12h`. Added after 'aggregate' when present. Ignored for raw time series. Default: False
             column_names (Literal['id', 'external_id', 'instance_id']): Use either instance IDs, external IDs or IDs as column names. Time series missing instance ID will use external ID if it exists then ID as backup. Default: "instance_id"
+            include_unit (bool): Include units as a second level in the columns. If availible unit external IDs will be used, if not the basic unit property of the time series will be used. If none of the time series have units, the columns will not have a second level.
```
haakonvt, line 1136: I think for simplicity I would like for the second level to always be added if `include_unit=True`. 
```suggestion
            include_unit (bool): Include units as a second level in the columns. If availible unit external IDs will be used, if not the basic unit property of the time series will be used.
```

cognite/client/data_classes/datapoints.py
```diff
@@ -980,6 +981,7 @@ def to_pandas(  # type: ignore [override]
             include_aggregate_name (bool): Include aggregate in the column name
             include_granularity_name (bool): Include granularity in the column name (after aggregate if present)
             include_status (bool): Include status code and status symbol as separate columns, if available.
+            include_unit (bool): Include units as a second level in the columns. If availible unit external IDs will be used, if not the basic unit property of the time series will be used. If the time series does not have a unit, the columns will not have a second level.
```
haakonvt, line 984: I also wonder if we should only support "proper" units as a way of saying "bye bye" to the previous free-text field. Wdyt?

tests/tests_integration/test_api/test_datapoints.py
```diff
@@ -2681,6 +2681,33 @@ def test_include_granularity_name_fails(self, cognite_client, one_mill_dps_ts):
                 id=one_mill_dps_ts[0].id, limit=5, granularity="1d", aggregates="min", include_granularity_name=None
             )
 
+    def test_include_unit(
```
haakonvt, line 2684: I would like a parametrize for include_status True/False as well since that logic seems to be touched

haakonvt (commented):
In general I like the addition of unit. If multi-level columns is the correct way, unsure (since we already pad the column with aggregate info and/or granularity as a string). But, you could also argue that these would be better suited to be also put into the multi-level columns. 

To be fair, I kind of like that - but that would mean `v8`. Wdyt?

MortGron:
> > I would say it is a general rule of thumb that one should not put multiple pieces of information into one string when that leads to the user needing to do potentially complex string manipulation. So very much in favor of also putting granularity and aggregate information into separate column levels.
> 
> Nice, then we agree! Do you have time to finish all these changes in the coming weeks?

Is there a deadline for V8 changes?

PR #2158, opened 2025-05-22
feat(simint-alpha): dev experience of simulation run data

evertoncolling:
## Description
Fetching results from a simulation is a bit more complex than we would like it to be, for the basic use cases we would like to provide a way for users to fetch data directly from the simulation run object (similar to how users can fetch response directly from a function call object). Also the format of the response from the API is not convenient for display in a notebook and for data scientists to work with the simulation results, so we're adding some convenience flattening into a pandas DataFrame in a similar way as it's done for Datapoints (source of inspiration within the SDK).

- added syntactic sugar to fetch simulation data for simulation runs
- added `to_pandas` method that transforms the simulation data object into a flatten representation, useful for rendering results in notebooks and for data manipulation

NOTE: we have not officially released this part of the SDK yet, we wait for all resources to be in place before bumping the version/updating changelog / docs.

CURRENT:
image

NEW:
image

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,7 @@ repos:
         entry: |
           poetry run dmypy run -- cognite tests/tests_integration/test_api/test_data_modeling
           tests/tests_unit/test_data_classes/test_data_models
+          tests/tests_unit/test_utils/test_time.py
         files: ^.*.(py|pyi)$
         language: system
         pass_filenames: false
--- a/cognite/client/_api/simulators/routine_revisions.py
+++ b/cognite/client/_api/simulators/routine_revisions.py
@@ -206,7 +206,7 @@ class SimulatorRoutineRevisionsAPI(APIClient):
                 ...         external_id="routine_rev_1",
                 ...         routine_external_id="routine_1",
                 ...         configuration=SimulatorRoutineConfiguration(
-                ...             data_sampling=SimulatorRoutineDataSampling(sampling_window=15, granularity="1m"),
+                ...             data_sampling=SimulatorRoutineDataSampling(sampling_window=15, granularity=1),
                 ...             inputs=[
                 ...                 SimulatorRoutineInputConstant(
                 ...                     name="Cold Water Temperature",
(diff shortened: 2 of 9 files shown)


cognite/client/data_classes/simulators/runs.py
```diff
@@ -424,10 +440,56 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        """Convert the simulation run data to a pandas DataFrame.
+
+        Returns:
+            pandas.DataFrame: The dataframe.
+        """
+        pd = local_import("pandas")
+
+        def _create_row(item: SimulationInput | SimulationOutput) -> dict:
+            """Create a row dictionary for an input or output item."""
+            item_type = (isinstance(item, SimulationInput) and "Input") or "Output"
```
doctrino, line 455: Any reason not to use a dump method? What happens if more properties are added to the simulation input?

cognite/client/data_classes/simulators/runs.py
```diff
@@ -424,10 +440,56 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+
+        return pd.DataFrame(rows)
+
 
 class SimulatorRunDataList(CogniteResourceList[SimulationRunDataItem], IdTransformerMixin):
     _RESOURCE = SimulationRunDataItem
 
+    def to_pandas(  # type: ignore [override]
+        self,
+    ) -> pandas.DataFrame:
```
doctrino, line 484: The goal of overwriting here is to simplify the signature?

doctrino (commented):
A few clarifications

cognite/client/data_classes/simulators/runs.py
polomani: I just followed what others done, but yeah, the base implementation has all these props which we don't need
```
    def to_pandas(
        self,
        camel_case: bool = False,
        expand_metadata: bool = False,
        metadata_prefix: str = "metadata.",
        convert_timestamps: bool = True,
    ) -> pandas.DataFrame:
```

doctrino (approved):
Good from me

PR #2160, opened 2025-05-27
[typing, tests] Datetime accepts a tzinfo parameter, use this directly instead of unpacking

qTipTip:
## Description

I had many typing errors in this file, so I decided to fix it. There is no new release or change in documentation; this is purely a change in tests.

We were constructing a dictionary with a single key `tzinfo` and then unpacking it for every call to datetime. The type linter complains that it does not recognize that this dict contains the `tzinfo` key, so it might as well be set explicitly.
 
## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,7 @@ repos:
         entry: |
           poetry run dmypy run -- cognite tests/tests_integration/test_api/test_data_modeling
           tests/tests_unit/test_data_classes/test_data_models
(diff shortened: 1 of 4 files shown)


dbrattli (approved):
:unicorn:

qTipTip:
> Please update the .pre-commit-config.yaml file to start running mypy on files where you fix typing.

Will do!

erlendvollset (approved):
LGTM

PR #2163, opened 2025-06-05
Maintain datatype retrieve rows

doctrino:
## Description

Avoid pandas converting integers to float in the method `client.raw.rows.retrieve_dataframe`.

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/raw.py
```diff
@@ -756,7 +756,10 @@ def retrieve_dataframe(
         else:
             idx = [r.key for r in rows]
         cols = [r.columns for r in rows]
-        return pd.DataFrame(cols, index=idx)
+        # We set dtype=object to allow for mixed types in the columns
+        # For example, if a column has both int and float values, we do not want pandas to convert
+        # the column to float64, as that would lose the int values.
+        return pd.DataFrame(cols, index=idx, dtype=object)
```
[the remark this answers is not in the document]
doctrino: What about introducing a new parameter, something like, `skip_dtype_conversion`? Then we set this to False by default and give a deprecation warning, then the user can set this to `True` and get this behavior.

tests/tests_unit/test_api/test_raw.py
```diff
@@ -378,6 +402,17 @@ def test_retrieve_dataframe_two_rows(self, cognite_client, mock_retrieve_raw_row
...
+    def test_retrieve_dataframe_integers(
+        self,
+        cognite_client: CogniteClient,
+        mock_retrieve_integer_rows,
+        integer_rows_response: dict,
+    ) -> None:
+        result = cognite_client.raw.rows.retrieve_dataframe(db_name="db1", table_name="table1", infer_dtypes=False)
+
+        actual = result.to_dict(orient="index")
+        assert actual == {row["key"]: row["columns"] for row in integer_rows_response["items"]}
```
finnag, line 414: Does this test actually test what you want to test?
It's python after all, so
```
>>> { "foo" : 1 } == {"foo" : 1.0}
True
```

doctrino: Adding an additional assert as well.

finnag:
:unicorn:

finnag (approved):

erlendvollset (approved):

PR #2165, opened 2025-06-09
Add argument to use instance IDs as column names in insert dataframe

MortGron:
## Description
The `insert_dataframe` methods does not accept instance IDs as column names. This PR adds this ability.

It adds the ability for the column names both to be of type `NodeId` or `tuple`. Dictionary style instnace IDs will not work as they are not hashable.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ Changes are grouped as follows
 ### Fixed
 - Fixes type annotations for Functions API. Adds new `FunctionHandle` type for annotating function handles.
 
+## [7.76.0] - 2025-05-15
+### Added
+- When ingesting datapoints, `insert_dataframe` now accepts instance IDs as column names.
+
 ## [7.75.1] - 2025-05-15
 ### Fixed
 - Fixes missing `instance_id` field in `Document` class returned from `client.documents.list()` and `client.documents.search()`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ Changes are grouped as follows
 
 ## [7.76.0] - 2025-05-15
 ### Added
-- When ingesting datapoints, `insert_dataframe` now accepts instance IDs as column names.
+- When ingesting datapoints, `insert_dataframe` now accepts instance IDs as column names when `instance_id_headers` is `True`. Note, in th next major release of the SDK, the behaviour of the column names will change and the ID type of the column will be determined based on the type of the column name. E.g. if the column name is of type `NodeId` it will automatically be interpreted as instance ID.
 
 ## [7.75.1] - 2025-05-15
 ### Fixed
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -2290,7 +2290,7 @@ class DatapointsAPI(APIClient):
                     dps.append({"datapoints": datapoints, "instanceId": NodeId.load(column_id)})
                 else:
                     raise ValueError(
-                        f"Could not find instance Ids in the coulnm header. InstanceId are given as NodeId or tuple. Got {type(column_id)}"
+                        f"Could not find instance IDs in the column header. InstanceId are given as NodeId or tuple. Got {type(column_id)}"
                     )
             else:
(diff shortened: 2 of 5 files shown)


MortGron:
> Just a fyi, planning to remove these args in next major, see eg. #2116

@haakonvt that looks great! There is an immediate practical need for being able to ingest dataframes to time series with instnace IDs, so the quesion becomes what is the timeline for V8?

haakonvt:
> > Just a fyi, planning to remove these args in next major, see eg. #2116
> 
> @haakonvt that looks great! There is an immediate practical need for being able to ingest dataframes to time series with instnace IDs, so the quesion becomes what is the timeline for V8?

The timeline is sept/oct this year so I’d say go ahead with this! But please update the documentation to reflect the coming changes already 👌

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ Changes are grouped as follows
 ### Fixed
 - Fixes type annotations for Functions API. Adds new `FunctionHandle` type for annotating function handles.
 
+## [7.76.0] - 2025-05-15
+### Added
+- When ingesting datapoints, `insert_dataframe` now accepts instance IDs as column names when `instance_id_headers` is `True`. Note, in th next major release of the SDK, the behaviour of the column names will change and the ID type of the column will be determined based on the type of the column name. E.g. if the column name is of type `NodeId` it will automatically be interpreted as instance ID.
```
MortGron, line 26: @haakonvt was it something like this you had in mind?

cognite/client/_api/datapoints.py
```diff
@@ -2280,6 +2285,11 @@ def insert_dataframe(self, df: pd.DataFrame, external_id_headers: bool = True, d
                 continue
             if external_id_headers:
                 dps.append({"datapoints": datapoints, "externalId": column_id})
+            elif instance_id_headers:
+                if isinstance(column_id, NodeId):
+                    dps.append({"datapoints": datapoints, "instanceId": column_id})
+                elif isinstance(column_id, tuple):
+                    dps.append({"datapoints": datapoints, "instanceId": NodeId.load(column_id)})
```
doctrino, line 2292: The `.load` method of `NodeId` can catch this already. In addition, if the user pass in `instance_id_headers` but don't have that type of headers, I think we should raise and not let it pass silently. 
```suggestion
                if isinstance(column_id, NodeId | tuple | dict):
                    dps.append({"datapoints": datapoints, "instanceId": NodeId.load(column_id)})
                else:
                    raise ValueError("Could not find instance Ids in the coulnm header. Please InstanceId are given as NodeId, tuple, or dict. Got {type(column_id)}"
```

cognite/client/_api/datapoints.py
```diff
@@ -2280,6 +2285,13 @@ def insert_dataframe(self, df: pd.DataFrame, external_id_headers: bool = True, d
                 continue
             if external_id_headers:
                 dps.append({"datapoints": datapoints, "externalId": column_id})
+            elif instance_id_headers:
+                if isinstance(column_id, NodeId | tuple):
+                    dps.append({"datapoints": datapoints, "instanceId": NodeId.load(column_id)})
```
larshagencognite, line 2290: How do we handle a load failure here?

cognite/client/_api/datapoints.py
```diff
@@ -2280,6 +2285,13 @@ def insert_dataframe(self, df: pd.DataFrame, external_id_headers: bool = True, d
...
+                else:
+                    raise ValueError(
```
larshagencognite, line 2292: This is not tested

cognite/client/_api/datapoints.py
MortGron: Are you thinking of the case when `column_id` is of type `tuple` and have the wrong number of elements?
Do you want a custom error message, not the one given by `NodeId.load` when load fails?

MortGron: @larshagencognite This PR will improve the error message for the user in this case https://github.com/cognitedata/cognite-sdk-python/pull/2179

PR #2173, opened 2025-06-17
[POFSP-1033] Simulator Integration docs & consistency fixes

evertoncolling:
## Description

This PR might look like a breaking change, but the SDK implementation is alpha, so we are allowed to break it until it matures (no usage at this point).

Changes:
1. Improves the doc strings and examples for multiple methods in simulator integration namespace
2. Enabled simulator docs in the SDK docs portal
2. The changes to method signatures and parameter names, such as standardizing on ids and external_ids to enhance the developer experience

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/simulators/logs.py
+++ b/cognite/client/_api/simulators/logs.py
@@ -54,11 +54,8 @@ class SimulatorLogsAPI(APIClient):
                 >>> logs = client.simulators.logs.retrieve(id=model.log_id)
 
             Get simulator logs by simulator integration id:
-                >>> res = client.simulators.integrations.list(
-                ...     filter=SimulatorIntegrationFilter(active=True)
-                ... )
-                >>> integration = res[0]
-                >>> logs = client.simulators.logs.retrieve(id=integration.log_id)
+                >>> integrations = client.simulators.integrations.list()
+                >>> logs = client.simulators.logs.retrieve(id=integrations[0].log_id)
 
             Get simulator logs by simulation run id:
                 >>> run = client.simulators.runs.retrieve(id=1)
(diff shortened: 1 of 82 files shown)


cognite/client/_api/simulators/logs.py
```diff
@@ -32,8 +32,7 @@ def retrieve(
     ) -> SimulatorLogList | None: ...
 
     def retrieve(self, id: int | Sequence[int]) -> SimulatorLogList | SimulatorLog | None:
```
polomani, line 34:
```suggestion
    def retrieve(self, ids: int | Sequence[int]) -> SimulatorLogList | SimulatorLog | None:
```
change to plural here as well, right?

polomani (approved):
LGTM

cognite/client/data_classes/simulators/routines.py
```diff
@@ -25,8 +25,6 @@ class SimulatorRoutineCore(WriteableCogniteResource["SimulatorRoutineWrite"], AB
     Each model can have multiple routines, each performing different objectives such as calculating optimal
     operation setpoints, forecasting production, benchmarking asset performance, and more.
 
-    Each simulator routine can have a maximum of 10 revisions
```
doctrino, line 28: Not true?

doctrino (approved):
no brainer

polomani (commented):
🦄

cognite/client/data_classes/simulators/routines.py
doctrino: Should probably be an API then that gives you the limit for your project?

PR #2210, opened 2025-07-25
Fix flaky workflow tests

doctrino:
## Description

The `retrieve_workflow` tests is flaky and has failed a few times
image

**The fix**: The fixture `workflow_list` is only used to test retrieve/list. So instead of recreating new workflows each run, I have made them permanent test data. Then, they should always be availalbe.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/tests/tests_integration/test_api/test_workflows.py
+++ b/tests/tests_integration/test_api/test_workflows.py
@@ -88,22 +88,27 @@ def new_workflow(cognite_client: CogniteClient, data_set: DataSet):
 
 
 @pytest.fixture(scope="class")
-def workflow_list(cognite_client: CogniteClient, data_set: DataSet):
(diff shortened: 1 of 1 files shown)


tests/tests_integration/test_api/test_workflows.py
```diff
@@ -88,22 +88,27 @@ def new_workflow(cognite_client: CogniteClient, data_set: DataSet):
 
 
 @pytest.fixture(scope="class")
-def workflow_list(cognite_client: CogniteClient, data_set: DataSet):
+def workflow_list(cognite_client: CogniteClient, data_set: DataSet) -> WorkflowList:
```
erlendvollset, line 91: To make it clear that this is not created on the fly.
```suggestion
def persisted_workflow_list(cognite_client: CogniteClient, data_set: DataSet) -> WorkflowList:
```

erlendvollset:
🦄

PR #2218, opened 2025-07-29
feat(simulators): Add support for external dependencies 

lpereiracgn:
## Description :pencil:

This PR introduces support for the external dependencies feature in simulator model revisions, this feature was recently added to the Simulator Integration API in this [PR](https://github.com/cognitedata/infrastructure/pull/22612), the spec can be found [here](https://github.com/cognitedata/service-contracts/pull/2886) 

## Checklist:
- [x] Tests added/updated.


cognite/client/data_classes/simulators/models.py
```diff
@@ -323,3 +354,60 @@ def _get_update_properties(cls, item: CogniteResource | None = None) -> list[Pro
...
+
+    @classmethod
+    def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
+        return cls(
+            id=resource["id"],
+        )
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        output = super().dump(camel_case=camel_case)
+        return output
```
abdullah-cognite, line 377: isn't this redundant?

tests/tests_integration/test_api/test_simulators/conftest.py
```diff
@@ -39,17 +39,47 @@ def seed_resource_names(cognite_client: CogniteClient) -> dict[str, str]:
     return resource_names.copy()
 
 
-@pytest.fixture(scope="session")
-def seed_file(cognite_client: CogniteClient, seed_resource_names: dict[str, str]) -> Iterator[FileMetadata | None]:
+def upload_file(
+    cognite_client: CogniteClient, filename: str, external_id: str, seed_resource_names: dict[str, str]
```
abdullah-cognite, line 43: why not just pass the `data_set_id`, instead of `seed_resource_names`?

tests/tests_integration/test_api/test_simulators/test_models.py
```diff
@@ -184,6 +190,54 @@ def test_create_model_and_revisions(
         assert len(multiple_model_revisions_created) == 2
         cognite_client.simulators.models.delete(external_ids=[model_external_id_1, model_external_id_2])
 
+    @pytest.mark.usefixtures("seed_model_revision_file", "seed_external_dependency_file", "seed_resource_names")
+    def test_create_model_and_revisions_with_external_dependencies(
```
abdullah-cognite, line 194: could we have used parameters with the original test to also test dependencies?
something like:
```
@pytest.mark.parametrize(
    "with_dependencies,expected_dependency_count", 
    [(True, 1), (False, 0)]
)
```

tests/tests_integration/test_api/test_simulators/test_models.py
```diff
@@ -184,6 +190,54 @@ def test_create_model_and_revisions(
...
+        ]
+
+        models_created = cognite_client.simulators.models.create(models_to_create)
+
+        assert models_created is not None
+        assert len(models_created) == 1
+        model_revision_external_id = model_external_id + "revision"
+        external_dependencies = [
+            SimulatorModelRevisionExternalDependency(
+                file=SimulatorModelExternalDependencyFileField(id=seed_external_dependency_file.id),
```
abdullah-cognite, line 219: should we also test for non-existent file ids?

tests/tests_integration/test_api/test_simulators/seed/ExtDependency.out
```diff
@@ -0,0 +1 @@
+External dependency file
```
abdullah-cognite, line 1: missing new line

tests/tests_integration/test_api/test_simulators/test_models.py
```diff
@@ -184,6 +190,54 @@ def test_create_model_and_revisions(
...
+            description="Test revision",
+            external_dependencies=external_dependencies,
+        )
+
+        model_revision_created = cognite_client.simulators.models.revisions.create(model_revision_to_create)
+
+        assert model_revision_created is not None
+        assert model_revision_created.external_id == model_revision_external_id
+        assert model_revision_created.external_dependencies == external_dependencies
+        cognite_client.simulators.models.delete(external_ids=[model_external_id])
```
abdullah-cognite, line 239: should we use an approach like this where the cleanup always happen in the finally block:
```

try:
    # Test code here
finally:
    # Cleanup code here
```

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [7.80.0] - 2025-08-04
```
abdullah-cognite, line 20: A comment related to documentation, in the `model_revisions.py` file, we could update the docstring with dependencies as well. 

Screenshot 2025-08-05 at 13 47 01

abdullah-cognite:
should we also add a convenience method to the `SimulatorModelRevisionsAPI` to get external dependencies?

like : 

````
def get_external_dependencies(
    self, 
    id: int | None = None, 
    external_id: str | None = None
) -> list[SimulatorModelRevisionExternalDependency]:
    """Get external dependencies for a simulator model revision."""
    revision = self.retrieve(ids=id, external_ids=external_id)
    if revision and revision.external_dependencies:
        return revision.external_dependencies
    return []
````

File: tests/tests_integration/test_api/test_simulators/test_models.py (hunk 3 above)
Comment (lpereiracgn) [reply tests/tests_integration/test_api/test_simulators/test_models.py:194]: It's not exactly what you're asking for but I tried to extend the original test here https://github.com/cognitedata/cognite-sdk-python/pull/2218/commits/a7b4eedd9e3b8e8766ad6957ea53723e7c8541e1

But I figured we already had too much happening already in that test. It's long enough as it is so I didn't want to make it larger.

File: tests/tests_integration/test_api/test_simulators/seed/data.py [hunk 8]
```diff
@@ -25,6 +25,7 @@
     "simulator_model_external_id": f"py_sdk_integration_tests_model_{random_str}",
     "simulator_model_revision_external_id": f"py_sdk_integration_tests_model_{random_str}_v1",
     "simulator_model_file_external_id": "ShowerMixer_simulator_model_file_5",
+    "simulator_model_external_dependency_file_external_id": "Prosper_simulator_model_external_dependency_file",
```
polomani, line 28: don't use Prosper references here. also use rename `.out` to something else (txt, csv, etc)

tests/tests_integration/test_api/test_simulators/conftest.py
```diff
@@ -39,17 +39,46 @@ def seed_resource_names(cognite_client: CogniteClient) -> dict[str, str]:
...
+def seed_external_dependency_file(
+    cognite_client: CogniteClient, seed_resource_names: dict[str, str]
+) -> Iterator[FileMetadata | None]:
+    data_set_id = seed_resource_names["simulator_test_data_set_id"]
+    file = upload_file(
+        cognite_client,
+        filename="ExtDependency.out",
+        external_id=seed_resource_names["simulator_model_external_dependency_file_external_id"],
+        data_set_id=data_set_id,
+    )
```
polomani, line 80: if fixture is only used for a single test -> call `upload_file` it inline from the test, don't need the fixture

cognite/client/_api/simulators/models_revisions.py
```diff
@@ -233,6 +234,15 @@ def create(
                 ...         external_id="revision2",
                 ...         file_id=2,
                 ...         model_external_id="a_2",
+                ...         external_dependencies = [
+                ...             SimulatorModelRevisionExternalDependency(
+                ...                 file=SimulatorModelExternalDependencyFileField(id=seed_external_dependency_file.id),
```
polomani, line 239:
```suggestion
                ...                 file=SimulatorModelExternalDependencyFileField(id=123),
```

tests/tests_integration/test_api/test_simulators/seed/ExtDependency.out
polomani, line 1: change the extension of this file

polomani (commented):
added some comments, try to not mention any particular simulator here

tests/tests_integration/test_api/test_simulators/conftest.py
lpereiracgn: it's being called from two tests: `test_create_model_and_revisions` ( existing) and `test_create_model_and_revisions_with_external_dependencies` ( the one I introduced )

tests/tests_integration/test_api/test_simulators/conftest.py
```diff
@@ -39,23 +39,54 @@ def seed_resource_names(cognite_client: CogniteClient) -> dict[str, str]:
...
+
     yield file
 
 
 @pytest.fixture(scope="session")
 def seed_simulator(cognite_client: CogniteClient, seed_resource_names: dict[str, str]) -> Iterator[None]:
     simulator_external_id = seed_resource_names["simulator_external_id"]
+    # seed data changed, adding this temporarily to clean up the simulator
+    cognite_client.simulators._post("/simulators/delete", json={"items": [{"externalId": simulator_external_id}]})
```
lpereiracgn, line 89: This needs to be removed once the CI is green, just leaving the comment here so I don't forget

tests/tests_integration/test_api/test_simulators/conftest.py
polomani: I am talking about `seed_external_dependency_file`, not the `seed_model_revision_file`.
it is used in a single test

lpereiracgn: Fixed https://github.com/cognitedata/cognite-sdk-python/pull/2218/commits/743024d291ac50fd96e27fdbc89368cbbea727be

polomani (approved):
LGTM

doctrino (approved):
Just a clarifying question from me, nothing blocking.

cognite/client/data_classes/simulators/models.py
```diff
@@ -323,3 +354,57 @@ def _get_update_properties(cls, item: CogniteResource | None = None) -> list[Pro
...
+@dataclass
+class SimulatorModelRevisionExternalDependency(CogniteObject):
+    """
+    Represents an external dependency for a simulator model revision.
+    Args:
+        file (SimulatorModelExternalDependencyFile): The CDF file associated with the external dependency.
+        arguments (dict[str, str]): A dictionary that contains the key-value pairs (fields) for the external dependency.
+    """
+
+    file: SimulatorModelExternalDependencyFile
```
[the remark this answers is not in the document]
polomani: @doctrino we will be working on DM support, this way it's easier to expand the API for that

cognite/client/data_classes/simulators/models.py
```diff
@@ -323,3 +354,57 @@ def _get_update_properties(cls, item: CogniteResource | None = None) -> list[Pro
...
+        )
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        output = super().dump(camel_case=camel_case)
+        output["file"] = self.file.dump(camel_case=camel_case)
+        return output
+
+    @classmethod
+    def _load_list(
+        cls, resource: dict[str, Any] | list[dict[str, Any]], cognite_client: CogniteClient | None = None
```
erlendvollset, line 402: same question on why this can't just be
```suggestion
        cls, resource: list[dict[str, Any]], cognite_client: CogniteClient | None = None
```

cognite/client/data_classes/simulators/models.py
```diff
@@ -41,8 +45,22 @@ def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None =
...
+            if "externalDependencies" in resource
+            else None,
         )
 
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        output = super().dump(camel_case=camel_case)
+        if isinstance(self.external_dependencies, list) and all(
+            isinstance(item, SimulatorModelRevisionExternalDependency) for item in self.external_dependencies
+        ):
+            output["externalDependencies"] = [item.dump(camel_case=camel_case) for item in self.external_dependencies]
```
erlendvollset, line 60: Typing already ensures that this is `list[SimulatorModelRevisionExternalDependency] | None`
```suggestion
        if self.external_dependencies is not None:
            output["externalDependencies"] = [item.dump(camel_case=camel_case) for item in self.external_dependencies]
```

cognite/client/data_classes/simulators/models.py
```diff
@@ -323,3 +354,57 @@ def _get_update_properties(cls, item: CogniteResource | None = None) -> list[Pro
...
+    @classmethod
+    def _load_list(
+        cls, resource: dict[str, Any] | list[dict[str, Any]], cognite_client: CogniteClient | None = None
+    ) -> list[SimulatorModelRevisionExternalDependency]:
+        if isinstance(resource, dict):
+            return [cls._load(resource, cognite_client)]
+        elif isinstance(resource, list):
+            return [cls._load(res, cognite_client) for res in resource if isinstance(res, dict)]
+        else:
+            class_name = cls.__class__.__name__
```
erlendvollset, line 409: When `cls` is already a class, `cls.__class__` will just be the class `type`.
```suggestion
            class_name = cls.__name__
```

cognite/client/data_classes/simulators/models.py
```diff
@@ -323,3 +354,57 @@ def _get_update_properties(cls, item: CogniteResource | None = None) -> list[Pro
...
+        return output
+
+    @classmethod
+    def _load_list(
+        cls, resource: dict[str, Any] | list[dict[str, Any]], cognite_client: CogniteClient | None = None
+    ) -> list[SimulatorModelRevisionExternalDependency]:
+        if isinstance(resource, dict):
+            return [cls._load(resource, cognite_client)]
+        elif isinstance(resource, list):
+            return [cls._load(res, cognite_client) for res in resource if isinstance(res, dict)]
```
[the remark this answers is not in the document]
lpereiracgn: Made it simpler for now, adjusted the typing and added the unit test.

tests/tests_integration/test_api/test_simulators/conftest.py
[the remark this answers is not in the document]
lpereiracgn: https://cognitedata.atlassian.net/browse/POFSP-1149

cognite/client/data_classes/simulators/models.py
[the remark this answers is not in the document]
erlendvollset: But how will it be extended? Won't these always just be different ID types? Extending this by just adding an `instance_id` field to `SimulatorModelExternalDependencyFile` is not a good way of doing it, since you won't be able to have the type system enforce which combinations are valid.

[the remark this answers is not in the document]
polomani: @erlendvollset so you prefer a breaking change & type safety over the simplicity? we do have API validation, so this is up to the SDK maintainers how this should be done here.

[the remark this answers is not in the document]
lpereiracgn: I'm getting this error from `test_base.TestCogniteObject `
```
loaded = instance.load(dumped, cognite_client=cognite_mock_client_placeholder)
loaded_dump = loaded.dump()
    
self._for_all_nested_dicts(loaded_dump, _remove_unknown_key)    
assert loaded_dump == instance.dump()
```

`.load` won't match `.dump` if we go with your approach, the API expects and returns `{"id": xxxx}` so we have to play with it in those methods to match the expected format. Do you know a workaround for this case? This test just goes through all the TestCogniteObjects.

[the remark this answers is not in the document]
erlendvollset: With that approach, it doesn't make sense to add more properties to FileReferenceInternalId, so you can type it as an `int` in the Python SDK, which we do for all other internal id references (even though we pass it as `{"id": 123}` to the API).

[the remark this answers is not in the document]
erlendvollset: @lpereiracgn That just needs to be addressed in the dump and load implementations?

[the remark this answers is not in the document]
lpereiracgn: It is already addressed and it works in our simulator tests.
https://github.com/cognitedata/cognite-sdk-python/pull/2218/commits/96a7dd7adfba65cc598f1bf02cf08d7cd5ba291a

The issue is that TestCogniteObjects it's expecting exactly the same thing for both and that's not possible ( load and dump ), we need to match what our API expects. The output of `.dump` will be different compared to `.load`

cognite/client/data_classes/simulators/models.py
```diff
@@ -323,3 +352,34 @@ def _get_update_properties(cls, item: CogniteResource | None = None) -> list[Pro
...
+        arguments (dict[str, str]): A dictionary that contains the key-value pairs (fields) for the external dependency.
+    """
+
+    file: int
+    arguments: dict[str, str]
+
+    @classmethod
+    def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
+        return cls(
+            file=resource["file"],
```
erlendvollset, line 372:
```suggestion
            file=resource["file"]["id"],
```

cognite/client/data_classes/simulators/models.py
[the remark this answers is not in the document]
erlendvollset: https://github.com/cognitedata/cognite-sdk-python/pull/2218#discussion_r2278572618

tests/tests_unit/test_data_classes/test_simulators.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+    def test_load_list(self, input_data):
+        result = SimulatorModelRevisionExternalDependency._load_list(input_data)
+        assert isinstance(result, list)
+        assert all(isinstance(item, SimulatorModelRevisionExternalDependency) for item in result)
+        assert len(result) == 2
+        assert result[0].file["id"] == 1111
+        assert result[0].arguments == {"fieldA": "valueA"}
+        assert result[1].file["id"] == 2222
+        assert result[1].arguments == {"fieldB": "valueB"}
```
polomani, line 26:
```suggestion
        assert isinstance(result[0].file, SimulatorExternalDependencyFileInternalId)
        assert result[0].file.id == 1111
        assert result[0].arguments == {"fieldA": "valueA"}
        assert isinstance(result[0].file, SimulatorExternalDependencyFileInternalId)
        assert result[1].file["id"] == 2222
        assert result[1].arguments == {"fieldB": "valueB"}
```

cognite/client/data_classes/simulators/models.py
```diff
@@ -354,6 +354,21 @@ def _get_update_properties(cls, item: CogniteResource | None = None) -> list[Pro
         ]
 
 
+@dataclass
+class SimulatorExternalDependencyFileReference(CogniteObject): ...
+
+
+@dataclass
+class SimulatorExternalDependencyFileInternalId(SimulatorExternalDependencyFileReference):
```
erlendvollset, line 362: Maybe just FileInternalId?
```suggestion
class FileInternalId(SimulatorExternalDependencyFileReference):
```

erlendvollset (approved):
🦄 Consider the suggested rename, otherwise looks great!

PR #2224, opened 2025-07-31
Make `CogniteObject` abstract

audunska:
It should not be instantiated itself, so it ought to be an abstract base class.


erlendvollset (approved):
LGTM

PR #2225, opened 2025-07-31
Only emit number of failed/successfull/unknown items in exception __str__ reprs

erlendvollset:
Previously we would emit the external id of the item, falling back to logging the whole request body if the external id doesn't exist. We often don't want to log the entire request body, as it may for example emit secrets, like we currently do for UserWrite. I think it may be better to just emit the number of each status for these exceptions, and let the user access the full objects by access the appropriate attributes

## Description

## Checklist:
- [ ] Tests added/updated.


ebrattli (approved):
lgtm!

PR #2235, opened 2025-08-07
docs: updated example to include sources for instances.list and one with to_pandas

andeplane:
## Description
The current list of examples for `client.data_modeling.instances.list` contains no examples where the user passes in `sources` to get properties from a view. In this PR I have added two examples:
1) one with sources being a list of one `ViewId`
2) one where we do `to_pandas` with `expand_properties=True` and `camel_case=True` since it is quite useful

## Checklist:


doctrino (dismissed):

andreavs (approved):
:unicorn:

PR #2244, opened 2025-08-12
Update files.py docstring

ivansurif:
## Description
Fix error in docstring, which referenced client.assets when it should read client.files

## Checklist:
- [ N/A ] Tests added/updated.
- [ ✓ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
- [ N/A ] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ N/A] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

### DIFF
--- a/cognite/client/_api/files.py
+++ b/cognite/client/_api/files.py
@@ -437,7 +437,7 @@ class FilesAPI(APIClient):
             Search for an asset with an attached label:
 
                 >>> my_label_filter = LabelFilter(contains_all=["WELL LOG"])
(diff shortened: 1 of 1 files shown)


haakonvt (changes requested):
Original looks correct to me. I would consider removing the example entirely or moving it to the similarly named asset method.

PR #2250, opened 2025-08-14
Make CogniteAPIError.response_code non-nullable again

erlendvollset:
This attribute was changed to nullable in a recent commit due to the fact that we are actually passing response_code=None in one context, so the type hint was inaccurate. This broke static type checking for some clients. A better solution is to not abuse the CogniteAPIError in this context and raise CogniteMultiException instead

## Description

## Checklist:
- [ ] Tests added/updated.


psalaberria002 (dismissed):
:unicorn:

amorken (approved):

PR #2262, opened 2025-08-21
test(integration): fix simulator object seed (attempt 2)

polomani:
the remote object doesn't have this field defined (on the project used by gh CD action), but it does have it in CI
This leads to an error:
> TestSimulatorRoutines::test_sort - KeyError: 'modelDependencies'


polomani (commented):
🦄

PR #2279, opened 2025-08-29
chore: [CDF-25629] Organization API

doctrino:
## Description

This is a preparation for introducing the principals API. This PR adds a `OrgAPI` that is intended to be subclassed by the REST endpoints that use the organization. 

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/tests/tests_integration/conftest.py
+++ b/tests/tests_integration/conftest.py
@@ -106,7 +106,7 @@ def make_cognite_client(beta: bool = False) -> CogniteClient:
 
 
 @pytest.fixture(scope="session")
-def cognite_client_cdf_authenticated() -> CogniteClient:
+def cognite_client_cog_idp() -> CogniteClient:
     """Some endpoints require a CDF authenticated client, for example, the principal endpoints:
     https://api-docs.cognite.com/20230101/tag/Principals#section/Authentication-for-this-API
 
--- a/tests/tests_integration/test_api/test_organization.py
+++ b/tests/tests_integration/test_api/test_organization.py
@@ -5,8 +5,8 @@ from cognite.client._api.organization import OrgAPI
 
 
 @pytest.fixture(scope="session")
-def org_api(cognite_client_cdf_authenticated: CogniteClient) -> OrgAPI:
-    client = cognite_client_cdf_authenticated
+def org_api(cognite_client_cog_idp: CogniteClient) -> OrgAPI:
+    client = cognite_client_cog_idp
     return OrgAPI(client.config, client._API_VERSION, client)
 
 


cognite/client/config.py
```diff
@@ -135,6 +135,7 @@ def __init__(
         self.credentials = credentials
         self.api_subversion = api_subversion or __api_subversion__
         self.base_url = (base_url or "https://api.cognitedata.com").rstrip("/")
+        self.auth_url = "https://auth.cognite.com"
```
haakonvt, line 138: I think this would be better suited as a class variable on `OrgAPI`. Unless there is an actual need to override this?

cognite/client/_api/organization.py
```diff
@@ -0,0 +1,38 @@
...
+    @cached_property
+    def _organization(self) -> str:
+        headers = self._configure_headers(
+            "application/json",
+            additional_headers=self._config.headers.copy(),
+            api_subversion=self._api_subversion,
+        )
+        # This is an internal endpoint, not part of the public API
+        full_url = urljoin(self._config.base_url, f"/api/v1/projects/{self._config.project}")
+        response = self._http_client.request(method="GET", url=full_url, headers=headers)
```
haakonvt, line 30: This should use the http client with retry

cognite/client/_api/organization.py
```diff
@@ -0,0 +1,38 @@
...
+        if response.status_code != 200:
+            raise CogniteAPIError(
+                "Could not look-up organization", response.status_code, response.headers.get("x-request-id")
+            )
+        try:
+            return response.json()["organization"]
+        except KeyError as e:
+            raise RuntimeError(f"Could not find 'organization' in response: {response.text}") from e
```
haakonvt, line 38: I vote we drop this check as I believe it isn't the job of the SDK to handle a 200 with a broken payload.
```suggestion
        return response.json()["organization"]
```

tests/tests_integration/conftest.py
```diff
@@ -103,3 +103,23 @@ def make_cognite_client(beta: bool = False) -> CogniteClient:
             **beta_configuration,
         )
     )
+
+
+@pytest.fixture(scope="session")
+def cognite_client_cdf_authenticated() -> CogniteClient:
```
haakonvt, line 109: `cognite_client`, defined in the fixture above is already an authenticated `CogniteClient` - what is the specific need for a different/extra one? 

If we need client credentials in particular (if yes, why?), I suggest you rename this fixture and explain why in a comment.

tests/tests_integration/test_api/test_organization.py
```diff
@@ -0,0 +1,23 @@
...
+    return OrgAPI(client.config, client._API_VERSION, client)
+
+
+class TestOrganizationAPI:
+    def test_get_organization(self, org_api: OrgAPI) -> None:
+        organization = org_api._organization
+        assert isinstance(organization, str)
+        assert organization != ""
+        if org_api._config.project in {"python-sdk-contributor", "python-sdk-test", "python-sdk-test-prod"}:
+            assert organization == "cog-python-sdk"
```
haakonvt, line 19: The else branch here should be marked with pytest skip.

tests/tests_unit/test_testing.py
```diff
@@ -10,7 +11,8 @@
 def test_ensure_all_apis_are_available_on_cognite_mock():
     mocked_apis = all_mock_children(CogniteClientMock())
     available = {v.__class__ for v in mocked_apis.values()}
-    expected = set(all_subclasses(APIClient))
+    # OrgAPI is a base API and should not mocked directly.
+    expected = set(all_subclasses(APIClient)) - {OrgAPI}
```
haakonvt, line 15: Consider adding an argument like `except` to the `all_subclasses` function

haakonvt (changes requested):

PR #2284, opened 2025-08-29
chore: stop warnings from being raised from our custom linter

haakonvt:
Description:
Stop warnings from being raised from our custom linter + fix a few type annotations.

### DIFF
--- a/scripts/custom_checks/docstr_examples.py
+++ b/scripts/custom_checks/docstr_examples.py
@@ -80,5 +80,5 @@ def fix_single_file(path: Path) -> str | None:
         return "\n".join(was_fixed)
 
 
-def format_docstring_examples(files: list[Path]) -> str:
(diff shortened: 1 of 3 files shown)


scripts/custom_checks/docstr_examples.py
```diff
@@ -80,5 +80,5 @@ def fix_single_file(path: Path) -> str | None:
         return "\n".join(was_fixed)
 
 
-def format_docstring_examples(files: list[Path]) -> str:
+def format_docstring_examples(files: tuple[Path, ...]) -> str:
```
audunska, line 83: Why is it a tuple? I'd think `Sequence[Path]` is more appropriate.

erlendvollset: ➕

haakonvt:
🦄

PR #2285, opened 2025-08-29
chore: Set max-parallell=4 for integration test workflow matrix

erlendvollset:
## Description

## Checklist:
- [ ] Tests added/updated.


.github/workflows/build.yml
```diff
@@ -52,6 +52,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         python-version: ["3.10", "3.11", "3.12", "3.13"]
+      max-parallel: 4
```
doctrino, line 55: Is this to reduce the chance of flakiness?

PR #2292, opened 2025-09-01
chore: bump pyodide version used in verify-jupyter.yml

haakonvt:
Updating our CI check that tries to avoid breaking Fusion Notebooks for new SDK releases to match the version that is in use.


erlendvollset (approved):
🦄

PR #2311, opened 2025-09-02
feat!: Make all types reflect reality and run mypy everywhere

erlendvollset:
`v8` overview: https://cognitedata.atlassian.net/browse/DM-3055
----

- removes templates
- fixes nullable vs required attributes on all types
- run mypy everywhere
- some special care to diagrams/vision/contextualization APIs since the code there had a high spaghetti factor


haakonvt:
Merged manually into #2273 due to github currently being set to squash-only

PR #2313, opened 2025-09-03
chore(functions): Actually make FunctionCall:wait() backoff exponential

erlendvollset:
With base 1, it's just random but constant at [0,1)

## Description

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/functions.py
+++ b/cognite/client/data_classes/functions.py
@@ -686,7 +686,7 @@ class FunctionCall(CogniteResource):
         return call_id, function_id
 
     def wait(self) -> None:
-        backoff = Backoff(max_wait=5, base=1)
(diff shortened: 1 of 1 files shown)


haakonvt (approved):

audunska (approved):
That's kinda funny 😁

audunska:
> Nice, this looks much better 👌
> 
> Avg. wait evolution of 1000 runs of n=15 backoff runs:
> 
> image

Off topic for this PR:
Seeing this I almost wonder if what we want is a [logistic function](https://en.wikipedia.org/wiki/Logistic_function) instead of an exponential, to get that nice smooth transition to constant :smile: 

Can't think of any pros or cons, it's just my physics brain that wants a nice smooth function.

PR #2318, opened 2025-09-05
chore: make ruff also format code examples in docstrings

haakonvt:
Note: We don't merge before v8 is in

## Description
This is nice to have because we can avoid the very tedious horizontal scrolling in the SDK docs.

### DIFF
--- a/cognite/client/_api/agents/agents.py
+++ b/cognite/client/_api/agents/agents.py
@@ -49,10 +49,9 @@ class AgentsAPI(APIClient):
                 ...     AgentUpsert,
                 ...     QueryKnowledgeGraphAgentToolUpsert,
                 ...     QueryKnowledgeGraphAgentToolConfiguration,
-                ...     DataModelInfo
+                ...     DataModelInfo,
                 ... )
                 >>> client = CogniteClient()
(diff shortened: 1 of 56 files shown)


pyproject.toml
```diff
@@ -18,6 +18,11 @@ packages = [{ include="cognite", from="." }]
 # let scripts use print statements
 "scripts/*" = ["T201"]
 
+[tool.ruff.format]
+# format code examples in docstrings to avoid users scrolling horizontally
+docstring-code-format = true
+docstring-code-line-length = 85
```
haakonvt, line 24: This is the PR, the rest is auto-fmt-work

sighol: Doesn't ruff already parse the `pyproject.toml` file already? I tried to check out the first commit in the PR and run `pre-commit --files SOME_PYTHON_FILE.PY`, and it did format the docstrings. If I do the same on the master branch, the docstrings are untouched.

sighol: If I checkout `f1c44291` and then run `pre-commit run --all`, I end up with the same state as `b57a65f1` :heavy_check_mark:

PR #2324, opened 2025-09-08
chore: include version configuration for agent TOOL

ronpal:
## Description

Adding version to the `QueryKnowledgeGraphAgentToolConfiguration` to be on par with the API. Chose not to go with enum since string is more future-proof.

Screenshot 2025-09-08 at 12 13 05

## Checklist:
- [] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/agents/agent_tools.py
+++ b/cognite/client/data_classes/agents/agent_tools.py
@@ -166,10 +166,12 @@ class QueryKnowledgeGraphAgentToolConfiguration(WriteableCogniteResource):
     Args:
         data_models (Sequence[DataModelInfo] | None): The data models and views to query.
         instance_spaces (InstanceSpaces | None): The instance spaces to query.
+        version (str | None): The version of the query generation strategy to use. A higher number does not necessarily mean a better query.
     """
 
     data_models: Sequence[DataModelInfo] | None = None
     instance_spaces: InstanceSpaces | None = None
+    version: str | None = None
 
     @classmethod
     def _load(
@@ -186,6 +188,7 @@ class QueryKnowledgeGraphAgentToolConfiguration(WriteableCogniteResource):
         return cls(
             data_models=data_models,
             instance_spaces=instance_spaces,
+            version=resource.get("version"),
         )
 
     def dump(self, camel_case: bool = True) -> dict[str, Any]:
@@ -196,6 +199,8 @@ class QueryKnowledgeGraphAgentToolConfiguration(WriteableCogniteResource):
         if self.instance_spaces:
             key = "instanceSpaces" if camel_case else "instance_spaces"
             result[key] = self.instance_spaces.dump(camel_case=camel_case)
+        if self.version:
+            result["version"] = self.version
         return result
 
     def as_write(self) -> QueryKnowledgeGraphAgentToolConfiguration:


cognite/client/data_classes/agents/agent_tools.py
```diff
@@ -166,10 +166,12 @@ class QueryKnowledgeGraphAgentToolConfiguration(WriteableCogniteResource):
     Args:
         data_models (Sequence[DataModelInfo] | None): The data models and views to query.
         instance_spaces (InstanceSpaces | None): The instance spaces to query.
+        version (str | None): The version of the query generation strategy to use. A higher number does not necessarily mean a better query.
```
andeplane, line 169: Would maybe like that we have string literals `v1` or `v2` here?

ronpal: @andeplane I considered that, but it gives a tight coupling that needs to be updated again for example if there comes a "v3". On the other hand, string literals is better guidance to the user. How frequently do we expect new versions?

andeplane (approved):
LGTM, nice

cognite/client/data_classes/agents/agent_tools.py
```diff
@@ -166,10 +166,12 @@ class QueryKnowledgeGraphAgentToolConfiguration(WriteableCogniteResource):
     Args:
         data_models (Sequence[DataModelInfo] | None): The data models and views to query.
         instance_spaces (InstanceSpaces | None): The instance spaces to query.
+        version (str | None): The version of the query generation strategy to use. A higher number does not necessarily mean a better query. Supported values are "v1" and "v2".
     """
 
     data_models: Sequence[DataModelInfo] | None = None
     instance_spaces: InstanceSpaces | None = None
+    version: str | None = None
```
doctrino, line 174: Any reason not to use literal to give the user a better hint?
```suggestion
    version: Literal["v1", "v2"]| None = None
```

doctrino (dismissed):

cognite/client/data_classes/agents/agent_tools.py
```diff
@@ -166,10 +166,12 @@ class QueryKnowledgeGraphAgentToolConfiguration(WriteableCogniteResource):
     Args:
         data_models (Sequence[DataModelInfo] | None): The data models and views to query.
         instance_spaces (InstanceSpaces | None): The instance spaces to query.
+        version (str | None): The version of the query generation strategy to use. A higher number does not necessarily mean a better query. Supported values are "v1" and "v2".
     """
 
     data_models: Sequence[DataModelInfo] | None = None
     instance_spaces: InstanceSpaces | None = None
+    version: Literal["v1", "v2"] | str | None = None
```
erlendvollset, line 174: Shouldn't have default values for read classes, it just makes their instantiation more error prone.
```suggestion
    data_models: Sequence[DataModelInfo] | None
    instance_spaces: InstanceSpaces | None
    version: Literal["v1", "v2"] | str | None
```

haakonvt:
@ronpal sorry for not getting back to you sooner, please ping me on slack for follow-up on risk review if it takes more than 1 hour 😜 

Re. @erlendvollset comment above:

> Shouldn't have default values for read classes, it just makes their instantiation more error prone.

When I take a look at the class:
`class QueryKnowledgeGraphAgentToolConfiguration(WriteableCogniteResource):`
...this looks like a write-version-class, no?

Also, from your screenshot, it seems like `data_models` is required, but `instance_spaces` and `version` are not.

If all of this is true, I think the correct solution is to add back default arg `= None` for the latter and require only the first.

haakonvt:
Risk review: 

My comment from 2 weeks ago, specifically this part:

> Also, from your screenshot, it seems like data_models is required, but instance_spaces and version are not.
> 
> If all of this is true, I think the correct solution is to add back default arg = None for the latter and require only the first.

needs addressing @ronpal 😄

ks93:
> @ks93 do not remove `risk-review ongoing`

Oops, sorry

haakonvt (approved):
🦄

PR #2325, opened 2025-09-08
Udpate toolkit modules

doctrino:
## Description

Currently the tests are failing for updating the contributor read/write groups. This is because there is a mismatch of the Toolkit modules and the CI/CD version of Toolkit in use. This update the toolkit modules.

## Checklist:
- [ ] Tests added/updated.


haakonvt:
🦄

PR #2328, opened 2025-09-08
chore: Only delete function, not schedule, in test cleanup

erlendvollset:
If the schedule delete fails, the function won't be cleaned up. The delete of the function cascades to its schedules.


audunska (commented):
Curious as to why

erlendvollset:
> Curious as to why

Updated PR description.

erlendvollset:
🦄

PR #2330, opened 2025-09-10
fix: make filters throw warning when used in a boolean context

haakonvt:
https://cognitedata.atlassian.net/browse/DM-3081

## Description
Fix: Make filters throw a warning when used in a boolean context to hinder users attempting to combine filters using and/or which do not work.

```py
>>> flt1 and flt2
warnings.warn(…)
flt1
>>> flt1 & flt2
✅
```

## Checklist:
- [x] Tests added/updated.


erlendvollset (approved):
🦄

PR #2331, opened 2025-09-10
fix(documents): stop paginating at limit

tommythorsen:
## Description
The documents search function does not appear to respect the limit parameter passed to it. While it does correctly pass it to the API, it also keeps paginating on the returned cursor until the API says stop, so it tends to return 1000 items regardless of the value of the limit parameter.

This commit changes the pagination loop to stop when we have received `limit` or more result items.

See slack thread:
https://cognitedata.slack.com/archives/CKY04V4CA/p1757496077990499

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/cognite/client/_api/documents.py
+++ b/cognite/client/_api/documents.py
@@ -616,6 +616,10 @@ class DocumentsAPI(APIClient):
             response = self._post(f"{self._RESOURCE_PATH}/search", json=body)
             json_content = response.json()
             results.extend(json_content["items"])
+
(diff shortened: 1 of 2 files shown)


haakonvt (approved):
LGTM

haakonvt:
🦄

haakonvt (changes requested):
Sorry, _just one question_ - do we need pagination here? Seems we can just remove it

PR #2214, opened 2025-07-26
[CDF-25309] Introduce service principals API

doctrino:
## Description

Stacked on #2212 #2213 

Adds the new Service Principal API.

## Checklist:
- [x] Tests added/updated.
- [x] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [x] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [x] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #2332, opened 2025-09-10
feat: [CDF-25309] Add service principal API

doctrino:
## Description

This is a move of the https://github.com/cognitedata/cognite-sdk-python/pull/2063 and https://github.com/cognitedata/cognite-sdk-python/pull/2214 PRs instead of rebasing them

This adds support for the following endpoints
* https://auth.cognite.com/api/v1/principals/me
* https://auth.cognite.com/api/v1/orgs/{org}/principals
* https://auth.cognite.com/api/v1/orgs/{org}/principals/byids

## Checklist:
- [x] Tests added/updated.


cognite/client/data_classes/_base.py
```diff
@@ -268,8 +268,11 @@ def _build_id_mappings(self) -> None:
         self._id_to_item, self._external_id_to_item, self._instance_id_to_item = {}, {}, {}
         if not self.data:
             return
-        if hasattr(self.data[0], "external_id"):
-            self._external_id_to_item = {item.external_id: item for item in self.data if item.external_id is not None}
+        self._external_id_to_item = {
+            item.external_id: item
+            for item in self.data
+            if hasattr(item, "external_id") and item.external_id is not None
+        }
```
doctrino, line 275: The principal breaks this assumption. There are service principal and user principals, were a service principal has an external ID while an internal does not.

cognite/client/_api/organization.py
```diff
@@ -35,3 +42,145 @@ def _organization(self) -> str:
...
+        ignore_unknown_ids: bool = False,
+    ) -> PrincipalList: ...
+
+    @overload
+    def retrieve(
+        self,
+        id: None = None,
+        external_id: None = None,
+        ignore_unknown_ids: bool = False,
+    ) -> PrincipalList: ...
```
erlendvollset, line 115: Is there any point in having this overload if we don't allow id and external_id to be provided at the same time?
```suggestion
```

erlendvollset (approved):
🦄

PR #2334, opened 2025-09-15
fix: filters now accept DirectRelationReference as instance identifier

olacognite:
## Description
When creating an edge, you need to use a DirectRelationReference for the type of the edge.
When you filter edges however, you can use InstanceId for the type, but DirectRelationReference is not covered by the type hint. Adding support for this, as I have seen that you can use them interchangeably in the filter. 

The annotation writer worker in the context-api is currently keeping two constants for the diagrams.FileLink, one direct relation reference and one node id. Being able to just have one is the motivation for this change. 

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/tests/tests_integration/test_api/test_data_modeling/test_instances.py
+++ b/tests/tests_integration/test_api/test_data_modeling/test_instances.py
@@ -509,7 +509,7 @@ class TestInstancesAPI:
         edge_type_class: type[NodeId, DirectRelationReference],
     ) -> None:
         edge_type = edge_type_class(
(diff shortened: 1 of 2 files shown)


tests/tests_integration/test_api/test_data_modeling/test_instances.py
```diff
@@ -139,6 +145,47 @@ def node_with_1_1_pressure_in_bar(
     return node
 
 
+@pytest.fixture
```
haakonvt, line 148: Let's make these fixtures session scoped

haakonvt (approved):
🦄

PR #2336, opened 2025-09-17
feat: data classes for debug notices

haakonvt:
https://cognitedata.atlassian.net/browse/DM-3058

## Description
Part one of adding debug notice support to the SDK: All necessary data classes.

Risk review note: Subclasses of CogniteResource are automatically picked up and tested. Don't believe me, run `pytest tests/tests_unit/test_base.py` after introducing a random error somewhere 😉

### DIFF
--- a/cognite/client/data_classes/data_modeling/debug.py
+++ b/cognite/client/data_classes/data_modeling/debug.py
@@ -1,0 +1,475 @@
+from __future__ import annotations
+
+from abc import ABC
+from dataclasses import dataclass
+from functools import cache
+from typing import TYPE_CHECKING, Any, Literal, NoReturn, cast, get_args, get_type_hints
+
+from typing_extensions import Self
+
+from cognite.client.data_classes._base import CogniteResource, CogniteResourceList
+from cognite.client.data_classes.data_modeling.ids import ContainerId
+from cognite.client.data_classes.data_modeling.instances import InstanceSort
+from cognite.client.utils._auxiliary import all_concrete_subclasses
+
+if TYPE_CHECKING:
+    from cognite.client import CogniteClient
+
+
+@dataclass(kw_only=True)
+class DebugParameters:
+    """
+    Debug parameters for debugging and analyzing queries.
+
+    Args:
+        emit_results (bool): Include the query result in the response. emit_results=False is required for advanced query analysis features.
+        timeout (int | None): Query timeout in milliseconds. Can be used to override the default timeout when analysing queries. Requires emit_results=False.
+        profile (bool): Most thorough level of query analysis. Requires emit_results=False.
+    """
+
+    emit_results: bool = True
+    timeout: int | None = None
+    profile: bool = False
+
+    def dump(self, camel_case: bool = True) -> dict[str, bool | int]:
+        res: dict[str, bool | int] = {
+            "emitResults" if camel_case else "emit_results": self.emit_results,
+            "profile": self.profile,
+        }
+        if self.timeout is not None:
+            res["timeout"] = self.timeout
+        return res
+
+
+@dataclass
+class DebugNotice(CogniteResource, ABC):
+    """
+    A notice that provides insight into the query's execution. It can highlight potential performance issues,
+    offer an optimization suggestion, or explain an aspect of the query processing. Each notice falls into a
+    category, such as indexing, sorting, filtering, or cursoring, to help identify areas for improvement.
+    """
+
+    category: str
+    code: str
+    hint: str
+    level: str
+
+    @classmethod
+    def _load(cls, data: dict[str, Any], cognite_client: CogniteClient | None = None) -> DebugNotice:
+        subclass_lookup = _create_debug_notice_subclass_map()
+        try:
+            key = (data["code"], data["category"])
+            return subclass_lookup[key]._load(data)
+        except KeyError:
+            return cast(DebugNotice, UnknownDebugNotice._load(data))
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        return {
+            "category": self.category,
+            "code": self.code,
+            "hint": self.hint,
+            "level": self.level,
+        }
+
+
+@dataclass
+class InvalidDebugOptionsNotice(DebugNotice, ABC):
+    pass
+
+
+@dataclass
+class ExcessiveTimeoutNotice(InvalidDebugOptionsNotice):
+    code: Literal["excessiveTimeout"]
+    category: Literal["invalidDebugOptions"]
+    level: Literal["warning"]
+    hint: Literal["Cannot specify too large timeout"]
+    timeout: int
+
+    @classmethod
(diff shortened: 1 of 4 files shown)


doctrino (approved):

cognite/client/data_classes/data_modeling/debug.py
```diff
@@ -0,0 +1,475 @@
...
+
+    Args:
+        emit_results (bool): Include the query result in the response. emit_results=False is required for advanced query analysis features.
+        timeout (int | None): Query timeout in milliseconds. Can be used to override the default timeout when analysing queries. Requires emit_results=False.
+        profile (bool): Most thorough level of query analysis. Requires emit_results=False.
+    """
+
+    emit_results: bool = True
+    timeout: int | None = None
+    profile: bool = False
```
einarmo, line 32: The api docs contains a few more parameters here, `includeTranslatedQuery` and `includePlan`. Are these excluded on purpose?

cognite/client/data_classes/data_modeling/debug.py
```diff
@@ -0,0 +1,475 @@
...
+        }
+
+
+@dataclass
+class InvalidDebugOptionsNotice(DebugNotice, ABC):
+    pass
+
+
+@dataclass
+class ExcessiveTimeoutNotice(InvalidDebugOptionsNotice):
```
einarmo, line 81: Should these have docstrings?

cognite/client/data_classes/data_modeling/debug.py
```diff
@@ -0,0 +1,475 @@
...
+
+@dataclass
+class IntractableDirectRelationsCursorNotice(CursoringNotice):
+    code: Literal["intractableDirectRelationsCursor"]
+    category: Literal["cursoring"]
+    level: Literal["warning"]
+    grade: Literal["D"]
+    hint: Literal[
+        "Cursoring combined with chaining to a list of direct relations requires a run-time sort and is therefore inherently intractable."
+    ]
```
einarmo, line 134: This is perhaps more a comment on the API spec, but including help strings in the API contract is _really_ weird. This does mean that if we ever wanted to change these it would require a year of deprecation for the old value.

cognite/client/data_classes/data_modeling/debug.py
```diff
@@ -0,0 +1,475 @@
...
+        )
+        return obj
+
+
+@dataclass
+class UnknownDebugNotice(DebugNotice):
+    category: str | None  # type: ignore [assignment]
+    code: str | None  # type: ignore [assignment]
+    hint: str | None  # type: ignore [assignment]
+    level: str | None  # type: ignore [assignment]
```
einarmo, line 422: If these are nullable, they should be nullable in the superclass. `type: ignore` is fine if we're working around a limitation in the type checker, or some incorrect types in something external, but if these are indeed nullable, then a user using an instance of a `DebugNotice` could run into issues when the typing tells them to assume that they aren't.

einarmo (commented):
I understand that this is split up due to line limits, but you will likely run into problems reproducing these debug notices from the real API anyway. You might as well write a test or two verifying that we can deserialize a list of debug notices correctly.

cognite/client/data_classes/data_modeling/debug.py
erlendvollset: Agree. This should just be a string (also in the spec).

erlendvollset: https://github.com/cognitedata/service-contracts/pull/2985

cognite/client/data_classes/data_modeling/debug.py
haakonvt: They are in alpha. Do you think we should add them right away? 🤔

cognite/client/data_classes/data_modeling/debug.py
haakonvt: I don't think so

cognite/client/data_classes/data_modeling/debug.py
haakonvt: Will update this to reflect the changes 👍

cognite/client/data_classes/data_modeling/debug.py
haakonvt: Please take a look at the PR this was split from (see the Risk Review note section in the PR description).

This is the "unknown" debug notice which is used to future-proof the SDK. It is the fallback in case new are added we don't know about yet. I originally had it separate from the `DebugNotice` class hierarchy but reviewers wanted it part of the tree + "optimistic" extraction of expected parameters which is why things are the way they are today.

einarmo:
> We do have automatic tests for serialization and deserialization.

I was not aware of this, where do I find these? If we have common tests that automatically deal with this, then we _do_ have tests.

cognite/client/data_classes/data_modeling/debug.py
einarmo: If you're aware, and it's intentional, I have no objections.

cognite/client/data_classes/data_modeling/debug.py
einarmo: I understand that, but you are lying to users in types, I really don't agree with this. IMO you should either:

 - Fail if this happens.
 - Set some default value in the fallback.
 - Change the type in the superclass (you can still narrow in each of the different cases).

There is, IMO, _never_ any use for `type: ignore` in production code except to work around limitations in external libraries.

einarmo: I still really disagree. In a language with real types, this is _impossible_ to express.

You do have another option: You can remove the base class entirely, or move this outside of the type hierarchy. Why does this need to inherit from `DebugNotice` at all? Some kind of union magic?

dimitryhoff: There is no need to change the type to `str | None` a string can be empty and still be a string.

My idea was to do as little special treatment as possible, however keep the hierarchy, the user can still do type matching or string matching if needed.

dimitryhoff: My opinions are more often colored by Strongly typed languages, Kotlin in this instance. So if someone with stronger Python background have a good explanation why we need to do X I'm eager to learn.

My logic was, as with Enums, you have `Unknown` as an escape hatch.
In code you can treat them all as Type `X` then either implement overloads/functions and use Polymorphism or switch on type (Very common pattern in Kotlin)
My main idea was this usage pattern:
```
match data:
        case UnknownDebugNotice():
            ...
        case SomeOtherDebugNotice()
```
I also though about Type Bounding like
```
T = TypeVar('T', bound=DebugNotice)
```

Pep/typing allows unions for types that makes in more open to use orthogonal classes for error handling

I don't know if something like this is possible and/or considered "good"/"bad"
```
T = TypeVar('T', bound=DebugNotice|UnknownDebugNotice)
```

einarmo: All I really want is a solution that avoids the `type: ignore` beyond that I think any solution is probably fine.

PR #2267, opened 2025-08-23
feat(simulators): Add support for running simulations with routine and model revision external IDs

evertoncolling:
## Description
This PR extends the Simulator Runs API to support two modes for running simulations:
  1. Routine-based mode (existing): Run by specifying only `routine_external_id`
  2. Revision-based mode (new): Run by specifying both `routine_revision_external_id` and `model_revision_external_id`

## Checklist:
- [x] Tests added/updated.
- [x] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [x] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ ] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #2148, opened 2025-04-15
fix(simulators): Consistency for filtering

lpereiracgn:
## Description
Some of our API listing methods currently require an import to perform filtering, while others allow users to pass arguments directly without additional imports. To improve the user experience, we are standardizing on the latter approach across all instances.

## Checklist:
- [x] Tests added/updated.
- [ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [ ] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ ] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #2339, opened 2025-09-19
feat(simulators): Add support for running simulations using revisions

lpereiracgn:
## Description
This PR extends the Simulator Runs API to support two modes for running simulations:
  1. Routine-based mode (existing): Run by specifying only `routine_external_id`
  2. Revision-based mode (new): Run by specifying both `routine_revision_external_id` and `model_revision_external_id`

Co-Authored: @evertoncolling . [Ref](https://github.com/cognitedata/cognite-sdk-python/pull/2267).

## Checklist:
- [x] Tests added/updated.


tests/tests_integration/test_api/test_simulators/test_runs.py
```diff
@@ -129,35 +99,61 @@ async def test_run_with_wait_and_retrieve(
...
+        ids=[
+            "with_routine_revision_and_model_revision",
+            "with_routine_only",
+        ],
+    )
     def test_create_run(
         self,
         cognite_client: CogniteClient,
-        seed_simulator_routine_revisions: tuple[SimulatorRoutineRevision, SimulatorRoutineRevision],
-        seed_resource_names: ResourceNames,
```
polomani, line 118: do not remove fixtures, you need routines to be avaible in order to create a run

tests/tests_integration/test_api/test_simulators/test_runs.py
```diff
@@ -129,35 +99,61 @@ async def test_run_with_wait_and_retrieve(
...
+        [
+            (
+                None,
+                f"{RESOURCES.simulator_routine_external_id}_v1",
+                RESOURCES.simulator_model_revision_external_id,
+            ),
+            (
+                RESOURCES.simulator_routine_external_id,
+                None,
+                None,
```
polomani, line 125: instead of `None, None` it would be cleaner to accept SimulationRunWrite as a parameter

tests/tests_unit/test_data_classes/test_simulators.py
```diff
@@ -464,3 +467,27 @@ def test_get_units_empty_units_list(self) -> None:
...
+        error_message = "Must specify either 'routine_external_id' alone, or both 'routine_revision_external_id' and 'model_revision_external_id' together."
+        with pytest.raises(
+            ValueError,
+            match=re.escape(error_message),
+        ):
+            SimulationRunWrite(
+                routine_external_id="routine_external_id_1",
+                routine_revision_external_id="routine_revision_external_id_1",
+                model_revision_external_id="model_revision_external_id_1",
+                run_type="external",
```
polomani, line 483:
```suggestion
```
nit, not relevant in this particular test

polomani (dismissed):

cognite/client/_api/simulators/routines.py
```diff
@@ -226,22 +226,58 @@ def list(
...
+        run_time: int | None = None,
+        queue: bool | None = None,
+        log_severity: Literal["Debug", "Information", "Warning", "Error"] | None = None,
+        wait: bool = True,
+        timeout: float = 60,
+    ) -> SimulationRun: ...
+
+    def run(
+        self,
+        routine_external_id: str | None = None,
```
doctrino, line 258: I think this should also allow you to take in a full run object.
```suggestion
        routine_external_id: str | SimulationRunWrite | None = None,
```

This will allow you to
```python
client.simulators.run(my_run_obj). 
```

doctrino (commented):
See suggestions, will review the rest when those are adrdessed.

cognite/client/_api/simulators/routines.py
polomani: we wanted to keep this one simple, also, can be added later, anyway?

cognite/client/data_classes/simulators/runs.py
```diff
@@ -87,20 +87,33 @@ class SimulationRunCore(WriteableCogniteResource["SimulationRunWrite"]):
...
+        model_revision_external_id: str | None = None,
     ) -> None:
         self.run_type = run_type
         self.run_time = run_time
         self.routine_external_id = routine_external_id
+        self.routine_revision_external_id = routine_revision_external_id
+        self.model_revision_external_id = model_revision_external_id
 
 
 class SimulationRunWrite(SimulationRunCore):
```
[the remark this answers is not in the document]
lpereiracgn: I tried this approach and IMO the UX doesn't look much cleaner than it is at the moment.

For example: Some time ago we had to do additional imports for filtering https://github.com/cognitedata/cognite-sdk-python/pull/2148/files and we removed them.

In this case, if the user reaches one of the errors based on the conditionals, fixing it would be much faster than looking for the right class to execute a simulation. It was getting confusing just writing the docstrings for this.

cognite/client/data_classes/simulators/runs.py
```diff
@@ -129,19 +159,35 @@ def __init__(
     @classmethod
     def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> SimulationRunWrite:
         inputs = resource.get("inputs", None)
+        routine_revision_external_id = resource.get("routineRevisionExternalId", None)
+        model_revision_external_id = resource.get("modelRevisionExternalId", None)
+        routine_external_id = resource.get("routineExternalId", None)
```
doctrino, line 164: None is default so no need to specify
```suggestion
        routine_revision_external_id = resource.get("routineRevisionExternalId")
        model_revision_external_id = resource.get("modelRevisionExternalId")
        routine_external_id = resource.get("routineExternalId")
```

doctrino (approved):
Just one nitpick

cognite/client/data_classes/simulators/runs.py
lpereiracgn: fixed! :)

cognite/client/data_classes/simulators/runs.py
```diff
@@ -110,15 +123,32 @@ class SimulationRunWrite(SimulationRunCore):
...
+            "Must specify either 'routine_external_id' alone, or both "
+            "'routine_revision_external_id' and 'model_revision_external_id' together."
+        )
+
+        if routine_external_id is not None:
+            if routine_revision_external_id is not None or model_revision_external_id is not None:
+                raise param_error
+        else:
+            if not (routine_revision_external_id is not None and model_revision_external_id is not None):
+                raise param_error
```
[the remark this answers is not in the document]
dbrattli: @gemini-code-assist I feel uncomfortable with the initializer being this complex/fragile, and raise exceptions for invalid input. Would it be better here to use e.g overloads to enable the type checker to at least detect invalid usage, or use a union type approach e.g `reference: RoutineReference | RevisionReference,  # Can't be wrong`. Another possibility is to use distinct factory methods to construct the type (but then python do not have private initializers). What do you think is the best solution here?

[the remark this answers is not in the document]
dbrattli: @lpereiracgn What do you think about these suggestions?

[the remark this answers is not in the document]
polomani: > Using Overloads for __init__: While @overload is fantastic for guiding type checkers and users of public APIs (like the run method you've already implemented), its utility for __init__ in a data class like SimulationRunWrite is primarily for type hinting. The underlying __init__ implementation would still need to handle the union of all possible parameters and perform runtime validation, potentially leading back to similar if/else complexity.

We had a large discussion on this, unfortunately we can't use __init__ overloads due to the bug in pydoclint
https://cognitedata.slack.com/archives/CG10VQPFX/p1756371280279699

We tried other approaches as well https://github.com/cognitedata/cognite-sdk-python/pull/2339#discussion_r2396360420 , but I'll let Luis answer on the other suggestions.

It might be the best to let the backend return the error and remove this validation alltogether, because this discussion is now taking more than a month and isn't really justifying the effort. Would you be fine with that?

cognite/client/data_classes/simulators/runs.py
```diff
@@ -110,15 +123,39 @@ class SimulationRunWrite(SimulationRunCore):
...
+        is_routine_mode = (
+            routine_external_id is not None
+            and routine_revision_external_id is None
+            and model_revision_external_id is None
+        )
+        is_revision_mode = (
+            routine_external_id is None
+            and routine_revision_external_id is not None
+            and model_revision_external_id is not None
+        )
```
dbrattli, line 150:
```suggestion
        is_revision_mode = (
            not routine_external_id
            and routine_revision_external_id
            and model_revision_external_id
        )
```

cognite/client/data_classes/simulators/runs.py
```diff
@@ -110,15 +123,39 @@ class SimulationRunWrite(SimulationRunCore):
...
+        param_error = ValueError(
+            "Must specify either 'routine_external_id' alone, or both "
+            "'routine_revision_external_id' and 'model_revision_external_id' together."
+        )
+
+        is_routine_mode = (
+            routine_external_id is not None
+            and routine_revision_external_id is None
+            and model_revision_external_id is None
+        )
```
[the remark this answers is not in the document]
lpereiracgn: Fixed via https://github.com/cognitedata/cognite-sdk-python/pull/2339/commits/ac996c20b7faf3420a2d1bb6110afa1b29f5c8f0

dbrattli (approved):
:unicorn:

doctrino (approved):

PR #2341, opened 2025-09-23
fix: do not throw UserWarning from SDK code

haakonvt:
(and make sure it doesn't happen again)

## Description
Follow-up from #2330 - apparently we use `if filter` a lot in the SDK codebase (and we don't want to do this of course).

Pytest will now error out if it ever sees this warning again when running tests to ensure we don't repeat this.

Also noted on slack: https://cognitedata.slack.com/archives/CG10VQPFX/p1758651269707159


erlendvollset:
🦄

PR #2346, opened 2025-09-25
fix(simulators): make SimulatorModelRevision.get_data() return single item for consistency

evertoncolling:
## Description
- Changes `SimulatorModelRevision.get_data()` return type from `SimulatorModelRevisionDataList` to `SimulatorModelRevisionData | None`
- Updates integration test to handle the new return type
- Makes the method consistent with `SimulationRun.get_data()` which already returns a single item

## Checklist:
- [x] Tests added/updated.


erlendvollset (approved):
🦄

PR #2348, opened 2025-09-26
fix(datapoints): fix rare bug when instance ID time series would repo…

haakonvt:
…rt as missing in datapoints queries

Found while porting datapoints code from threading to async. Probably something nobody has ever hit, but still worth fixing.

Added integration test that fails on `master`.


erlendvollset (approved):
🦄

doctrino (dismissed):

PR #2349, opened 2025-10-01
fix(agents): replace agentId with agentExternalId in chat API

ks93:
[alpha] Align SDK with API specification by replacing agentId parameter with agentExternalId in the chat() method and AgentChatResponse class.

This is a breaking change for alpha users, but alpha features are subject to breaking changes without prior notice.

## Description

## Checklist:
- [x] Tests added/updated.
- [x] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change. If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.

### DIFF
--- a/cognite/client/_api/agents/agents.py
+++ b/cognite/client/_api/agents/agents.py
@@ -242,7 +242,7 @@ class AgentsAPI(APIClient):
 
     def chat(
         self,
-        agent_id: str,
+        agent_external_id: str,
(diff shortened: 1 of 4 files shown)


erlendvollset (approved):
🦄

PR #2354, opened 2025-10-01
refactor: disallow python 3.9 for functions api

jonasblanc:
## Description
- disable python 3.9 for function
- bump test to test valid runtime ie: "py310", "py311", "py312"
- type runtime as RunTime instead of str

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/functions.py
+++ b/cognite/client/data_classes/functions.py
@@ -196,7 +196,7 @@ class Function(FunctionCore):
         self.id: int = id  # type: ignore
         self.created_time: int = created_time  # type: ignore
         self.status: str = status  # type: ignore
-        self.runtime_version: RunTime | None = runtime_version
+        self.runtime_version: str | None = runtime_version
         self.error = error
         self._cognite_client = cast("CogniteClient", cognite_client)
 


cognite/client/data_classes/functions.py
```diff
@@ -146,7 +146,7 @@ class Function(FunctionCore):
         env_vars (dict[str, str] | None): User specified environment variables on the function ((key, value) pairs).
         cpu (float | None): Number of CPU cores per function. Allowed range and default value are given by the `limits endpoint. <https://developer.cognite.com/api#tag/Functions/operation/functionsLimits>`_, and None translates to the API default. On Azure, only the default value is used.
         memory (float | None): Memory per function measured in GB. Allowed range and default value are given by the `limits endpoint. <https://developer.cognite.com/api#tag/Functions/operation/functionsLimits>`_, and None translates to the API default. On Azure, only the default value is used.
-        runtime (str | None): Runtime of the function. Allowed values are ["py39","py310", "py311", "py312"]. The runtime "py312" resolves to the latest version of the Python 3.12 series.
+        runtime (str | None): Runtime of the function. Allowed values are ["py310", "py311", "py312"]. The runtime "py312" resolves to the latest version of the Python 3.12 series.
```
haakonvt, line 149: Small suggestion 
```suggestion
        runtime (Runtime | None): Runtime of the function. Allowed values are ["py310", "py311", "py312"]. The runtime "py312" resolves to the latest version of the Python 3.12 series.
```

haakonvt (changes requested):
This is not a breaking change in the SDK. This would push a new major version, which we really don’t want 😅

qTipTip (approved):
LGTM! 

What change in this PR would cause a major release, @haakonvt?

haakonvt:
> LGTM!
> 
> What change in this PR would cause a major release, @haakonvt?

The `!` that previously was in the PR title

erlendvollset:
This is a breaking change though 🤔 should wait for v8?

doctrino (approved):

PR #2357, opened 2025-10-03
feat(functions): Relax handle validation

dbrattli:
## Description

Support function handle as variable assignment needed by Cognite Typed Functions

### Overview

Relaxes function handle validation to support more flexible patterns for defining the entry point of Cognite Functions. Previously, only function definitions were accepted - now variable assignments are also supported.

### New Supported Patterns

**Before:** Only function definitions
```python
def handle(data, client, secrets):
    return process_data(data)
```

**After:** Function definitions + variable assignments
```python
# Pattern 1: Simple assignment
def create_handler():
    def inner_function(data, client, secrets):
        return {"result": "processed"}
    return inner_function

handle = create_handler()

# Pattern 2: Annotated assignment  
from collections.abc import Callable

handle: Callable = create_handler()
```

### Use Cases

This enhancement enables several useful development patterns:

- **Factory Functions**: Create handlers dynamically based on configuration
- **Decorator Patterns**: Wrap handlers with middleware or instrumentation
- **Conditional Logic**: Select different handlers based on environment
- **Code Organization**: Separate handler creation logic from the entry point

### Implementation Details

- Extends AST parsing to recognize `ast.Assign` and `ast.AnnAssign` nodes
- Maintains validation that the target must be named "handle"
- Relaxes argument validation for callable assignments (validation happens at runtime)
- Only accepts simple name assignments (e.g., `handle = ...`), not complex expressions
- We do not validate arguments of callables. They are expected to do runtime validation themselves.

### Backwards Compatibility

✅ Fully backwards compatible - all existing function definitions continue to work unchanged

### Testing

Added comprehensive test coverage for:
- Simple variable assignments (`handle = callable`)
- Annotated assignments (`handle: Callable = callable`) 
- Integration with folder-based function deployment
- Validation of assignment targets (must be simple names)

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/tests/tests_unit/test_api/test_functions.py
+++ b/tests/tests_unit/test_api/test_functions.py
@@ -17,6 +17,7 @@ from cognite.client._api.functions import (
     _extract_requirements_from_file,
     _get_fn_docstring_requirements,
     _validate_and_parse_requirements,
+    _validate_function_handle,
     get_handle_function_node,
     validate_function_folder,
 )
@@ -559,6 +560,22 @@ class TestFunctionsAPI:
 
         assert isinstance(res, Function)
         assert mock_functions_create_response.calls[3].response.json()["items"][0] == res.dump(camel_case=True)
+
+    def test_validate_function_handle_with_complex_assignment_target_raises(self):
+        # Create AST node for: obj.handle = some_callable
+        code = "obj.handle = lambda: None"
+        tree = ast.parse(code)
+        assign_node = tree.body[0]
+        with pytest.raises(TypeError, match="Assignment target must be a simple name"):
+            _validate_function_handle(assign_node)
+
+    def test_validate_function_handle_with_complex_annotated_assignment_target_raises(self):
+        # Create AST node for: obj.handle: Callable = some_callable
+        code = "obj.handle: Callable = lambda: None"
+        tree = ast.parse(code)
+        ann_assign_node = tree.body[0]
+        with pytest.raises(TypeError, match="Assignment target must be a simple name"):
+            _validate_function_handle(ann_assign_node)
 
     def test_create_with_handle_function_and_file_id_raises(
         self, mock_functions_create_response, function_handle, cognite_client


cognite/client/_api/functions.py
```diff
@@ -246,7 +246,7 @@ def create(
...
             cpu (float | None): Number of CPU cores per function. Allowed range and default value are given by the `limits endpoint. <https://developer.cognite.com/api#tag/Functions/operation/functionsLimits>`_, and None translates to the API default. On Azure, only the default value is used.
             memory (float | None): Memory per function measured in GB. Allowed range and default value are given by the `limits endpoint. <https://developer.cognite.com/api#tag/Functions/operation/functionsLimits>`_, and None translates to the API default. On Azure, only the default value is used.
-            runtime (RunTime | None): The function runtime. Valid values are ["py39", "py310", "py311", "py312", `None`], and `None` translates to the API default which will change over time. The runtime "py312" resolves to the latest version of the Python 3.12 series.
+            runtime (RunTime | None): The function runtime. Valid values are ["py39", "py310", "py311", `None`], and `None` translates to the API default which will change over time. The runtime "py311" resolves to the latest version of the Python 3.11 series.
```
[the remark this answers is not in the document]
dbrattli: Reverted

cognite/client/_api/functions.py
```diff
@@ -792,14 +793,18 @@ def validate_function_folder(root_path: str, function_path: str, skip_folder_val
...
-    else:
-        name = handle_obj.__name__
-        accepts_args = set(signature(handle_obj).parameters)
+    match handle_obj:
+        case ast.FunctionDef():
+            name = handle_obj.name
+            accepts_args = {arg.arg for arg in handle_obj.args.args}
+        case ast.Name():
+            name = handle_obj.id
+            accepts_args = set()
```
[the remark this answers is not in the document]
dbrattli: I think this is acceptable. If you generate a handle function at runtime you will not get validation of arguments.

erlendvollset (approved):
LGTM

cognite/client/_api/functions.py
```diff
@@ -716,15 +716,16 @@ def status(self) -> FunctionsStatus:
...
+def get_handle_function_node(file_path: Path) -> ast.FunctionDef | ast.Name | None:
+    tree = ast.parse(file_path.read_text())
+    for node in tree.body:
+        if isinstance(node, ast.FunctionDef) and node.name == "handle":
+            return node
+        if isinstance(node, ast.Assign):
+            for target in node.targets:
+                if isinstance(target, ast.Name) and target.id == "handle":
+                    return target
+    return None
```
haakonvt, line 728: I wonder (as a follow-up) if this function should not short-circuit on first found and rather raise if > 1 is found. Wdyt @dbrattli ?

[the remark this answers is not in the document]
HMEiding: I agree that we should not support anything else. However, we don't know what user's might have already done. Should we consider it a breaking change?

cognite/client/_api/functions.py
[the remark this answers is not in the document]
HMEiding: Yes, I agree that giving a warning is a good idea. And in cases where the callable variable `handle` is generated by `cognite-typed-functions`, the validation can be done there (or already is, am I right?)

HMEiding (commented):
Great to get this in 👍

cognite/client/_api/functions.py
dbrattli: I think we should not pretend we can statically validate the Python code being deployed. Anything can be wrong (or right) with the function. Checking that it has at least a `handle` is fine with me, but trying to evaluate here if the code will fly or not is over-engineering imo. I.e the existing logic will reject valid code like a callable `handle` since it's not a `FunctionDef`. Why this strict requirement to begin with? If the user deploys a function with two things called `handle`, then I'm good if it works. E.g if a route is also called `handle`, ref: https://github.com/cognitedata/cognite-typed-functions/blob/main/README.md

dbrattli:
So every time we deploy a new style `Cognite Typed Function` we need to warn the user that we have not validated that the arguments is a subset of the allowed? That will confuse much more than it helps.

cognite/client/_api/functions.py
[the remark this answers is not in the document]
dbrattli: @gemini-code-assist For Functions v2 as this PR is trying to fix, this warning will then happen for every function being deployed. I cannot see how that would be valuable, For the new framework we will produce the handle like this:

from cognite_typed_functions import create_function_handle

# for single app
handle = create_function_handle(app)

# or, for multiple apps (composition)
handle = create_function_handle(mcp_app, introspection_app, main_app)

haakonvt:
@dbrattli & @HMEiding Thanks for the context on "Functions V2" or "Cognite Typed Function" (I guess?).

With that in mind, I have a different suggestion. I agree that if this pattern is going to see hefty future use, it makes no sense to keep warning people over and over.

However, what I don't like is this PR not preserving backwards compatibility. I would suggest we go for a third option: add a "skip_validation" parameter instead and use that by default in your Typed Functions, then leave all other changes as is.

cognite/client/_api/functions.py
haakonvt: `inspect.getsource` could grab the wrong code, that's the point I'm trying to make! Anyway, I agree on the point re. validation - that can never be bullet-proof.

However, as recent incidents have shown, the FunctionsAPI is subjective to heavy load, and thus keeping errors from triggering a new deployment is for sure valuable.

> Why this strict requirement to begin with?

YAGNI - _until you do_, which seems to be now!

Let me end this reply with a friendly tease regrading over-engineering; Cognite Typed Functions sure looks nice 😉  ( ❤️ )

dbrattli: Reverted changes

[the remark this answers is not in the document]
dbrattli: Reverted changes

cognite/client/_api/functions.py
```diff
@@ -251,6 +252,7 @@ def create(
...
             extra_index_urls (list[str] | None): Extra Index URLs for Python Package Manager to use. Be aware of the intrinsic security implications of using the `extra_index_urls` option. `More information can be found on official docs, <https://docs.cognite.com/cdf/functions/#additional-arguments>`_
             skip_folder_validation (bool): When creating a function using the 'folder' argument, pass True to skip the extra validation step that attempts to import the module. Skipping can be useful when your function requires several heavy packages to already be installed locally. Defaults to False.
+            skip_validation (bool): Pass True to skip argument validation of the handle function. This should be used if the handler does its own validation at runtime (e.g., frameworks like Cognite Typed Functions). When enabled, only checks that the function/callable is named 'handle', but skips validation of function arguments. Defaults to False.
```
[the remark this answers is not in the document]
dbrattli: Fixed

dbrattli:
@haakonvt Thanks, that is probably the best way forward for now.

haakonvt (dismissed):
:unicorn: looking forward to try out V2!

cognite/client/_api/functions.py
```diff
@@ -214,6 +214,7 @@ def create(
         index_url: str | None = None,
         extra_index_urls: list[str] | None = None,
         skip_folder_validation: bool = False,
+        skip_validation: bool = False,
```
HMEiding, line 217: Should we instead call it `skip_handle_validation` to avoid confusing it with the `skip_folder_validation` argument? What do you think?
```suggestion
        skip_handle_validation: bool = False,
```

cognite/client/_api/functions.py
```diff
@@ -288,6 +290,15 @@ def create(
                 >>>
                 >>> function = client.functions.create(name="myfunction", function_handle=handle)
 
+            Create function with validation skipped (for handlers that do their own validation):
```
HMEiding, line 293:
```suggestion
            Create function with handle validation skipped (for handles that do their own validation):
```

tests/tests_unit/test_api/test_functions.py
```diff
@@ -1315,3 +1317,169 @@ def validate_file_upload_call(*args, **kwargs):
     folder = Path(__file__).parent / "function_test_resources" / "good_absolute_import"
     file_id = fns_api_with_mock_client._zip_and_upload_folder(folder, name="name", external_id=xid)
     assert file_id == 123
+
+
+class TestSkipValidationFunctionality:
+    """Test skip_validation parameter functionality"""
+
+    def test_validate_function_handle_with_skip_validation_true(self):
+        """Test that we don't call _validate_function_handle when skip_validation=True"""
```
HMEiding, line 1326: This test name and docstring is wrong?

tests/tests_unit/test_api/test_functions.py
```diff
@@ -1315,3 +1317,169 @@ def validate_file_upload_call(*args, **kwargs):
...
+            body=[],
+            decorator_list=[],
+            returns=None,
+        )
+
+        # Should raise TypeError when called directly (which happens when skip_validation=False)
+        with pytest.raises(TypeError, match=r"Arguments .* must be a subset of"):
+            _validate_function_handle(func_def)
+
+    def test_validate_function_handle_name_node_with_skip_validation(self):
```
HMEiding, line 1370: This name is also wrong.

HMEiding (changes requested):
I think this solution is a good way forward. However, put in a couple of comments worth considering.

tests/tests_unit/test_api/test_functions.py
```diff
@@ -1315,3 +1317,169 @@ def validate_file_upload_call(*args, **kwargs):
...
+    ):
+        """Test that invalid handle args fail validation when skip_validation=False"""
+
+        # Create a handle function that would fail validation (invalid args)
+        def handle(custom_arg1, custom_arg2, invalid_arg):
+            """This handle has invalid arguments that don't match ALLOWED_HANDLE_ARGS"""
+            return {"result": "success"}
+
+        # Should raise TypeError when skip_validation=False (default)
+        with pytest.raises(TypeError, match=r"Arguments .* must be a subset of"):
```
HMEiding, line 1430: These two can also be parametrized with `skip_validation=True` and `skip_validation=False`

tests/tests_unit/test_api/test_functions.py
```diff
@@ -1315,3 +1317,169 @@ def validate_file_upload_call(*args, **kwargs):
...
+
+        def wrong_name_function(data, client, secrets):
+            """This function has the wrong name and should fail with skip_validation=False"""
+            return {"result": "success"}
+
+        # Should raise TypeError when skip_validation=False (default) because of wrong name
+        with pytest.raises(TypeError, match="Function is named 'wrong_name_function' but must be named 'handle'"):
+            cognite_client.functions.create(
+                name="typed_function_wrong_name", function_handle=wrong_name_function, skip_validation=False
+            )
```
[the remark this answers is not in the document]
dbrattli: We can do that, but then each unit-test becomes a big if-test 🤔 Not sure that is cleaner, wdyt? E.g:

```py
    @pytest.mark.parametrize("skip_handle_validation, should_raise", [(True, False), (False, True)])
    def test_create_function_with_handle_assignment(
        self, mock_functions_create_response, cognite_client, skip_handle_validation, should_raise
    ):
        """Test folder-based deployment with handle assignment and skip_handle_validation parameter"""
        folder = os.path.join(os.path.dirname(__file__), "function_test_resources", "function_with_handle_assignment")

        if should_raise:
            # Test validate_function_folder directly - should fail when skip_handle_validation=False
            with pytest.raises(TypeError, match="must contain a function named 'handle'"):
                validate_function_folder(
                    folder, "handler.py", skip_folder_validation=True, skip_handle_validation=skip_handle_validation
                )
        else:
            # Should work with skip_handle_validation=True even though get_handle_function_node returns None
            res = cognite_client.functions.create(
                name="typed_function_folder",
                folder=folder,
                function_path="handler.py",
                skip_handle_validation=skip_handle_validation,
            )
            assert isinstance(res, Function)
            assert mock_functions_create_response.calls[3].response.json()["items"][0] == res.dump(camel_case=True)
```

haakonvt:
FYI @dbrattli , I'm fixing merge conflict for `cognite/client/_api/functions.py` now (which I am the root cause for, see #2358 ).

PR #2361, opened 2025-10-08
feat(data_modeling): DM-3146 Support constraint and index state fields

audunska:
## Description
Constraints and indexes gain a new field `state`, which describes whether the constraint or index is valid.
Container properties gain a new field `constraint_state` with state fields for each of the property-level constraints `nullable`, `max_list_size`, and `max_text_size`.

All these new fields are `init = False`, since they are not meant to be set by the user. They get populated when loading from the service, and get stripped when `as_apply()` is called on the container.

## Checklist:
- [x] Tests added/updated.


audunska:
Not sure if we want to be forward-compatible with respect to the state enum, or not.

PR #2246, opened 2025-08-12
feat(records): alpha support for streams and records

erlendvollset:
- **Add support for streams**
- **Add basic support for records**
- **Raise FeaturePreviewWarning on all invocations of stream/record methods**
- **Use new warn_on_all_method_invocations decorator across the board**

## Description
Please describe the change you have made.

## Checklist:
- [ ] Tests added/updated.
- [ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [ ] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ ] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #2380, opened 2025-10-24
feat: add support for the Streams API

evertoncolling:
## Description
Adds support for the Streams API in the Data Modeling module, enabling users to create, retrieve, list, and delete streams.

## Checklist:
- [x] Tests added/updated.


haakonvt:
I kindly ask we do this post v8 unless there is a very sudden need!

haakonvt:
Btw, seen this? #2246

evertoncolling:
> Btw, seen this? #2246

Yes, this one is based on that one, but focusing only on streams (plus making sure we keep single item operations on top of streams and changing the stream templates from enum to a simple string).

The idea is to get this one in, as it's simpler, and then have a second PR focused only on records.

PR #2387, opened 2025-10-30
release `8.0.0a1` to PyPI (DM-3062)

haakonvt:
## Note to reviewer:
1. Commit history is clean - please review commit-by-commit!
2. Github only shows the most recent 250 commits, and this PR boasts 300+ ( :hurtrealbad: ), so you need to review those outside the UI 🕺

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -8,6 +8,78 @@ Changes are grouped as follows:
 - `Changed` for changes that do not fall into any other category
 - `Optional` for new, optional methods/features that you should be aware of - *and could take advantage of*
 
+## From v7 to v8
+### Dependency
+- **httpx**: The SDK now uses `httpx` instead of `requests` for HTTP operations. This is a new required dependency that
+  enables the new async interface.
+
+### Optional
+- **Async Support**: The SDK now provides full async support. The main client is now `AsyncCogniteClient`, but the synchronous `CogniteClient` is still available for backward compatibility. An important implementation detail is that it just wraps `AsyncCogniteClient`.
+- All helper/utility methods on data classes now have an async variant. A few examples: Class `Asset` has `children` and now also `children_async`, `subtree` and `subtree_async`, class `Function` now has `call` and `call_async`, class `TimeSeries` now has `latest` and `latest_async` etc.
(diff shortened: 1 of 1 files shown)


Aasnaess (dismissed):
LGTM

cognite/client/_cognite_client.py
```diff
@@ -92,27 +92,31 @@ def __init__(self, config: ClientConfig | None = None) -> None:
         # APIs just using base_url:
         self._api_client = APIClient(self._config, api_version=None, cognite_client=self)
 
-    def get(self, url: str, params: dict[str, Any] | None = None, headers: dict[str, Any] | None = None) -> Response:
+    async def get(
+        self, url: str, params: dict[str, Any] | None = None, headers: dict[str, Any] | None = None
+    ) -> httpx.Response:
```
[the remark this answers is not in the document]
erlendvollset: Same for the sync client

cognite/client/_api/time_series.py
```diff
@@ -178,29 +175,19 @@ def __call__(
...
             method="POST",
             chunk_size=chunk_size,
             filter=filter.dump(camel_case=True),
             advanced_filter=advanced_filter,
             limit=limit,
-            partitions=partitions,
             sort=prep_sort,
-        )
-
-    def __iter__(self) -> Iterator[TimeSeries]:
```
erlendvollset: Why have the `__iter__` methods been removed? - not mentioned in migration guide.

haakonvt: It is mentioned there, below `### Removed`.

Imo it should never have been added, having `__call__` as the way to iterate through data is perfect - and enough. Take instances for example, using `__iter__` can never return anything else than nodes. It also encourages iterating through all available data which we don't want.

cognite/client/utils/_concurrency.py
```diff
@@ -1,43 +1,127 @@
...
+    from cognite.client import global_config
+
+    return asyncio.BoundedSemaphore(global_config.max_workers)
+
+
+@cache
+def get_global_datapoints_semaphore() -> asyncio.BoundedSemaphore:
+    from cognite.client import global_config
+
+    return asyncio.BoundedSemaphore(global_config.max_workers)
```
erlendvollset: Why is this separate from the other global semaphore?

PR #2388, opened 2025-10-31
feat(simint-alpha): add kind=long routines support

polomani:
## Description
This pull request introduces support for kind="long" routines in the simulators API.
API docs: https://pr-3015.specs.preview.cogniteapp.com/20230101-alpha.json.html#tag/Simulator-Routines/operation/create_simulator_routine_simulators_routines_post

"Kind" for routines / routine revisions is an optional field and currently only supports "long" as a single avalable option. "Kind=long" routines have different limits on the API, which allows running heavier computations with more inputs, outputs, steps.

New unit tests are added to validate the new functionality, and existing integration tests are refactored for improved stability.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/tests/tests_integration/test_api/test_simulators/test_integrations.py
+++ b/tests/tests_integration/test_api/test_simulators/test_integrations.py
@@ -3,26 +3,7 @@ import time
 import pytest
 
 from cognite.client._cognite_client import CogniteClient
-from cognite.client.exceptions import CogniteAPIError
-from cognite.client.utils._text import random_string
-from tests.tests_integration.test_api.test_simulators.seed.data import SIMULATOR_INTEGRATION, ResourceNames
-
-
-def cleanup_inactive_integrations(cognite_client: CogniteClient, external_id_to_keep: str) -> None:
-    """Cleanup simulator integrations except the one with the given external_id."""
-    try:
-        all_integrations = cognite_client.simulators.integrations.list(limit=None)
-        integrations_to_delete = [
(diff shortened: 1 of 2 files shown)


tests/tests_unit/test_api/test_simulators/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+import re
+from collections.abc import Callable
+
+from requests import PreparedRequest
+from responses import RequestsMock
+
+from tests.utils import jsgz_load
+
+
+def add_mocked_request(rsps: RequestsMock, url: str, request_callback: Callable[[dict], tuple[int, dict]]) -> None:
```
haakonvt: Using the callback makes this logic a bit hard to follow to be honest. I wonder if we should remove this function entirely?

I use the `callback` when the return value is dynamic, it doesn't seem to be that here? As far as I can see you return:

```py
def request_callback(_: dict) -> tuple[int, dict]:
    response_item = {
        **TEST_ROUTINE_REV_ITEM_RESPONSE_FIELDS,
        **mock_response_fields,
    }
    return (200, response_item)
```

This should allow you to just do this directly:
```py
responses.add(
    "POST",
    url=...,
    json=PAYLOAD,
    status=STATUS,
)
```

polomani: done

haakonvt (approved):
🦄

PR #2394, opened 2025-11-13
fix: Device code authentication fails.

ronpal:
## Description

Fixes error due to lack of `.json()` attribute on object returned from msal. 

Also fixes `OAuthDeviceCode` failing when using `oauth_discovery_url` instead of `authority_url`. The code assumed the MSAL authority object would always have `device_authorization_endpoint`, but this isn't the case for non-Azure identity providers.

Manual tests (since we don't have integration tests):

### Before: Fails immediately

Screenshot 2025-11-25 at 15 29 31

### After: Device code request successful

Screenshot 2025-11-25 at 15 36 18

### Finally! Authentication successful
Screenshot 2025-11-25 at 15 36 44

## Checklist:
- [x] Tests added/updated.


haakonvt (commented):
This looks a lot better! A few last comments:

cognite/client/credentials.py
```diff
@@ -371,6 +372,67 @@ def scopes(self) -> list[str]:
     def scope_string(self) -> str:
         return " ".join(self.__scopes)
 
+    def _get_device_authorization_endpoint(self) -> str:
+        """Get device authorization endpoint."""
```
haakonvt:
```suggestion
```

tests/tests_unit/test_credential_providers.py
```diff
@@ -111,6 +113,145 @@ def test_access_token_generated(self, mock_public_client, expires_in):
...
+                None,
+                "device_authorization_endpoint not found",
+                id="oidc_discovery_missing_endpoint",
+            ),
+            # OIDC discovery network error
+            pytest.param(
+                None,
+                None,
+                "https://auth0.example.com/oauth",
+                requests.exceptions.RequestException("Network error"),
```
haakonvt: nit: preferable to test with something that's _not-the-base-class-exception_

trygveu (approved):

PR #2395, opened 2025-11-19
fix: configure release-please to use manifest mode

dbrattli:
## Problem

The current setup has `release-type: python` specified in the workflow file, which causes release-please to run in **simple mode**. This means the `release-please-config.json` file is ignored, and any settings in it (like `extra-files`) don't take effect. Ref: https://github.com/googleapis/release-please-action/issues/1088#issuecomment-2797894923

This creates a confusing situation where developers think their config file is being used, but it's actually ignored.

## Solution

Remove `release-type` from the workflow and use **manifest mode** instead, which properly reads settings from `release-please-config.json`.

## Changes

- **Workflow**: Remove `release-type: python` and add explicit `config-file` and `manifest-file` parameters
- **Config**: Move `release-type` into the `packages["."]` section (proper manifest format)
- **Manifest**: Create `.release-please-manifest.json` with current version (7.89.0)
- **Config**: Fix `package-name` to match actual package name

## Impact

- ✅ Config file settings now actually work
- ✅ Future config changes (custom changelog sections, etc.) will be respected
- ✅ No behavior change for current releases (already at v7.89.0)
- ✅ `extra-files` setting for `cognite/client/_version.py` now properly applied

## References

- [Release Please Action: Manifest Mode](https://github.com/googleapis/release-please-action#configuration)
- Similar fix in [cognite-typed-functions#115](https://github.com/cognitedata/cognite-typed-functions/pull/115)

---


haakonvt (approved):
🦄  - thanks for fixing!

PR #2414, opened 2025-11-26
fix: dont send null for system-managed props CogniteFile

haakonvt:
https://cognitedata.atlassian.net/browse/DM-3369

## Description
Unfortunately, the SDK sends full request bodies for all "unset" fields (or, they could have been set to None but we can't tell before v8).

This is a stop-gap to at least skip system-managed fields when null.

Raised from slack: https://cognitedata.slack.com/archives/C03G11UNHBJ/p1764156830780839


haakonvt (commented):
🦄

PR #2426, opened 2025-12-16
fix: handle missing id in TimeSeriesID for broken subscription references

### LINKED ISSUE #2418: `DatapointsSubscriptionAPI` fails when a subscribed `TimeSeries` changes `external_id`
**System information:**
 - OS: Ubuntu 24.04.3
 - Python Version: 3.13.7
 - SDK Version: 7.90.1 (latest) and 7.88.0

**Describe the bug**
When utilizing `DatapointSubscription`, updating the `external_id` of a `TimeSeries` that is currently a member of a subscription causes:

1. Updates to the Time Series are silently no longer written to the subscription, without as much as a warning
2. `DatapointsSubscriptionAPI.list_member_time_series()` crashes with an obscure error message, as illustrated in reproducer below
3. Operations like adding, removing and setting the member Time Series for the subscription crashes after a while

**To Reproduce**
The following code snippet reproduces the crash mentioned in point 2 above. It creates a dummy time series and subscription, verifies the membership, updates the time series external_id, and then attempts to list the members again.

```python
import cognite.client as ccl
import cognite.client.data_classes as cdc

# --- CONFIG ---
# Please insert a valid data_set_id and client
data_set_id: int = 123456789 
client: ccl.CogniteClient = ccl.CogniteClient()
# --------------

dummy_ts: None | cdc.TimeSeries = None
dummy_sub: None | cdc.DatapointSubscription = None
try:
    # 1. Create a dummy time series
    dummy_ts = client.time_series.create(
        cdc.TimeSeriesWrite(
            external_id="DELETE_ME",
            data_set_id=data_set_id,
        )
    )
    assert dummy_ts.external_id is not None

    # 2. Create a dummy subscription and add the dummy time series to it
    dummy_sub = client.time_series.subscriptions.create(
        cdc.DataPointSubscriptionWrite(
            "DELETE_ME",
            partition_count=1,
            time_series_ids=[dummy_ts.external_id],
            # NOTE^^: Contrary to what the parameter name suggests, this has to be the external_id, not the id
            instance_ids=[],
            data_set_id=data_set_id,
        )
    )

    # 3. Verify that the dummy time series is in the subscription
    member_ts = client.time_series.subscriptions.list_member_time_series(
        dummy_sub.external_id
    )
    assert len(member_ts) == 1
    assert member_ts[0].id == dummy_ts.id

    # 4. Update the external_id of the dummy time series
    dummy_ts.external_id = "NEW_DELETE_ME"
    client.time_series.update(dummy_ts)

    # 5. This is now crashing with an obscure `KeyError: 'id'` error, originating from deep inside the Cognite SDK
    member_ts = client.time_series.subscriptions.list_member_time_series(
        dummy_sub.external_id
    )

finally:
    # 6. Cleanup
    if dummy_ts is not None:
        # We need to delete by internal ID in case the external_id update failed or succeeded partway
        client.time_series.delete(dummy_ts.id, ignore_unknown_ids= True)
    if dummy_sub is not None:
        client.time_series.subscriptions.delete(dummy_sub.external_id, ignore_unknown_ids= True)
```

Full traceback:
```
  File ".../reproducer.py", line 63, in <module>
    member_ts = client.time_series.subscriptions.list_member_time_series(
        dummy_sub.external_id
    )
  File ".../python3.13/site-packages/cognite/client/_api/datapoints_subscriptions.py", line 197, in list_member_time_series
    return self._list(
           ~~~~~~~~~~^
        method="GET",
        ^^^^^^^^^^^^^
    ...<4 lines>...
        other_params={"externalId": external_id},
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File ".../python3.13/site-packages/cognite/client/_api_client.py", line 673, in _list
    list(itertools.chain.from_iterable(resource_fetcher)),
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "YY/python3.13/site-packages/cognite/client/_api_client.py", line 487, in _list_generator
    yield list_cls._load(unprocessed_items, cognite_client=self._cognite_client)
          ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "YY/python3.13/site-packages/cognite/client/data_classes/_base.py", line 410, in _load
    resources = [cls._RESOURCE._load(resource, cognite_client=cognite_client) for resource in resource_list]
                 ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "YY/python3.13/site-packages/cognite/client/data_classes/datapoints_subscriptions.py", line 287, in _load
    id=resource["id"],
       ~~~~~~~~^^^^^^
KeyError: 'id'
```

**Expected behavior**
Updating the `external_id` of a `TimeSeries` that is a member of a `DatapointSubscription` should not have any effect on the subscription at all. Updates to the data of a `TimeSeries` should still be written to the subscription as before the change in `external_id`.

**Additional context** 
**Design Observation:** The crash appears to stem from the fact that `DatapointSubscription` link to `TimeSeries` via the mutable `external_id` rather than the immutable internal `id`.

Since the API forces the use of `external_id` when creating a subscription (passing an integer `id` is not supported/accepted), the user has no way to create a stable link that survives an `external_id` update.

Ideally, the Subscription service should either internally resolve and store the immutable `id`, or allow the user to pass the immutable `id`, so that metadata updates to the `TimeSeries` do not break the subscription contract.

evertoncolling:
## Description
When a time series `external_id` is changed or the time series is deleted, the `/timeseries/subscriptions/members` API returns members without an `id` field. The SDK was expecting `id` to always be present, causing a KeyError. 

This PR:
- Makes `TimeSeriesID.id` optional to align with the API contract
- Adds `is_resolved` property to check if a reference points to an existing time series

Fixes #2418

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/tests/tests_unit/test_data_classes/test_datapoint_subscriptions.py
+++ b/tests/tests_unit/test_data_classes/test_datapoint_subscriptions.py
@@ -1,7 +1,12 @@
 import pytest
 
 from cognite.client.data_classes import filters
-from cognite.client.data_classes.datapoints_subscriptions import DatapointSubscription, DataPointSubscriptionWrite
+from cognite.client.data_classes.datapoints_subscriptions import (
+    DatapointSubscription,
+    DataPointSubscriptionWrite,
+    TimeSeriesID,
+    TimeSeriesIDList,
+)
 
 
 class TestDataPointSubscription:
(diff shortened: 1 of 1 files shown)


sighol (approved):
LGTM!

haakonvt (approved):

PR #2429, opened 2025-12-22
fix: restore log flow to root logger on disable (DM-3439)

haakonvt:
## Description
https://cognitedata.atlassian.net/browse/DM-3439

In the SDK, we accept the “debug: bool = False” parameter, which also can be turned on or off later. It has 2 main issues:

## 1. Critical Issue: The "Silent Logger" Bug
`_disable_debug_logging()` leaves the logger in a broken state.

#### The Flow:

1. `_configure_logger_for_debug_mode()` sets logger.propagate = False.
2. `_disable_debug_logging()` clears all handlers (`logger.handlers = []`) but does not reset propagate to `True`.

**The Result**: After you enable and then disable debug mode, the logger has no handlers and does not propagate to the root logger. Any subsequent logs (even errors or warnings) will be silently swallowed and effectively lost.

Fix: Restore propagation in the disable function.

## 2. High Risk: Direct List Manipulation
Directly assigning `logger.handlers = []` is generally discouraged in Python logging for two reasons:

1. Thread Safety: The logging module has internal locks. Modifying the list directly bypasses these safety mechanisms, which could cause race conditions in a multi-threaded application.
2. Destructiveness: This wipes out any handler attached to that logger. If a user had manually attached a FileHandler to the Cognite logger to save logs to a disk, this function would silently delete it.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1736,14 +1736,14 @@ testing = ["docopt", "pytest"]
 
 [[package]]
 name = "pathspec"
-version = "1.0.1"
+version = "1.0.2"
 description = "Utility library for gitignore style pattern matching of file paths."
 optional = false
 python-versions = ">=3.9"
 groups = ["dev"]
 files = [
-    {file = "pathspec-1.0.1-py3-none-any.whl", hash = "sha256:8870061f22c58e6d83463cfce9a7dd6eca0512c772c1001fb09ac64091816721"},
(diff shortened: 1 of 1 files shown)


vvb16 (approved):
Looks good :+1:

haakonvt:
🦄

doctrino (approved):

PR #2436, opened 2026-01-07
Krishna/python sdk suport for limits api

krishna1633:
## Description

Adds SDK support for the Limits API to retrieve limit values from Cognite Data Fusion. Version Bumped to  8.0.0a3

API Docs:

- `retrieve`: <https://api-docs.cognite.com/20230101-alpha/tag/Limits/operation/fetchLimitById/>
- `list`: <https://api-docs.cognite.com/20230101-alpha/tag/Limits/operation/listLimits/>

## Key Features

- Data classes: `Limit` and `LimitList`
- API methods:
  - `retrieve(id: str) -> Limit | None` - retrieves a single limit by its ID
  - `list(filter: Filter | None = None, limit: int | None = 1000) -> LimitList` - lists all limits, optionally filtered by prefix
- Integrated into `CogniteClient` as `client.limits`
- Uses `cdf-version: {api_subversion}-alpha` header for all requests (dynamically constructed from client config)
- Feature preview warning for alpha API endpoints

## Implementation Details

- `retrieve()` uses GET `/limits/values/{id}` endpoint
  - Returns `None` if the limit is not found (does not raise an error)
- `list()` supports two modes:
  - Without filter: Uses GET `/limits/values` endpoint
  - With filter: Uses POST `/limits/values/list` endpoint with a `Prefix` filter on `limitId` property
  - Only `Prefix` filters are supported for filtering limits
- Automatic pagination support via `_list` helper method (handles `nextCursor` automatically)

## Additional Changes

- Unit and integration tests for `retrieve()` and `list()` methods
- Auto-generated sync API file (`cognite/client/_sync_api/limits.py`) via pre-commit hook
- Documentation added to `docs/source/limits.rst`

## Checklist:
- [x] Tests added/updated.


haakonvt (approved):
🦄 LGTM!

PR #2456, opened 2026-01-29
refactor: concurrency settings in the SDK (DM-3473)

haakonvt:
Introduce configurable, per-API, _and_ method (read/write/delete) concurrency settings.

Context / Problem
* v7 uses a singleton `ThreadPoolExecutor`. Most calls bypass it → effective “unlimited” concurrency.
* v8 moved to async + bounded semaphores at the HTTP boundary. Correct but strict.
* Result: regressions where users previously relied on implicit concurrency (e.g. anyio.to_thread, or just massive thread pools), now throttled for the first time.
* Question: bug or feature? Enforcement is correct, but configurability is missing.

What this PR does
* Adds explicit, configurable concurrency settings.
* Preserves v8’s safety guarantees while allowing informed overrides.
* Enables separate concurrency budgets per API and per operation.

Why this is better
* Fixes perceived v8 “slowdown” without reverting to unsafe behavior.
* Makes concurrency a first-class, properly documented concept.
* Allows internal services to scale concurrency _deliberately_, not _accidentally_.

### DIFF
--- a/tests/tests_integration/test_api/test_assets.py
+++ b/tests/tests_integration/test_api/test_assets.py
@@ -490,7 +490,7 @@ def create_hierarchy_with_cleanup(
 
 @pytest.fixture(scope="class")
 def set_create_lim(async_client: AsyncCogniteClient) -> Iterator[None]:
-    with override_semaphore(2, target="basic"), pytest.MonkeyPatch.context() as mp:
+    with override_semaphore(2, target="general"), pytest.MonkeyPatch.context() as mp:
         # We set a low limit to hopefully detect bugs in how resources are split
         # without unnecessarily overloading the API with many thousand assets/request:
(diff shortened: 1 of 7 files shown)


erlendvollset:
Another approach could be to set concurrency limits based on regex patterns evaluated against the url path, ala what we do with `NON_IDEMPOTENT_POST_ENDPOINT_REGEX_PATTERN`. Then we don't need to explicitly use the correct semaphore all over the code base. But maybe explicit is good?

erlendvollset:
should we also be able to associate an endpoint with a set of semaphores? i.e. if we have `read=3`, `query(postgres)=2`, `search(elastic)=2`, then you pass `read(3), query(2)` for postgres-backed endpoints and `read(3), search(2)` for elastic-backed endpoints and you can have 2 concurrent queries, 2 concurrent searches, but never more than 3 concurrent reads in total.

PR #2457, opened 2026-02-02
Make `.get()` helper mappings initialize lazily

haakonvt:
The SDK has "since forever" supported quick item lookup on its list-classes via the .get-method that keeps a mapping internally to avoid linear search. Building this mapping should not be done unless when needed to avoid unneccessary slowdown and memory usage. This PR does exactly that.

### DIFF
--- a/cognite/client/data_classes/user_profiles.py
+++ b/cognite/client/data_classes/user_profiles.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from collections.abc import Sequence
+from collections.abc import Iterable
+from functools import cached_property
 from typing import Any
 
 from typing_extensions import Self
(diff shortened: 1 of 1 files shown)


cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -1035,11 +1035,36 @@ def as_ids(self) -> list[NodeId]:
...
+    def _instance_id_to_item(self) -> dict[tuple[str, str], T_Instance]:  # type: ignore [override]
+        if self.__instance_id_to_item is None:
+            self.__instance_id_to_item = {(inst.space, inst.external_id): inst for inst in self.data}
+        return self.__instance_id_to_item
+
+    @property
+    def _ext_id_to_item(self) -> dict[str, T_Instance]:
+        if self.__ext_id_to_item is None:
+            self.__ext_id_to_item = {inst.external_id: inst for inst in self.data}
+        return self.__ext_id_to_item
```
erlendvollset: Can use `@cached_property` here?
```suggestion
    def __init__(self, resources: Sequence[T_Instance]) -> None:
        super().__init__(resources)
        self.__ambiguous_xids: set[str] | None = None

    @cached_property
    def _instance_id_to_item(self) -> dict[tuple[str, str], T_Instance]:  # type: ignore [override]
        return {(inst.space, inst.external_id): inst for inst in self.data}

    @cached_property
    def _ext_id_to_item(self) -> dict[str, T_Instance]:
        return {inst.external_id: inst for inst in self.data}
```

PR #2460, opened 2026-02-02
docs: improve readability of API method signatures

MortGron:
## Description
In the docs long object signatures are very hard to read. Splitting the arguments of long function signatures on separate lines will greatly help on readablility. 

Example of current look:
image

Example of new look:
image

## Checklist:
- [ ] Tests added/updated.


haakonvt:
I also found a way to go from:

--------

Screenshot 2026-02-16 at 19 11 28

erlendvollset (approved):
🦄

PR #2464, opened 2026-02-06
Concurrency limiting stop gap (separate on CDF project, DM-3473)

haakonvt:
While we discuss the final solution (see https://github.com/cognitedata/cognite-sdk-python/pull/2456), we need a stop-gap in place. This just ensures we separate concurrency limiting on CDF project, which drastically reduces problems for backend services.


erlendvollset (dismissed):
🦄

PR #2469, opened 2026-02-13
feat: Add deprecation warning for Vision API

mallpunk:
Added deprecation warning for Vision API.

Ref https://docs.cognite.com/cdf/deprecated#deprecated-and-retired-features

### DIFF
--- a/docs/source/contextualization.rst
+++ b/docs/source/contextualization.rst
@@ -43,8 +43,8 @@ Convert to an interactive SVG where the provided annotations are highlighted
 .. automethod:: cognite.client._api.diagrams.DiagramsAPI.convert
 
 
-Vision
-------
+Vision (deprecated)
+-------------------
 
 .. warning::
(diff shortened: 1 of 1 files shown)


cognite/client/_api/vision.py
```diff
@@ -18,6 +18,16 @@
 class VisionAPI(APIClient):
     _RESOURCE_PATH = "/context/vision"
 
+    @staticmethod
+    def _deprecation_warning() -> None:
+        warnings.warn(
+            "The Vision API will be removed in a future version of the SDK. "
+            "Please migrate to the recommended alternative. "
+            "Read more at: https://docs.cognite.com/cdf/deprecated#deprecated-and-retired-features",
+            FutureWarning,
```
haakonvt: Although I agree `FutureWarning` is most correct, ref docs:

> Base class for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

The problem is this "...ignored by default, unless triggered by code in `__main__`".

I think we really want all of our users to see this warning, hence I suggest changing to `UserWarning`

PR #2470, opened 2026-02-13
fix: Retrieving routine revision with zero valued input or empty time series external ID in the configuration

MortGron:
## Description
client.simulators.routines.revisions.list or retrieve no longer returns a corrupted routine revision if the revision has a zero valued input or empty time series external ID in the configuration

## Checklist:
- [x] Tests added/updated.


haakonvt (dismissed):
Please add a comment to the relevant test on this format:

```
test_api_client.py:378-380 — "Between SDK version 7.0.0 and 7.33.1, ordering of results was broken... This bug came back in version 8.0.0, and was fixed in 8.0.3."
test_assets.py:607 — "SDK 5.10.1 and earlier versions had a bug that could run upsert even when False."
test_assets.py:622 — "SDK 6.24.0 and earlier versions had a bug when using upsert..."
test_datapoints.py:818 — "Bug prior to 7.74.1, when iterating using instance ID..."
test_datapoints.py:845 — "Bug prior to 7.74.3, when iterating object aggregate datapoints..."
test_datapoints.py:884 — "Bug prior to 7.86.0, when mixing instance IDs with other identifier types..."
test_datapoints.py:1366 — "Bug prior to 7.32.4, several methods on DatapointsArray would fail..."
test_datapoints.py:1578 — "SDK bug v<5, id mapping would not exist..."
test_datapoints.py:2094 — "SDK bug v<5, would be all None"
test_documents.py:201 — "Bug prior to 7.70.3 would convert user given property..."
test_identifier.py:29 — "Bug prior to 7.70.7, empty string external ID would fail..."
test_raw.py:440 — "Bug in 8.0.0 to 8.0.7: limit=-1 was not normalised to None..."
test_datapoints.py:367 — "Bug in 8.0.0 to 8.1.0: all asyncio tasks were created at once..."
test_capabilities.py:614 — "Would fail... prior to 7.2.1 due to a bug in 'as_tuples'."
test_filters.py:58 — "Bug in 8.0.0 to 8.0.5: eq was added to Filter..."
test_graphql.py:7 — "Bug prior to 7.20.1 would try to convert non-integer timestamps..."
test_filters.py (data_models):308 — "Bug prior to 6.32.4 would dump user given keys in camelCase"
test_filters.py (data_models):318 — "Bug prior to 7.63.10: Not-filter would accept and ignore all-but-the-first filter..."
test_filters.py (data_models):326 — "Bug prior to 7.63.10: CompoundFilters (and/or/not) would accept no filters given..."
test_instances.py:577 — "Bug prior to 7.62.6 made to_pandas(expand_properties=True) fail..."

```

haakonvt:
There's also two linter errors from mypy:

```
cognite/client/data_classes/simulators/routine_revisions.py:89: error: Unused "type: ignore" comment  [unused-ignore]
                return UnknownCogniteObject(resource)  # type: ignore[return-value]
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cognite/client/data_classes/simulators/routine_revisions.py:89: error: Name "UnknownCogniteObject" is not defined  [name-defined]
                return UnknownCogniteObject(resource)  # type: ignore[return-value]
                       ^~~~~~~~~~~~~~~~~~~~
cognite/client/data_classes/simulators/routine_revisions.py:89: note: Error code "name-defined" not covered by "type: ignore" comment
Found 2 errors in 1 file (checked 514 source files)
```

cognite/client/data_classes/simulators/routine_revisions.py
```diff
@@ -79,9 +79,15 @@ def _load(cls, resource: dict[str, Any] | SimulatorRoutineInput) -> Self:
 
         is_constant = resource.get("value")
         is_timeseries = resource.get("sourceExternalId")
```
polomani: simpler was to just change these two lines
```
is_constant = resource.get("value") is not None
is_timeseries = resource.get("sourceExternalId") is not None
```

haakonvt (dismissed):

haakonvt:
Consider @polomani suggestion btw

cognite/client/data_classes/simulators/routine_revisions.py
MortGron: Thanks

haakonvt (approved):
🦄

PR #2480, opened 2026-02-18
docs: Automatically update copyright years

MortGron:
## Description
For many years it has been like this:

image

With this change, the copyright year will automatially be set to the current year:

image

The start year is the year of the first tagged version of the SDK: 2017

To use a year range would be consistent with the Cognite home page: 

image

## Checklist:
- [ ] Tests added/updated.


haakonvt (approved):
:unicorn: great!

PR #2483, opened 2026-02-20
docs: Remove duplicate type hints in docstrings and annotations

MortGron:
## Description
The duplication of type hints in both docstring and as annotations makes the docstrings unnecessarily hard to read as a developer. It is possible to remove type hints in the docstrings without changing the structure of the documentation. This PR removes all type hints from the docstrings, but enables a Sphinx extension to keep the type hints in the documentation.

Current documentation: 

image

New documentation:

image

A few points to discuss:

* Should it still be allowed to provide type hints in the docstrings? It might make sense for very long type hints to have a custom type hint instead
* Is the change in documentation look accepptable? All type hints are now stylized as inline code in the documentation, which in my opinion is slightly less readable than how it was. 

## Checklist:
- [ ] Tests added/updated.


erlendvollset (commented):
I think most (sane) modern tooling relies on signature annotations, so this is a good change.

@haakonvt does this affect the v8 sync app generator?

PR #2485, opened 2026-02-22
feat!: Change Pandas timestamp resolution to ms for timestamp attributes

MortGron:
## Description
Currently when using `to_pandas` on list like objects a timestamp resolution of 'ns' is enforced for both Pandas v2 and v3. For single Cognite resources that have only timestamp attributes, resolution will be 'ns' in Pandas v2 and 'ms' in v3. For single Cognite resources that have at least on timestamp attribute and at least one non-timestamp attribute, time resolution will be 'ms' in both v2 and v3.

In Pandas v3 the timestamp resolution is to a much larger extent [inferred from the input data](https://pandas.pydata.org/docs/whatsnew/v3.0.0.html#datetime-timedelta-resolution-inference), where Pandas v2 would automatically set resolution 'ns'. CDF has a resolution of 'ms', so it would make sense for the `to_pandas` methods to also convert timestamp columns to `ms`.

An example where keeping resolution 'ns' would be inconenient in Pandas v3 is when using the `from pandas.testing import assert_frame_equal` where the resolution of timestamp columns are compared.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -163,14 +163,19 @@ class CogniteResource(ABC):
         for element in ignore or []:
             dumped.pop(element, None)
 
+        has_time_attribute = False
         if convert_timestamps:
             for k in TIME_ATTRIBUTES.intersection(dumped):
                 dumped[k] = pd.Timestamp(dumped[k], unit="ms")
+                has_time_attribute = True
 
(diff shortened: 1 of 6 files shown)


haakonvt:
Closing in favor of https://github.com/cognitedata/cognite-sdk-python/pull/2724

PR #2486, opened 2026-02-23
chore: update all links to the official Cognite API reference (DM-3537)

haakonvt:
Ideally, the old would redirect correctly, but as discussed on Slack, we should update the links anyway:
https://cognitedata.slack.com/archives/CN9QYQRSR/p1770909541673619


haakonvt:
🦄

PR #2488, opened 2026-02-23
fix: avoid flaky workflow execution list test by comparing IDs

trygveu:
## Summary
- Replaces `assertCountEqual` (full object equality) with an ID-based subset check in `test_list_workflow_executions`.
- The `cancel()` response can return before async fields like `end_time` are populated server-side, causing the subsequent `list()` to return objects that don't compare equal despite being the same execution.

## Background

In PR #2487 I saw two flaky test runs with errors like this:

```
tests/tests_integration/test_api/test_workflows.py:581: 
/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/unittest/case.py:1216: in assertCountEqual
    self.fail(msg)

self = <unittest.case.TestCase testMethod=runTest>
msg = 'Element counts were not equal:\nFirst has 1, Second has 0:  <cognite.client.data_classes.workflows.WorkflowExecution ...6e510>\nFirst has 0, Second has 1:  <cognite.client.data_classes.workflows.WorkflowExecution object at 0x7f0004c6f3e0>'

    def fail(self, msg=None):
        """Fail immediately, with the given message."""
>       raise self.failureException(msg)
E       AssertionError: Element counts were not equal:
E       First has 1, Second has 0:  <cognite.client.data_classes.workflows.WorkflowExecution object at 0x7f0004c6e510>
E       First has 0, Second has 1:  <cognite.client.data_classes.workflows.WorkflowExecution object at 0x7f0004c6f3e0>

/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/unittest/case.py:715: AssertionError
=============================== warnings summary ===============================
```

I believe this connects to the timing of the fixtures, and the async nature of cancel, so that it will be more robust to compare it by IDs instead of object equality.


haakonvt (approved):
Sounds good to me, thanks for taking the time to address 😄

PR #2489, opened 2026-02-23
feat: Enable Python 3.14 in CI and reduce destinations

trygveu:
Python 3.14 is stable and should be part of our CI.


tests/tests_unit/test_utils/test_time.py
```diff
@@ -789,7 +787,7 @@ def test_month_aligner__ceil(self, dt: datetime, expected: datetime) -> None:
         assert expected == MonthAligner.ceil(dt)
 
     def test_month_aligner_ceil__invalid_date(self) -> None:
-        with pytest.raises(ValueError, match=r"^day is out of range for month$"):
+        with pytest.raises(ValueError, match=r"day.+range"):
```
haakonvt: Why these two random changes?

PR #2498, opened 2026-03-02
update v8 with latest from v7

haakonvt:
It may look a bit weird with the 3 merge commits, but the reason is simple. Merge up-to-but-not-including 1e95c0e9788c32c6dbf3e404bd018f1fed7a95d7, merge that using `-s ours` (to avoid 650 merge errors when we can easily apply the same changes on v8 with a single regex), then fast-forward and merge the last commits.


erlendvollset (approved):
🦄

PR #2528, opened 2026-03-25
chore: handle pandas v2 and v3 by enforcing datetime freq of millisec

haakonvt:
Quite a few changes in pandas v3, specifically around the inference of frequency for datetime-like data: https://pandas.pydata.org/docs/whatsnew/v3.0.0.html#datetime-timedelta-resolution-inference

This PR unifies frequency to always be milliseconds, matching the API closely.


haakonvt:
Moved to https://github.com/cognitedata/cognite-sdk-python/pull/2724

PR #2553, opened 2026-03-31
docs: encourage higher level tools for interacting with data modeling

MortGron:
## Description
Discourage the use of the pure SDK for dealing with data modeling

I am very open to changing the exact wording.

## Checklist:
- [ ] Tests added/updated.


haakonvt (commented):
LGTM, but let's get the opinion of @arnekv on this before merging 👍

PR #2563, opened 2026-04-09
docs: Show correct accessors in the documentation

MortGron:
## Description
Currently the docs does not show how the users access all the methods of the SDK because it shows the methods as beloning to classes the useres do not directly interact with. With this PR the docs will show how the users actually can access the methods of the SDK.

Currently: 

image

Future:

image

To achieve this, a few changes had to be applied:

* The API methods are now displayed using autosummary. This means all API methods gets its own page in the docs.
* Because of this many docstrings had to be fixed for Sphinx not to throw errors.
* The `AsyncCogniteClient` has to get the correct accessors. This is only done if a special envionrmental variable is set.

How autosummary looks like:

image

## Checklist:
- [x] Tests added/updated.


erlendvollset:
There's a large amount of formatting changes in the same commit as the main change here, which makes this a lot harder to review than it needs to be. Please split that into a dedicated commit.

The "autosummary and one page per method" setup did not work great the last time we tried it, but will click around a bit tomorrow to see what it's like.

haakonvt:
Thanks for keeping the PRs coming for docs improvements! While the change in yellow rectangle looks great, I think the change I've marked in the red rectangle is not so great. Can this be modified?

----

image

MortGron:
> There's a large amount of formatting changes in the same commit as the main change here, which makes this a lot harder to review than it needs to be. Please split that into a dedicated commit.
> 
> The "autosummary and one page per method" setup did not work great the last time we tried it, but will click around a bit tomorrow to see what it's like.

This PR has only one goal, and that is to remove the references to the internal classes in the documenation. All other changes are consequences of the way this is achieved. Let me elaborate on this:

* The fundamental approach is to use the package https://sphinx-autosummary-accessors.readthedocs.io/en/stable/ . This is the same approach other major Python packages, like pandas, use to solve the same problem.
* This necceitates the use of `autosummary` where we have the problem with accessors. Note we do not need to use this for the data classes, as they do already have correct accessors. So this PR does not change the way data classes are doumented.
* When usuing `autosummary`, Sphinx will throw errors if the summary line of the docstring is a hyperlink with a dot before the link address. This is the main formatting change of dostrings that is needed for Sphinx to not fail. 
* In addition, autosummary requires the summary line of dosctrings to not be too long. In particular, docstring that did not have an empty line after the summary line, or that have summary lines of multiple sentences, have been changes to make the autosummary output look accepptable.

What do you mean it did not work great the last time we tried? The major problem with the last attempt was that it had multiple goals, while thie PR only has one goal. I am afraid to not go for this solution, but also not have any clear other solution that solves this problem. I imagine it possible to dig into Sphinx and do more major modifications to achieve correct accessors, but I do not know how to do it, and I am afraid no one else will do it.

MortGron:
> Thanks for keeping the PRs coming for docs improvements! While the change in yellow rectangle looks great, I think the change I've marked in the red rectangle is not so great. Can this be modified?
> 
> image

Thank you for the quick feedback. Is this better? The page heading will end up in the table of contents on the left.

image

So this is the template showing how the above example was generated. It takes the method name, removes underscores and capitalizes the words.

image

In particular for this AI section, we should probably make a new heading to indicate this is dealing with documents.

haakonvt:
@MortGron I'm still not entirely sold on this change - would you mind inviting to a quick 15 min meeting with myself + @erlendvollset and showcase some before and after; what gets better and what potentially gets worse? 😄

MortGron:
> @MortGron I'm still not entirely sold on this change - would you mind inviting to a quick 15 min meeting with myself + @erlendvollset and showcase some before and after; what gets better and what potentially gets worse? 😄

Sold in the sense that you do not want to change the way the accessors are displayed in the documentation, or in the sense that you do not like the side effects of this particular solution?

MortGron:
@haakonvt @erlendvollset all headings from the current documentation have been restored. So now the table of contents in the left menu will show the same headings as the current documentation. The same applies to the stub page for every API method.
image

Compare below example with screenshot above:
image

haakonvt (dismissed):
🦄 - did a comparison, and I think this is great! One page per method is very nice, making it easier on the eyes. The full "attribute path" is ofc also a great change. Good stuff!

cognite/client/_sync_api/data_modeling/instances.py
```diff
@@ -530,7 +530,9 @@ def subscribe(
         throttle_seconds: float = 1,
     ) -> SubscriptionContext:
         """
-        Subscribe to a query and get updates when the result set changes. This runs the sync() method in a background task.
+        Subscribe to a query and get updates when the result set changes.
+
+        This runs the sync() method in a background task.
```
haakonvt, line 535: What happened here?

cognite/client/data_classes/contextualization.py
```diff
@@ -1234,7 +1234,7 @@ async def save_predictions_async(
         See https://docs.cognite.com/api/v1/#operation/annotationsSuggest
 
         Args:
-            creating_user (str): (str, optional): A username, or email, or name.
+            creating_user (str): A username, or email, or name.
```
haakonvt, line 1237: Nice catch

haakonvt:
Sorry -  is this test necessary/correct?

```
diff --git a/tests/tests_unit/test_cognite_client.py b/tests/tests_unit/test_cognite_client.py
index f9917c792..aa926440d 100644
--- a/tests/tests_unit/test_cognite_client.py
+++ b/tests/tests_unit/test_cognite_client.py
@@ -1,6 +1,8 @@
 
+import importlib
 import logging
+import os
 import ssl
 from typing import Any
 
@@ -158,3 +160,11 @@ class TestCogniteClient:
 
         assert isinstance(sync_client, CogniteClient)
         assert sync_client.config is async_client.config
+
+    def test_client_has_accessors(self) -> None:
+        os.environ["BUILD_COGNITE_SDK_DOCS"] = "true"
+        from cognite.client import _cognite_client
+
+        importlib.reload(_cognite_client)
+        client_class = _cognite_client.AsyncCogniteClient
+        assert hasattr(client_class, "time_series")
```

haakonvt:
🦄

doctrino (dismissed):

haakonvt:
🦄

doctrino (approved):

PR #2566, opened 2026-04-10
feat(files): Support automatic upload of files larger than 5GB

msusag:
## Description
The SDK functions `upload()` for single or directory upload, and `upload_content()` for CogniteFile based files, is currently limited to only upload files smaller than 5GB in AWS and Azure CDF clusters.

With the changes in this PR the `upload()` and `upload_content` functions will raise this to support upload of files up to (about) 1000 GiB in size, as documented in the Files API docs (https://api-docs.cognite.com/20230101/tag/Files/operation/initMultiPartUpload).

Internally the functions will now use the multipart upload functions. Multipart upload enables upload of files larger than 5 GB in all CDF Cloud environments (Azure, AWS, and Google). In addition upload performance can be increased, especially for single large files.

With this change, client scripts or notebooks using upload() and upload_content() will benefit implicitly and not have to port their scripts to use the (more low level) multipart upload functions in the SDK. 
Note that the other non-file related upload functions are not updated, as they are meant for streaming data, where data is not parallell readable from different offsets, and size is not generally known up front.

A new global max_open_files semaphore was added to ensure that the async parallel uploads adhere to the concurrency setting in the configuration. The current implementation can fail with "too many open files" when uploading directories with many files.

To test the functionality, a custom Python test script was created to test upload performance for few large, and also many small files. The main test case uploads a folder with 4 files, where 2 of the files were large, above 5 GB.
In total 30703.83 MB of data is uploaded, per test run.
The file upload script mainly calls `client.files.upload(DIRECTORY_NAME, recursive=True)`.

With the current latest released 8.0.7 version, the upload() of large files works in CDF Google clusters, but fails in AWS and Azure clusters.
With the fixes in this PR, the script succeeds in all 3 CDF cloud environments.
The upload script achieves high upload rate on all the cloud environments, with the 1000 Mbps Ethernet connection I am using.

Comparison of old and new behaviour and performance for upload of folder with a few large files.
| CDF cloud env, cluster | Before<br>(SDK 8.0.7) | After<br>(with this PR) |
|----------------------------|-------------------|----------------|
| Google cloud, greenfield     | 706 Mbps              | 838 Mbps         |
| Azure                                     | Fails                        | 839 Mbps         |
| AWS, orangefield                 | Fails                        | 835 Mbps         |

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/files.py
```diff
@@ -484,7 +492,7 @@ async def upload(
         recursive: bool = False,
         overwrite: bool = False,
     ) -> FileMetadata | FileMetadataList:
-        """`Upload a file <https://api-docs.cognite.com/20230101/tag/Files/operation/initFileUpload>`_
+        """`Upload a file or directory`_
```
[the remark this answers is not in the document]
msusag: Fixed in later commit

cognite/client/_api/files.py
```diff
@@ -588,8 +596,55 @@ async def upload(
...
+        part_no: int,
+        part_size: int,
+        file_size: int,
+    ) -> None:
+        offset = part_no * part_size
+        read_size = min(part_size, file_size - offset)
         with path.open("rb") as fh:
-            return await self.upload_bytes(fh, overwrite=overwrite, **file_metadata.dump(camel_case=False))
+            fh.seek(offset)
+            content = fh.read(read_size)
```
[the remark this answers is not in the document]
msusag: Using AsyncFileChunker for reading in later commits.

cognite/client/_api/files.py
```diff
@@ -588,8 +596,55 @@ async def upload(
...
+        for more details on multipart upload and the constraints on part size and count.
+
+        Args:
+            file_size (int): The total file size in bytes.
+
+        Returns:
+            tuple[int, int]: A tuple of (part_size, num_parts).
+        """
+        if file_size < MIN_MULTIPART_SIZE:
+            return MIN_MULTIPART_SIZE, 1
```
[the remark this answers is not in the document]
msusag: Fixed in later commit.

cognite/client/_api/files.py
```diff
@@ -588,8 +596,55 @@ async def upload(
...
+        session = await self.multipart_upload_session(
+            parts=num_parts,
+            overwrite=overwrite,
+            **file_metadata.dump(camel_case=False),
+        )
+
+        async with session:
+            await asyncio.gather(
+                *(self._upload_file_part(session, path, part_no, part_size, file_size) for part_no in range(num_parts))
+            )
```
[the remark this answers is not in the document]
msusag: Semaphore added in later commit.

tests/tests_unit/test_api/test_files.py
```diff
@@ -95,6 +96,33 @@ def example_file(mock_geo_location: GeoLocation) -> dict[str, Any]:
...
+        json={},
+        is_reusable=True,
+    )
+    yield example_file
+
+
+@pytest.fixture
+def mock_upload_bytes_response(
+    httpx_mock: HTTPXMock, example_file: dict[str, Any], async_client: AsyncCogniteClient
 ) -> Any:
```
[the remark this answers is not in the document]
msusag: Fixed in later commits.

tests/tests_unit/test_api/test_files.py
```diff
@@ -95,6 +96,33 @@ def example_file(mock_geo_location: GeoLocation) -> dict[str, Any]:
 @pytest.fixture
 def mock_file_upload_response(
     httpx_mock: HTTPXMock, example_file: dict[str, Any], async_client: AsyncCogniteClient
+) -> Any:
```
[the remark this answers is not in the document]
msusag: Fixed in later commit.

cognite/client/_api/files.py
```diff
@@ -564,32 +572,90 @@ async def upload(
...
+        path: Path,
+        part_no: int,
+        part_size: int,
+        file_size: int,
+    ) -> None:
+        offset = part_no * part_size
+        read_size = min(part_size, file_size - offset)
         with path.open("rb") as fh:
-            return await self.upload_bytes(fh, overwrite=overwrite, **file_metadata.dump(camel_case=False))
+            await session.upload_part_async(part_no, AsyncFileChunker(fh, offset=offset, size=read_size))  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
msusag: Fixed in later commits.

cognite/client/_constants.py
```diff
@@ -33,3 +33,9 @@ def __bool__(self) -> Literal[False]:
 DATA_MODELING_DEFAULT_LIMIT_READ = 10
 DEFAULT_DATAPOINTS_CHUNK_SIZE = 100_000
 _RUNNING_IN_BROWSER = IN_BROWSER
+
+# Files API constants
+MIN_MULTIPART_SIZE = 5 * 1024 * 1024  # 5 MiB
+MAX_MULTIPART_SIZE = 4000 * 1024 * 1024  # 4000 MiB
+DEFAULT_MULTIPART_SIZE = 50 * 1024 * 1024  # 50 MiB
+MAX_MULTIPART_PARTS = 250
```
haakonvt: Let's add "FILE" into the constant names to make them self-explanatory

cognite/client/_api/files.py
```diff
@@ -448,22 +456,45 @@ async def upload_content(
         external_id: str | None = None,
         instance_id: NodeId | None = None,
     ) -> FileMetadata:
-        """`Upload a file content <https://api-docs.cognite.com/20230101/tag/Files/operation/getUploadLink>`_
+        """`Upload file content <https://api-docs.cognite.com/20230101/tag/Files/operation/getMultiPartUploadLink>`_
```
haakonvt, line 462: Are we sure we want to point this away from the "old" get upload link? The multi-part is a bit more complex.

cognite/client/_api/files.py
```diff
@@ -448,22 +456,45 @@ async def upload_content(
...
-        elif path.is_dir():
+        if path.is_dir():
             raise IsADirectoryError(path)
-        raise FileNotFoundError(path)
+        if not path.is_file():
+            raise FileNotFoundError(path)
+
+        from cognite.client import global_config
+
+        upload_semaphore = asyncio.Semaphore(global_config.concurrency_settings.general.write)
```
haakonvt, line 482: The docs for multi-part say:

> The parts can optionally be uploaded in parallel, preferably on a subset of parts at a time, for example maximum 3 concurrent PUT operations.

Should we introduce a separate `concurrency_settings` for file upload/download concurrency?

cognite/client/utils/_uploading.py
```diff
@@ -49,28 +49,36 @@ class AsyncFileChunker(AsyncIterator[bytes]):
...
     CHUNK_SIZE = 64 * 1024  # 64 KiB chunks by default, copying httpx default
 
-    def __init__(self, file_handle: BinaryIO) -> None:
+    def __init__(self, file_handle: BinaryIO, *, offset: int = 0, size: int | None = None) -> None:
         from cognite.client import global_config
 
         self._file_handle = file_handle
         self._chunk_size = global_config.file_upload_chunk_size or self.CHUNK_SIZE
+        self._remaining = size
+        self.size = size  # exposed so prepare_content_for_upload can set Content-Length
```
haakonvt, line 66: Exposing `size` is a great addition

cognite/client/utils/_uploading.py
```diff
@@ -21,7 +21,7 @@ def prepare_content_for_upload(
             content = content.encode("utf-8")
             file_size = len(content)
         case AsyncIterable():
-            file_size = None
+            file_size = getattr(content, "size", None)
```
haakonvt: Suggestion:
```py
case AsyncFileChunker():
    file_size = content.size
case AsyncIterable():
    file_size = None
```

cognite/client/_api/files.py
```diff
@@ -484,7 +515,13 @@ async def upload(
         recursive: bool = False,
         overwrite: bool = False,
     ) -> FileMetadata | FileMetadataList:
-        """`Upload a file <https://api-docs.cognite.com/20230101/tag/Files/operation/initFileUpload>`_
+        """`Upload a file or directory <https://api-docs.cognite.com/20230101/tag/Files/operation/initMultiPartUpload>`_
+
+        Creates files in files API with metadata and uploads file content.
+
+        If path is a directory, this method will upload all files in that directory. Use `recursive=True` for subdirectories as well.
```
haakonvt:
```suggestion
        Tip:
            If path is a directory, this method will upload all files in that directory. Use `recursive=True` for subdirectories as well.
```
or `Note:`

cognite/client/_api/files.py
```diff
@@ -564,32 +601,87 @@ async def upload(
...
+        session = await self.multipart_upload_session(
+            parts=num_parts,
+            overwrite=overwrite,
+            **file_metadata.dump(camel_case=False),
+        )
+
+        async with session:
+
+            async def upload_part(part_no: int) -> None:
+                async with upload_semaphore:
```
haakonvt, line 644: Don't need this semaphore, already being auto-used.

cognite/client/_api/files.py
```diff
@@ -564,32 +601,87 @@ async def upload(
             source_modified_time=source_modified_time,
             security_categories=security_categories,
         )
+        from cognite.client import global_config
```
andersfylling: I'm seeing a few lazy evaluated imports, could we keep these at the top of the file - or is there something preventing that?

cognite/client/_api/files.py
```diff
@@ -564,32 +601,87 @@ async def upload(
...
+        path: Path,
+        part_no: int,
+        part_size: int,
+        file_size: int,
+    ) -> None:
+        offset = part_no * part_size
+        read_size = min(part_size, file_size - offset)
         with path.open("rb") as fh:
-            return await self.upload_bytes(fh, overwrite=overwrite, **file_metadata.dump(camel_case=False))
+            await session.upload_part_async(part_no, AsyncFileChunker(fh, offset=offset, size=read_size))
```
haakonvt, line 685: I guess the actual file read is blocking, but I think we should just accept that.

haakonvt (commented):
Excellent changes - great QOL improvement! Consider if we can reuse `execute_async_tasks_with_fail_fast` or `execute_async_tasks` from `cognite/client/utils/_concurrency.py`. A few other suggestion/remarks:

cognite/client/_api/files.py
msusag: I also noticed that during development, as I was following a pattern from other code that used global_config.

If the same line is moved to top of the file along with other imports, it leads to a circular import error:

`ImportError: cannot import name 'global_config' from partially initialized module 'cognite.client' (most likely due to a circular import)
`

msusag: I saw the `from cognite.client import global_config` is used in a lot of places within functions in the SDK. So I thought that was the practice to avoid the circular imports.

But I see that using the more specific `from cognite.client.config import global_config` (adding .config) solves the circular import problem.

cognite/client/_api/files.py
msusag: The upload() function will with the changes in this PR use the files/multiuploadlink (and files/completemultipartupload) endpoint.
So it seems more correct to reference that endpoint in the REST API documentation. 
The API docs there contains further details about the file size restrictions, that it provides a uniform API across all cloud environments for CDF, and the steps needed to use multipart upload API endpoints.

However, all those details are really not that relevant for users of upload(), as the function here takes care of the part size calculation, upload parts in parallel, etc, and completing the operation with the completemultipartupload call.

cognite/client/_api/files.py
msusag: That could be relevant yes, but then probably in a separate PR?
Or should the httpx resource limits be used for that?
https://www.python-httpx.org/advanced/resource-limits/

I think it could be potentially also relevant to support network bandwidth throttling, 
for throttling the upload of files (also for download, and for other data?).
Throttling can perhaps be implemented with existing async or httpx extensions.

cognite/client/_api/files.py
msusag: Yes, the synchronous `path.open` is still used. 
mmap or asyncio could be used, but I doubt it really would improve things.

cognite/client/_api/files.py
msusag: Moved the import to top of the file in later commits.

cognite/client/_api/files.py
msusag: Added Note:

cognite/client/_api/files.py
msusag: Without the semaphore the _upload_file_part calls gets out of control, running more calls than concurrency_limit.

I added more test cases checking for this, see the new `test_upload_content_semaphore_limits_concurrent_parts` 
and 
`test_upload_semaphore_limits_concurrent_parts` 
in test_files.py.

cognite/client/utils/_uploading.py
msusag: Fixed in later commits.

cognite/client/_constants.py
msusag: Fixed in newer commits

cognite/client/_api/files.py
haakonvt: That's interesting 🤔 Every API call in the SDK will at some point hit `AsyncHTTPClientWithRetry._with_retry`:
```py
async with concurrency_control:
    if headers is not None:
        self.refresh_auth_header(headers)
    response = await coro_factory()
```
From how I read your current implementation, each successful upload / PUT is actually consuming 2 X from the semaphore.

haakonvt: I had misread the code, discussed on slack: https://cognitedata.slack.com/archives/CG10VQPFX/p1776087716226049

haakonvt (dismissed):
🦄  - really great quality-of-life improvements!

cognite/client/_api/files.py
msusag: Separate files concurrency settings has been introduced in later commits.

haakonvt (dismissed):
🦄

haakonvt:
🦄

haakonvt (approved):

Aasnaess (approved):

PR #2577, opened 2026-04-19
chore(ci): surface pyodide test errors via .catch() + fix CI check

ks93:
## Summary

`scripts/test-pyodide.js` wraps `test_cognite_sdk()` with a bare `.then(...)` and no `.catch(...)`. When the pyodide promise rejects, Node's default unhandled-rejection handler dumps the 69KB minified `pyodide.asm.js` source instead of the actual Python traceback, and GitHub Actions' log collector then truncates everything after the blob. The real error never reaches the build log.

This PR adds a `.catch()` that logs `err.stack` and exits 1, so future failures surface the actual traceback.

## Reproduction

The first commit on this branch is an empty commit off master. It reproduces the original failure on a clean master-based PR — demonstrating that the `build_and_test_streamlit_pyodide` failure is repo-wide (not caused by any specific feature PR).

With the `.catch()` change applied, the failing step log prints the actual error instead of a minified-JS dump:

```
Pyodide test failed: PythonError: Traceback (most recent call last):
  ...
  File ".../micropip/transaction.py", line 281, in find_wheel
```

Root cause (addressed in the separate authlib-pin PR): `authlib 1.7.0` added `joserfc` as a transitive dependency; `joserfc 1.6.4` requires `cryptography>=45.0.1`, which pyodide 0.26 (used by Fusion Streamlit / stlite) cannot satisfy.

## Test plan
- [ ] `build_and_test_streamlit_pyodide` still fails on this PR (expected — this change only improves error surfacing, not the underlying cause). Once the authlib pin PR lands on master, this PR rebased on master should go green.

### DIFF
--- a/scripts/test-pyodide.js
+++ b/scripts/test-pyodide.js
@@ -37,8 +37,21 @@ server.listen(PORT, () => {
     let pyodide = await loadPyodide();
     await pyodide.loadPackage("micropip");
     const micropip = pyodide.pyimport("micropip");
-    // Read packages to install from environment variable as JSON
 
+    // TEMPORARY WORKAROUND (added 2026-04-19, auto-disables 2026-05-03):
+    // authlib 1.7+ requires cryptography>=45.0.1, which has no pure-Python wheel;
+    // the streamlit-pinned Pyodide 0.26.2 only ships cryptography 43.x, so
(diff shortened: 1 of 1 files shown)


haakonvt (dismissed):
:unicorn: great!

haakonvt:
🦄

doctrino (approved):

PR #1892, opened 2024-08-28
Refactor OAuthDeviceCode to support non-Entra IdPs

gregertw:
## Description

1. Added support for non-Microsoft Entra identity providers:

* New oauth_discovery_url parameter for standard OAuth discovery URLs
* Disabled instance discovery and authority validation for non-Entra authorities

2. Enhanced token caching:

* Added clear_cache option to clear existing token cache
* Introduced mem_cache_only flag for in-memory-only token caching

3. Improved token refresh logic:

* Now checks for valid refresh tokens before initiating a new device code flow

4. Added new convenience methods:

* OAuthDeviceCode.default() for generic identity providers
* OAuthDeviceCode.default_for_azure_ad() for Azure AD with pre-configured settings

5. Updated error handling and user prompts for device code flow

## Checklist:
- [ ] Tests added/updated.
- [ ] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
  If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.
- [ ] Changelog updated in [CHANGELOG.md](https://github.com/cognitedata/cognite-sdk-python/blob/master/CHANGELOG.md).
- [ ] Version bumped. If triggering a new release is desired, bump the version number in [_version.py](https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_version.py) and [pyproject.toml](https://github.com/cognitedata/cognite-sdk-python/blob/master/pyproject.toml) per [semantic versioning](https://semver.org/).

PR #2585, opened 2026-04-21
fix(credentials): reuse OAuthDeviceCode token cache across processes

ronpal:
## Description

`OAuthDeviceCode._refresh_access_token` had two bugs that together made the on-disk MSAL token cache unusable across process boundaries — the device code prompt appeared on every CLI invocation (e.g. every `cdf modules upgrade` call), even seconds apart.

**Bug 1 — refresh token never used.** The RT branch was guarded by `expires_on > time.time()`, but MSAL's `SerializableTokenCache` doesn't store `expires_on` on refresh-token entries, so the condition was always False and the cached RT was silently skipped.

**Bug 2 — cached access token expiry ~1s (walrus precedence).** `if expiry := int(...) - time.time() > 0:` parses as `expiry := ((a - b) > 0)`, so `expires_in` became a `bool` and the token was treated as expiring in ~1 second, causing `__should_refresh_token` to fire on every API call.

## Fix

- Drop the `expires_on` guard and try any RT found in the cache; if the IdP rejects it, fall through to the device code flow (matches how MSAL handles revoked RTs).
- Split the walrus into two statements so `expires_in` is the actual remaining seconds.

Change is scoped to `OAuthDeviceCode._refresh_access_token`; public API and the device code fallback from #1892 / #2394 are unchanged. See the reviewer-notes comment for more detail.

## Tests

- Fixed `test_refresh_token_from_cache` to match MSAL's real cache shape (no `expires_on` on RT) — this is what masked bug 1 originally.
- Added `test_refresh_token_invalid_falls_back_to_device_flow` and `test_cached_access_token_expiry_is_seconds_not_bool` as direct regression tests.

All 47 tests in `tests/tests_unit/test_credential_providers.py` pass.

## Checklist:
- [x] Tests added/updated.
- [x] Documentation updated. Documentation is generated from docstrings — no docstring changes required for this bug fix.


ronpal:
### Reviewer notes (context for the code changes)

The fix is intentionally minimal — just a few lines — so I pulled the explanatory comments out of the diff and am leaving them here instead:

**Refresh-token branch (bug 1).** MSAL's `SerializableTokenCache` does not persist `expires_on` on refresh-token entries (see `msal/token_cache.py`'s `__add`: only `credential_type, secret, home_account_id, environment, client_id, target, last_modification_time` are stored). Guarding the RT lookup with `expires_on > now` therefore always evaluated to False and the cached refresh token was silently skipped on every process start. The new code just tries any RT it finds; if the IdP rejects it (no `access_token` in the response), we fall through to the device code flow, which matches how MSAL itself handles revoked refresh tokens.

**Access-token branch (bug 2).** The previous code was `if expiry := int(...) - time.time() > 0:`. Python operator precedence parses this as `expiry := ((a - b) > 0)`, so `expiry` was a `bool`, not the remaining seconds. `credentials["expires_in"] = True` then became `time.time() + 1.0` downstream, i.e. the cached access token was treated as expiring in ~1s and `__should_refresh_token` fired on every subsequent API call within the process. The fix splits it into two statements.

**Tests.** `test_refresh_token_from_cache` was previously putting `expires_on` on the RT mock, which didn't reflect the real on-disk cache shape and masked bug 1 — updated to omit it and also assert the device flow is not triggered. Two new tests (`test_refresh_token_invalid_falls_back_to_device_flow`, `test_cached_access_token_expiry_is_seconds_not_bool`) pin down the two behaviours directly.

ronpal:
Addressed @gemini-code-assist's review in c924e567:

- **`break` inside `if "access_token" in result:`** — fixed. We'll now keep trying additional cached refresh tokens if the first one is revoked before falling back to the device code flow.
- **Scope the RT / AT cache searches by `client_id` (and `target` for AT)** — fixed, to avoid picking up tokens belonging to a different app or scope from a shared MSAL cache.
- **`assert "Authorization", "Bearer ..." == ...` always-truthy tuple** — fixed in the two tests added in this PR. The same pattern appears on ~10 pre-existing assertions in `test_credential_providers.py`; leaving those for a separate cleanup PR to keep this one scoped to the cache bug.

tests/tests_unit/test_credential_providers.py
```diff
@@ -353,9 +349,54 @@ def test_refresh_token_from_cache(self, mock_public_client: MagicMock) -> None:
...
+        mock_public_client().client.obtain_token_by_device_flow.return_value = {
+            "access_token": "fresh_access_token",
+            "expires_in": 3600,
+        }
+
+        creds = OAuthDeviceCode(**self.DEFAULT_PROVIDER_ARGS)
+        creds._refresh_access_token()
+
+        mock_public_client().client.obtain_token_by_refresh_token.assert_called_once_with("stale_refresh_token")
+        mock_public_client().http_client.post.assert_called_once()
```
polomani: I think it's better to check `assert_called_once_with` or match url when mocking

cognite/client/credentials.py
```diff
@@ -467,18 +467,24 @@ def _refresh_access_token(self) -> tuple[str, float]:
...
         # - A valid refresh token, and if so, use it automatically to redeem a new access token.
         credentials = None
-        for token in self.__app.token_cache.search(self.__app.token_cache.CredentialType.REFRESH_TOKEN):
-            if "expires_on" in token and token["expires_on"] > time.time():
-                credentials = token
+        for token in self.__app.token_cache.search(
+            self.__app.token_cache.CredentialType.REFRESH_TOKEN,
+            query={"client_id": self.client_id},
+        ):
+            result = self.__app.client.obtain_token_by_refresh_token(token.get("secret", ""))
```
polomani: obtain_token_by_refresh_token can throw exceptions, what should be done in that case? loop further?

tests/tests_unit/test_credential_providers.py
ronpal: Agreed!

cognite/client/credentials.py
ronpal: Nice catch!

cognite/client/credentials.py
```diff
@@ -471,7 +471,10 @@ def _refresh_access_token(self) -> tuple[str, float]:
             self.__app.token_cache.CredentialType.REFRESH_TOKEN,
             query={"client_id": self.client_id},
         ):
-            result = self.__app.client.obtain_token_by_refresh_token(token.get("secret", ""))
+            try:
+                result = self.__app.client.obtain_token_by_refresh_token(token.get("secret", ""))
```
polomani: shouldn't we delete token from cache if it's invalid now? previously we wouldn't call `obtain_token_by_refresh_token`, but now this this in the loop it'll get expensive looping through potentially a few outdated tokens

ronpal: True! Updated now, so three possible outcomes each iteration:

1. Exception raised → except catches it, continue jumps to next token — remove_rt never reached.
2. Server returns a valid token ("access_token" in result) → break exits the loop — remove_rt never reached.
3. Server returns an error (e.g. {"error": "invalid_grant"}, no "access_token") → neither except nor break fires, so execution falls through to remove_rt(token), then the loop continues to the next token.

so we remove_rt is only reachable in case 3.

polomani (dismissed):
LGTM

cognite/client/credentials.py
```diff
@@ -467,18 +467,28 @@ def _refresh_access_token(self) -> tuple[str, float]:
         # - A valid access token.
         # - A valid refresh token, and if so, use it automatically to redeem a new access token.
         credentials = None
```
haakonvt, line 470: Is the 3-line comment here still correct?

tests/tests_unit/test_credential_providers.py
```diff
@@ -353,9 +349,86 @@ def test_refresh_token_from_cache(self, mock_public_client: MagicMock) -> None:
...
+
+        mock_device_response = Mock()
+        mock_device_response.json.return_value = {
+            "user_code": "ABCDEF",
+            "message": "Follow the link and enter the code",
+        }
+        expected_device_auth_url = "https://login.microsoftonline.com/xyz/oauth2/v2.0/devicecode"
+        mock_public_client().authority.device_authorization_endpoint = expected_device_auth_url
+        mock_public_client().http_client.post.return_value = mock_device_response
+        mock_public_client().client.obtain_token_by_device_flow.return_value = {
```
haakonvt: Although correct usage, it looks a bit weird for untrained mock-eyes (that calling returns the same instance each time). I would suggest you bind it to a local variable:

```py
mock_pca_instance = mock_public_client()

mock_pca_instance.token_cache.search.side_effect = search_side_effect
mock_pca_instance.client.obtain_token_by_refresh_token.return_value = { ... }
# ... later ...
mock_pca_instance.client.obtain_token_by_refresh_token.assert_called_once_with(...)
```

cognite/client/credentials.py
ronpal: Good catch — replaced with a one-liner that matches the actual flow in 87cdff01.

tests/tests_unit/test_credential_providers.py
ronpal: Done in 87cdff01 — bound `mock_public_client()` to `mock_pca_instance` in the four tests touched by this PR. Left the other tests in the file untouched to keep the diff focused on the bug fix.

haakonvt (changes requested):
Not solving the current toolkit issue

haakonvt:
Merged in #2602

PR #2586, opened 2026-04-22
feat(Data Modeling): Add /sync utility method to store cache in Files API (DM-3759)

doctrino:
## Description

Adds the utility function `sync_with_file_cache`. 

From docstring:

```python
>>> from cognite.client.data_classes.data_modeling.query import (
...     QuerySync,
...     SelectSync,
...     NodeResultSetExpressionSync,
... )
>>> from cognite.client._api.data_modeling.instances import FileCacheConfig
>>> from cognite.client.data_classes.data_modeling.ids import ViewId
>>> client = CogniteClient()
>>> pump_view = ViewId("mySpace", "Pump", "v1")
>>> query = QuerySync(
...     with_={"pumps": NodeResultSetExpressionSync()},
...     select={"pumps": SelectSync()},
... )
>>> cache_config = FileCacheConfig(
...     external_id_prefix="pump_sync_cache",
...     security_category_name="my_security_category",
... )
>>> # First call: syncs all pumps and saves data to file
>>> result = client.data_modeling.instances.sync_with_file_cache(
...     query, cache_config=cache_config
... )
>>> # Subsequent calls: returns all instances (cached + new changes)
>>> result2 = client.data_modeling.instances.sync_with_file_cache(
...     query, cache_config=cache_config
... )
```

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/data_modeling/instances.py
+++ b/cognite/client/_api/data_modeling/instances.py
@@ -41,6 +41,7 @@ from cognite.client.data_classes.data_modeling.instances import (
     EdgeApplyResult,
     EdgeApplyResultList,
     EdgeList,
+    EdgeListWithCursor,
     InstanceAggregationResultList,
     InstanceInspectResultList,
     InstanceInspectResults,
@@ -55,6 +56,7 @@ from cognite.client.data_classes.data_modeling.instances import (
     NodeApplyResult,
     NodeApplyResultList,
     NodeList,
+    NodeListWithCursor,
     T_Edge,
     T_Node,
     TargetUnit,
@@ -1835,11 +1837,16 @@ class InstancesAPI(APIClient):
         include_typing: bool = False,
         debug: DebugParameters | None = None,
     ) -> QueryResult:
-        """Sync instances with cursor state cached in a CDF file.
+        """Sync instances with cursor state and instances cached in a CDF file.
 
-        This is a utility method that combines the sync endpoint with persistent cursor
-        storage in CDF Files. The cursor state is automatically saved after each successful
-        sync and restored on subsequent calls, enabling incremental syncs across sessions.
+        This is a utility method that combines the sync endpoint with persistent
+        storage of both cursor state and instance data in CDF Files. The method:
+        1. Loads cached cursors and instances from the file (if available)
+        2. Performs an incremental sync using the cached cursors
+        3. Merges the sync result with cached instances (adding/updating new items,
+           removing items with deleted_time set)
+        4. Saves the merged data and updated cursors back to the cache file
+        5. Returns the full merged result (all instances, not just incremental changes)
 
         The cache file external_id is generated as `{cache_config.external_id_prefix}_{query_hash}`
         where query_hash is derived from the query structure (or just `{query_hash}` if the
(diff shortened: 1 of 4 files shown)


cognite/client/_api/data_modeling/instances.py
```diff
@@ -1746,6 +1835,289 @@ async def _query_or_sync(
...
+        if cache_config.security_category is not None:
+            return [cache_config.security_category]
+
+        if cache_config.security_category_name is None:
+            return None
+
+        security_categories = await self._cognite_client.iam.security_categories.list(limit=-1)
+        for sc in security_categories:
+            if sc.name == cache_config.security_category_name:
+                return [sc.id]
```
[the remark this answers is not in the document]
doctrino: No, that is another cache to maintain.

erlendvollset:
Nice! I think we should go for sqlite from the get-go. Having a json mode is redundant if we have sqlite and it has multiple benefits: don't need to fit everything into memory, much more expressive queries, can have indexes.

We should also make it very clear in the docs that `sync_with_file_cache` needs to be invoked sufficiently frequently. Cursors expire after 3 days, but with sufficient amounts of changes the sync query can start timing out long before that.

chessnocow:
> with sufficient amounts of changes the sync query can start timing out long before that

How does it apply to the first run @erlendvollset ? We need to get the full load from time to time at least

erlendvollset:
> > with sufficient amounts of changes the sync query can start timing out long before that
> 
> How does it apply to the first run @erlendvollset ? We need to get the full load from time to time at least

This needs to use a [two-phase sync](https://docs.cognite.com/cdf/dm/dm_concepts/dm_querying#sync-modes), where the backfill phase can use a different index to efficiently fetch all historical data on the first run.

chessnocow:
> This needs to use a [two-phase sync](https://docs.cognite.com/cdf/dm/dm_concepts/dm_querying#sync-modes), where the backfill phase can use a different index to efficiently fetch all historical data on the first run.

That should be implemented as a minimal functionality, I guess. @doctrino

cognite/client/_sync_api/data_modeling/instances.py
```diff
@@ -1236,6 +1236,99 @@ def sync(self, query: QuerySync, include_typing: bool = False, debug: DebugParam
...
+                ...     QuerySync,
+                ...     SelectSync,
+                ...     NodeResultSetExpressionSync,
+                ... )
+                >>> from cognite.client._api.data_modeling.instances import FileCacheConfig
+                >>> from cognite.client.data_classes.data_modeling.ids import ViewId
+                >>> client = CogniteClient()
+                >>> pump_view = ViewId("mySpace", "Pump", "v1")
+                >>> query = QuerySync(
+                ...     with_={"pumps": NodeResultSetExpressionSync()},
```
alexbrasetvik, line 1310: Let's make the examples use `mode: twoPhase` by default. https://api-docs.cognite.com/20230101/tag/Instances/operation/syncContent#!path=with/property%20name*/0/mode&t=request

If we shipped the sync API new today, that'd be the default behaviour.

haakonvt: Made a separate PR for this, and will rebase/update this one afterwards: #2595

haakonvt:
I have refactored this implementation into a context manager + a bit more strict design here: https://github.com/cognitedata/cognite-sdk-python/pull/2599

haakonvt:
Now available in `8.4.0` 🚀

PR #2588, opened 2026-04-23
fix(raw): limit=-1 silently returned only the first chunk of rows (DM-3733)

haakonvt:
## Bug summary
`limit=-1` (and `math.inf`) are documented as "return everything" sentinels, equivalent to `None`. In the concurrent row fetcher (`_list_generator_concurrent`), the consumer loop checked `if limit is None` — so `-1` fell into the finite-limit branch where `chunk[:-1]` dropped the last row and triggered an early exit.

**Fix:** normalise `-1`/`inf` to `None` at the top of `_list_generator_concurrent`. Adds a parametrized regression test covering all three unlimited sentinels.

Found by @dmhv


amorken (approved):
risk review ok or not ok who knows but the bot should be happy

cognite/client/_api/raw/rows.py
```diff
@@ -226,7 +226,7 @@ async def _fetch_and_enqueue(iterator: AsyncIterator[RowList]) -> None:
                     await asyncio.sleep(0.5)
                     continue
 
-                if limit is None:
+                if not is_finite(limit):
```
amorken, line 229: on second thought, why not `is_unlimited(limit)` ?

haakonvt: Making mypy happy: is_finite is a TypeGuard[int], so mypy narrows limit to int in the arithmetic on lines 234 and 238. is_unlimited has no such guard, so those lines would still fail type checking.

PR #2594, opened 2026-04-27
fix(datapoints): restore backpressure in insert_multiple to fix memory regression vs v7

haakonvt:
In v8, all insert tasks are created at once via asyncio.create_task(). Because dp.dump() is synchronous (no await), every task serialized its full payload before any of them could block on the write semaphore, causing all datapoints to materialize in memory simultaneously.

Fix: acquire the write semaphore before serialization so that payload preparation is throttled to the same concurrency as the HTTP requests (default: 5), matching v7 behavior where the ThreadPoolExecutor(max_workers=5) provided this naturally.


haakonvt:
🦄

erlendvollset (approved):

audunska (approved):
Great catch!

haakonvt:
> Great catch!

Gotta give kudos to customer Lyse 😅

PR #2595, opened 2026-04-27
feat(Data Modeling): change data modeling /sync default mode to two_phase

haakonvt:
Changes the default sync mode to `two_phase`. Was supposed to be part of the v8 release but forgotten sometime during parental leave.

Is this breaking? Yes, but...
1. Mixpanel usage data still shows very low adoption from v7 to v8
2. Unknowing users should only see a performance improvement during backfill phase, with no real downside apart from maybe 1 empty response body during switching from backfill to "live".


cognite/client/data_classes/data_modeling/query.py
```diff
@@ -497,7 +495,7 @@ class NodeResultSetExpressionSync(ResultSetExpressionSync):
...
         skip_already_deleted (bool): If set to False, the API will return instances that have been soft deleted before sync was initiated. Soft deletes that happen after the sync is initiated and a cursor generated, are always included in the result. Soft deleted instances are identified by having deletedTime set.
-        sync_mode (Literal['one_phase', 'two_phase', 'no_backfill'] | None): Specify whether to sync instances in a single phase; in a backfill phase followed by live updates, or without any backfill. Only valid for sync operations.
+        sync_mode (Literal['one_phase', 'two_phase', 'no_backfill']): Specify whether to sync instances in a single phase (slow); in a backfill phase followed by live updates (fast), or without any backfill. Defaults to "two_phase".
```
audunska: Not sure we can promise that it will unconditionally be fast!

audunska (dismissed):
I think this is hardly a breaking change at all. If people are syncing, they need to be live-tailing after non-full batches anyway.

cognite/client/data_classes/data_modeling/query.py
alexbrasetvik: I think we shouldn't have slow or fast in there.

cognite/client/_api/data_modeling/instances.py
```diff
@@ -874,7 +874,11 @@ async def subscribe(
                 >>> view_id = ViewId("someSpace", "someView", "v1")
                 >>> filter = Equals(view_id.as_property_ref("myAsset"), "Il-Tempo-Gigante")
                 >>> query = QuerySync(
-                ...     with_={"work_orders": NodeResultSetExpressionSync(filter=filter)},
+                ...     with_={
+                ...         "work_orders": NodeResultSetExpressionSync(
+                ...             filter=filter, sync_mode="two_phase"
```
audunska, line 879: Why include the explicit default here?

audunska (approved):
Very nice!

cognite/client/_api/data_modeling/instances.py
haakonvt: Maximize future LLM knowledge 😅

haakonvt:
🦄

PR #2599, opened 2026-04-30
feat(Data Modeling): add alpha `instances.sync_with_file_cache` for /sync jobs with file cache (DM-3759)

haakonvt:
An evolution of https://github.com/cognitedata/cognite-sdk-python/pull/2586, adding or changing a few things:

- **Context manager pattern** — `SyncSessionWithCache` holds state across multiple syncs; first PR did a full load/sync/save roundtrip on every call
- **Fewer CDF file API calls** — state is kept in memory and only written on a timer (`backup_every`) or on exiting from the context manager, instead of every call
- **`sync_until_live()`** — built-in loop to catch up with all pending changes; previously up to the caller to handle
- **Typed accessors** — `get_nodes()` / `get_edges()`, lifted from `QueryResults`.
- **`invalidate()`** — explicit reset to trigger a fresh backfill
- **Stricter permissions** — only security category allowed (and enforced)

Usage examples from docstring:

```py
# One-off job that loads state from the Files API, syncs until it has caught up with all
# live changes, then does some work that require huge amounts of instance data, then backs
# up the progress on exit (from the context manager):

from cognite.client import AsyncCogniteClient
from cognite.client.data_classes.data_modeling.query import (
    QuerySync,
    SelectSync,
    NodeResultSetExpressionSync,
)
client = AsyncCogniteClient()
query = QuerySync(
    with_={"my_nodes": NodeResultSetExpressionSync(limit=1000)},
    select={"my_nodes": SelectSync()},
)
session = client.data_modeling.instances.sync_with_file_cache(
    query,
    file_external_id="my_backup_file",
    security_category=123,
    backup_every=None,  # Only backup on exit
    backup_on_exit=True,
)
def do_work(nodes: NodeList) -> None:
    print(len(nodes))  # ¯\_(ツ)_/¯

async with session:
    await session.sync_until_live()
    do_work(session.get_nodes("my_nodes"))

# Longer-running job with periodic backups, e.g. regularly computing statistics for a
# dashboard:

import asyncio
session = client.data_modeling.instances.sync_with_file_cache(
    query,
    file_external_id="my_backup_file",
    security_category=123,
    backup_every=timedelta(minutes=15),
)
async with session:
    while True:
        await session.sync_until_live()
        do_work(session.get_nodes("my_nodes"))
        await asyncio.sleep(60)
```

### DIFF
--- a/cognite/client/_api/data_modeling/instances.py
+++ b/cognite/client/_api/data_modeling/instances.py
@@ -5,7 +5,7 @@ import inspect
 import logging
 import random
 from collections.abc import AsyncIterator, Awaitable, Callable, Iterable, Sequence
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -68,10 +68,10 @@ from cognite.client.data_classes.data_modeling.query import (
     QuerySync,
     SourceSelector,
 )
-from cognite.client.data_classes.data_modeling.sync import SubscriptionContext
+from cognite.client.data_classes.data_modeling.sync import SubscriptionContext, SyncSessionWithCache
 from cognite.client.data_classes.data_modeling.views import View
 from cognite.client.data_classes.filters import _BASIC_FILTERS, Filter, _validate_filter
-from cognite.client.utils._auxiliary import is_unlimited, load_yaml_or_json, unpack_items
+from cognite.client.utils._auxiliary import at_least_one_is_not_none, is_unlimited, load_yaml_or_json, unpack_items
 from cognite.client.utils._experimental import FeaturePreviewWarning
 from cognite.client.utils._identifier import DataModelingIdentifierSequence
 from cognite.client.utils._retry import Backoff
@@ -175,6 +175,12 @@ class InstancesAPI(APIClient):
             sdk_maturity="alpha",
             feature_name="Data modeling debug parameters 'includeTranslatedQuery' and 'includePlan'",
             pluralize=True,
+        )
+        self._warn_on_sync_with_file_cache = FeaturePreviewWarning(
+            api_maturity="General Availability",
+            sdk_maturity="alpha",
+            feature_name="Sync sessions with file caching for the Instances API",
+            pluralize=False,
         )
 
     def _get_semaphore(self, operation: Literal["read", "write", "delete", "search"]) -> asyncio.BoundedSemaphore:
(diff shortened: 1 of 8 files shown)


cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -72,6 +73,9 @@
 
     from cognite.client.data_classes.data_modeling.debug import DebugInfo
 
+
+logger = logging.getLogger(__name__)
```
doctrino: This is unused

cognite/client/_api/data_modeling/instances.py
```diff
@@ -1776,6 +1782,136 @@ async def _query_or_sync(
...
+
+            If you *really* don't care about missing deleted instances, there's
+            ``allow_expired_cursors_and_accept_missed_deletes=True`` on :class:`QuerySync`
+            that will allow you to use an older cursor. **Be careful**, you can not change this
+            setting mid-session, as the hash of your query would no longer match the existing,
+            and a full backfill would be triggered.
+
+        Warning:
+            This functionality is in **alpha** and only the async context manager interface is
+            currently available. The API and behaviour may change without notice.
```
erlendvollset, line 1818: I see a version in `_sync_api`. Is this outdated?

cognite/client/data_classes/data_modeling/sync.py
```diff
@@ -25,3 +40,206 @@ def is_running(self) -> bool:
 
     def is_alive(self) -> bool:
         return self.is_running()
+
+
+class SyncSessionWithCache(AbstractAsyncContextManager["SyncSessionWithCache"]):
```
erlendvollset, line 43: Should this live in a dedicated module instead of in `data_classes`? It's not a data class.

cognite/client/data_classes/data_modeling/sync.py
```diff
@@ -25,3 +40,206 @@ def is_running(self) -> bool:
...
+    async def sync(self) -> None:
+        """Perform one incremental sync, updating the internal snapshot state."""
+        await self._sync_one()
+
+    async def sync_until_live(self) -> None:
+        """Call :meth:`sync` repeatedly until we have caught up with all live changes."""
+        while True:
+            result = await self._sync_one()
+            # We have enforcement on limit being given (although mypy doesn't know that):
+            if all(len(result[k]) < self._query.with_[k].limit for k in result):  # type: ignore[operator]
```
erlendvollset: If this is a two-phase sync, DMS can return a batch with fewer than `limit` instances when the backfill phase completes. That doesn't mean we're live-tailing yet, since more data can have been ingested since the backfill started. We essentially need to hit two batches with fewer than `limit` instances.

cognite/client/_api/data_modeling/instances.py
```diff
@@ -1776,6 +1782,136 @@ async def _query_or_sync(
...
+        Args:
+            query (QuerySync): The sync query.
+            file_external_id (str): External ID of the CDF file used as the durable backup store.
+                The file is created automatically on the first backup if it does not yet exist.
+            security_category (int): The security category to apply to the backup file. The file is
+                created (and re-uploaded on each backup) with this security category set, preventing
+                users who lack access to the underlying instances from reading the backup directly.
+            backup_every (timedelta | None): How often to upload state to CDF during a
+                session (when active). ``None`` uploads only on context-manager exit.
+            backup_on_exit (bool): Whether to upload state to CDF on context-manager exit.
```
erlendvollset, line 1829: Do we ever want to set both `backup_every` and `backup_on_exit=true`, or should we just disallow that?

cognite/client/_api/data_modeling/instances.py
```diff
@@ -1776,6 +1782,136 @@ async def _query_or_sync(
...
+                "The 'query' must not have cursors set. The session manages cursors internally "
+                "and restores them from the CDF backup file on entry."
+            )
+        if missing := [k for k in query.select if query.with_[k].limit is None]:
+            raise ValueError(
+                f"All result-set expressions referenced in 'select' must have an explicit 'limit'. Missing: {missing}"
+            )
+        categories = await self._cognite_client.iam.security_categories.list(limit=None)
+        if security_category not in categories.as_ids():
+            raise ValueError(f"Security category {security_category!r} does not exist in this project.")
```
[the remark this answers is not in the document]
haakonvt: @gemini-code-assist nice retrieve method you got there, too bad it doesnt exist

cognite/client/_api/data_modeling/instances.py
haakonvt: It is available on both clients, but the session returned has to be entered as an async context manager.

cognite/client/data_classes/data_modeling/sync.py
haakonvt: Right... I don't actually know how we can solve this without peeking inside the cursor 🤔

cognite/client/_api/data_modeling/instances.py
haakonvt: I think it is fine to allow both. We can consider changing the default `backup_on_exit` to `False` however

cognite/client/data_classes/data_modeling/sync.py
erlendvollset: Can't we just loop until we've hit two batches fewer than limit instances?

haakonvt: How would you know that after restoring from cache, that may or may not have finished the backfill phase?

haakonvt: I guess we would have to store `"sync_mode"` and `"backfill_completed"` in the cache 🤔

alexbrasetvik: Two subsequent empty responses would be a pretty safe and cheap bet, and we can have a conversation on whether we expose a "caught_up" thing in the future. I'm specifically thinking about the "Here's a sync cursor with progress even though the batch is empty, but we're not through yet" thing which can be handy if you're way behind on sync and a lot of unrelated changes have happened since.

haakonvt: @alexbrasetvik that is one excellent idea!

haakonvt: (If that was what you meant Erlend I'm sorry 😂 )

cognite/client/data_classes/data_modeling/sync.py
haakonvt: I was unsure - but since your sync-context-class for the subscribe method lives here I thought I'd just use it as well.

erlendvollset: SGTM.

erlendvollset (approved):
LGTM

haakonvt:
🦄

PR #2600, opened 2026-04-30
fix: Transformation notification response class

doctrino:
## Description

`transfomrationExternalId` is not required in the response:
image

Seems like the test that should have caught this was take out by an inexperienced developer 😝: https://github.com/cognitedata/cognite-sdk-python/pull/888. I am guessing that the service was instable, and the transformation team did not follow up. 

## Checklist:
- [x] Tests added/updated.


cognite/client/data_classes/transformations/notifications.py
```diff
@@ -79,6 +79,7 @@ def as_write(self) -> TransformationNotificationWrite:
         return TransformationNotificationWrite(
             destination=self.destination,
             transformation_id=self.transformation_id,
+            transformation_external_id=self.transformation_external_id,
```
doctrino:
```suggestion
```

haakonvt (approved):
🦄

PR #2606, opened 2026-05-06
fix(workflows): route unlisted tasks through unknown parameter/output types

JustAnotherNormalDev:
Problem

Listing workflow versions and loading executions could fail when the API returns task types that are not first-class in the SDK (for example new or beta task kinds in responses).

Solution

Route unlisted task types through UnknownWorkflowTaskParameters and UnknownWorkflowTaskOutput so the wire JSON is preserved and load no longer throws. Keep known task types on their existing dedicated classes.

UnknownWorkflowTaskParameters is constructed from the task type / taskType plus the parameters / input blob (not the full task envelope), so WorkflowTask.dump() stays aligned with the API shape.

For generic CogniteResource JSON/YAML round-trips, dump() must always yield a top-level dict. When _parameters is not a dict (for example a string from the fake resource generator in unit tests), dump() wraps the value in a single-key dict using _OPAQUE_WRAPPER_KEY ("__cogniteSdkUnknownWorkflowTaskParametersOpaque__"). That satisfies CogniteResource.load → load_resource_to_dict, which only accepts parsed JSON/YAML that resolves to a dict. _load detects that exact one-key shape and unwraps the inner value so the in-memory payload matches what was stored before serialization. Normal API dict payloads are returned unchanged and never use this wrapper.

Files

cognite/client/data_classes/workflows.py — unknown task parameters/output handling, load_parameters wiring, _OPAQUE_WRAPPER_KEY wrap/unwrap for non-dict unknown parameter bodies.
cognite/client/data_classes/__init__.py — export surface for updated workflow types.
tests/tests_unit/test_data_classes/test_workflows.py — coverage for unknown task serialization and related behavior.
tests/tests_unit/test_base.py — UnknownWorkflowTaskParameters included in generic CogniteResource round-trip tests where applicable.
cognite/client/_cognite_client.py — generated/synced client updates.
cognite/client/_sync_cognite_client.py — generated/synced sync client updates.
scripts/sync_client_codegen/sync_client_template.txt — codegen template updates matching generated client changes.


cognite/client/data_classes/workflows.py
```diff
@@ -257,6 +267,63 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        type_ = resource.get("type", resource.get("taskType"))
+        inner = resource.get("parameters", resource.get("input")) or {}
+        return cls(type_ if isinstance(type_, str) else "unknown", inner)
+
+    @property
+    def task_type(self) -> str:  # type: ignore[override]
+        return self.dynamic_task_type
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        return convert_all_keys_recursive(self._parameters, camel_case=camel_case)
```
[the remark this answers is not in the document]
JustAnotherNormalDev: taken care

cognite/client/data_classes/workflows.py
```diff
@@ -664,6 +733,42 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+    def __init__(self, output: dict[str, Any]) -> None:
+        self._output = output
+
+    @classmethod
+    def load(cls, data: dict[str, Any]) -> Self:
+        raw = data.get("output")
+        return cls(raw if isinstance(raw, dict) else {})
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        return convert_all_keys_recursive(self._output, camel_case=camel_case)
```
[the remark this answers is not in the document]
JustAnotherNormalDev: taken care

cognite/client/data_classes/workflows.py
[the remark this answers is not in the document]
JustAnotherNormalDev: there is a validation. this is not relevant

cognite/client/data_classes/workflows.py
```diff
@@ -257,6 +267,66 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
         return output
 
 
+class FunctionAppTaskParameters(WorkflowTaskParameters):
+    task_type = "functionApp"
+
+    def __init__(
```
VerstraeteBert: Not the correct definition for the FunctionApp task, missing a few fields from the api spec. 

After verifying with @jonasblanc, they are only in private beta. We should probably not support this just yet, let's only keep the Unknown parameters for now.

VerstraeteBert (changes requested):

cognite/client/data_classes/workflows.py
JustAnotherNormalDev: I removed the function app now

VerstraeteBert (dismissed):
LGTM

cognite/client/data_classes/workflows.py
```diff
@@ -179,7 +187,7 @@ def load_parameters(cls, data: dict) -> WorkflowTaskParameters:
         elif type_ == "simulation":
             return SimulationTaskParameters._load(parameters)
         else:
-            raise ValueError(f"Unknown task type: {type_}. Expected {ValidTaskType}")
+            return UnknownWorkflowTaskParameters(type_ if isinstance(type_, str) else "unknown", parameters)
```
haakonvt: Please make it work so that we delegate all the loading logic to the unknown class:
```suggestion
            return UnknownWorkflowTaskParameters._load(parameters)
```

cognite/client/data_classes/workflows.py
```diff
@@ -257,6 +265,28 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        return cls(type_ if isinstance(type_, str) else "unknown", inner)
+
+    @property
+    def task_type(self) -> str:  # type: ignore[override]
+        return self.dynamic_task_type
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        return convert_all_keys_recursive(
+            self._parameters if isinstance(self._parameters, dict) else {},
+            camel_case=camel_case,
```
haakonvt: We have had incidents in the past due to usage of `convert_all_keys_recursive`, please don't use it. Dump will for the purposes of calling the API never use camel_case=False, so feel free to silently ignore that - possibly with a warning

cognite/client/data_classes/workflows.py
```diff
@@ -664,6 +694,28 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        self._output = output
+
+    @classmethod
+    def load(cls, data: dict[str, Any]) -> Self:
+        task_type = data.get("taskType", "unknown")
+        output = data.get("output")
+        return cls(task_type, output if isinstance(output, dict) else {})
+
+    @property
+    def task_type(self) -> str:  # type: ignore[override]
```
haakonvt: This should pretend to return a `ValidTaskType`

cognite/client/data_classes/workflows.py
```diff
@@ -828,8 +880,8 @@ def __init__(
         self.reason_for_incompletion = reason_for_incompletion
 
     @property
-    def task_type(self) -> ValidTaskType:
-        return self.input.task_type
+    def task_type(self) -> str:
+        return cast(str, self.input.task_type)
```
haakonvt: I think you should revert this

haakonvt (changes requested):
A few comments, otherwise looks good!

cognite/client/data_classes/workflows.py
JustAnotherNormalDev: taken care

cognite/client/data_classes/workflows.py
JustAnotherNormalDev: added a warning and ignored it.

cognite/client/data_classes/workflows.py
JustAnotherNormalDev: reverted

cognite/client/data_classes/workflows.py
JustAnotherNormalDev: Let me know if this works

cognite/client/data_classes/workflows.py
```diff
@@ -562,8 +598,8 @@ def __init__(
         self.depends_on = depends_on
 
     @property
-    def type(self) -> ValidTaskType:
-        return self.parameters.task_type
+    def type(self) -> str:
+        return cast(str, self.parameters.task_type)
```
haakonvt: Did you forget this one?

JustAnotherNormalDev: my bad. missed it.

VerstraeteBert (dismissed):

VerstraeteBert (dismissed):

cognite/client/data_classes/workflows.py
```diff
@@ -257,6 +267,32 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        return cast(ValidTaskType, self.dynamic_task_type)
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        if not camel_case:
+            warnings.warn(
+                "camel_case=False is ignored for UnknownWorkflowTaskParameters.dump(); API payloads use camelCase keys.",
+                UserWarning,
+                stacklevel=2,
+            )
+        params = self._parameters if isinstance(self._parameters, dict) else {}
```
haakonvt: You ignore parameters if it is not a dict. That means your init should say dict rather than Any?

cognite/client/data_classes/workflows.py
```diff
@@ -664,6 +700,32 @@ def dump(self, camel_case: bool = True) -> dict[str, Any]:
...
+        return cast(ValidTaskType, self.dynamic_task_type)
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        if not camel_case:
+            warnings.warn(
+                "camel_case=False is ignored for UnknownWorkflowTaskOutput.dump(); API payloads use camelCase keys.",
+                UserWarning,
+                stacklevel=2,
+            )
+        out = self._output if isinstance(self._output, dict) else {}
```
haakonvt: Same question here

haakonvt: I also dont think we should have any kind of such logic? The payload is unkown, so we should return what we got instantiated with

cognite/client/data_classes/workflows.py
JustAnotherNormalDev: agree. Its dict now. dict[str, Any]:

cognite/client/data_classes/workflows.py
JustAnotherNormalDev: i changed it to dict. But do you mean , remove the warn message as well?

haakonvt: No the warning is great! I mean, if self._output is not dict we ignore it in the output. I don't think we should

haakonvt: Reason: The unknown class should just accept what gets passed to it - maybe in the future you want this to be a list or whatever, then this would break

tests/tests_unit/test_base.py
haakonvt, line 1: If you are exempting your data classes from automatic testing, you need to add a better explanation please

JustAnotherNormalDev: please ignore this one. i reverted it back. Pushed it by mistake . I had added it for testing.

PR #2618, opened 2026-05-11
fix: deprecate public api_client property. warn on async api_client footgun on sync client

haakonvt:
The API client that is returned has only private methods. It does not make sense to keep when we have top-level methods to do arbitrary GET/POST/PUT etc. calls.


haakonvt:
🦄

audunska (approved):

vvb16 (approved):

PR #2621, opened 2026-05-12
feat(Data Modeling): support usedFor=record on containers

evertoncolling:
## Description
This PR adds support for creating and querying records containers via the SDK.

- Widen `ContainerApply.used_for` and `Container.used_for` to also accept `"record"`, aligning with the v1 API spec (`UsedFor` enum = `node | edge | record | all`). Exposes a new public `ContainerUsedFor` type alias.
- Add a `used_for` filter to `containers.list()` and `containers(...)` (iterate), forwarded as the `usedFor` query parameter on `GET /models/containers`. Accepts a single value or a sequence.

## Checklist:
- [x] Tests added/updated.


evertoncolling:
This one would close https://github.com/cognitedata/cognite-sdk-python/pull/2391 (which was based on v7)

cognite/client/data_classes/data_modeling/containers.py
```diff
@@ -226,9 +228,20 @@ def as_write(self) -> ContainerApplyList:
 
 
 class _ContainerFilter(CogniteFilter):
-    def __init__(self, space: str | None = None, include_global: bool = False) -> None:
+    def __init__(
+        self,
+        space: str | None = None,
+        include_global: bool = False,
+        used_for: ContainerUsedFor | Sequence[ContainerUsedFor] | None = None,
```
erlendvollset, line 236: I think we should default this to return all container types. If we pass `null` to the API, it will not return records containers. That was to ensure we didn't break clients that didn't support `usedFor: records` yet, the SDK now supports it, so there's no reason to not return them by default.

cognite/client/data_classes/data_modeling/containers.py
```diff
@@ -226,9 +228,20 @@ def as_write(self) -> ContainerApplyList:
...
+    ) -> None:
         self.space = space
         self.include_global = include_global
+        if used_for is None:
+            self.used_for: list[ContainerUsedFor] | None = None
+        elif isinstance(used_for, str):
+            self.used_for = [used_for]
+        else:
+            self.used_for = list(used_for)
```
erlendvollset: The final else block should not also make this a list, since it will be a Sequence then. Also best to raise an exception if we get an unexpected value.
```suggestion
        if used_for is None:
            self.used_for: list[ContainerUsedFor] | None = None
        elif isinstance(used_for, str):
            self.used_for = [used_for]
        elif isinstance(used_for, Sequence):
            self.used_for = used_for
        else:
            raise ValueError("Invalid value for 'used_for': {used_for}")
```

erlendvollset: Or maybe a TypeError is more appropriate, let's add a test in any case.

cognite/client/_sync_api/data_modeling/containers.py
```diff
@@ -171,6 +184,7 @@ def list(
         space: str | None = None,
         limit: int | None = DATA_MODELING_DEFAULT_LIMIT_READ,
         include_global: bool = False,
+        used_for: ContainerUsedFor | Sequence[ContainerUsedFor] | None = None,
```
erlendvollset, line 187: Same note on defaulting to list all container types.

cognite/client/_sync_api/data_modeling/containers.py
```diff
@@ -38,19 +39,30 @@ def __init__(self, async_client: AsyncCogniteClient) -> None:
...
+        limit: int | None = None,
     ) -> Iterator[ContainerList]: ...
 
     def __call__(
         self,
         chunk_size: int | None = None,
         space: str | None = None,
         include_global: bool = False,
+        used_for: ContainerUsedFor | Sequence[ContainerUsedFor] | None = None,
```
erlendvollset, line 65: Same note on defaulting to list all container types.

evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/0ef2e15a3424e244e094c354b52ef4e6a4058fa1 and https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/bfe0ca8db0e7e1e85282940745ed9a3abf57f3eb

cognite/client/_sync_api/data_modeling/containers.py
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/0ef2e15a3424e244e094c354b52ef4e6a4058fa1 and https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/bfe0ca8db0e7e1e85282940745ed9a3abf57f3eb

cognite/client/data_classes/data_modeling/containers.py
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/0ef2e15a3424e244e094c354b52ef4e6a4058fa1 and https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/bfe0ca8db0e7e1e85282940745ed9a3abf57f3eb

cognite/client/data_classes/data_modeling/containers.py
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/0ef2e15a3424e244e094c354b52ef4e6a4058fa1 and https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/bfe0ca8db0e7e1e85282940745ed9a3abf57f3eb

erlendvollset:
🦄

erlendvollset (dismissed):

cognite/client/_api/data_modeling/containers.py
```diff
@@ -246,6 +253,11 @@ async def list(
                 >>> # async_client = AsyncCogniteClient()  # another option
                 >>> container_list = client.data_modeling.containers.list(limit=5)
 
+            Filter containers by `used_for`:
+
+                >>> record_containers = client.data_modeling.containers.list(used_for="record")
+                >>> all_containers = client.data_modeling.containers.list(used_for=["all", "record"])
```
haakonvt: @evertoncolling Won't this exclude "node only" and "edge only" containers? If yes, then we can't name this `all_containers`, it's just too confusing imo.

Can we, in the SDK, go with `node_and_edge` instead?

evertoncolling: Yes, that's true!!!

Even I get confused by this from time to time. But I dont think it would be a good idea to "patch" the name of the filter in the SDK. I would rather align with the API, and document it clearly in the docstrings/comments.

I will fix this in a new commit :)

Irrespective of this PR, we may be able to improve this in the API in the future, paving a way for less confusion. That would need some good discussions, based on how do we see the evolution of DMS and Records, the recommended usage patterns, etc.

cognite/client/_api/data_modeling/containers.py
```diff
@@ -246,6 +253,17 @@ async def list(
...
+            Filter containers by `used_for`:
+
+                >>> # containers that can be used for records only
+                >>> record_only_containers = client.data_modeling.containers.list(used_for="record")
+                >>> # containers that can be used either for nodes or edges
+                >>> node_or_edge_containers = client.data_modeling.containers.list(
+                ...     used_for=["node", "edge"]
+                ... )
+                >>> # containers that can be used for both nodes and edges
+                >>> node_and_edge_containers = client.data_modeling.containers.list(used_for=["all"])
```
haakonvt:
```suggestion
            Filter containers by `used_for`. Note that "all" refers to containers that stores *both*
            nodes and edges:

                >>> # High-volume data containers (records)
                >>> record_containers = client.data_modeling.containers.list(used_for="record")
                >>> 
                >>> # Containers that store ONLY nodes or ONLY edges (excludes "all"):
                >>> containers = client.data_modeling.containers.list(
                ...     used_for=["node", "edge"]
                ... )
                >>> # All containers that can store nodes:
                >>> containers = client.data_modeling.containers.list(
                ...     used_for=["node", "all"]
                ... )
```

haakonvt (commented):
I have done my utmost to make it clear now 😅

cognite/client/_api/data_modeling/containers.py
evertoncolling: I like it! 

The options are still more complicated than I would like them to be, but given what we have to work with, I think your suggestion works well.

evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/c53e7d4e689f9d5cd8fca0881307c4a6c9c955d9

cognite/client/_api/data_modeling/containers.py
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-python/pull/2621/commits/c53e7d4e689f9d5cd8fca0881307c4a6c9c955d9

haakonvt (approved):
🦄

PR #2624, opened 2026-05-15
docs(simulators): improve type hints and default values for some arguments in the simulator module

MortGron:
## Description
Improve some type hints and default values in the simulators module. Some None type hints have been removed, but this should be ok since simulators is in alpha.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/cognite/client/_api/simulators/runs.py
+++ b/cognite/client/_api/simulators/runs.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import AsyncIterator, Sequence
-from typing import TYPE_CHECKING, overload
+from typing import TYPE_CHECKING, Literal, overload
 
 from cognite.client._api_client import APIClient
 from cognite.client._constants import DEFAULT_LIMIT_READ
@@ -47,8 +47,8 @@ class SimulatorRunsAPI(APIClient):
         self,
         chunk_size: int,
         limit: int | None = None,
-        status: str | None = None,
-        run_type: str | None = None,
+        status: Literal["queued", "ready", "running", "success", "failure"] | None = None,
+        run_type: Literal["external", "manual", "scheduled"] | None = None,
         model_external_ids: SequenceNotStr[str] | None = None,
         simulator_integration_external_ids: SequenceNotStr[str] | None = None,
         simulator_external_ids: SequenceNotStr[str] | None = None,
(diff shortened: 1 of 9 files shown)


cognite/client/_api/simulators/runs.py
```diff
@@ -47,8 +47,8 @@ def __call__(
         self,
         chunk_size: int,
         limit: int | None = None,
-        status: str | None = None,
-        run_type: str | None = None,
+        status: Literal["queued", "ready", "running", "success", "failure"] | None = None,
```
haakonvt: Consider somthing like this:
```py
SimulatorRunStatus = Literal....

        status: SimulatorRunStatus | None = None,
```

haakonvt (commented):
LGTM! Let's just do a quick check-in with @evertoncolling before we merge

cognite/client/data_classes/simulators/runs.py
```diff
@@ -149,7 +146,7 @@ def _load(cls, resource: dict[str, Any]) -> SimulationRunWrite:
         return cls(
             run_type=resource.get("runType"),
             run_time=resource.get("runTime"),
-            queue=resource.get("queue"),
+            queue=resource.get("queue", False),
```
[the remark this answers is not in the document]
haakonvt: Fields in API responses are either missing or have the correct type.

cognite/client/_api/simulators/runs.py
MortGron: For some reason I thougth type aliases would not be expanded in the docs, but they are, so then there is no reason not to do as you suggest
image

evertoncolling:
> LGTM! Let's just do a quick check-in with @evertoncolling before we merge

All looks good to me! Its a nice pattern to have the type hints match the API defaults

cognite/client/_sync_api/simulators/runs.py
```diff
@@ -74,8 +75,8 @@ def __call__(
         self,
         chunk_size: int | None = None,
         limit: int | None = None,
-        status: str | None = None,
-        run_type: str | None = None,
+        status: SimulatorRunStatus | None = None,
+        run_type: Literal["external", "manual", "scheduled"] | None = None,
```
haakonvt: Consider if you also want to SimulatorRunType

haakonvt (dismissed):
🦄

cognite/client/_sync_api/simulators/runs.py
MortGron: Done

haakonvt (approved):
🦄

PR #2632, opened 2026-05-20
fix: respect HTTP(S)_PROXY env vars for httpx client (DM-3823)

haakonvt:
## Summary

The v8 migration from `requests` to `httpx` introduced a custom `AsyncHTTPTransport` passed via `transport=` to `httpx.AsyncClient`. httpx only checks proxy env vars when no custom transport is provided (`allow_env_proxies = trust_env and transport is None`), so `HTTPS_PROXY` / `HTTP_PROXY` were silently ignored breaking users behind corporate proxies.

Fix: stop creating a custom transport and pass all settings (`proxy`, `verify`, `limits`) directly to `AsyncClient`. All behavior is preserved, env proxy detection works again.

### DIFF
--- a/cognite/client/_http_client.py
+++ b/cognite/client/_http_client.py
@@ -53,30 +53,17 @@ def get_global_async_httpx_client() -> httpx.AsyncClient:
         return _global_async_httpx_clients[loop]
     except KeyError:
         pass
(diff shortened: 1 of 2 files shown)


erlendvollset (approved):

haakonvt:
🦄

PR #2634, opened 2026-05-20
feat(agents): support expanded reasoning response in agent chat

ks93:
## Summary

- Adds the `data` field to `AgentReasoningItem`, which the Agents API now returns to expose tool-call details from the agent's reasoning trace
- Introduces `ToolCallDetail`, `ReasoningDataItem`, `ToolCallReasoningDataItem`, and `UnknownReasoningDataItem` following the same type-dispatch pattern as `MessageContent`/`TextContent`/`UnknownContent`
- All four new classes are exported from `cognite.client.data_classes.agents`

API reference: https://api-docs.cognite.com/20230101-beta/tag/Agents/operation/agent_session_ai_agents_chat_post#!c=200&path=response/2/messages/reasoning/data&t=response

## Test plan

Test manually locally.

- [ ] `pytest tests/tests_unit/test_api/test_agents_chat.py -v` -- all 11 tests pass
- [ ] Verify `data=None` is preserved when field is absent from the API response

### DIFF
--- a/cognite/client/data_classes/agents/chat.py
+++ b/cognite/client/data_classes/agents/chat.py
@@ -465,24 +465,129 @@ class AgentDataItem(CogniteResource):
 
 
 @dataclass
+class ToolCallDetail(CogniteResource):
+    """Details of a tool call made during agent reasoning.
+
+    Args:
+        id (str): The id of the tool call.
+        name (str): The name of the tool that was called.
+        tool_type (str): The type of the tool that was called.
+        input (dict[str, Any]): The parameters that were passed to the tool.
+        result (dict[str, Any]): The results that were returned by the tool.
+    """
+
+    id: str
+    name: str
+    tool_type: str
+    input: dict[str, Any] = field(default_factory=dict)
+    result: dict[str, Any] = field(default_factory=dict)
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        key = "toolType" if camel_case else "tool_type"
+        return {"id": self.id, "name": self.name, key: self.tool_type, "input": self.input, "result": self.result}
+
+    @classmethod
+    def _load(cls, data: dict[str, Any]) -> ToolCallDetail:
+        return cls(
+            id=data["id"],
+            name=data["name"],
+            tool_type=data["toolType"],
+            input=data.get("input", {}),
+            result=data.get("result", {}),
+        )
+
+
+@dataclass
+class ReasoningDataItem(CogniteResource, ABC):
+    """Base class for reasoning data item types."""
+
+    _type: ClassVar[str]
+
+    @classmethod
+    def _load(cls, data: dict[str, Any]) -> ReasoningDataItem:
+        item_type = data.get("type", "")
+        klass = _REASONING_DATA_CLS_BY_TYPE.get(item_type, UnknownReasoningDataItem)
+        return klass._load_item(data)
+
+    @classmethod
+    @abstractmethod
+    def _load_item(cls, data: dict[str, Any]) -> ReasoningDataItem: ...
+
+
+@dataclass
+class ToolCallReasoningDataItem(ReasoningDataItem):
+    """Reasoning data item for a tool call.
+
+    Args:
+        tool_call (ToolCallDetail | None): Details of the tool call.
+    """
+
+    _type: ClassVar[str] = "toolCall"
+    tool_call: ToolCallDetail | None = None
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        key = "toolCall" if camel_case else "tool_call"
+        result: dict[str, Any] = {"type": self._type}
+        if self.tool_call is not None:
+            result[key] = self.tool_call.dump(camel_case=camel_case)
+        return result
+
+    @classmethod
+    def _load_item(cls, data: dict[str, Any]) -> ToolCallReasoningDataItem:
+        tool_call_data = data.get("toolCall")
+        return cls(tool_call=ToolCallDetail._load(tool_call_data) if tool_call_data else None)
+
+
+@dataclass
+class UnknownReasoningDataItem(ReasoningDataItem):
+    """Unknown reasoning data item type for forward compatibility.
+
+    Args:
+        type (str): The item type.
+        data (dict[str, Any]): The raw item data.
+    """
+
+    type: str = ""
+    data: dict[str, Any] = field(default_factory=dict)
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        result = self.data.copy()
+        result["type"] = self.type
+        return result
+
+    @classmethod
+    def _load_item(cls, data: dict[str, Any]) -> UnknownReasoningDataItem:
+        return cls(type=data.get("type", ""), data=data)
+
+
+_REASONING_DATA_CLS_BY_TYPE: dict[str, type[ReasoningDataItem]] = {
+    ToolCallReasoningDataItem._type: ToolCallReasoningDataItem,
+}
+
+
+@dataclass
 class AgentReasoningItem(CogniteResource):
     """Reasoning item in agent response.
 
     Args:
         content (list[MessageContent]): The reasoning content.
+        data (list[ReasoningDataItem] | None): The data of the reasoning.
     """
 
     content: list[MessageContent]
+    data: list[ReasoningDataItem] | None = None
 
     def dump(self, camel_case: bool = True) -> dict[str, Any]:
-        return {
-            "content": [item.dump(camel_case=camel_case) for item in self.content],
-        }
+        result: dict[str, Any] = {"content": [item.dump(camel_case=camel_case) for item in self.content]}
+        if self.data is not None:
+            result["data"] = [item.dump(camel_case=camel_case) for item in self.data]
+        return result
 
     @classmethod
     def _load(cls, data: dict[str, Any]) -> AgentReasoningItem:
         content = [MessageContent._load(item) for item in data.get("content", [])]
-        return cls(content=content)
+        data_items = [ReasoningDataItem._load(item) for item in data.get("data", [])] or None
+        return cls(content=content, data=data_items)
 
 
 @dataclass
--- a/cognite/client/data_classes/agents/chat.py
+++ b/cognite/client/data_classes/agents/chat.py
@@ -482,10 +482,6 @@ class ToolCallDetail(CogniteResource):
     input: dict[str, Any] = field(default_factory=dict)
     result: dict[str, Any] = field(default_factory=dict)
 
-    def dump(self, camel_case: bool = True) -> dict[str, Any]:
-        key = "toolType" if camel_case else "tool_type"
-        return {"id": self.id, "name": self.name, key: self.tool_type, "input": self.input, "result": self.result}
-
     @classmethod
     def _load(cls, data: dict[str, Any]) -> ToolCallDetail:
         return cls(
@@ -502,6 +498,11 @@ class ReasoningDataItem(CogniteResource, ABC):
     """Base class for reasoning data item types."""
 
     _type: ClassVar[str]
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        output = super().dump(camel_case=camel_case)
+        output["type"] = self._type
+        return output
 
     @classmethod
     def _load(cls, data: dict[str, Any]) -> ReasoningDataItem:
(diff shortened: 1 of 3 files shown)


cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,130 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+
+    @classmethod
+    def _load(cls, data: dict[str, Any]) -> ToolCallDetail:
+        return cls(
+            id=data["id"],
+            name=data["name"],
+            tool_type=data["toolType"],
+            input=data.get("input", {}),
+            result=data.get("result", {}),
+        )
```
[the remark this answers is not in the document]
ks93: /gemini addressed in c5c126103

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,130 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+
+    @classmethod
+    def _load(cls, data: dict[str, Any]) -> ReasoningDataItem:
+        item_type = data.get("type", "")
+        klass = _REASONING_DATA_CLS_BY_TYPE.get(item_type, UnknownReasoningDataItem)
+        return klass._load_item(data)
+
+    @classmethod
+    @abstractmethod
+    def _load_item(cls, data: dict[str, Any]) -> ReasoningDataItem: ...
```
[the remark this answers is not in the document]
ks93: /gemini addressed in c5c126103

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,130 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+        key = "toolCall" if camel_case else "tool_call"
+        result: dict[str, Any] = {"type": self._type}
+        if self.tool_call is not None:
+            result[key] = self.tool_call.dump(camel_case=camel_case)
+        return result
+
+    @classmethod
+    def _load_item(cls, data: dict[str, Any]) -> ToolCallReasoningDataItem:
+        tool_call_data = data.get("toolCall")
+        return cls(tool_call=ToolCallDetail._load(tool_call_data) if tool_call_data else None)
```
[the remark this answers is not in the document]
ks93: /gemini partial — `dump` refactored as suggested; snake_case fallback in `_load_item` not applied since the API only sends camelCase

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,131 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        output = super().dump(camel_case=camel_case)
+        output["type"] = self._type
+        return output
+
+    @classmethod
+    def _load(cls, data: dict[str, Any]) -> ReasoningDataItem:
+        item_type = data.get("type", "")
+        klass = _REASONING_DATA_CLS_BY_TYPE.get(item_type, UnknownReasoningDataItem)
```
haakonvt: Is this field required? Unless a future "reasoning data type" is expected to not have it, lets:
```suggestion
        item_type = data["type"]
        klass = _REASONING_DATA_CLS_BY_TYPE.get(item_type, UnknownReasoningDataItem)
```

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,131 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+        key = "toolCall" if camel_case else "tool_call"
+        result: dict[str, Any] = {"type": self._type}
+        if self.tool_call is not None:
+            result[key] = self.tool_call.dump(camel_case=camel_case)
+        return result
+
+    @classmethod
+    def _load_item(cls, data: dict[str, Any]) -> ToolCallReasoningDataItem:
+        tool_call_data = data.get("toolCall")
+        return cls(tool_call=ToolCallDetail._load(tool_call_data) if tool_call_data else None)
```
haakonvt: I think this works:
```suggestion
        return cls(tool_call=ToolCallDetail._load_if(data.get("toolCall"))
```

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,131 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+@dataclass
+class UnknownReasoningDataItem(ReasoningDataItem):
+    """Unknown reasoning data item type for forward compatibility.
+
+    Args:
+        type (str): The item type.
+        data (dict[str, Any]): The raw item data.
+    """
+
+    type: str = ""
```
haakonvt: Make this an instance property instead
```suggestion
```

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,131 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+    data: dict[str, Any] = field(default_factory=dict)
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        result = self.data.copy()
+        result["type"] = self.type
+        return result
+
+    @classmethod
+    def _load_item(cls, data: dict[str, Any]) -> UnknownReasoningDataItem:
+        return cls(type=data.get("type", ""), data=data)
```
haakonvt:
```suggestion
        data = data.copy()
        item_type = data.pop("type")
        return cls(type=item_type, data=data)
```

haakonvt (commented):
A few comments, mostly LGTM

cognite/client/data_classes/agents/chat.py
ks93: Done.

cognite/client/data_classes/agents/chat.py
ks93: Done.

cognite/client/data_classes/agents/chat.py
ks93: Done.

cognite/client/data_classes/agents/chat.py
ks93: Done.

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,132 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+@dataclass
+class UnknownReasoningDataItem(ReasoningDataItem):
+    """Unknown reasoning data item type for forward compatibility.
+
+    Args:
+        type (str): The item type.
+        data (dict[str, Any]): The raw item data.
+    """
+
+    type: str
```
haakonvt, line 550: I was first thinking `@property`, but this is prob fine

haakonvt (dismissed):
LGTM

haakonvt:
@nithinb ping 😄

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,132 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+        result: dict[str, Any] = {"content": [item.dump(camel_case=camel_case) for item in self.content]}
+        if self.data is not None:
+            result["data"] = [item.dump(camel_case=camel_case) for item in self.data]
+        return result
 
     @classmethod
     def _load(cls, data: dict[str, Any]) -> AgentReasoningItem:
         content = [MessageContent._load(item) for item in data.get("content", [])]
-        return cls(content=content)
+        data_items = [ReasoningDataItem._load(item) for item in data.get("data", [])] or None
```
nithinb, line 591: Can data.get("data", []) be an empty list? 
Because this silently converts an empty list to None. If it doesn't matter, I think a comment would be ideal in such scenarios

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,132 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        result = self.data.copy()
+        result["type"] = self.type
+        return result
+
+    @classmethod
+    def _load_item(cls, data: dict[str, Any]) -> UnknownReasoningDataItem:
+        data = data.copy()
+        item_type = data.pop("type")
+        return cls(type=item_type, data=data)
```
nithinb: nit and not blocking but generally it is better to store `data.copy()` in a separate variable instead of storing it in the same variable.

Like same name used for parameter and copy of the parameter.

tests/tests_unit/test_api/test_agents_chat.py
```diff
@@ -113,11 +139,11 @@ def test_chat_simple_message(
...
-        assert isinstance(agent_msg.reasoning[0], AgentReasoningItem)
-        assert len(agent_msg.reasoning[0].content) == 1
-        content = agent_msg.reasoning[0].content[0]
-        assert isinstance(content, TextContent)
-        assert content.text == "The user is asking about capabilities"
+        reasoning_item = agent_msg.reasoning[0]
+        assert isinstance(reasoning_item, AgentReasoningItem)
+        assert isinstance(reasoning_item.content[0], TextContent)
+        assert reasoning_item.data is not None
+        assert isinstance(reasoning_item.data[0], ToolCallReasoningDataItem)
```
nithinb, line 147: I think you should also add validation of actual data in reasoning_item.data and reasoning_item.content

nithinb (commented):
I had a few comments, feel free to reach out once the changes are made.

cognite/client/data_classes/agents/chat.py
ks93: Good catch! The API docs mark `data` as optional with no `required` flag, so the API omits the field entirely rather than sending an empty list. The `or None` is safe here.

tests/tests_unit/test_api/test_agents_chat.py
ks93: The base test `test_json_serialize` in `test_base.py` runs a full dump/load round-trip on every concrete `CogniteResource` subclass, including the new ones here. Field values are validated there generically.

cognite/client/data_classes/agents/chat.py
ks93: Fixed in cd1b4735a.

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,132 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
...
+        type (str): The item type.
+        data (dict[str, Any]): The raw item data.
+    """
+
+    type: str
+    data: dict[str, Any] = field(default_factory=dict)
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        result = self.data.copy()
+        result["type"] = self.type
```
nithinb, line 555: I have a question related to this, the parent type ReasoningDataItem has a _type defined as an attirbute. Which is not going to be used when you do a dump on UnknownReasoningDataItem object. I wanted to understand if this is expected?

cognite/client/data_classes/agents/chat.py
```diff
@@ -464,25 +464,132 @@ def _load(cls, data: dict[str, Any]) -> AgentDataItem:
         return cls(type=item_type, data=item_data)
 
 
+@dataclass
+class ToolCallDetail(CogniteResource):
```
nithinb, line 468: Question: this does not have a dump method, is that expected?

nithinb (commented):
I had 2 questions. The changes in general look good. 
I think once they are answered, I am good to approve this

cognite/client/data_classes/agents/chat.py
ks93: Yes, intentional. All fields on `ToolCallDetail` are plain values, so the inherited `CogniteResource.dump()` already camelCases `tool_type` to `toolType` correctly on its own. This was actually the original ask from gemini-code-assist earlier in this thread, adding a manual override back would just duplicate what the base class already does.

cognite/client/data_classes/agents/chat.py
ks93: Yes, by design. `_type` is a `ClassVar` because known subtypes like `ToolCallReasoningDataItem` have a fixed type string. `UnknownReasoningDataItem` is the catch-all for types we don't recognize yet, so its type string varies per instance and lives in the regular `type` field instead. That's why it overrides `dump`/`_load_item` rather than relying on the base class's `_type`-based implementation.

nithinb (commented):
risk review ok

haakonvt (approved):

PR #2639, opened 2026-05-22
docs(Functions): clarify type annotation restrictions when using function_handle

jonasblanc:
## Description
 Using non-builtin type annotations (e.g. client: CogniteClient) in a function_handle causes a NameError at deploy time, because only the handle function's source code is uploaded and annotations are evaluated at module import time, before the function body runs. Updated the FunctionHandle docstring and create() docstring to reflect this, and updated the example to use string annotation form ("CogniteClient | None")

## Checklist:
- [x] Tests added/updated.


haakonvt (dismissed):
LGTM

haakonvt:
There is a function test with a TODO - consider stringifying that annotation with an explainer

jonasblanc:
> There is a function test with a TODO - consider stringifying that annotation with an explainer

I believe it's fine as is. It's limited to protocol definition and never actually deployed. Ruff is not happy about string annotation combined with lazy evaluation. I added a comment.

cognite/client/_api/functions/__init__.py
```diff
@@ -244,6 +244,11 @@ async def create(
             .. note:
                 When using a predefined function object, you can list dependencies between the tags `[requirements]` and `[/requirements]` in the function's docstring.
                 The dependencies will be parsed and validated in accordance with requirement format specified in `PEP 508 <https://peps.python.org/pep-0508/>`_.
+
+            .. note:
+                Only the source code of the handle function itself is deployed. Non-builtin type
+                annotations (e.g. ``client: CogniteClient``) will cause a ``NameError`` at deploy
+                time. Either omit the annotation or use string form (e.g. ``"CogniteClient | None"``).
```
dbrattli: Not sure why you would like `CogniteClient` to be optional i.e `| None`? AFAIK, it will never be optional. If you declare a client you will always get a `CogniteClient`  injected (you will never get `None`). If you don't declare it will not be injected. So declaring it as optional will force the rest of the code to narrow it before using it.

haakonvt:
> > There is a function test with a TODO - consider stringifying that annotation with an explainer
> 
> I believe it's fine as is. It's limited to protocol definition and never actually deployed. Ruff is not happy about string annotation combined with lazy evaluation. I added a comment.

Oh my, haha, I had apparently not even merged/pushed the comment I was thinking about 😂  Sorry about that (proof I'm not clinically insane)

image

cognite/client/data_classes/functions.py
```diff
@@ -62,6 +67,8 @@ async def __call__(
         self,
         *,
         # TODO(haakonvt): change to CogniteClient | AsyncCogniteClient when/if functions start supporting it
+        # Note: in user-defined handle functions, only the function source is deployed, so
+        # CogniteClient must be a string annotation or omitted (see FunctionHandle docstring).
```
haakonvt: Feel free to revert this - as this was (from above post) not the comment I was referring to 😄

haakonvt (dismissed):

dbrattli (dismissed):
:unicorn:

haakonvt (approved):
🦄

PR #2642, opened 2026-05-23
feat(Data Modeling): Unite Files, Time Series and Datapoints directly in DM

haakonvt:


haakonvt:
Split into several smaller pieces

PR #2645, opened 2026-05-25
chore: move to standardized pyproject.toml format

MortGron:
## Description
The pyproject toml file has been updated to use standardized sections as described in https://packaging.python.org/en/latest/guides/writing-pyproject-toml/. This would make it easier to change away from Poetry in the future if desirable. Poetry also recommends this new format, and has deprecated parts of the old format, e.g. the `optional = true` currently used.

Note, poetry has an issue with dependencies that has dropped support for older Python versions (Python 3.10). It in some cases specifies the same dependency multiple times for the same Pythonv version in poetry.lock, so some workarounds have been applied.

## Checklist:
- [ ] Tests added/updated.


pyproject.toml
```diff
@@ -1,18 +1,96 @@
...
-documentation = "https://cognite-sdk-python.readthedocs-hosted.com"
 authors = [
-    "Erlend Vollset <erlend.vollset@cognite.com>",
-    "Håkon Treider <hakon.treider@cognite.com>",
-    "Anders Albert <anders.albert@cognite.com>"
+    {name = "Erlend Vollset", email = "erlend.vollset@cognite.com>"},
+    {name = "Håkon Treider", email = "hakon.treider@cognite.com"},
+    {name = "Anders Albert", email = "anders.albert@cognite.com"}
 ]
 license = "Apache-2.0"
```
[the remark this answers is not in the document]
MortGron: The page https://packaging.python.org/en/latest/guides/writing-pyproject-toml/ uses PEP 639 as the reference

pyproject.toml
```diff
@@ -1,18 +1,96 @@
...
+    "numpy >=1.25; python_version > '3.10'"
+]  # poetry 2.4.1 does not correctly limit numpy 2.2 to only python 3.10
+sympy = ["sympy"]
+pandas = ["pandas >=2.1"]
+geo = [
+    "geopandas >=0.14",
+    "shapely >=1.7.0"
+]
+yaml = ["PyYAML (>=6, <7)"]
+functions = ["pip >=20"]
```
haakonvt: This is part of main deps, so this does nothing:
```suggestion
```

haakonvt (dismissed):
One suggestion, otherwise LGTM

pyproject.toml
haakonvt: (This is prob referred to in other parts of the documentation as well)

MortGron: Also removed from "all", and updated README where the extras ("functions" and "all") were mentioned

haakonvt (dismissed):
LGTM

Khyat-Cognite:
Missing docs update — docs/source/extensions_and_optional_dependencies.rst still references removed functions extra

https://github.com/cognitedata/cognite-sdk-python/blob/6904cd7ac1f715017054460d30c78dc7f86a4ae7/docs/source/extensions_and_optional_dependencies.rst?plain=1#L19

Users following the docs will try pip install "cognite-sdk[functions]" which will either silently do nothing or error depending on pip version.

poetry.lock
```diff
@@ -3990,10 +3897,10 @@ files = [
 name = "tzdata"
 version = "2026.2"
 description = "Provider of IANA time zone data"
-optional = false
+optional = true
```
Khyat-Cognite: optional = true, markers ONLY reference extras (extra == "pyodide" or ...)

The pyproject.toml correctly has "tzdata >=2024.1; sys_platform == 'win32'" in dependencies, but Poetry's lock resolution appears to have lost this.

This only affects poetry install workflows (not pip install which reads [project].dependencies directly). CI passes because tests install extras. 
But any Windows developer doing bare poetry install for the first time may hit timezone failures.

That said should it be tested on windows?

poetry.lock
```diff
@@ -4070,16 +3977,15 @@ test = ["big-O", "jaraco.functools", "jaraco.itertools", "jaraco.test", "more_it
 type = ["pytest-mypy (>=1.0.1) ; platform_python_implementation != \"PyPy\""]
 
 [extras]
-all = ["PyYAML", "geopandas", "numpy", "pandas", "pip", "shapely", "sympy"]
-functions = ["pip"]
+all = ["PyYAML", "geopandas", "numpy", "pandas", "shapely", "sympy"]
 geo = ["geopandas", "shapely"]
-numpy = ["numpy"]
+numpy = ["numpy", "numpy"]
```
Khyat-Cognite, line 3982: Why duplicate numpy here?
The [project.optional-dependencies].numpy has two entries with different python_version markers; Poetry resolves them as two separate packages (numpy 2.2 for 3.10, numpy 2.4 for 3.11+) and emits both names? -- Is this the reason why?

This is generated by Poetry — re-running poetry lock may be needed after Poetry fixes this, or add a comment explaining the quirk.

pyproject.toml
```diff
@@ -1,18 +1,95 @@
...
+    "geopandas >=0.14",
+    "shapely >=1.7.0"
+]
+yaml = ["PyYAML (>=6, <7)"]
+pyodide = [
+    "tzdata >=2024.1",
+    "anyio"
+]  # keep pyodide related dependencies outside of 'all'
+all = [
+    "numpy >=1.25",
```
Khyat-Cognite: numpy extra: "numpy (>=1.25, <2.3); python_version == '3.10'" (correctly capped)
all extra: "numpy >=1.25" (no cap for Python 3.10)

Khyat-Cognite: Also flagged by gemini..

pyproject.toml
```diff
@@ -1,18 +1,95 @@
...
+    "sympy",
+    "pandas >=2.1",
+    "geopandas >=0.14",
+    "shapely >=1.7.0",
+    "PyYAML (>=6, <7)",
+ ]
```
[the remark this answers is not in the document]
MortGron: image
It does not work with Poetry. In which PEP is this feature included? I believe it works with pip, but it is not standard.

poetry.lock
MortGron: I think I found the issue. Poetry does not like that 'tzdata' is both in the main dependencies and in an extra group. My suggestion is therefore to only have 'tzdata' in the main group, and restrict it by sys_platform to work on pyodide.

MortGron: @haakonvt do you have any objection to this change?

poetry.lock
MortGron: @Khyat-Cognite Is there any action we should take? The poetry lock file should not be manually modified.

Khyat-Cognite: You're right just finished reading up on this myself — no manual edit needed. This is a Poetry quirk from having two numpy entries with different python_version markers in [project.optional-dependencies].numpy. Poetry emits both resolved package names into the [extras] section, resulting in numpy = ["numpy", "numpy"].

It's cosmetically ugly but functionally harmless (pip/installers deduplicate). No action required from our side — this should resolve itself when Poetry improves PEP 621 support.

poetry.lock
Khyat-Cognite: That makes sense — having tzdata in both main deps (with sys_platform == 'win32') and in the pyodide extra is what's confusing Poetry's resolver.

Your proposed fix sounds right: keep tzdata only in the main dependencies section but broaden the marker to cover both use cases:
`"tzdata >=2024.1; sys_platform == 'win32' or sys_platform == 'emscripten'"`

Or if pyodide doesn't reliably report sys_platform == 'emscripten', a simpler approach:
```
# In [project] dependencies:
"tzdata >=2024.1"  # Lightweight, needed for Windows + pyodide
```

Since tzdata is tiny (~500KB wheel, pure Python) and has no transitive deps, making it unconditional is also acceptable — it's the approach pandas itself takes.

Either way, removing it from the pyodide extra resolves the Poetry conflict. +1 to this approach.

MortGron: Since we already have decided to clean up `functions` as an extra, what about `pyodide` ? `anyio` is a dependency of  `httpx`, so the extra `pyodide` will now do nothing. Another question for you @haakonvt

pyproject.toml
```diff
@@ -1,18 +1,93 @@
...
+    "numpy >=1.25; python_version > '3.10'"
+]  # poetry 2.4.1 does not correctly limit numpy 2.2 to only python 3.10
+sympy = ["sympy"]
+pandas = ["pandas >=2.1"]
+geo = [
+    "geopandas >=0.14",
+    "shapely >=1.7.0"
+]
+yaml = ["PyYAML (>=6, <7)"]
+pyodide = ["anyio"]  # keep pyodide related dependencies outside of 'all'
```
haakonvt: Ref other discussion @MortGron, feel free to entirely remove category "pyodide".

poetry.lock
haakonvt: Go ahead!

Khyat-Cognite:
Will wait for CI to finish but yeah.. Looks Good!

Khyat-Cognite (approved):
🐴

pyproject.toml
```diff
@@ -1,18 +1,92 @@
...
 license = "Apache-2.0"
+license-files = ["LICENSE"]
+requires-python = ">= 3.10, <4"
+dependencies = [
+    "httpx (>=0, <1)",
+    "msal (>=1.31, <2)",
+    "authlib (>=1, <2)",
+    "protobuf >=4",
+    "packaging >=20",
+    "pip >=20", # make optional once poetry doesn't auto-remove it on "simple install"
```
nithinb, line 21: nit: the general convention is to use either `<library> (<version range>)` or `<library> <version range>`, the combination is not something that is not generally recommended but its not a blocker.

pyproject.toml
```diff
@@ -1,18 +1,92 @@
...
+]
+docs = [
+    "sphinx ==8.1.*",  # Increase when Python 3.10 support is dropped
+    "sphinx-rtd-theme >=1",
+    "sphinx-copybutton (>=0.5.2, <0.6)",
+    "sphinx-autosummary-accessors (>=2025.3.1, <2026)"
+]
+
+[project.urls]
+Documentation = "https://cognite-sdk-python.readthedocs-hosted.com"
```
nithinb, line 90: nit: Future addition, I think we can add more metadata to actually point to the repository, the changelog, etc.

nithinb (approved):
:unicorn:

pyproject.toml
MortGron: Good suggestion! I kept it out of this PR to have to keep this PR focused on one thing

haakonvt (approved):

PR #2647, opened 2026-05-26
feat(Data Modeling, Files): Add files api to data modeling with /retrieve and /list (DM-3842)

haakonvt:
This is the first of 3 or 4 parts, adding a "Data Modeling specific Files API interface" to the SDK that should significantly reduce the current friction.

In this part 1 I've added:
- New `files` API to to `client.data_modeling`
- Added the full interface (currently raising NotImplemented)
- Code to guarantee and handle only file nodes are fetched
- Some needed updates to the sync codegen scripts
- Tests


cognite/client/_api/data_modeling/files.py
```diff
@@ -0,0 +1,202 @@
...
+
+    async def __call__(self) -> NoReturn:
+        raise NotImplementedError("This method is not implemented yet!")
+
+    @overload
+    async def retrieve(
+        self,
+        nodes: NodeId | tuple[str, str],
+        *,
+        source: View | ViewId | tuple[str, str, str] = COGNITE_FILE_VIEW_ID,
```
haakonvt, line 79: Note for reviewer: `ViewId`s are immutable so they are fine to use as default args

cognite/client/_api/data_modeling/files.py
```diff
@@ -0,0 +1,200 @@
...
+        case _:
+            raise TypeError(f"Expected View, ViewId, or a (space, external_id, version) tuple, got {type(source)}")
+
+    if source_as_id == COGNITE_FILE_VIEW_ID:
+        return [source_as_id], False
+
+    # User has passed a custom source, we include CogniteFile source to guarantee only file nodes
+    # are returned. We will later strip them (hence the 'True' flag) to avoid returning nodes with
+    # properties from multiple sources as they are very annoying to work with in the SDK.
+    return [source_as_id, COGNITE_FILE_VIEW_ID], True
```
polomani, line 39: nit: can we document the return tuple type? hard to say what this `True` bool means here..

cognite/client/_api/data_modeling/files.py
```diff
@@ -0,0 +1,200 @@
...
+            Fetch properties from a custom view (note, only files will be returned):
+
+                >>> from cognite.client.data_classes.data_modeling import ViewId
+                >>> res = client.data_modeling.files.retrieve(
+                ...     NodeId("my-space", "my-file"),
+                ...     source=ViewId("my-space", "MyFileExtension", "v1"),
+                ... )
+        """
+        sources, strip = _resolve_source(source)
+        result = await self._cognite_client.data_modeling.instances.retrieve_nodes(nodes=nodes, sources=sources)  # type: ignore[arg-type]
```
polomani: can _resolve_source() return Sequence[ViewId] so we get rid of type ignore?

cognite/client/_api/data_modeling/files.py
```diff
@@ -0,0 +1,200 @@
...
+            instance_type="node",
+            sources=sources,
+            space=space,
+            sort=sort,
+            filter=filter,
+            limit=limit,
+        )
+        if strip:
+            for node in results:
+                node.drop_source(COGNITE_FILE_VIEW_ID)
```
polomani, line 200: 😅

cognite/client/_api/data_modeling/files.py
haakonvt: What more do you need 😂

Image

cognite/client/_api/data_modeling/files.py
haakonvt: Is strip a bad word? 😆

cognite/client/_api/data_modeling/files.py
haakonvt: I’ll widen what retrieve_nodes accept 👍

haakonvt: @polomani ✅

cognite/client/_api/data_modeling/files.py
polomani: yeah, you documented it referring to it as "True flag" (which is what it exactly is), without explaining what this boolean is for: it's a flag that tells the caller whether to strip CogniteFIle source or not

cognite/client/_api/data_modeling/files.py
polomani: nono, it's just I found the approach itself quite witty

polomani (approved):
🦄

PR #2656, opened 2026-05-29
chore: remove unused dev dependency pep8-naming

MortGron:
## Description
pep8-naming is not used. If we want to use it, the same functionality is in ruff. If you try to run it, it flags at least 8 non-pep8 compliant names in the `cognite/` folder

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -749,23 +749,6 @@ files = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


haakonvt (approved):
🦄

PR #2657, opened 2026-05-30
chore: Add linkcheck to github actions. Fix broken external links.

MortGron:
## Description
There is a number of broken external hyperlinks in the code base. The goal of this PR is to fix them, and to add an automated check of all links in the docstrings and the documentation..

Some urls which previously where formatted as links, have been reformatted as code. This applies to `localhost` urls and some other links such as `login.microsoftonline.com` which it does not make sense that users click on.

## Checklist:
- [ ] Tests added/updated.


haakonvt (commented):
Does this spam «try fetch» all these URLs again and again? 😬

haakonvt:
Also, completely unrelated, you are on a PR rampage; mind updating Gemini styleguide to let ruff deal with formatting?

MortGron:
> Does this spam «try fetch» all these URLs again and again? 😬

All URLs are tried every time Github Actions runs, if that is what you mean. If you refer to retry policy, it can be controlled, primarily with these config options:
image

haakonvt:
@MortGron lets only do this with `--hook-stage manual` then

haakonvt:
I think there may be 400+ links

MortGron:
> @MortGron lets only do this with `--hook-stage manual` then

Anyone who has cloned this repo and installed the dependencies can even without this PR do all the link checks by simply typing `make -C docs/ linkcheck` in the terminal. I am not sure what would be gained by wrapping that command in another tool, assuming we want to keep the status quo where link checking is done manually on demand.

So if no automatic link checking is desired, I can remove the Github Actions part, and let this PR be a one-time fixining of broken links. What do you think @haakonvt ?

haakonvt:
Nice, well, I missed that this was only part of the singular "build docs" action. I think that sounds good then 👍

cognite/client/_api/data_modeling/instances.py
```diff
@@ -771,7 +771,7 @@ async def inspect(
         involved_views: InvolvedViews | None = None,
         involved_containers: InvolvedContainers | None = None,
     ) -> InstanceInspectResults:
-        """`Reverse lookup for instances <https://developer.cognite.com/api/v1/#tag/Instances/operation/instanceInspect>`_.
+        """`Reverse lookup for instances <https://docs.cognite.com/20230101/instances/inspect-instances>`_.
```
haakonvt: I'm unsure about linking to `https://docs.cognite.com/20230101/instances/inspect-instances`, it has a bit of a manual step now e.g. new API endpoints require a human to become visible.

haakonvt: I think we should use `https://api-docs.cognite.com/20230101/`. Maybe @arnekv can make this decision for us?

MortGron:
> Nice, well, I missed that this was only part of the singular "build docs" action. I think that sounds good then 👍

But are we really sure we want no automatic check of external links? We then completely rely on users reporting broken links, insteady of proactively finding issues. Maybe there is a better trigger that could be used: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows

cognite/client/_api/data_modeling/instances.py
MortGron: I am not sure I fully understand what you mean by that they require a human. 

On a related note, my biggest general concern about https://docs.cognite.com/20230101 is that it suggest users should use pure `requests`  to interact with CDF:

image

haakonvt: Oh that is bad. Real bad…

haakonvt:
> > Nice, well, I missed that this was only part of the singular "build docs" action. I think that sounds good then 👍
> 
> But are we really sure we want no automatic check of external links? We then completely rely on users reporting broken links, insteady of proactively finding issues. Maybe there is a better trigger that could be used: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows

Oh you misunderstood - this PR is fine to merge since only the one/singular job will fetch and check links 👌

cognite/client/_api/data_modeling/instances.py
haakonvt: Yeah, lets not use `docs.cognite.com`. Ready to approve once replaced 😄

MortGron: Now all links point to `api-docs.cognite.com`

MortGron:
> > > Nice, well, I missed that this was only part of the singular "build docs" action. I think that sounds good then 👍
> > 
> > 
> > But are we really sure we want no automatic check of external links? We then completely rely on users reporting broken links, insteady of proactively finding issues. Maybe there is a better trigger that could be used: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows
> 
> Oh you misunderstood - this PR is fine to merge since only the one/singular job will fetch and check links 👌

Ok, sorry, my bad.

Still a question if the default 30 s timeout limit should be decreased. If suddenly `api-docs.cognite.com` becomes slow, than with a 30 s timeout it can take a long time for this Github Action to complete. But that is more of a hypothetical scenario.

cognite/client/_api/data_modeling/instances.py
haakonvt: Seems to still be a few left like: https://docs.cognite.com/20230101/function-calls/call-a-function-asynchronously

I see that eg. this one was already using docs.cognite.com from before so maybe its ok...

haakonvt (approved):
Good job! Let's get an extra sanity check in RR 👍 

(See my one other comment)

thorkildcognite (commented):
🦄

PR #2663, opened 2026-06-03
feat(metering): add Python SDK support for the Metering API

krishna1633:
## Description

Adds SDK support for the Metering API to retrieve consumption data for a CDF project. Follows the same pattern as the Limits API (#2436).

API spec: `versions/v1/metering.preproc.yml`

## Key Features

- Data classes: `MeteringData`, `MeteringDataList`, `MeteringDataPoint`
- API methods on `client.metering`:
  - `retrieve(id, start, end, number_of_datapoints)` — GET `/metering/meters/{meterId}`
  - `retrieve_multiple(ids, start, end, number_of_datapoints)` — POST `/metering/meters/byids`
  - `list(filter, limit, start, end, number_of_datapoints)` — GET/POST `/metering/meters[/list]`
- All methods support optional time-series parameters (`start` + `number_of_datapoints` together enable historical data)
- Uses `cdf-version: {api_subversion}-alpha` header (alpha API)
- Feature preview warning for alpha endpoint

## Implementation Details

- `retrieve()` — GET with optional query params for time range
- `retrieve_multiple()` — direct `_post` to `/byids` (custom `meterId` field, not standard `id`/`externalId`)
- `list()` — GET without filter, POST to `/list` with `Prefix` filter; time-range params go as query params (GET) or body (POST) via `other_params`
- Added `MeterId` identifier class in `utils/_identifier.py` (mirrors `LimitId`)
- Auto-generated sync client files via pre-commit `sync-client-codegen` hook

## Checklist
- [x] Tests added (unit + integration)
- [x] Documentation added (`docs/source/metering.rst`)
- [x] All pre-commit hooks pass (ruff, mypy, custom checks, pydoclint, codespell, sync codegen)
- [x] `"metering"` added to idempotent POST whitelist in `test_meta.py`
- [x] Metering URL retryability test cases added to `test_api_client.py`


cognite/client/_api/metering.py
```diff
@@ -0,0 +1,210 @@
...
+    def __init__(self, config: ClientConfig, api_version: str | None, cognite_client: AsyncCogniteClient) -> None:
+        super().__init__(config, api_version, cognite_client)
+        self._warning = FeaturePreviewWarning(
+            api_maturity="alpha",
+            sdk_maturity="alpha",
+            feature_name="Metering API",
+        )
+
+    def _alpha_headers(self) -> dict[str, str]:
+        return {"cdf-version": f"{self._config.api_subversion}-alpha"}
```
haakonvt: Lets move this to utilities + make it check that "alpha" is not already in `self._config.api_subversion`

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,210 @@
...
+            params["end"] = end
+        if number_of_datapoints is not None:
+            params["numberOfDatapoints"] = number_of_datapoints
+        return params
+
+    async def retrieve(
+        self,
+        id: str,
+        start: int | None = None,
+        end: int | None = None,
```
haakonvt: See what e.g. DatapointsAPI supports for start/end (at least tz aware datetime)

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,210 @@
...
+
+                >>> from cognite.client import CogniteClient, AsyncCogniteClient
+                >>> client = CogniteClient()
+                >>> # async_client = AsyncCogniteClient()  # another option
+                >>> res = client.metering.retrieve_multiple(
+                ...     ids=["atlas.monthly_ai_tokens", "files.storage_bytes"]
+                ... )
+        """
+        self._warning.warn()
+        body: dict[str, Any] = {"items": [{"meterId": id_} for id_ in ids]}
```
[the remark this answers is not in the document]
krishna1633: trusts the type annotations

tests/tests_unit/test_api/test_metering.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def test_metering_data_list_as_ids(self) -> None:
+        from cognite.client.data_classes.metering import MeteringData, MeteringDataList
+
+        items = [
+            MeteringData(meter_id="atlas.monthly_ai_tokens"),
+            MeteringData(meter_id="files.storage_bytes"),
+        ]
+        lst = MeteringDataList(items)
+        assert lst.as_ids() == ["atlas.monthly_ai_tokens", "files.storage_bytes"]
```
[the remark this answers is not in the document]
krishna1633: No empty guarding other sdk

cognite/client/_api/metering.py
haakonvt: ...and please, change existing uses like `headers = {"cdf-version": f"{self._config.api_subversion}-alpha"}` ❤️

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,210 @@
...
+
+        Retrieves consumption data by its ``meterId``.
+
+        Args:
+            id (str): Meter ID to retrieve.
+                Metering is identified by an id containing the service name and a service-scoped metering name.
+                For instance ``atlas.monthly_ai_tokens`` is the id of the ``atlas`` service metering ``monthly_ai_tokens``.
+                Service and metering names are always in ``lower_snake_case``.
+            start (int | None): Start timestamp (inclusive) for historical data, in milliseconds since epoch.
+                **Must be provided together with** ``number_of_datapoints`` to get time-series data.
```
haakonvt: > **Must be provided together with**

Should you validate this in the code?

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,210 @@
...
+        self._warning.warn()
+        params = self._time_range_params(start, end, number_of_datapoints) or None
+        return await self._retrieve(
+            identifier=MeterId(id),
+            cls=MeteringData,
+            headers=self._alpha_headers(),
+            params=params,
+        )
+
+    async def retrieve_multiple(
```
haakonvt: We don't use `retrieve_multiple` anymore; it is from the time when we did not have `typing.overload` 😄

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,210 @@
...
+        end: int | None = None,
+        number_of_datapoints: int | None = None,
+    ) -> MeteringDataList:
+        """`List all meters <https://api-docs.cognite.com/20230101-alpha/tag/Metering/operation/listConsumptionData/>`_.
+
+        Lists all available meters for a specific project. Optionally filter by meter ID prefix using a ``Prefix`` filter.
+
+        Args:
+            filter (Prefix | None): Optional ``Prefix`` filter to apply on the ``meterId`` property (only ``Prefix`` filters are supported).
+            limit (int | None): Maximum number of meters to return. Defaults to 1000. Set to ``None`` or ``-1`` to return all meters.
```
haakonvt: Defaults to 1000?

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,210 @@
...
+            start (int | None): Start timestamp (inclusive) for historical data, in milliseconds since epoch.
+                **Must be provided together with** ``number_of_datapoints`` to get time-series data.
```
haakonvt: > **Must be provided together with**

Maybe validate?

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,210 @@
...
+                If omitted, only meter metadata is returned without time-series data.
+            end (int | None): End timestamp (inclusive) for historical data, in milliseconds since epoch.
+                Only relevant when ``start`` is provided. Defaults to the current time on the server if omitted.
+            number_of_datapoints (int | None): Number of equal-width time buckets to return between ``start`` and ``end``.
+                **Must be provided together with** ``start`` to get time-series data.
+                Valid range: 0-600. API server default is 0 (metadata only).
```
haakonvt: Maybe validate as well?

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,210 @@
...
+                ...     start=1764547200000,
+                ...     end=1767225599000,
+                ...     number_of_datapoints=30,
+                ... )
+        """
+        self._warning.warn()
+        other_params = self._time_range_params(start, end, number_of_datapoints) or None
+
+        return await self._list(
+            method="GET" if filter is None else "POST",
```
haakonvt: Why can't you just always use GET here? Passing no filter is kind of like passing a filter matching everything, right?
```py
    method="GET" if filter is None else "POST",
```

cognite/client/data_classes/metering.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @classmethod
+    def _load(cls, resource: dict[str, Any]) -> MeteringDataPoint:
+        return cls(timestamp=resource["timestamp"], average=resource["average"])
+
+    def dump(self) -> dict[str, Any]:
+        return {"timestamp": self.timestamp, "average": self.average}
+
+    def __eq__(self, other: Any) -> bool:
+        return isinstance(other, MeteringDataPoint) and self.dump() == other.dump()
```
haakonvt: Why have you implemented dunder eq?

cognite/client/data_classes/metering.py
```diff
@@ -0,0 +1,73 @@
+from __future__ import annotations
+
+from typing import Any
+
+from cognite.client.data_classes._base import CogniteResource, CogniteResourceList, basic_instance_dump
+
+
+class MeteringDataPoint:
```
haakonvt: Should inherit from `CogniteResource`

cognite/client/data_classes/metering.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def __init__(self, timestamp: int, average: float) -> None:
+        self.timestamp = timestamp
+        self.average = average
+
+    @classmethod
+    def _load(cls, resource: dict[str, Any]) -> MeteringDataPoint:
+        return cls(timestamp=resource["timestamp"], average=resource["average"])
+
+    def dump(self) -> dict[str, Any]:
```
haakonvt: Missing "camel_case" arg (although it doesnt do anything here)

haakonvt (commented):
A few comments - did not make it all the way through 😄 A lot looks great btw, much nits here:

krishna1633:
> A few comments - did not make it all the way through 😄 A lot looks great btw, much nits here:

Sure. Once I address these comments, will ask you for another iteration so you can all the way through

cognite/client/_api/metering.py
krishna1633: Done — moved to BasicAsyncAPIClient._alpha_version_header() with a double-alpha guard. Updated LimitsAPI inline uses as well.

cognite/client/_api/metering.py
krishna1633: Done — start and end now accept int (epoch ms), datetime, or relative strings like "2w-ago". Converted to epoch ms via timestamp_to_ms before hitting the API.

cognite/client/_api/metering.py
krishna1633: Adding a client-side guard would diverge from the SDK pattern of not duplicating server-enforced rules. Caller needs an error to know. It may not be concerned if error is from sdk or API as long as it is clear.
Per the server spec, providing only start without numberOfDatapoints silently returns metadata (server default is 0) — this is documented, expected behavior. The docstring already states **Must be provided together with**.

cognite/client/_api/metering.py
krishna1633: Done — removed retrieve_multiple, retrieve now uses @overload (single str → MeteringData | None, list → MeteringDataList). Same pattern as UnitsAPI

cognite/client/_api/metering.py
krishna1633: Fixed — docstring now correctly says 25 (DEFAULT_LIMIT_READ)

cognite/client/_api/metering.py
krishna1633: Same as above — skipping for the same reason

cognite/client/_api/metering.py
krishna1633: the server returns a clear 400 with "Datapoints must be a valid integer in the range of 0 and 600"

cognite/client/data_classes/metering.py
krishna1633: Done — removed. CogniteResource provides __eq__ via dump()

cognite/client/data_classes/metering.py
krishna1633: Done

cognite/client/_api/metering.py
krishna1633: did not follow.
API supports both . GET is used for the no-filter case since the endpoint doesn't accept a body.

krishna1633:
@haakonvt This is ready again

cognite/client/data_classes/metering.py
krishna1633: Done

cognite/client/_basic_api_client.py
```diff
@@ -219,6 +219,11 @@ def __setstate__(self, state: dict[str, Any]) -> None:
     def _select_async_http_client(self, is_retryable: bool) -> AsyncHTTPClientWithRetry:
         return self._http_client_with_retry if is_retryable else self._http_client
 
+    def _alpha_version_header(self) -> dict[str, str]:
+        subversion = self._config.api_subversion
+        version = subversion if subversion.endswith("-alpha") else subversion + "-alpha"
```
haakonvt: Just to be on the safe side
```suggestion
        version = subversion if "alpha" in subversion else subversion + "-alpha"
```

cognite/client/_api/limits.py
```diff
@@ -50,7 +50,7 @@ async def retrieve(self, id: str) -> Limit | None:
         """
         self._warning.warn()
 
-        headers = {"cdf-version": f"{self._config.api_subversion}-alpha"}
+        headers = self._alpha_version_header()
```
haakonvt, line 53: Great, thanks!

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,182 @@
...
+            )
+        body: dict[str, Any] = {"items": [MeterId(id_).as_dict() for id_ in id]}
+        body.update(self._time_range_params(start, end, number_of_datapoints))
+        res = await self._post(
+            url_path=self._RESOURCE_PATH + "/byids",
+            json=body,
+            headers=self._alpha_version_header(),
+            semaphore=self._get_semaphore("read"),
+        )
+        return MeteringDataList._load(res.json()["items"])._maybe_set_client_ref(self._cognite_client)
```
haakonvt: See if you can use `_retrieve_multiple`

cognite/client/_api/metering.py
haakonvt: Then why can't you always use POST?

tests/tests_integration/test_api/test_metering.py
```diff
@@ -0,0 +1,42 @@
...
+from cognite.client import CogniteClient
+from cognite.client.data_classes import MeteringData, MeteringDataList
+
+
+class TestMeteringAPI:
+    def test_list(self, cognite_client: CogniteClient) -> None:
+        res = cognite_client.metering.list()
+
+        assert isinstance(res, MeteringDataList)
+        assert all(meter.meter_id is not None for meter in res)
```
haakonvt, line 15: Don't use `all` in tests as they give false positives for empty (unless you first assert on non-empty) :
```py
>>> all(x is None for x in())
True
```

tests/tests_integration/test_api/test_metering.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_retrieve_single(self, cognite_client: CogniteClient) -> None:
+        meters = cognite_client.metering.list()
+        if len(meters) == 0:
+            pytest.skip("No meters found in project - skipping retrieve test")
```
haakonvt, line 20: Can we add a fixture that automatically populates data? Something like try-retrieve-except-create-and-return

tests/tests_integration/test_api/test_metering.py
```diff
@@ -0,0 +1,42 @@
...
+        meters = cognite_client.metering.list(limit=2)
+        if len(meters) == 0:
+            pytest.skip("No meters found in project - skipping retrieve_multiple test")
+
+        ids = meters.as_ids()
+        res = cognite_client.metering.retrieve(id=ids)
+
+        assert isinstance(res, MeteringDataList)
+        assert len(res) == len(ids)
+        assert [m.meter_id for m in res] == ids
```
haakonvt: as_ids?

tests/tests_unit/test_api/test_metering.py
```diff
@@ -0,0 +1,316 @@
...
+        metering_url: str,
+    ) -> None:
+        httpx_mock.add_response(
+            method="GET",
+            url=f"{metering_url}/{ATLAS_METER['meterId']}",
+            status_code=200,
+            json=ATLAS_METER,
+        )
+
+        meter_id = str(ATLAS_METER["meterId"])
```
haakonvt: Already str

tests/tests_unit/test_api/test_metering.py
```diff
@@ -0,0 +1,316 @@
...
+        httpx_mock: HTTPXMock,
+        metering_url: str,
+    ) -> None:
+        url_pattern = re.compile(re.escape(f"{metering_url}/{ATLAS_METER['meterId']}") + r"\?.*")
+        httpx_mock.add_response(
+            method="GET",
+            url=url_pattern,
+            status_code=200,
+            json=ATLAS_METER_WITH_DATA,
+        )
```
haakonvt: Looks repeated? Fixture

tests/tests_unit/test_api/test_metering.py
```diff
@@ -0,0 +1,316 @@
...
+            start=start_dt,
+            end=end_dt,
+            number_of_datapoints=10,
+        )
+
+        request_url = str(httpx_mock.get_requests()[0].url)
+        assert f"start={timestamp_to_ms(start_dt)}" in request_url
+        assert f"end={timestamp_to_ms(end_dt)}" in request_url
+
+    def test_retrieve_with_relative_string_start(
```
haakonvt: So many tests are slightly similar here; please manually make a selection. Consider parametrize.

tests/tests_unit/test_api/test_metering.py
```diff
@@ -0,0 +1,316 @@
...
+            number_of_datapoints=10,
+        )
+
+        request_url = str(httpx_mock.get_requests()[0].url)
+        assert "start=" in request_url
+        # value is an epoch ms integer, not the string "4w-ago"
+        start_val = int(request_url.split("start=")[1].split("&")[0])
+        assert start_val > 0
+
+    def test_dump_and_load_roundtrip(self) -> None:
```
haakonvt: Dump load roundtrips are automatically tested.

haakonvt (commented):
Please resolve comments you have addressed

cognite/client/_api/metering.py
haakonvt: We regularly duplicate such logic in the SDK for immediate user feedback, so please add here. I agree that for the number of allowed datapoints, that might change in the future so it should not be validated.

krishna1633: Done

cognite/client/_api/metering.py
krishna1633: done

cognite/client/_basic_api_client.py
krishna1633: Done

tests/tests_integration/test_api/test_metering.py
krishna1633: added `assert len(res) > 0`

tests/tests_integration/test_api/test_metering.py
krishna1633: Meters are created automatically by the platform when services consume them , there's no create endpoint in the metering API, so we can't populate test data. pytest.skip is the right fallback.

tests/tests_integration/test_api/test_metering.py
krishna1633: Done

tests/tests_unit/test_api/test_metering.py
krishna1633: Done

tests/tests_unit/test_api/test_metering.py
krishna1633: yup. added

tests/tests_unit/test_api/test_metering.py
krishna1633: refactored

tests/tests_unit/test_api/test_metering.py
krishna1633: Removed

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,184 @@
...
+
+                >>> from cognite.client import CogniteClient, AsyncCogniteClient
+                >>> client = CogniteClient()
+                >>> # async_client = AsyncCogniteClient()  # another option
+                >>> meters = client.metering.list(limit=None)
+
+            List meters filtered by prefix (e.g., all meters for the 'atlas' service):
+
+                >>> from cognite.client.data_classes.filters import Prefix
+                >>> prefix_filter = Prefix("meterId", "atlas.")
```
haakonvt:
```suggestion
                >>> prefix_filter = Prefix("meter_id", "atlas.")
```

cognite/client/_api/metering.py
krishna1633: Had to look into it. It was doable by adding one more indentifier MeterIdSequence

docs/source/index.rst
```diff
@@ -167,6 +167,9 @@ Contents
 
    hosted_extractors
    postgres_gateway
+   data_organization
+   limits
+   metering
```
haakonvt: What happened here?

tests/tests_unit/test_api/test_metering.py
```diff
@@ -0,0 +1,316 @@
...
+        meters_data = [
+            {"meterId": "atlas.monthly_ai_tokens", "datapoints": []},
+            {"meterId": "atlas.monthly_ai_prompts", "datapoints": []},
+        ]
+        httpx_mock.add_response(
+            method="POST",
+            url=f"{metering_url}/list",
+            status_code=200,
+            json={"items": meters_data},
+        )
```
haakonvt: Please move repeated mocked repsonses like this one to fixtures

docs/source/index.rst
krishna1633: My bad. I took rebase and this happened during conflict resolve. Fixed

cognite/client/_api/metering.py
krishna1633: Also, using _retrieve_multiple with MeterIdSequence. One caveat: the metering /byids endpoint returns a plain 404 instead of the SDK-standard 400 + missing field, so _retrieve_multiple doesn't catch it natively. Added a try/except around the call to return None for singleton 404s. For the mixed case (one valid + one invalid ID), the server returns 404 for the whole request — so that raises CogniteAPIError, consistent with how other SDK APIs behave.

cognite/client/_api/metering.py
```diff
@@ -0,0 +1,182 @@
...
+        identifiers = MeterIdSequence.load(id)
+        try:
+            return await self._retrieve_multiple(
+                list_cls=MeteringDataList,
+                resource_cls=MeteringData,
+                identifiers=identifiers,
+                other_params=self._time_range_params(start, end, number_of_datapoints) or None,
+                headers=self._alpha_version_header(),
+            )
+        except CogniteAPIError as e:
```
haakonvt: In the `_retrieve_multiple` method, this logic is already defined:
```py
try:
    tasks_summary.raise_compound_exception_if_failed_tasks(
        task_unwrap_fn=unpack_items_in_payload,
        task_list_element_unwrap_fn=identifiers.extract_identifiers,
    )
except CogniteNotFoundError:
    if identifiers.is_singleton():
        return None
    raise
```

So I tried running `client.metering.retrieve` with a list of one invalid id nad the wrong error was raised: I expected `CogniteNotFoundError`, but got the generic CogniteAPIError. 

This was unexpected tbh and led me down a small rabbit hole in the code base until I found this:

```py
async def raise_api_error(self, cognite_client: AsyncCogniteClient) -> NoReturn:
    cluster = cognite_client._config.cdf_cluster
    project = cognite_client._config.project


    match self.status_code, self.duplicated, self.missing:
        case 401, *_:
            await self._raise_no_project_access_error(cognite_client, cluster, project)
        case 409, list(), None:
            self._raise_api_error(CogniteDuplicatedError, cluster, project)
        case 400 | 422, None, list():
            self._raise_api_error(CogniteNotFoundError, cluster, project)
        case _:
            self._raise_api_error(CogniteAPIError, cluster, project)
```

It seems 404 is not expected for "ids not found" (why? No idea). I think it is worth adding to the above method.

Also, this "400 vs 404" is not a new issue, I found this quite recent Slack thread: https://cognitedata.slack.com/archives/C85QG3UBU/p1763387804040699

krishna1633: This PR can wait if you are pushing a fix around this.

docs/source/index.rst
```diff
@@ -167,6 +167,7 @@ Contents
 
    hosted_extractors
    postgres_gateway
+   metering
```
haakonvt, line 170: Consider moving this to `:caption: Specialized`

haakonvt (approved):
🦄  see the one comment ^

cognite/client/_api/metering.py
haakonvt: Fixed in #2689

docs/source/index.rst
krishna1633: Sure. Will do a small follow up PR

PR #2668, opened 2026-06-04
feat: Update hosted extractor API maturity

MortGron:
## Description
It has been a long time since the API of hosted extractors moved to general availability

## Checklist:
- [ ] Tests added/updated.


haakonvt (approved):
🦄  - thanks!!

PR #2673, opened 2026-06-08
chore(gitignore): don't track git worktrees

andersfylling:
## Description

Git worktrees allows us to have multiple "workspaces" of the same repo. Essentially just branches of the same repo data, that we shouldn't search through or commit as part of our work.


andersfylling:
:unicorn:

haakonvt (approved):

PR #2678, opened 2026-06-09
feat: Raw api polars integration

christoffer-hk:
## Description
Started on an integration with Polars, as an alternative to Pandas. Beginning with RAW rows (RowCore and RowListCore) and a `retrieve_polars_dataframe` function in the RawRowsAPI. Currently implemented the bare minimum to have mostly feature parity with how we integrate with Pandas. 

This only scratches the surface with what is technically possible to do with Polars. I believe there is a lot of potential in using Polars, especially now in version 8> with running things async. A lot of cool topics are mentioned here in a comparison written by Polars https://docs.pola.rs/user-guide/migration/pandas/ . 

## Checklist:
- [x] Tests added/updated.


christoffer-hk:
A couple comments from myself. 
- The sync-client-codegen does not add the polars import properly into the _sync_api; it falls outside of the `TYPE_CHECKING` check. 
- I have not ensured symmetry on purpose, by not making an implementation for something like `insert_polars_dataframe` to keep the size of the PR low. 
- I have to use the runtime compatibility version of polars (rtcompat) due to running under rosetta. While gemini suggests it should be ok for other devs to work with that, it'd be nice if someone could verify. Otherwise I suppose we need a different dev-dependency group to handle the different versions of polars.

PR #2680, opened 2026-06-11
feat(records): add list (filter) endpoint with read data classes

andersfylling:
Adds `list()` (filter) to the Records API.

`POST /streams/{id}/records/filter` → `RecordList`, with support for filtering by time range, DM filter expressions, source selection, sorting, and optional type information.

Added stream type as future rate limits will depend on it. For now I've implemented the lowest rate limit, which is related to immutable streams.

https://cognitedata.atlassian.net/browse/HVD-1262


cognite/client/_api/data_modeling/records.py
```diff
@@ -23,11 +36,14 @@ def __init__(self, config: ClientConfig, api_version: str | None, cognite_client
         )
 
     _OPERATION_TO_RATE_LIMIT: ClassVar[dict[str, RecordsConcurrencyOperation]] = {
+        "read": RecordsConcurrencyOperation.READ,
         "write": RecordsConcurrencyOperation.WRITE,
         "delete": RecordsConcurrencyOperation.WRITE,
     }
```
haakonvt, line 38: I still think this is an unnecessary indirection 😄

cognite/client/_api/data_modeling/records.py
```diff
@@ -44,6 +60,7 @@ async def delete(
         items: RecordId | Sequence[RecordId],
         *,
         stream_id: str,
+        stream_type: StreamType = _DEFAULT_STREAM_TYPE,
```
haakonvt: I don't understand the usage here at all - it seems to only be used by `_get_semaphore`? 🤔

cognite/client/_api/data_modeling/records.py
```diff
@@ -76,13 +94,15 @@ async def delete(
             identifiers=RecordIdSequence.load(items),
             wrap_ids=True,
             resource_path=self._records_url(stream_id),
+            override_semaphore=self._get_semaphore("delete", stream_type),
```
haakonvt, line 93: If you can't infer `stream_type`, you need to collapse this to a single semaphore setting that works in all cases

cognite/client/_api/data_modeling/records.py
```diff
@@ -182,4 +206,77 @@ async def upsert(
...
+        sort: Sequence[InstanceSort] | InstanceSort | None = None,
+        limit: int = 10,
+        include_typing: bool = False,
+    ) -> RecordList:
+        """`Filter records in a stream <https://api-docs.cognite.com/20230101/tag/Records/operation/filterRecords>`_.
+
+        Returns records matching the given filters, sorted by ``lastUpdatedTime`` unless a custom
+        ``sort`` is given. This endpoint is not cursor-paged: it returns at most ``limit`` records
+        (max 1000). To page over a large time window, issue multiple calls with partitioned
+        ``last_updated_time`` ranges.
```
haakonvt: We need to support pagination, it is one of the key features of the SDK 😄

cognite/client/_api/data_modeling/records.py
```diff
@@ -182,4 +206,77 @@ async def upsert(
...
+    async def list(
+        self,
+        stream_id: str,
+        *,
+        stream_type: StreamType = _DEFAULT_STREAM_TYPE,
+        last_updated_time: TimeRange | None = None,
+        filter: Filter | None = None,
+        sources: Sequence[RecordSourceSelector] | None = None,
+        sort: Sequence[InstanceSort] | InstanceSort | None = None,
+        limit: int = 10,
```
haakonvt, line 217: We have a default constant that should be used

cognite/client/_api/data_modeling/records.py
```diff
@@ -182,4 +206,77 @@ async def upsert(
...
+
+        Args:
+            stream_id (str): External ID of the stream to query.
+            stream_type (StreamType): Type of the stream ("immutable" or "mutable"). Defaults to "immutable".
+            last_updated_time (TimeRange | None): Filter by last-updated time. **Required for
+                immutable streams** (must include a lower bound).
+            filter (Filter | None): Filter expression (see :mod:`cognite.client.data_classes.filters`).
+            sources (Sequence[RecordSourceSelector] | None): Which container properties to return.
+            sort (Sequence[InstanceSort] | InstanceSort | None): Sort specification(s); up to 5.
+            limit (int): Maximum number of records to return (1-1000). Defaults to 10.
```
haakonvt, line 233:
```suggestion
            limit (int): Maximum number of records to return (1-1000).
```

cognite/client/_api/data_modeling/records.py
```diff
@@ -182,4 +206,77 @@ async def upsert(
...
             resource_path=self._records_url(stream_id, "/upsert"),
             no_response=True,
+            override_semaphore=self._get_semaphore("write", stream_type),
+        )
+
+    async def list(
+        self,
+        stream_id: str,
+        *,
+        stream_type: StreamType = _DEFAULT_STREAM_TYPE,
```
haakonvt: Same argument as above, this will confuse users. If they specify the wrong value here, the list method will still work.
```suggestion
```

haakonvt (commented):
Reviewed the API implementation, will try to cover data classes + tests tomorrow

cognite/client/_api/data_modeling/records.py
andersfylling: I'm cleaning it up here: https://github.com/cognitedata/cognite-sdk-python/pull/2688

the concurrency setup is a bit complicated, so wrote something dumbed down in this PR.

cognite/client/_api/data_modeling/records.py
andersfylling: it will be related to an upcoming concurrency config: https://github.com/cognitedata/cognite-sdk-python/pull/2688

Right now it's a noop. but all operations are stream type bound.

cognite/client/_api/data_modeling/records.py
andersfylling: I don't think I quite understand, could you elaborate?

cognite/client/_api/data_modeling/records.py
andersfylling: Pagination is not a concept for this api endpoint, sadly.

haakonvt: The SDK should still do it to avoid 100 customers doing it in 100 wrong ways xD

andersfylling: Are you saying we should postponed adding streams and records to the SDK until we can add paging in the backend?

haakonvt: No 😅 I have not used records - but I just assumed you can/need to paginate by passing ie last timestamp+1 and so on

haakonvt: I read your docstring now lol. I want the SDK to handle this automatically:

> To page over a large time window, issue multiple calls with partitioned…

andersfylling:
I'll remove the stream_type parameter for now, and make it part of the follow up PR. please ignore these for now

cognite/client/_api/data_modeling/records.py
andersfylling: yeah, my view is we can do it. But I don't think we should support niche pagination. So unless the API supports it, I don't want to implement this for every SDK.

PR #2684, opened 2026-06-15
fix: don't inject null cursor in _list_generator_raw_responses

andersfylling:
Mirrors the conditional guard already present in `_list_generator` (lines 257–258 of `_api_client.py`):

```python
# _list_generator — correct
params["limit"] = current_limit
if next_cursor is not None:
    params["cursor"] = next_cursor

# _list_generator_raw_responses — before this fix
params.update(limit=current_limit, cursor=next_cursor)  # always injects "cursor": null
```

Sending `"cursor": null` on the first request breaks cursorless POST endpoints that reject unknown body fields.


haakonvt (approved):
🦄  a bit of overlap between tests

haakonvt (approved):
🦄

Issue #2682, opened 2026-06-11
Conflicting Sequence parameterizations in IdTransformerMixin break type inference for list iteration


rhuanbarreto:
## Summary

The `IdTransformerMixin` class hierarchy introduces conflicting `Sequence` parameterizations that break type inference for `__iter__` return types in strict type checkers like [ty](https://github.com/astral-sh/ty) (Astral's type checker).

When calling `list(cdf_client.files.list(...))`, the result should be `list[FileMetadata]`, but ty infers `list[HasInternalId | Unknown | HasExternalId]` due to inconsistent generic bases.

## Root Cause

Two independent issues in `cognite/client/data_classes/_base.py`:

### 1. Conflicting `Sequence` parameterizations in mixins

```python
# _base.py
class ExternalIDTransformerMixin(Sequence[HasExternalId], ABC): ...
class InternalIdTransformerMixin(Sequence[HasInternalId], ABC): ...
class IdTransformerMixin(ExternalIDTransformerMixin, InternalIdTransformerMixin): ...
```

`IdTransformerMixin` inherits from both `Sequence[HasExternalId]` and `Sequence[HasInternalId]`, creating conflicting parameterizations of the same generic base. When combined with `CogniteResourceList` (which is also iterable), the type checker unions all element types.

### 2. Unparameterized `UserList`

```python
class CogniteResourceList(UserList, Generic[T_CogniteResource]):
```

`UserList` is not parameterized with `T_CogniteResource`, contributing an `Unknown` element type to the union.

## Impact

Any code doing:
```python
files: list[FileMetadata] = list(cdf_client.files.list(data_set_ids=[...], limit=None))
```
requires a `# type: ignore` or `# ty: ignore[invalid-assignment]` suppression.

This affects all `*List` classes that inherit from `IdTransformerMixin` (e.g., `FileMetadataList`, `AssetList`, `EventList`, `TimeSeriesList`, etc.).

## Suggested Fix

**Minimal fix - remove `Sequence` from mixin bases** (the mixins only iterate via `for x in self`, which they get from the concrete `CogniteResourceList`/`UserList` base, so the `Sequence` base is unnecessary):

```python
# Before:
class ExternalIDTransformerMixin(Sequence[HasExternalId], ABC):
class InternalIdTransformerMixin(Sequence[HasInternalId], ABC):

# After:
class ExternalIDTransformerMixin(ABC):
class InternalIdTransformerMixin(ABC):
```

[...]

PR #2687, opened 2026-06-15
fix: parameterize UserList with T_CogniteResource in CogniteResourceList

rhuanbarreto:
## Summary

Parameterize the `UserList` base class in `CogniteResourceList` so that `self.data` is properly typed as `list[T_CogniteResource]` instead of `list[Any]`. This eliminates the `Unknown` element type that the unparameterized `UserList` contributed to type inference in strict type checkers.

Relates to #2682

## Change

```python
# Before
class CogniteResourceList(UserList, Generic[T_CogniteResource]): ...

# After
class CogniteResourceList(UserList[T_CogniteResource], Generic[T_CogniteResource]): ...
```

## Mypy regression fixes

The stricter `self.data` typing (`list[T_CogniteResource]` instead of `list[Any]`) surfaced 11 pre-existing type-safety gaps. Each is fixed with a targeted suppression:

| File | Error | Fix |
|------|-------|-----|
| `_base.py` (3 errors) | `T_CogniteResource` has no attribute `id`/`external_id`/`instance_id` | `cast(list[Any], self.data)` - accesses are runtime-guarded by `hasattr()` but `T_CogniteResource` is not bound to these protocols |
| `datapoints.py` (3 errors) | `_T_DPS` type variable cannot be `LatestDatapoint` | `# type: ignore [type-var]` - `LatestDatapoint` is not `Datapoints | DatapointsArray` but has the same interface |
| `_api/datapoints.py` (2 errors) | `filter(None, dps_lst)` incompatible with `Iterable[T | None]` | `# type: ignore [arg-type]` - filter correctly removes None values |
| `_api/geospatial.py` (1 error) | `list(feature)` narrows to `list[FeatureCore]`, incompatible with broader union | `# type: ignore [assignment]` - narrowing is correct within the `isinstance` guard |
| `_pandas_helpers.py` (2 errors) | `.timestamp`/`.timezone` not on `CogniteResource` base | `# type: ignore [attr-defined]` - both `Datapoints` and `DatapointsArray` have these attributes |

## Non-breaking

- All unit tests pass (2,755 passed, 1 skipped)
- mypy reports zero new errors on `cognite/`
- No runtime behavior changes - `UserList[T]` is erased at runtime; only type checkers see the difference

### DIFF
--- a/cognite/client/_api/datapoints.py
+++ b/cognite/client/_api/datapoints.py
@@ -718,7 +718,7 @@ class DatapointsAPI(APIClient):
 
             # We should never yield an empty chunk, so we filter out empty or exhausted time series from result
             # (need to rebuild to not keep references to those empty in various private "id lookups")
-            dps_lst = dps_lst_cls(list(filter(None, dps_lst)))
+            dps_lst = dps_lst_cls(list(filter(None, dps_lst)))  # type: ignore [arg-type]
             if not any(dps_lst):
                 if alive_queries:
                     continue
--- a/cognite/client/_api/geospatial.py
+++ b/cognite/client/_api/geospatial.py
@@ -296,7 +296,7 @@ class GeospatialAPI(APIClient):
         if chunk_size is not None and (chunk_size < 1 or chunk_size > self._CREATE_LIMIT):
             raise ValueError(f"The chunk_size must be strictly positive and not exceed {self._CREATE_LIMIT}")
         if isinstance(feature, (FeatureList, FeatureWriteList)):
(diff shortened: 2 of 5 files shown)


rhuanbarreto:
Closing this PR. The UserList parameterization forces stricter typing on self.data, but the existing code intentionally uses hasattr() guards because T_CogniteResource is not bound to HasInternalId/HasExternalId protocols. Fixing the resulting mypy regressions requires casts and type: ignore comments that make the code worse, not better. The cost outweighs the benefit.

PR #2689, opened 2026-06-15
fix(exceptions): Raise specific error CogniteNotFoundError also on 404

haakonvt:
## Summary
`raise_api_error` previously only raised `CogniteNotFoundError` for `400`/`422` responses with a {"missing": [...]} body. Plain `404`s (used by newer APIs like Hosted Extractors and Postgres Gateway) fell through to the generic `CogniteAPIError`.

## Changes:
  - Added case 404, None, None to `raise_api_error` so plain 404s now raise `CogniteNotFoundError` with an empty missing list
  - `_retrieve()` and workflow APIs now catch `CogniteNotFoundError` instead of manually checking `e.code == 404`
  - 5 APIs with `sdk_maturity="alpha"` (hosted extractors: sources, destinations, mappings, jobs; postgres gateway: users) now return None for a not-found singleton rather than raising; type annotations updated accordingly
  - Integration tests updated to assert is None after deletion, covering both the default and `ignore_unknown_ids=True` paths
  - Fixed a codegen bug that generated duplicate TypeAlias definitions in sync wrappers for nested APIs


cognite/client/_basic_api_client.py
```diff
@@ -121,7 +121,9 @@ async def raise_api_error(self, cognite_client: AsyncCogniteClient) -> NoReturn:
                 await self._raise_no_project_access_error(cognite_client, cluster, project)
             case 409, list(), None:
                 self._raise_api_error(CogniteDuplicatedError, cluster, project)
-            case 400 | 422, None, list():
+            case 400 | 404 | 422, None, list():
+                self._raise_api_error(CogniteNotFoundError, cluster, project)
+            case 404, *_:
```
haakonvt: Should prob merge these two `case`s together

cognite/client/_basic_api_client.py
```diff
@@ -121,7 +121,10 @@ async def raise_api_error(self, cognite_client: AsyncCogniteClient) -> NoReturn:
                 await self._raise_no_project_access_error(cognite_client, cluster, project)
             case 409, list(), None:
                 self._raise_api_error(CogniteDuplicatedError, cluster, project)
-            case 400 | 422, None, list():
+            case 400 | 404 | 422, None, list():
+                self._raise_api_error(CogniteNotFoundError, cluster, project)
+            case 404, None, None:
+                self.missing = []  # Some recent APIs return 404 without a missing list when a resource is not found
```
haakonvt, line 127: ^These are the real changes

cognite/client/_api/hosted_extractors/jobs.py
```diff
@@ -68,22 +68,22 @@ async def __call__(
...
-    async def retrieve(self, external_ids: str, ignore_unknown_ids: bool = False) -> Job | None: ...
+    async def retrieve(self, external_ids: str, ignore_unknown_ids: bool = False) -> Job: ...
 
     @overload
     async def retrieve(self, external_ids: SequenceNotStr[str], ignore_unknown_ids: bool = False) -> JobList: ...
 
     async def retrieve(
         self, external_ids: str | SequenceNotStr[str], ignore_unknown_ids: bool = False
-    ) -> Job | None | JobList:
+    ) -> Job | JobList:
```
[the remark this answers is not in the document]
haakonvt: @gemini-code-assist, it can't. then it would be an empty list which is already what is typed

cognite/client/_api_client.py
```diff
@@ -88,11 +105,10 @@ async def _retrieve(
                 headers=headers,
                 semaphore=self._get_semaphore("read"),
             )
-            return cls._load(res.json())._maybe_set_client_ref(self._cognite_client)
-        except CogniteAPIError as e:
-            if e.code != 404:
-                raise
-        return None
+        except CogniteNotFoundError:
+            return None
```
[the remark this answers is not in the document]
haakonvt: @gemini-code-assist we have the following in `raise_api_error`:

```py
  case 404, None, None:
      self.missing = []
      self._raise_api_error(CogniteNotFoundError, ...)
```

scripts/sync_client_codegen/codegen_utils.py
```diff
@@ -121,13 +121,9 @@ def get_all_imports(tree: ast.Module, source_code: str, source_path: Path) -> tu
...
-        # Typically there are type aliases defined in the async API module that we need to import:
-        if maybe in source_code:
-            to_import = maybe.split(": ")[0]
-            import_path = path_as_importable(source_path).replace(".__init__", "")
-            extras.append(f"from {import_path} import {to_import}")
-        # For a lot of 'nested' APIs, these are in parent / init:
-        elif maybe in parent_source:
+        # If the alias is defined in this file, it will be included via module_constants — no import needed.
+        # For nested APIs, the alias lives in the parent init and must be imported explicitly:
+        if maybe in parent_source and maybe not in source_code:
```
audunska, line 126: Don't really understand what's going on here

audunska (approved):

scripts/sync_client_codegen/codegen_utils.py
haakonvt: Deep part of the automatic codegen 😅 Was a subtle bug causing imports + definitions of the same variable

haakonvt:
🦄

PR #2698, opened 2026-06-21
feat(records): sync endpoint

andersfylling:
See https://docs.cognite.com/20230101/records/sync-records-from-a-stream

Since we don't have hierarchical rate limiting I've chosen a number that doesn't cause issue for any configuration: https://docs.cognite.com/api-reference/concepts/20230101/rate-limits#records-request-budgets-detailed-tables

PR #2699, opened 2026-06-22
feat(records): filter endpoint

andersfylling:
Adds the filter records endpoint to the SDK.
documentation: https://api-docs.cognite.com/20230101/tag/Records/operation/filterRecords

---
> [!NOTE]
> Original PR here: https://github.com/cognitedata/cognite-sdk-python/pull/2680

@haakonvt made a few comments that are still relevant:
 - `review comment`: [We need to support pagination, it is one of the key features of the SDK 😄](https://github.com/cognitedata/cognite-sdk-python/pull/2680#discussion_r3421420535)
   `response`: I want to push back on this, and say we shouldn't offer pagination to begin with. At least not until the api has implemented it. There is also the sync endpoint where users can a large volume of records (https://github.com/cognitedata/cognite-sdk-python/pull/2698).


cognite/client/utils/_concurrency.py
```diff
@@ -425,7 +430,7 @@ def __init__(self) -> None:
             write_schema=1,
         )
         self._files = FileConcurrencyConfig(self, read=4, write=2, upload=5, download=5, delete=2, open_files=15)
-        self._records = RecordsGlobalConcurrencyConfig(self, write=20)
+        self._records = RecordsGlobalConcurrencyConfig(self, read=4, write=20)
```
andersfylling, line 429: `4` is the lowest bucket used across all query endpoints https://api-docs.cognite.com/20230101/tag/Records#section/Rate-and-concurrency-limits

cognite/client/utils/_concurrency.py
erlendvollset, line 429: Should the write budget be a lot smaller here? It says the overall concurrency budget is 20 requests, so letting a single client eat all that by default doesn't seem ideal.

Image

andersfylling: sure, but this pr only introduces the filter endpoint. it only changes some comments related to concurrency so I'd say it's off topic for this PR

erlendvollset: alright

erlendvollset (approved):
🦄

PR #2705, opened 2026-06-25
feat(records): aggregate endpoint

andersfylling:
# Description

Introduces support for the aggregate endpoint for the Records collection.


cognite/client/_api/data_modeling/records.py
```diff
@@ -235,6 +237,106 @@ async def upsert(
...
+
+            Aggregate average temperature using typed helpers:
+
+                >>> from cognite.client import CogniteClient
+                >>> from cognite.client.data_classes.data_modeling.records import Avg
+                >>> client = CogniteClient()
+                >>> res = client.data_modeling.records.aggregate(
+                ...     stream_id="my-stream",
+                ...     aggregates={
+                ...         "avg_temp": Avg(["my-space", "sensor", "temperature"]),
```
haakonvt: Let's make it more clear that this is a property reference.

Ref docs:
> the first segment is the space that the container belongs to,
the second segment is the container external id,
the third segment is the property id inside the container.

cognite/client/data_classes/data_modeling/__init__.py
haakonvt, line 1: I think we should keep these out of `cognite/client/data_classes/data_modeling/__init__.py` and only in `...records.py`

cognite/client/_api/data_modeling/records.py
```diff
@@ -235,6 +237,106 @@ async def upsert(
...
+            filter (Filter | dict[str, Any] | None): Filter expression.
+            target_units (RecordTargetUnits | Sequence[RecordTargetUnit] | None): Unit conversion specification.
+            include_typing (bool): Include property type metadata in the response.
+
+        Returns:
+            RecordsAggregation: Aggregate results keyed by the requested aggregate IDs.
+
+        Examples:
+
+            Aggregate average temperature using typed helpers:
```
haakonvt: I also wonder a bit about these examples - they feel like examples of what kind of data that belongs in the Datapoints service, not the Records service. Could you ask Claude to come up with different examples? 😅

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -30,12 +42,224 @@
...
 
+def _dump_aggregate_value(value: Any) -> Any:
+    if isinstance(value, Mapping):
+        return {key: _dump_aggregate_value(val) for key, val in value.items()}
+    if isinstance(value, list | tuple):
+        return [_dump_aggregate_value(item) for item in value]
+    dump = getattr(value, "dump", None)
+    if callable(dump):
+        return _dump_aggregate_value(dump())
+    return value
```
haakonvt: Suggestion: Instead of accepting any dumpable objects here, let's only expect `RecordsAggregate` subclasses.

Another suggestion: `match/case`

```suggestion
def _dump_aggregate_value(value: Any) -> Any:
    match value:
        case Mapping() as m:
            return {key: _dump_aggregate_value(val) for key, val in m.items()}
        case [*items]:
            return [_dump_aggregate_value(item) for item in items]
        case RecordsAggregate() as agg:
            return _dump_aggregate_value(agg.dump())
        case _:
            return value
```

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -30,12 +42,224 @@
...
+
+
+class RecordsAggregate(CogniteResource):
+    """Base class for typed Records aggregate request builders."""
+
+    _aggregate_name: ClassVar[str]
+
+    @classmethod
+    def _load(cls, resource: dict[str, Any]) -> Self:
+        raise NotImplementedError(f"{cls.__name__} is a request builder and cannot be loaded from API responses")
```
haakonvt: In v7 of the SDK, `load` was made public so that all request and response classes could be dumped (and loaded back) into config files, thus the `_load` method here need to be supported.

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -30,12 +42,224 @@
...
+
+class _PropertyAggregate(RecordsAggregate):
+    def __init__(self, property: Sequence[str]) -> None:
+        self.property = list(property)
+
+    def _dump_body(self) -> dict[str, Any]:
+        return {"property": self.property}
+
+
+class Avg(_PropertyAggregate):
```
haakonvt: Suggestion:
```suggestion
class Average(_PropertyAggregate):
```

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -30,12 +42,224 @@
...
+            body["aggregates"] = self.aggregates
+        return body
+
+
+class UniqueValues(_NestedAggregate):
+    """Bucket records by unique property values."""
+
+    _aggregate_name = "uniqueValues"
+
+    def __init__(self, property: Sequence[str], aggregates: Mapping[str, Any] | None = None, size: int | None = None):
```
haakonvt: Just gonna mention this once, but `property: Sequence[str]` accepts `str` due to strings being (infinitely) iterable.

I know we have `SequenceNotStr`, but I don't like that either - for instance it blocks `tuple` of strings... 🤷

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -30,12 +42,224 @@
...
+    def __init__(
+        self,
+        buckets_path: str,
+        window: int,
+        function: Literal[
+            "MovingFunctions.max",
+            "MovingFunctions.min",
+            "MovingFunctions.sum",
+            "MovingFunctions.unweightedAvg",
+            "MovingFunctions.linearWeightedAvg",
```
haakonvt: Is the reason for keeping the `"MovingFunctions."` prefix that we might get more later that we would need to distinguish?

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -127,6 +351,140 @@ def as_ids(self) -> list[RecordId]:
...
 
+class RecordsAggregateResult(CogniteResource):
+    """Base class for typed Records aggregate results."""
+
+    _raw_result: dict[str, Any]
+
+    @classmethod
+    def _load(cls, resource: dict[str, Any]) -> RecordsAggregateResult:
+        for aggregate in ("avg", "count", "min", "max", "sum"):
+            if aggregate in resource:
```
haakonvt: Method returns on first key found, so this O(N) loop seem bad? Are there expected to be more than exactly one top-level key?

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -127,6 +351,140 @@ def as_ids(self) -> list[RecordId]:
...
+        if "fnValue" in resource:
+            return MovingFunctionAggregateResult(fn_value=resource["fnValue"], raw_result=resource)
+        for result_cls in (
+            UniqueValuesAggregateResult,
+            NumberHistogramAggregateResult,
+            TimeHistogramAggregateResult,
+            FilterAggregateResult,
+        ):
+            if result_cls._buckets_key in resource:
+                return result_cls._load(resource)
```
haakonvt: Same comment here, I would rather have you build a mapping as a module-level constant for constant-time lookup

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -127,6 +351,140 @@ def as_ids(self) -> list[RecordId]:
...
+        return cls(resource)
+
+    def __init__(self, raw_result: dict[str, Any]) -> None:
+        self._raw_result = raw_result
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        return self._raw_result
```
haakonvt: This seem a bit lazy tbh 😆 This `camel_case` param is being ignored

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -127,6 +351,140 @@ def as_ids(self) -> list[RecordId]:
...
+        self.aggregates = raw_bucket.get("aggregates", {})
+        self.results = {
+            aggregate_id: RecordsAggregateResult._load(result)
+            for aggregate_id, result in self.aggregates.items()
+            if isinstance(result, dict)
+        }
+
+    @classmethod
+    def _load(cls, resource: dict[str, Any]) -> Self:
+        return cls(resource)
```
haakonvt: This init looks like load

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -127,6 +351,140 @@ def as_ids(self) -> list[RecordId]:
...
+class FilterAggregateResult(_BucketAggregateResult):
+    _buckets_key = "filterBuckets"
+
+
+class RecordsAggregation(CogniteResource):
+    """Aggregate results returned from the Records aggregate endpoint.
+
+    Args:
+        aggregates (dict[str, Any]): Aggregate results keyed by the client-defined aggregate IDs.
+        typing (TypeInformation | None): Optional property typing metadata.
```
haakonvt, line 121: Feels weird to me to have this typing info randomly put here, but I guess you are just mimicking DMS? Should we consider dropping that from /aggregate until someone requests it?

tests/tests_unit/test_api/test_data_modeling/test_records.py
```diff
@@ -262,6 +279,227 @@ def test_upsert_chunks(
...
+                    "filters": [
+                        {"range": {"property": ["createdTime"], "gte": 1}},
+                        {"matchAll": {}},
+                    ],
+                    "aggregates": {"total": {"count": {}}},
+                }
+            },
+        }
+
+    def test_records_aggregation_dump_round_trip(self) -> None:
```
haakonvt: Tested automatically, unless you opted out

cognite/client/data_classes/data_modeling/records.py
andersfylling: `avg` is what is used in the documentation.
image
image

cognite/client/data_classes/data_modeling/records.py
andersfylling: it's because that is what we have in the docs: 

Image

I'm not sure it's worth creating an abstraction for it

cognite/client/data_classes/data_modeling/records.py
andersfylling: it's part of the documentation, so lets keep it so users get a familiar sense when using the python sdk.

image

cognite/client/data_classes/data_modeling/records.py
andersfylling: [6c37d01](https://github.com/cognitedata/cognite-sdk-python/pull/2705/commits/6c37d01e2e121eb503d9d40df19530dbc8abe54c)

Maybe it could be worth doing AST based testing to enforce this?

cognite/client/_api/data_modeling/records.py
andersfylling: specified the property key in the examples

cognite/client/_api/data_modeling/records.py
andersfylling: re-used the examples from the cognite api docs

cognite/client/data_classes/data_modeling/__init__.py
andersfylling: just a heads up it will be inconsistent with the other data classes that are accessible there. Do you want purely aggregate related classes, or all records/streams classes to be in their own module?

cognite/client/data_classes/data_modeling/records.py
andersfylling: [100c3c1](https://github.com/cognitedata/cognite-sdk-python/pull/2705/commits/100c3c14f23e3fb9af1a50a11120aedde5821377)

cognite/client/data_classes/data_modeling/records.py
andersfylling: [100c3c1](https://github.com/cognitedata/cognite-sdk-python/pull/2705/commits/100c3c14f23e3fb9af1a50a11120aedde5821377)

cognite/client/data_classes/data_modeling/records.py
andersfylling: [23d7777](https://github.com/cognitedata/cognite-sdk-python/pull/2705/commits/23d77773d7f1d4ad68357e17ac3396b060202e20)

tests/tests_unit/test_api/test_data_modeling/test_records.py
andersfylling: [23d7777](https://github.com/cognitedata/cognite-sdk-python/pull/2705/commits/23d77773d7f1d4ad68357e17ac3396b060202e20)

cognite/client/data_classes/data_modeling/__init__.py
andersfylling: I've prefixed the response agg types with records, and then i moved the builder types into the records module. Do you want me to move the response types into records as well?

cognite/client/data_classes/data_modeling/records.py
erlendvollset: A tuple is a Sequence, if SequenceNotStr blocks that it needs to be fixed, but I don't believe it does. It was specifically added to address this.

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -30,12 +42,325 @@
...
+
+class _PropertyAggregate(RecordsAggregate):
+    def __init__(self, property: list[str] | tuple[str, ...]) -> None:
+        self.property = list(property)
+
+    def _dump_body(self) -> dict[str, Any]:
+        return {"property": self.property}
+
+
+class Avg(_PropertyAggregate):
```
erlendvollset: Any reason we can't reuse the existing classes in data_classes/aggregations?

cognite/client/data_classes/data_modeling/records.py
haakonvt: Ah, I misremembered, it is the `runtime_checkable` check that fails with tuples. We could probably just remove that (+ any isinstance usage) and otherwise keep `SequenceNotStr` 👍 

```py
from cognite.client.utils.useful_types import SequenceNotStr, is_sequence_not_str

some_tuple = ("a", "b", "c")
some_list = ["a", "b", "c"]
some_string = 'abc'

print(f"{isinstance(some_tuple, SequenceNotStr) = }")
print(f"{isinstance(some_list, SequenceNotStr) = }")
print(f"{isinstance(some_string, SequenceNotStr) = }")
print()
print(f"{is_sequence_not_str(some_tuple) = }")
print(f"{is_sequence_not_str(some_list) = }")
print(f"{is_sequence_not_str(some_string) = }")
```
Prints:
```
isinstance(some_tuple, SequenceNotStr) = False
isinstance(some_list, SequenceNotStr) = True
isinstance(some_string, SequenceNotStr) = False

is_sequence_not_str(some_tuple) = True
is_sequence_not_str(some_list) = True
is_sequence_not_str(some_string) = False
```

cognite/client/data_classes/data_modeling/records.py
haakonvt: Agreed, no need

cognite/client/data_classes/data_modeling/records.py
haakonvt: I still think we should use human readable class names

cognite/client/data_classes/data_modeling/records.py
andersfylling: Those seem to be designed around views, and have different logic for refering to a property. We also have aggregates that aren't handled there++. So to me it got a bit messy, and I went with our own. Though I hope we can consolidate over time.

cognite/client/data_classes/data_modeling/records.py
andersfylling: [6385ad3](https://github.com/cognitedata/cognite-sdk-python/pull/2705/commits/6385ad391f4255605f85d2e694377e55e0288343)

cognite/client/data_classes/data_modeling/records.py
andersfylling: I just copied what I saw other places:
```py
property: list[str] | tuple[str, ...],
```

Fine if we just stick with that for now? The can always be follow up prs for cleaning things up

erlendvollset: Considering how easy it is to fix, let's just do it right away.

cognite/client/data_classes/data_modeling/records.py
erlendvollset: > Those seem to be designed around views

There's no mention of views there, so not sure what you're referring to.

> have different logic for refering to a property.

I thought we were referencing properties in the same way everywhere? What's the deviation?

> We also have aggregates that aren't handled there

You can add new aggregates to that module

andersfylling: image

The property there is a single str, not a list like I use. and whenever you do aggregates you specify a view:
image

If we introduce an abstraction like that before we get view support in, then we limiting the api features. Which I would like to avoid. Given we marked this as an alpha in the sdk, I assume it's fine to introduce a breaking change later on. Wdyt?

cognite/client/data_classes/data_modeling/records.py
andersfylling: [1c96875](https://github.com/cognitedata/cognite-sdk-python/pull/2705/commits/1c96875f8978584d2193d5b21d00d15b3bab729e)

cognite/client/data_classes/data_modeling/records.py
erlendvollset: The view id is not mentioned in the the aggregation data classes, it's only passed to the aggregate method in the dm aggregate API. But the difference in how properties are referenced is enough to keep these separate.

We do need to consider that we'll soon introduce the new search/aggregate endpoints for instances though, and those will mirror the records implementation. I would like us to reuse data classes there where it makes sense, so moving these generic aggregate data classes into a shared `data_classes/data_modeling/aggregations.py` module makes sense.

andersfylling: Alright, I'll wait for a complete review before I make more changes.

andersfylling: I'm against this. I agree on re-use, but I don't want to start implementing DMS support in this PR.. I think it's better to make those changes later on when DMS is ready to implement it. As @haakonvt already stated he wanted these in the records module, I feel we are going a bit back and fourth, and now the scope is increasing to trying to compensate for what DMS might use in the future.

erlendvollset: We know that the instances endpoint will support the same set of features as the records endpoint since it's just a thin wrapper. 

These therefore don't inherently belong to just records, they will be shared between records and instances. 

We have to expect a little bit of back and forth when we're trying to land the right API design. These are largely one-way door decisions, so we have to make an effort to get it right on the first try.

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -111,6 +133,194 @@ def as_ids(self) -> list[RecordId]:
...
+        typing (TypeInformation | None): Optional property typing metadata.
+    """
+
+    def __init__(self, aggregates: dict[str, Any], typing: TypeInformation | None = None) -> None:
+        self.aggregates = aggregates
+        self.results = {
+            aggregate_id: RecordsAggregateResult._load(result)
+            for aggregate_id, result in aggregates.items()
+            if isinstance(result, dict)
+        }
```
erlendvollset: This constructor should be properly typed and the mapping logic moved into the _load method.
```suggestion
    def __init__(self, aggregates: dict[str, RecordsAggregateResult], typing: TypeInformation | None = None) -> None:
        self.aggregates = aggregates
```

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -111,6 +133,194 @@ def as_ids(self) -> list[RecordId]:
...
+
+_BUCKET_RESULT_BY_KEY: dict[str, type[_RecordsBucketAggregateResult]] = {
+    result_cls._buckets_key: result_cls
+    for result_cls in (
+        RecordsUniqueValuesAggregateResult,
+        RecordsNumberHistogramAggregateResult,
+        RecordsTimeHistogramAggregateResult,
+        RecordsFilterAggregateResult,
+    )
+}
```
erlendvollset: I think these AggregateResult classes also should live in the shared `data_classes/data_modeling/aggregates` module. They will be the same for instances and records.

cognite/client/data_classes/data_modeling/records.py
```diff
@@ -111,6 +133,194 @@ def as_ids(self) -> list[RecordId]:
...
+        aggregates: dict[str, Any] | None = None,
+    ) -> None:
+        self.count = count
+        self.value = value
+        self.interval_start = interval_start
+        self.aggregates = aggregates or {}
+        self.results = {
+            aggregate_id: RecordsAggregateResult._load(result)
+            for aggregate_id, result in self.aggregates.items()
+            if isinstance(result, dict)
```
erlendvollset: I might be missing something, but I don't understand why we would want to keep the dictly-typed aggregates here.
```suggestion
        aggregates: dict[str, RecordsAggregateResult] | None = None,
    ) -> None:
        self.count = count
        self.value = value
        self.interval_start = interval_start
        self.aggregates = aggregates
```

cognite/client/data_classes/data_modeling/aggregates.py
```diff
@@ -0,0 +1,307 @@
...
+                hard_bounds=body.get("hardBounds"),
+            )
+        if name == "filters":
+            return Filters(filters=body["filters"], aggregates=body.get("aggregates"))
+        if name == "movingFunction":
+            return MovingFunction(buckets_path=body["bucketsPath"], window=body["window"], function=body["function"])
+        return UnknownAggregate(resource)
+
+
+class _PropertyAggregate(Aggregate):
```
erlendvollset: Super classes of classes that are part of the public API should also be part of the public API.
```suggestion
class PropertyAggregate(Aggregate):
```

cognite/client/data_classes/data_modeling/aggregates.py
```diff
@@ -0,0 +1,307 @@
...
+    _aggregate_name = "max"
+
+
+class Sum(_PropertyAggregate):
+    """Sum aggregate over a container property."""
+
+    _aggregate_name = "sum"
+
+
+class _NestedAggregate(Aggregate):
```
erlendvollset: Super classes of classes that are part of the public API should also be part of the public API.
```suggestion
class NestedAggregate(Aggregate):
```

cognite/client/data_classes/data_modeling/aggregates.py
```diff
@@ -0,0 +1,307 @@
...
+        self,
+        buckets_path: str,
+        window: int,
+        function: MovingFunctions
+        | Literal[
+            "MovingFunctions.max",
+            "MovingFunctions.min",
+            "MovingFunctions.sum",
+            "MovingFunctions.unweightedAvg",
+            "MovingFunctions.linearWeightedAvg",
```
erlendvollset: Better to just append this prefix when dumping?
```suggestion
            "max",
            "min",
            "sum",
            "unweightedAvg",
            "linearWeightedAvg",
```

I would also consider just having either the enum or the literal -- why both?

cognite/client/data_classes/data_modeling/records.py
andersfylling: moved.

cognite/client/data_classes/data_modeling/aggregates.py
andersfylling: chose to simply inline it, instead of documenting a class that doesn't add anything for the users.

cognite/client/data_classes/data_modeling/aggregates.py
andersfylling: chose to simply inline it, instead of documenting a class that doesn't add anything for the users.

cognite/client/data_classes/data_modeling/aggregates.py
andersfylling: I missed that literal part. I've moved to enum only, and then a prefix constant for each member. just to be sure i dont run into surprises down the line.

andersfylling: actually... I'm dropping the prefix constant. writing them out fully makes it easier to grep for. I'll instead add a test case to be sure there arent any typos on the prefix.

cognite/client/data_classes/data_modeling/aggregates.py
erlendvollset: If you're implement something which wants to traverse the aggregate tree it is probably useful to have this abstraction, but can always be added later.

cognite/client/data_classes/data_modeling/aggregates.py
```diff
@@ -0,0 +1,520 @@
...
+    def _dump_body(self) -> dict[str, Any]:
+        body: dict[str, Any] = {
+            "filters": [filter.dump() if isinstance(filter, Filter) else filter for filter in self.filters]
+        }
+        if self.aggregates is not None:
+            body["aggregates"] = self.aggregates
+        return body
+
+
+class MovingFunctions(str, Enum):
```
haakonvt, line 275: (Just an idea 😄 ) Would maybe be nice if this also accepted non-prefixed strings?

```py
>>> MovingFunctions("sum") is MovingFunctions.SUM
True
```

I think you could quite easily get this behavour by implementing `_missing_`.

haakonvt (approved):
Great work here! May want to check in with @erlendvollset before merging, I did not verify that all his previous comments were resolved

erlendvollset:
Lgtm2

andersfylling:
:unicorn:

cognite/client/data_classes/data_modeling/aggregates.py
andersfylling: it's a nice addition, but ill drop it for now. Need to get this SDK out asap :smile:

PR #2708, opened 2026-06-30
chore(records): reduce concurrency limits

andersfylling:
# Description

Start from a more conservative stand and instead increase the concurrency limits over time as we get more traffic.


erlendvollset (approved):

erlendvollset:
🦄

PR #2711, opened 2026-07-07
fix: support Pyodide 314 IN_PYODIDE rename for browser detection

andeplane:
## Summary
Fixes browser/Pyodide detection under **Pyodide 314**, which broke `from cognite.client import CogniteClient` in JupyterLite / browser environments.

Pyodide 314 renamed `pyodide.ffi.IN_BROWSER` → `IN_PYODIDE` ([pyodide#5916](https://github.com/pyodide/pyodide/pull/5916)). `cognite/client/_constants.py` imported the old name and only guarded against `ModuleNotFoundError`:

```python
try:
    from pyodide.ffi import IN_BROWSER
except ModuleNotFoundError:
    IN_BROWSER = False
```

Because `pyodide.ffi` still exists and only the *name* is gone, Python raises `ImportError` (not `ModuleNotFoundError`), so it propagates and crashes at import time:

```
ImportError: cannot import name 'IN_BROWSER' from 'pyodide.ffi'
```

## Fix
Import the `pyodide.ffi` module and read the flag via `getattr`, preferring the new `IN_PYODIDE` name and falling back to the old `IN_BROWSER`, all under a single `ImportError` guard:

```python
try:
    import pyodide.ffi

    IN_BROWSER = bool(getattr(pyodide.ffi, "IN_PYODIDE", getattr(pyodide.ffi, "IN_BROWSER", False)))
except ImportError:
    IN_BROWSER = False
```

This works across old and new Pyodide versions and outside Pyodide entirely.

## Regression coverage
This slipped through because the real Pyodide smoke test (`verify-jupyter` / `scripts/test-pyodide.js`) was pinned to `pyodide@0.29.0`. Added two layers of coverage:

1. **Unit test** `tests/tests_unit/test_constants.py` — reloads `_constants` against simulated `pyodide.ffi` modules exposing:
   | Simulated environment | Expected `IN_BROWSER` |
   | --- | --- |
   | Pyodide 314 (`IN_PYODIDE=True` / `False`) | `True` / `False` |
   | Older Pyodide (`IN_BROWSER=True` / `False`) | `True` / `False` |
   | Both names present (transitional) | `True` (prefers `IN_PYODIDE`) |
   | `pyodide.ffi` present, neither name | `False` |
   | Pyodide not installed | `False` |

   Runs across the full Python **3.10–3.14** unit-test matrix.

[...]

PR #2719, opened 2026-07-13
chore: make pyodide import into a util fn to simplify testing

haakonvt:
In https://github.com/cognitedata/cognite-sdk-python/pull/2711, the change was added to support the new variable name for "are we running with pyodide". It added tests that relied on `importlib.reload` which should never be used in testing.

This PR changes this to a simple function that is easy to test instead.


tests/tests_unit/test_constants.py
```diff
@@ -46,22 +23,12 @@ class TestBrowserDetection:
...
-        monkeypatch: pytest.MonkeyPatch,
-        restore_constants: None,
-        ffi_attrs: dict[str, bool],
-        expected: bool,
-    ) -> None:
-        _reload_with_fake_pyodide_ffi(monkeypatch, ffi_attrs)
-        assert constants.IN_BROWSER is expected
-        assert constants._RUNNING_IN_BROWSER is expected
+    def test_detects_pyodide_across_versions(self, ffi_attrs: dict[str, bool], expected: bool) -> None:
+        ffi_mod = types.ModuleType("pyodide.ffi")
```
polomani, line 27: cool!

polomani (approved):
looks even more good to me

haakonvt:
🦄

PR #2720, opened 2026-07-13
feat(config): Add `ssl_config` to `global_config` (DM-3969)

haakonvt:
**feat(config): add `ssl_context` to `GlobalConfig`**

Adds `global_config.ssl_context` (`ssl.SSLContext | None`) so users can configure custom certificate verification without relying on environment variables.

**Motivation:** The SDK migrated from `requests` to `httpx`, which dropped `REQUESTS_CA_BUNDLE` support. Users with corporate/internal CAs previously had to fall back to `SSL_CERT_FILE`/`SSL_CERT_DIR`. This gives them a clean in-code alternative.

### DIFF
--- a/cognite/client/config.py
+++ b/cognite/client/config.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import getpass
 import pprint
 import re
+import ssl
 import warnings
 from typing import Any, ClassVar, NoReturn, overload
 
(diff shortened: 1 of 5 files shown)


cognite/client/config.py
```diff
@@ -156,6 +166,10 @@ def apply_settings(self, settings: dict[str, Any] | str) -> None:
                 "Cannot apply 'concurrency_settings' via apply_settings. Modify the individual attributes on "
                 "'global_config.concurrency_settings' instead."
             )
+        if "ssl_context" in loaded:
+            raise ValueError(
+                "Cannot apply 'ssl_context' via apply_settings. Set 'global_config.ssl_context' directly instead."
+            )
```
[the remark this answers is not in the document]
haakonvt: @gemini-code-assist The round-trip argument doesn't hold, vars(global_config) includes _concurrency_settings and _max_workers (private names), so it already can't be round-tripped through apply_settings. The explicit error is intentional and consistent with how concurrency_settings is handled.

The real reason to keep it: ssl_context can't meaningfully come from YAML/JSON (you can't serialize an SSLContext), so if someone passes it via apply_settings it's almost certainly a bug. The clear error message is better than silently trying to setattr a dict or string onto it.

cognite/client/config.py
```diff
@@ -30,6 +31,9 @@ class GlobalConfig:
         max_connection_pool_size (int): The maximum number of connections which will be kept in the SDKs connection pool.
             Defaults to 20.
         disable_ssl (bool): Whether or not to disable SSL. Defaults to False
+        ssl_context (ssl.SSLContext | None): Custom SSL context for certificate verification. Overrides the
+            default certifi bundle. Ignored when ``disable_ssl`` is True. Must be set before the first API
+            request. Defaults to None. See https://cognite-sdk-python.readthedocs-hosted.com/en/latest/settings.html#ssl-certificate-configuration
```
haakonvt, line 36: Local render of docs:
Image

vvb16 (approved):

haakonvt:
🦄

cognite/client/config.py
haakonvt, line 36: @yangru96 oops I merged a new /latest/ link 😅

PR #2724, opened 2026-07-20
fix(pandas): version-conditional timestamp resolution for to_pandas() (ns on v2, ms on v3)

haakonvt:
Pandas v3 removed the implicit nanosecond default for datetimes, so `to_pandas()` output would get arbitrary precision depending on input. This pins resolution explicitly: ns on pandas v2 (matching old behavior), ms on pandas v3 (matching CDF's native resolution), via shared helpers in `_pandas_helpers.py` used across every timestamp-producing code path.

"Single" resource objects (aka not the _list_ version), now delegate to the list-class `to_pandas` method to avoid having duplicated logic. This relies on `_LIST_CLASS` / `_RESOURCE` back-references staying in sync, so `test_meta.py` now enforces it: every `CogniteResourceList._RESOURCE` must point to a resource class whose `_LIST_CLASS` points right back. A class added without wiring up its counterpart now fails CI instead of silently duplicating conversion logic later. Also adds a `test_standalone_to_pandas_allowlist` test documenting the deliberate exceptions (e.g. `Datapoint`, `RowCore`) that don't follow this pattern.

For the pandas versioning itself, added `tests/tests_unit/test_utils/test_pandas_helpers.py` with a `force_pandas_major_version` fixture that monkeypatches the version check, so both the ns and ms branches run in every CI invocation regardless of which pandas is actually installed.


haakonvt:
Merged in #2780

PR #2725, opened 2026-07-27
feat(SimInt): Add `delete_oldest` option to model revision creation

mohammad-arif662:
## Description
Adds a `delete_oldest` parameter to `client.simulators.models.revisions.create()`, matching the `deleteOldest` flag added on the create model revision API. When `True`, the API deletes the oldest revision(s) for a model if the per-model revision limit has been reached, enabling rolling revision cleanup instead of failing the create call.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/tests/tests_unit/test_api/test_simulators/test_model_revisions.py
+++ b/tests/tests_unit/test_api/test_simulators/test_model_revisions.py
@@ -20,65 +20,11 @@ TEST_MODEL_REVISION_ITEM_RESPONSE_FIELDS = {
 
 class TestModelRevisions:
     @pytest.mark.parametrize(
-        "write_input,delete_oldest,expected_request_body",
+        "delete_kw",
         [
-            pytest.param(
-                SimulatorModelRevisionWrite(
-                    external_id="sdk-test-revision",
-                    model_external_id="sdk-test-model",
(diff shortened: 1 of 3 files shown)


tests/tests_unit/test_api/test_simulators/test_model_revisions.py
```diff
@@ -0,0 +1,115 @@
...
+                    "deleteOldest": False,
+                    "items": [
+                        {
+                            "externalId": "sdk-test-revision",
+                            "modelExternalId": "sdk-test-model",
+                            "fileId": 1,
+                        }
+                    ],
+                },
+                id="create_model_revision_default_delete_oldest",
```
haakonvt: Few things - almost the entire payload is constant and repeated here. How about a simple `pytest.mark.parametrize` using:
```
"delete_kw",
(
    ({"delete_oldest": True}),
    ({"delete_oldest": False}),
    ({}),
)
```
Then you can `**delete_kw` in the _single_ unit test.

haakonvt (commented):
Looks like a very handy feature. Only one comment from me on the testing, otherwise LGTM!

haakonvt (approved):
🦄

PR #2730, opened 2026-07-30
feat(credentials): let users choose the account for interactive login (DM-3512)

yangru96:
## Description
OAuthInteractive silently used the first account it found in the token cache, so users signed in with several accounts had no way to choose and often ended up on the wrong one.

Add `prompt` and `login_hint` to OAuthInteractive, its `default_for_entra_id` helper, `(Async)CogniteClient.default_oauth_interactive` and the YAML/JSON config accepted by `load()`. Pass `prompt="select_account"` to pick the account in the browser, or `login_hint` to name it up front and skip being asked. Once signed in, later token refreshes stay on that account without prompting again.

Behavior change: when the token cache holds several accounts and neither argument is given, the browser now opens with an account picker instead of silently using an arbitrary account. A single cached account is still reused silently, so users with one account are unaffected.

## Checklist:
- [x Tests added/updated.
- [x] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change. If a new method has been added it should be referenced in [cognite.rst](https://github.com/cognitedata/cognite-sdk-python/blob/master/docs/source/cognite.rst) in order to generate docs based on its docstring.


cognite/client/credentials.py
```diff
@@ -27,6 +27,9 @@
 
 _TOKEN_EXPIRY_LEEWAY_SECONDS_DEFAULT = 30  # Do not change without also updating all the docstrings using it
 
+InteractivePrompt = Literal["select_account", "login", "consent"]
```
haakonvt, line 31: Good coverage on what `"select_account"` does, but `"login"` and `"consent"` are a bit lacking in this regard.

cognite/client/credentials.py
```diff
@@ -649,6 +652,8 @@ class OAuthInteractive(_OAuthCredentialProviderWithTokenRefresh, _WithMsalSerial
         redirect_port (int): Redirect port defaults to 53000.
         token_cache_path (Path | None): Location to store token cache, defaults to os temp directory/cognitetokencache.{client_id}.bin.
         token_expiry_leeway_seconds (int): The token is refreshed at the earliest when this number of seconds is left before expiry. Default: 30 sec
+        prompt (InteractivePrompt | None): What you are asked when signing in, e.g. ``"select_account"`` to pick which account to use. Setting it always takes you to the browser, even when you are already signed in, which is what you want when switching account. Default: None, meaning you go straight in with the account you signed in with before, and are only asked to pick when there are several.
+        login_hint (str | None): Username (e.g. email) of the account to sign in with. Set it to always use that account without being asked. Default: None
```
haakonvt: When is `login_hint` really needed? I think the interplay between existing behaviour + prompt + login_hint would be a good Note-section in the docstring that you can refer to instead of having a "short and dense" explanation here.

cognite/client/credentials.py
```diff
@@ -701,18 +732,40 @@ def client_id(self) -> str:
...
+        return self.__prompt
+
+    @property
+    def login_hint(self) -> str | None:
+        return self.__login_hint
+
     def _refresh_access_token(self) -> tuple[str, float]:
-        # Try the in-memory token cache silently (MSAL checks AT first, then RT automatically).
-        # Falls through to interactive flow if nothing usable is found.
         credentials = None
```
haakonvt, line 771: There are some existing comments in this function that have been removed (I assume because they are no longer correct). 

It seems to me that the logic is growing in complexity with this PR and that having some explanation of "what is going on" would be appropriate 😄

cognite/client/credentials.py
```diff
@@ -701,18 +732,40 @@ def client_id(self) -> str:
...
+                scopes=self.__scopes,
+                port=self.__redirect_port,
+                prompt=prompt,
+                login_hint=login_hint,
+            )
         self._verify_credentials(credentials)
+
+        self.__has_signed_in = True
+        claims = credentials.get("id_token_claims", {})
+        if username := claims.get("preferred_username") or claims.get("upn"):
```
haakonvt, line 802: These look wildly magical - where do `preferred_username` and `upn` come from?

tests/tests_unit/test_credential_providers.py
haakonvt, line 1: I am a bit sceptical to the real value of several of these tests (and existing tbh).

To what extent have you tested this new logic yourself? Especially with the caching it is super easy to screw something up (speaking from experience here 😅 )

haakonvt (commented):
A few late night comments from me, overall looks solid. I will attempt to test it out in the morning and connect you with the guy that reported this initially, cc @spex66

cognite/client/credentials.py
yangru96: Fair - I agree they look magical without context. They are from id_token_claims from MSAL response. `id_token_claims` is the decoded id token in the response payload. Here is the [id token claims reference](https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference) and code location `masl/oauth2cli/oidc.py:165-172` in msal sdk. 

The reason this is hard to look up is that MSAL never defines a schema for the
response — it's an untyped dict, and the docstring only promises that it "typically
contains an `access_token` key". Happy to add a short comment pointing at the docs
if that helps the next reader.

cognite/client/credentials.py
yangru96: Comments are back and updated :)

cognite/client/credentials.py
yangru96: Good call, moved to a Note-section and added a pointer in the the parameter docs.

cognite/client/credentials.py
yangru96: Added a comment on those lines with the docs link, so the next reader
doesn't have to go digging through msal.

tests/tests_unit/test_credential_providers.py
yangru96: Fair. The reason I'd keep them is that the logic branches on three things at once: how many accounts are cached (none / one / several), whether prompt was passed, and whether we already signed in this session. The "several accounts" branch also needs a second real account, so it's hard to hit by hand.

I just tested it for real though: a temp script running locally against the "quick start sdk" app registration in Azure (Graph as the scope), with a real on-disk cache and a fresh provider each step.

1. Fresh cache, no args — browser opens, no picker.
2. Run again, no args — no browser, cached account reused.
3. login_hint for that same account — no browser, reused silently.
4. login_hint for an account I've never used — browser opens with login_hint=... in the authorize URL; whether Entra pre-fills it or shows the picker depends on what sessions the browser already has.
5. prompt="select_account" — picker appears even though I'd signed in.

All as expected.

cognite/client/credentials.py
```diff
@@ -649,6 +653,34 @@ class OAuthInteractive(_OAuthCredentialProviderWithTokenRefresh, _WithMsalSerial
         redirect_port (int): Redirect port defaults to 53000.
         token_cache_path (Path | None): Location to store token cache, defaults to os temp directory/cognitetokencache.{client_id}.bin.
         token_expiry_leeway_seconds (int): The token is refreshed at the earliest when this number of seconds is left before expiry. Default: 30 sec
+        prompt (InteractivePrompt | None): What the browser asks you the first time you sign in, e.g. ``"select_account"`` to pick which account to use. See the note on account selection below. Default: None
+        login_hint (str | None): Username (e.g. email) of the account to sign in with. See the note on account selection below. Default: None
+
+    Note:
```
haakonvt, line 659: This is great! 💯

scripts/sync_client_codegen/sync_client_template.txt
haakonvt, line 1: Did you render docs locally and verify that these links work as intended? 
```
:class:`~cognite.client.credentials.OAuthInteractive`.
```

haakonvt (approved):
Great job with this! One comment re. docs otherwise LGTM

scripts/sync_client_codegen/sync_client_template.txt
yangru96: yes, I have rendered the docs locally and verified that the links work as intended

andreavs (commented):
🦄

PR #2733, opened 2026-07-31
fix(tests): use real function id in workflow version list fixture

aaditijoag-cognite:
## Description
The backend now validates that a FunctionTaskParameters external_id references an existing function on workflow version upsert, so the dummy "non-existing-function" id caused test_list_workflow_versions to fail. Use the session-scoped `a_function` fixture instead, matching the pattern already used by _new_async_workflow_version.

## Checklist:
- [X] Tests added/updated.
- [] Documentation updated. Documentation is generated from docstrings - these must be updated according to your change.
- [X] The PR title follows the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) spec.


polomani (approved):
🦄

PR #2732, opened 2026-07-31
feat(transformations): add external data sources client API

Khyat-Cognite:
[CDF-28251](https://cognitedata.atlassian.net/browse/CDF-28251)

Stacked on #2729 (`feat/transformations-externaldata-models`) — **review #2729 first**; this PR's diff is client-API only, no data class file is touched.

Adds `client.transformations.external_data_sources` for the `/transformations/externaldata` endpoint.

## Endpoint surface

| Method | Endpoint | Capability |
|---|---|---|
| `list(limit=25)` / `__call__(chunk_size, limit)` | `GET /transformations/externaldata` | `transformationsExternalDataSourcesAcl:READ` |
| `upsert(data_source)` | `POST /transformations/externaldata` | `transformationsExternalDataSourcesAcl:WRITE` |
| `delete(external_id)` | `POST /transformations/externaldata/delete` | `transformationsExternalDataSourcesAcl:WRITE` |
| `verify_usability(external_id)` | `POST /transformations/externaldata/usability` | `transformationsExternalDataSourcesAcl:USE` |

There is no `/byids` endpoint, so there is no `retrieve()`.

## Tests

Unit (`tests/tests_unit/test_api/test_transformations_external_data.py`, 8 tests) — asserts the wire, not dump/load round trips:

- `list()` parses into `OneLakeExternalDataSource` with typed nested settings
- `__call__()` yields one-by-one, and `__call__(chunk_size=25)` yields 25 + 5 for a 30-item response
- `upsert()` single and sequence — request items carry `"format": "one_lake"` and complete nested settings including `clientSecret`
- `delete()` sends only `items`
- `verify_usability()` sends a flat `externalId` and unwraps the nested response; the missing-`usableVersion` case gives `is_usable is False`

Integration (`tests/tests_integration/test_api/test_transformations/test_externaldata.py`) — upsert → list → verify_usability → delete against a live Fabric lakehouse, skipped unless `ONELAKE_CLIENT_ID`, `ONELAKE_TENANT_ID`, `ONELAKE_CLIENT_SECRET`, `ONELAKE_WORKSPACE_ID` and `ONELAKE_CONTAINER_ID` are set.

Also verified locally: sync codegen `verify` clean, `pre-commit` clean (the only mypy errors are pre-existing `tests/utils.py` numpy-typing ones, present on the base commit too), full unit suite failure set identical to the base commit, and `make linkcheck`/`make html` with `-W` both succeed with the new autosummary pages generated.

[...]

PR #2729, opened 2026-07-30
feat(transformations): add external data source data classes

Khyat-Cognite:
[CDF-28228](https://cognitedata.atlassian.net/browse/CDF-28228)

## Summary

Data-class layer for the transformations external data sources API — Fabric OneLake connections for `ext_onelake()` in transform SQL. **Data models only**, no client API or docs guide (see follow-up plan below).

**Supersedes #2714.** That PR targeted `/transformations/external_data`; the contract has since moved to **`/transformations/externaldata`** with a materially different payload:
- `OneLakeLocationDescription` is now `workspaceId`/`containerId` (UUIDs), not `workspaceName`/`containerName`
- `settings`, `credentials`, `locationDescription`, `format` are all required on both read and write
- `clientSecret` is required on every write (full replace); never returned on read
- The usability response wraps the id: `{"externalId": {"externalId": "..."}, "usableVersion": "..."}`
- The endpoint is public beta

### Endpoints / ACL

| Endpoint | Method | ACL |
|----------|--------|-----|
| `/transformations/externaldata` | GET list, POST upsert (201) | READ / WRITE |
| `/transformations/externaldata/delete` | POST | WRITE |
| `/transformations/externaldata/usability` | POST | USE |

### Production code
- `cognite/client/data_classes/transformations/externaldata.py` — read/write/list models, OneLake credentials/settings, usability
- `cognite/client/data_classes/capabilities.py` — `TransformationsExternalDataSourcesAcl`
- `cognite/client/data_classes/transformations/__init__.py` — re-exports (kept out of the top-level `data_classes/__init__.py` per review)
- `docs/source/transformations.rst` — one `automodule` line for the new module

### Tests
- `tests/tests_unit/test_base.py` — two exclusion lines (mirroring the existing `Source`/`SourceList` exclusions) so the generic `as_write` tests correctly skip this hierarchy; round-trip dump/load is otherwise covered automatically by that same generic suite
- `tests/tests_unit/test_data_classes/test_capabilities.py` — new ACL entries in `all_acls()`
- `tests/tests_unit/test_data_classes/test_transformations/test_externaldata.py` — format dispatch, unknown-format fallback (including the order-dependent behavior inherited from `CogniteResourceList` when an unknown-format item is first in a list — a known trade-off already present for `SourceList`, not something new here), `as_write` raising, secret redaction in `repr`/`str`, usability status

### What comes next
| PR | Scope |
|----|-------|
[...]

PR #2734, opened 2026-07-31
docs(transformations): add external data sources guide

Khyat-Cognite:
Third and last PR in the stack: #2729 (data classes) → #2732 (client API) → this one. **Review #2732 first**; this diff is docs only.

Adds `docs/source/transformations_external_data_sources.rst`, reached from a nested toctree on the Transformations page — the feature is one sub-API of transformations, not a family of its own, so it belongs under that page rather than as a peer entry in `index.rst`.

## Contents

Concepts (format, credentials, location, data set scoping), a per-operation permission table, then task-oriented sections: register a data source, verify it is usable, read its tables from transformation SQL, list data sources, rotate the client secret, delete a data source.

### DIFF
--- a/docs/source/transformations.rst
+++ b/docs/source/transformations.rst
@@ -57,6 +57,11 @@ Transformation External Data Sources
 
    AsyncCogniteClient.transformations.external_data_sources
 
+.. toctree::
+   :maxdepth: 2
+
+   transformations_external_data_sources
+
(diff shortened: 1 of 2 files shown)


haakonvt (approved):

Khyat-Cognite:
🦄

PR #2747, opened 2026-08-06
fix(records): validate property paths instead of splitting bare strings

andersfylling:
A property path is a sequence of segments, but `SequenceNotStr` is a typing-only protocol — `str` satisfies it at runtime. So a bare string is accepted and expanded into one segment per character:

```python
Count(property="temperature_c").property
# ['t', 'e', 'm', 'p', 'e', 'r', 'a', 't', 'u', 'r', 'e', '_', 'c']
# CogniteAPIError: Property has incorrect number='13' of segments. Expected 1 or 3
```

Reported in review of the records aggregate work.

`validate_property_path` is added to `data_modeling/_validation.py` and applied everywhere a path or a list of property identifiers is taken: all eight aggregate builders, `RecordTargetUnit.property` and `RecordSourceSelector.properties` (same hole, and the API requires `minItems: 1` there). Paths are at most three segments, so every segment is type checked, and an empty path is rejected. The error names the argument and shows the expected shape.

Nothing that previously produced a valid request changes behaviour — every input this rejects was already rejected by the API.

First commit is the failing tests, then the fix, then the review follow-up.


cognite/client/data_classes/data_modeling/_validation.py
```diff
@@ -43,8 +45,38 @@
...
+
 def validate_data_modeling_identifier(space: str | None, external_id: str | None = None) -> None:
     if space and space in RESERVED_SPACE_IDS:
         raise ValueError(f"The space ID: {space!r} is reserved. Please use another ID.")
     if external_id and external_id in RESERVED_EXTERNAL_IDS:
         raise ValueError(f"The external ID: {external_id!r} is reserved. Please use another ID.")
+
+
+def validate_property_path(
+    property: SequenceNotStr[str], argument: str = "property", hint: str = PROPERTY_PATH_HINT
```
haakonvt: nit: I would suggest not shadowing built-in `property`. Not a big deal tbh
```suggestion
    prop: SequenceNotStr[str], argument: str = "property", hint: str = PROPERTY_PATH_HINT
```

cognite/client/data_classes/data_modeling/_validation.py
```diff
@@ -43,8 +45,38 @@
...
+) -> list[str]:
```
haakonvt, line 64: The `str` in the "inner" return annotation `list[str]` is not actually validated, just the sequence part is validated. Also accepts empty sequences:

```py
>>> validate_property_path([])
[]
>>> validate_property_path(["a"])
['a']
>>> validate_property_path([1, 2])
[1, 2]
```
Since the expected number of elements here is low, like 3, I think doing isinstance checks on all of them is ok.

cognite/client/data_classes/data_modeling/_validation.py
```diff
@@ -43,8 +45,38 @@
 )
 
 
+PROPERTY_PATH_HINT = (
```
haakonvt, line 48: Great error message btw

cognite/client/data_classes/data_modeling/_validation.py
andersfylling: Done, renamed to `prop`. All call sites pass it positionally, so nothing else changed.

cognite/client/data_classes/data_modeling/_validation.py
andersfylling: Good catch, both fixed in d650c125 — every segment is now `isinstance`-checked and an empty path raises `ValueError`. Empty was never valid anyway: paths need 1 or 3 segments, and `RecordSourceSelector.properties` is `minItems: 1`.

```py
>>> validate_property_path([])
ValueError: 'property' must not be empty. A property is addressed by its path: ...
>>> validate_property_path([1, 2])
TypeError: 'property' must be a sequence of strings, but 1 is of type int. ...
```

cognite/client/data_classes/data_modeling/_validation.py
```diff
@@ -43,8 +48,56 @@
...
+
+
 def validate_data_modeling_identifier(space: str | None, external_id: str | None = None) -> None:
     if space and space in RESERVED_SPACE_IDS:
         raise ValueError(f"The space ID: {space!r} is reserved. Please use another ID.")
     if external_id and external_id in RESERVED_EXTERNAL_IDS:
         raise ValueError(f"The external ID: {external_id!r} is reserved. Please use another ID.")
+
+
+def validate_str_sequence(value: Any, argument: str, hint: str) -> list[str]:
```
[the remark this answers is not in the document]
andersfylling: Outdated — `validate_str_sequence` was folded into `validate_property_path` before the current head, and that signature takes `SequenceNotStr[str]`, no `Any`.

cognite/client/data_classes/data_modeling/_validation.py
```diff
@@ -43,8 +48,56 @@
...
+        raise TypeError(f"{argument!r} must be a sequence of strings, not the string {value!r}. {hint}")
+    if not isinstance(value, Sequence):
+        raise TypeError(f"{argument!r} must be a sequence of strings, not {type(value).__name__}. {hint}")
+    if not value:
+        raise ValueError(f"{argument!r} cannot be empty. {hint}")
+    if invalid := [item for item in value if not isinstance(item, str)]:
+        raise TypeError(
+            f"{argument!r} must be a sequence of strings, but got {invalid[0]!r} "
+            f"of type {type(invalid[0]).__name__}. {hint}"
+        )
```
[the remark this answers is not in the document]
andersfylling: Outdated — that comprehension is gone. The per-segment check added in d650c125 is a `for` loop that raises on the first non-string segment.

haakonvt (approved):

andersfylling:
:unicorn:

PR #2751, opened 2026-08-06
fix(records): reject a filter limit above 1000 instead of truncating

andersfylling:
The filter endpoint answers with a single page of at most 1000 records and returns no cursor, so nothing can page past it. A larger limit is silently capped:

```python
res = client.data_modeling.records.filter(stream_id=STREAM_ID, limit=5000)
len(res)  # <= 1000, no warning, no error
```

`None` and `-1` — unlimited elsewhere in the SDK — behave the same way. This now raises, and the message says what to do instead: narrow `last_updated_time` and repeat, or follow the change feed with `sync()`, which does paginate.

**Worth a second opinion:** this is the only PR in the batch that turns a currently-succeeding call into an error. Silently returning a fraction of what was asked for seems worse than failing, and Records is alpha on the SDK side, so it is filed as a `fix`. The alternative is capping with a warning — say the word and I will switch it.

`sync`'s limit docstring is corrected in passing: it does follow the cursor, so it genuinely serves limits above 1000.

First commit is the failing tests, second is the fix, including the regenerated sync client.


cognite/client/_api/data_modeling/records.py
```diff
@@ -29,6 +29,20 @@
     from cognite.client import AsyncCogniteClient
     from cognite.client.config import ClientConfig
 
+# The filter endpoint answers with a single page and returns no cursor, so nothing can page past it:
+_FILTER_MAX_LIMIT = 1000
+
+
+def _validate_filter_limit(limit: Any) -> None:
```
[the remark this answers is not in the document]
erlendvollset: Should just be typed as `int`

cognite/client/_api/data_modeling/records.py
```diff
@@ -29,6 +29,20 @@
...
+    hint = (
+        "The filter endpoint returns a single page and does not paginate; to read more records, narrow "
+        "'last_updated_time' and repeat, or follow the change feed with sync()."
+    )
+    if isinstance(limit, bool) or not isinstance(limit, int):
+        raise TypeError(f"'limit' must be an int between 1 and {_FILTER_MAX_LIMIT}, not {type(limit).__name__}. {hint}")
```
haakonvt: Although `isinstance(True, int) -> True`, I feel like just an `if not isinstance(limit, int):` is sufficient here. But no big deal. Assuming int subclasses are exceedingly rare doing `type(limit) is not int` is also an option 😅

cognite/client/_api/data_modeling/records.py
```diff
@@ -29,6 +29,20 @@
...
+    if not 1 <= limit <= _FILTER_MAX_LIMIT:
+        raise ValueError(f"'limit' must be between 1 and {_FILTER_MAX_LIMIT}, got {limit}. {hint}")
```
haakonvt:
```suggestion
        raise ValueError(f"'limit' must be between 1 and {_FILTER_MAX_LIMIT} (inclusive), got {limit}. {hint}")
```

tests/tests_unit/test_api/test_data_modeling/test_records.py
```diff
@@ -947,3 +947,18 @@ def test_sync_record_deleted_tombstone(self) -> None:
         assert record.status == "deleted"
         assert record.properties is None
         assert "properties" not in record.dump()
+
+
+class TestRecordsAPIFilterLimit:
+    """The filter endpoint returns a single page of at most 1000 records and has no cursor, so a
+    bigger limit is silently truncated to 1000 instead of being reported."""
+
```
haakonvt: These tests seem self-explanatory
```suggestion
```

haakonvt (approved):
Nice!

andersfylling:
:unicorn:

PR #2753, opened 2026-08-06
feat(streams): rename StreamTemplateWriteSettings to StreamWriteSettings

andersfylling:
`StreamTemplateWriteSettings` is the settings object of `StreamWrite`; naming it after the template described only the one settings shape that exists today. Renamed to `StreamWriteSettings`, with the docstring examples that import it updated alongside.

Streams are alpha on the SDK side, so the old name is removed rather than kept as a deprecated alias — say the word if you would rather ship an alias for one release first.

Typed as `feat` deliberately: it is a visible rename of a public class, so it belongs in the changelog, but marking it breaking would make release-please cut a major version.

Single commit — a rename has no failing-test story, so the test updates are part of it.


haakonvt (approved):
LGTM given alpha status. Consider doing `StreamTemplateWriteSettings = StreamWriteSettings`.

andersfylling:
:unicorn:

PR #2754, opened 2026-08-07
refactor: update ruff version and refactor based on new linter rules

MortGron:
## Description
Update ruff version. Some new linter rules, first and foremost the new rule none-not-at-end-of-union (RUF036) introduced in v 0.16, causes reformatting of a number of files.

## Checklist:
- [ ] Tests added/updated.


pyproject.toml
```diff
@@ -73,7 +73,7 @@ dev = [
     "types-simplejson >=3.19.0.20240801",
     "types-pytz (>=2025.2.0.20250516, <2026)",
     "tenacity (>=9.1.2, <10)",
-    "ruff ==0.13.2",
+    "ruff ==0.16.2",
```
haakonvt: Uhm, why is this pinned? That must be a mistake? 🤔

haakonvt (dismissed):
Nice! One comment (pre-existing issue):

pyproject.toml
MortGron: Removed the pin of the patch specifier

haakonvt (approved):
🦄  - dont think we need to pin the minor version but this is at least better 👍

PR #2755, opened 2026-08-07
feat(data modeling): merge records sync and sync_resume into a chunk-yielding generator

evertoncolling:
## Description
`client.data_modeling.records.sync()` and `sync_resume()` never returned whenever a page held fewer records than `limit`.

`RecordsAPI._sync` delegated to the generic `_list` paging helper, whose loop terminates on:

```python
# cognite/client/_api_client.py:328
if total_retrieved == limit or next_cursor is None:
    break
```

Neither condition can be reached for this endpoint:

- The sync endpoint **always** returns a `nextCursor` — that is precisely what makes the change feed resumable. It signals end-of-feed with `hasNext: false`, which the paging loop never read.
- So the loop could only exit when the feed happened to return exactly as many records as were asked for.

## The new API
Following the review discussion, the fix grew into a redesign: `sync` and `sync_resume` are merged into a **single generator method** that issues one POST per chunk and yields each `SyncRecordList` as it arrives, until the feed is exhausted (`hasNext: false`). Nothing is accumulated in memory.

```python
# Start from a relative time, iterate the whole feed:
for chunk in client.data_modeling.records.sync(stream_id="my-stream", initialize_cursor="7d-ago"):
    process(chunk)
    last_cursor = chunk.cursor  # persist after processing each chunk

# Later, resume from the stored cursor:
for chunk in client.data_modeling.records.sync(stream_id="my-stream", cursor=last_cursor):
    ...

# Manual single-chunk control:
chunk = next(client.data_modeling.records.sync(stream_id="my-stream", cursor=last_cursor))
```

- Exactly one of `initialize_cursor` / `cursor` is required; `@overload` stubs enforce this at type-check time (a lazy runtime `ValueError` backstops untyped callers).
- `chunk_size` (1–1000, default 1000) replaces `limit` as the per-request page size. It is deliberately **not** named `limit`: SDK-wide, `limit` means "total items, `None` fetches everything", while `chunk_size` is the established name for per-yielded-chunk size (see the `__call__` iterators). There is no `chunk_size=None` single-record mode since each chunk carries the cursor needed to resume.
- The generator terminates at end-of-feed rather than polling forever — the endpoint has no server-side poll timeout (unlike datapoint subscriptions), so tailing is done by calling `sync(cursor=...)` again later.
- The sync client gets the equivalent `Iterator[SyncRecordList]` via codegen (`"sync"` added to `SYNC_METHODS_TO_KEEP` because mypy requires async-generator overload stubs to be plain `def`).

## Breaking changes
Records is an alpha feature behind `FeaturePreviewWarning`.

- `sync_resume()` is removed — pass `cursor=` to `sync()` instead.
- `sync()` now returns a generator of `SyncRecordList` chunks instead of a single list.
- The `limit` parameter is replaced by `chunk_size`; the unlimited (`limit=None`) accumulate-everything mode is gone — iterate the generator instead.

## Checklist:
- [x] Tests added/updated.


cognite/client/_api/data_modeling/records.py
```diff
@@ -69,28 +69,32 @@ async def _sync(
         initialize_cursor: str | None = None,
         cursor: str | None = None,
     ) -> SyncRecordList:
-        other_params: dict[str, Any] = {}
+        # Sync returns exactly one page of the change feed: 'limit' is the page size and
+        # 'hasNext' tells the caller whether to ask for more. We must not use the generic
+        # paging helpers here - they stop on an exhausted cursor, but the sync endpoint always
+        # returns a 'nextCursor' (that is what makes the feed resumable), so a page that does
+        # not fill 'limit' would make them request forever.
+        verify_limit(limit)
```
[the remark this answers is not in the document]
evertoncolling: Added in `76167b3`.

Note this is defensive only: `_sync` is private, and the two public entry points each pass exactly one of the pair — `sync()` forwards `initialize_cursor`, `sync_resume()` forwards `cursor` — so the branch is not reachable from the public API. Keeping it as a guard against future callers of the internal helper.

tests/tests_integration/test_api/test_data_modeling/test_records.py
```diff
@@ -0,0 +1,269 @@
...
+        properties = {**target.sources[0].properties, "value": 99.0}
+        cognite_client.data_modeling.records.upsert(
+            RecordWrite(
+                space=target.space,
+                external_id=target.external_id,
+                sources=[RecordSource(source=container_ref, properties=properties)],
+            ),
+            stream_id=mutable_stream.external_id,
+        )
+        time.sleep(2)
```
andersfylling: I think we should place the filter call + assert in a function we can retry instead of depending on a timeout - that way we our tests are a bit more robust.

tests/tests_integration/test_api/test_data_modeling/test_records.py
```diff
@@ -0,0 +1,269 @@
...
+                    source=container_ref,
+                    properties={"name": tag, "value": float(i), "processed": i % 2 == 0},
+                )
+            ],
+        )
+        for i in range(3)
+    ]
+    cognite_client.data_modeling.records.ingest(records, stream_id=mutable_stream.external_id)
+    # Records are not queryable the instant ingest returns.
+    time.sleep(2)
```
andersfylling: Can we instead of doing a timeout query the api until the entries appear? that way we are certain the backend state is what we expect.

andersfylling (commented):
Left two comments to make the test suite more robust. @asosnovski and I were just talking about the sync endpoint so this was a nice change to see!

cognite/client/_api/data_modeling/records.py
```diff
@@ -69,28 +69,34 @@ async def _sync(
         initialize_cursor: str | None = None,
         cursor: str | None = None,
     ) -> SyncRecordList:
-        other_params: dict[str, Any] = {}
+        # Sync returns exactly one page of the change feed: 'limit' is the page size and
+        # 'hasNext' tells the caller whether to ask for more. We must not use the generic
+        # paging helpers here - they stop on an exhausted cursor, but the sync endpoint always
+        # returns a 'nextCursor' (that is what makes the feed resumable), so a page that does
+        # not fill 'limit' would make them request forever.
```
haakonvt: Lets remove this rationale or make it drastically shorter - not using one of the existing helpers does not need an explanation imo!

cognite/client/_api/data_modeling/records.py
```diff
@@ -69,28 +69,34 @@ async def _sync(
...
             url_path=self._records_url(stream_id, "/sync"),
-            limit=limit,
-            filter=filter.dump(camel_case_property=False) if isinstance(filter, Filter) else filter,
-            other_params=other_params,
-            initial_cursor=cursor,
-            settings_forcing_raw_response_loading=["records_sync_cursor"],
+            json=body,
+            semaphore=self._get_semaphore("read"),
         )
+        return SyncRecordList._load_raw_api_response([res.json()])._maybe_set_client_ref(self._cognite_client)
```
haakonvt: `SyncRecordList` doesn't use the client reference, so no need to call the "maybe set it" method
```suggestion
        return SyncRecordList._load_raw_api_response([res.json()])
```

haakonvt: The old code was arguable wrong using: `settings_forcing_raw_response_loading=["records_sync_cursor"]`. There is no such option called `records_sync_cursor`.

The correct would be to set it to `include_typing`. When True, we need to load it from the raw API response.

In this new code, I guess we can always just use the "raw load" call, instead of using normal `_load` when not including typing.

tests/tests_integration/test_api/test_data_modeling/test_records.py
```diff
@@ -0,0 +1,269 @@
...
+    RecordContainerId,
+    RecordSource,
+    RecordSourceSelector,
+    RecordWrite,
+    TimeRange,
+)
+from cognite.client.data_classes.data_modeling.streams import Stream
+
+# Streams cannot be hard deleted and their external IDs can never be reused, so these tests reuse
+# a stream that already exists in the project, plus one container, for every run.
```
haakonvt: Streams cannot be hard deleted ever? 👀

tests/tests_integration/test_api/test_data_modeling/test_records.py
```diff
@@ -0,0 +1,269 @@
...
+CONTAINER_EXTERNAL_ID = "PythonSdkIntegrationTestRecords"
+
+
+def an_hour_ago() -> str:
+    """A lower bound for 'lastUpdatedTime'.
+
+    TimeRange forwards its bounds to the API untouched, and the API only accepts ISO-8601 there -
+    not the "1h-ago" shorthand that the rest of the SDK understands.
+    """
+    return (datetime.now(timezone.utc) - timedelta(hours=1)).strftime("%Y-%m-%dT%H:%M:%SZ")
```
haakonvt, line 49: The `TimeRange` class should be taught how to deal with relative time strings 🚀

tests/tests_integration/test_api/test_data_modeling/test_records.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+@pytest.fixture(scope="session")
+def mutable_stream(cognite_client: CogniteClient) -> Stream:
+    """Reuse the shared mutable test stream that already exists in the project.
+
+    These tests never create a stream: a project may only hold a few active streams, and a stream
+    can neither be deleted nor have its external ID reused - so creating one per suite would
+    permanently burn a quota slot and eventually make every run fail on the quota. Fall back to
+    any other mutable stream in case the shared one is ever replaced.
```
haakonvt: So much prose from your LLM, trim it! 😂 
```suggestion
```

tests/tests_integration/test_api/test_data_modeling/test_records.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    stream = cognite_client.data_modeling.streams.retrieve(STREAM_EXTERNAL_ID)
+    if stream is not None:
+        return stream
+    for stream in sorted(cognite_client.data_modeling.streams.list(), key=lambda s: s.external_id):
+        if stream.type == "Mutable":
+            return stream
+    pytest.skip(f"No mutable stream in the project (looked for {STREAM_EXTERNAL_ID!r}) to ingest records into.")
```
haakonvt: I don't agree with this logic which may end up causing writes to a "random" stream in the users project.

If the fixed test stream does not exist you just need to create it (right?) 🤔

tests/tests_integration/test_api/test_data_modeling/test_records.py
```diff
@@ -0,0 +1,269 @@
...
+    return [RecordSourceSelector(source=container_ref, properties=["*"])]
+
+
+@pytest.fixture
+def ingested_records(
+    cognite_client: CogniteClient,
+    mutable_stream: Stream,
+    container_ref: RecordContainerId,
+) -> Iterator[list[RecordWrite]]:
+    """Ingest a small, uniquely tagged batch and clean it up afterwards."""
```
haakonvt:
```suggestion
```

tests/tests_integration/test_api/test_data_modeling/test_records.py
haakonvt: I agree with this. Adding a simple helper with a low, fixed, number of retries using `Backoff` (`cognite/client/utils/_retry.py`) sounds like the way to go.

tests/tests_integration/test_api/test_data_modeling/test_records.py
```diff
@@ -0,0 +1,269 @@
...
+        mutable_stream: Stream,
+        container_ref: RecordContainerId,
+        sources: list[RecordSourceSelector],
+        ingested_records: list[RecordWrite],
+    ) -> None:
+        """Regression test: sync must return one page, even when it holds fewer than 'limit' records.
+
+        The endpoint always returns a 'nextCursor' - that is what makes the feed resumable - so
+        paging until the cursor runs out never terminates. Before this was fixed, asking for more
+        records than the feed can deliver made the SDK request in a loop until the test timed out.
```
haakonvt:
```suggestion
```

tests/tests_unit/test_api/test_data_modeling/test_records.py
```diff
@@ -793,6 +793,128 @@ def test_sync_target_units_body_shape(
...
+        cognite_client: CogniteClient,
+        httpx_mock: HTTPXMock,
+        sync_url_pattern: re.Pattern,
+        record_response: dict,
+        stream_id: str,
+        has_next: bool,
+    ) -> None:
+        # The sync endpoint always returns a 'nextCursor' - that is what makes the feed
+        # resumable - so a page holding fewer records than 'limit' must not make the SDK
+        # keep requesting. Only one response is registered, so any extra request fails.
```
haakonvt: Nice, _this_ is where this context belongs - keep it! ❤️

tests/tests_unit/test_api/test_data_modeling/test_records.py
```diff
@@ -793,6 +793,128 @@ def test_sync_target_units_body_shape(
...
+            url=sync_url_pattern,
+            status_code=200,
+            json={"items": items, "nextCursor": "p2", "hasNext": False},
+        )
+        page = cognite_client.data_modeling.records.sync_resume(stream_id=stream_id, cursor="p1", limit=10)
+        assert len(page) == 1
+        assert page.cursor == "p2"
+        assert len(httpx_mock.get_requests()) == 1
+
+    def test_sync_resume_does_not_send_initialize_cursor(
```
haakonvt: Uhm, is this test meaningful?

tests/tests_unit/test_api/test_data_modeling/test_records.py
```diff
@@ -793,6 +793,128 @@ def test_sync_target_units_body_shape(
...
+            ],
+        }
+
+    def test_sync_rejects_non_positive_limit(
+        self,
+        cognite_client: CogniteClient,
+        stream_id: str,
+    ) -> None:
+        with pytest.raises(ValueError, match="limit must be strictly positive"):
+            cognite_client.data_modeling.records.sync(stream_id=stream_id, initialize_cursor="c", limit=0)
```
haakonvt: This helper util is already covered by its own tests, not sure it makes sense to test here as well

haakonvt (commented):
Great fix, I fully agree to move away from the existing helper method 👌 Some comments:

cognite/client/_api/data_modeling/records.py
haakonvt: When I think about it, please make sure normal  `_load` works as intended. It is a bug otherwise as `load` is public and we have e.g. Toolkit relying on it.

tests/tests_integration/test_api/test_data_modeling/test_records.py
evertoncolling: That's a bit harsh indeed. It takes a couple of weeks for the external Id to be reusable again, while the stream is soft deleted. I'll clarify this 😅

tests/tests_integration/test_api/test_data_modeling/test_records.py
evertoncolling: That's fair! I was discussing something around time anchors in the API like what data points support ("1h-ago"), but until that lands we could handle something like this in the SDK. That would be a future dedicated PR though.

tests/tests_integration/test_api/test_data_modeling/test_records.py
evertoncolling: Indeed! Fixed in https://github.com/cognitedata/cognite-sdk-python/pull/2755/commits/a7a82705e1d8874f91f93e50d2b6912f01b79ba0

tests/tests_integration/test_api/test_data_modeling/test_records.py
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-python/pull/2755/commits/27c889121135b2d30ad3a5ed9ae9c3ec267fac85

tests/tests_integration/test_api/test_data_modeling/test_records.py
evertoncolling: Addressed in e7186b4

tests/tests_integration/test_api/test_data_modeling/test_records.py
evertoncolling: Addressed in e7186b4

cognite/client/_api/data_modeling/records.py
evertoncolling: Addressed in dbed804

tests/tests_unit/test_api/test_data_modeling/test_records.py
evertoncolling: Fair! Removed in 7060d65

tests/tests_unit/test_api/test_data_modeling/test_records.py
evertoncolling: Fair! Removed in 7060d65

cognite/client/_api/data_modeling/records.py
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-python/pull/2755/commits/2e2ab24647ce5a471ec5faa8063a003e791302de

cognite/client/_api/data_modeling/records.py
```diff
@@ -64,33 +66,46 @@ async def _sync(
...
+        url_path = self._records_url(stream_id, "/sync")
+        responses: list[dict[str, Any]] = []
+        while True:
+            res = await self._post(url_path=url_path, json=body, semaphore=semaphore)
+            responses.append(response := res.json())
+            if not drain or not response["hasNext"]:
+                break
+            body.pop("initializeCursor", None)
+            body["cursor"] = response["nextCursor"]
+        return SyncRecordList._load_raw_api_response(responses)
```
andersfylling: should we consider a generator to reduce memory?

haakonvt: As in continuously yielding chunks back to the user? I’m all in favor

evertoncolling: Then maybe we should add it via a separate method that has this pattern, no? We can refactor the existing method to sync a single page (either start the sync or resume) with manual control and add a dedicated method that exhausts the cursor until `hasNext` is false with the ability to yield chunks. Kind of like [iterate_data](https://cognite-sdk-python.readthedocs-hosted.com/en/latest/generated/cognite.client.AsyncCogniteClient.time_series.subscriptions.iterate_data.html) in datapoints. WDYT?

Also, since we are changing a lot here, we could also merge `sync` and `sync_resume` as suggested. I guess we can do some overload so a single method accepts the two signatures.

evertoncolling: Or just a single "generator" method that handles everything? Both manual pagination control by yeidling one page at the time and iterating over the whole dataset?

evertoncolling: Ended up testing out the latter :)

scripts/sync_client_codegen/constants.py
```diff
@@ -29,6 +29,7 @@
 SYNC_METHODS_TO_KEEP = {
     "compare_capabilities",
     "__call__",
+    "sync",  # async generator with overloads, so its overload stubs are plain 'def' (see above)
```
haakonvt: Please add a `TODO(haakont): link API class + method here; sync could match an unwanted method`

tests/tests_integration/test_api/test_data_modeling/test_records.py
```diff
@@ -0,0 +1,281 @@
...
+            stream_id=mutable_stream.external_id,
+            initialize_cursor="1m-ago",
+            sources=sources,
+            filter=tagged,
+            chunk_size=2,
+        ):
+            assert page.cursor is not None
+            seen.extend(record.external_id for record in page)
+            pages += 1
+            assert pages < 20, "sync did not exhausted the feed"
```
haakonvt:
```suggestion
            assert pages < 20, "sync did not exhaust the feed"
```

cognite/client/_api/data_modeling/records.py
```diff
@@ -445,98 +411,126 @@ async def filter(
...
 
-        Returns:
-            SyncRecordList: One page of change records, with ``cursor`` and ``has_next`` set.
+            Fetch a single chunk with manual control, e.g. to poll at your own cadence:
+
+                >>> chunk = next(
+                ...     client.data_modeling.records.sync(
+                ...         stream_id="my-stream", cursor="previously-stored-cursor"
+                ...     )
+                ... )
```
haakonvt: Let's show that we store the records iterator to a variable so that the user can call next multiple times

cognite/client/_api/data_modeling/records.py
```diff
@@ -445,98 +411,126 @@ async def filter(
...
-            sources (Sequence[RecordSourceSelector] | None): Which container properties to return.
-            target_units (RecordTargetUnits | Sequence[RecordTargetUnit] | None): Properties to convert
-                to another unit.
-            limit (int): Maximum number of records to return in this page (1-1000). Defaults to 10.
-            include_typing (bool): If True, include property type information on the returned
-                list's ``typing`` attribute.
+                >>> for chunk in client.data_modeling.records.sync(
+                ...     stream_id="my-stream", cursor="previously-stored-cursor"
+                ... ):
+                ...     pass  # process chunk, then persist chunk.cursor
```
haakonvt:
```suggestion
                ...     pass
```

cognite/client/_api/data_modeling/records.py
```diff
@@ -445,98 +411,126 @@ async def filter(
...
+                Mutually exclusive with ``cursor``.
+            cursor (str | None): Resume from a cursor from a previously yielded chunk. Mutually
+                exclusive with ``initialize_cursor``.
             filter (Filter | None): Filter expression (see :mod:`cognite.client.data_classes.filters`).
             sources (Sequence[RecordSourceSelector] | None): Which container properties to return.
             target_units (RecordTargetUnits | Sequence[RecordTargetUnit] | None): Properties to convert
                 to another unit.
-            limit (int): Maximum number of records to return in this page (1-1000). Defaults to 10.
-            include_typing (bool): If True, include property type information on the returned
+            chunk_size (int): Number of records per yielded chunk, between 1 and 1000. Defaults to 1000.
```
haakonvt, line 491: Since we are sending a request for each chunk (no "intelligent" fetching behind the scenes) lets add a warning that the using e.g. `chunk_size=1` comes at a really _hefty performance penalty_ and should arguable not be used.

cognite/client/_api/data_modeling/records.py
```diff
@@ -31,6 +30,8 @@
 
 
 class RecordsAPI(APIClient):
+    _SYNC_LIMIT = 1000  # Maximum page size of the sync endpoint
```
haakonvt: Let's move this to `__init__` like similar limits are stored in the SDK

cognite/client/_api/data_modeling/records.py
```diff
@@ -445,98 +411,126 @@ async def filter(
...
-            limit=limit,
-            filter=filter,
-            sources=sources,
-            target_units=target_units,
-            include_typing=include_typing,
-        )
+        if (initialize_cursor is None) == (cursor is None):
+            raise ValueError("Pass exactly one of 'initialize_cursor' or 'cursor'.")
+        if not (isinstance(chunk_size, int) and 1 <= chunk_size <= self._SYNC_LIMIT):
+            raise ValueError(f"'chunk_size' must be an integer between 1 and {self._SYNC_LIMIT}.")
```
haakonvt:
```suggestion
            raise ValueError(f"'chunk_size' must be an integer between 1 and {self._SYNC_LIMIT} (inclusive).")
```

cognite/client/_api/data_modeling/records.py
haakonvt, line 491: Another comment on the `chunk_size`, elsewhere in the SDK it typically defaults to `None`, which returns one-item-at-the-time. 

I think I like having this set to 1000 since we don't do any clever fetching behind the scenes (and then returning only `chunk_size` at the time to the user). Tagging @erlendvollset in case he wants consistency with previous `__call__` methods.

haakonvt (commented):
Great combination into one sync-method; a few comments:

scripts/sync_client_codegen/constants.py
evertoncolling: done in https://github.com/cognitedata/cognite-sdk-python/pull/2755/commits/4c12564d8eae49355e88435b3b8726d96692d3e6

cognite/client/_api/data_modeling/records.py
evertoncolling: done in https://github.com/cognitedata/cognite-sdk-python/pull/2755/commits/4c12564d8eae49355e88435b3b8726d96692d3e6

cognite/client/_api/data_modeling/records.py
evertoncolling: updated the docstring in https://github.com/cognitedata/cognite-sdk-python/pull/2755/commits/4c12564d8eae49355e88435b3b8726d96692d3e6

cognite/client/_api/data_modeling/records.py
evertoncolling: addressed in https://github.com/cognitedata/cognite-sdk-python/pull/2755/commits/4c12564d8eae49355e88435b3b8726d96692d3e6

cognite/client/_api/data_modeling/records.py
```diff
@@ -458,6 +457,12 @@ async def sync(
         carries the ``cursor`` to persist for resuming later; this is also why records are always
         yielded in chunks rather than one by one.
 
+        Warning:
+            Every chunk is fetched with a separate API request - there is no smart fetching
+            behind the scenes - so a small ``chunk_size`` multiplies the number of requests
+            and comes at a hefty performance penalty. Keep the default of 1000 (the API
+            maximum) unless your per-chunk processing genuinely needs smaller checkpoints.
```
haakonvt: A little too honest maybe 😂 
```suggestion
        Warning:
            Every chunk is fetched with a separate API request, so a small ``chunk_size`` increases
            the number of requests (i.e. comes at a high performance penalty). Keep the default of
            1000 (the API maximum) unless your per-chunk processing genuinely needs smaller batches.
```

haakonvt (dismissed):
Nice! One comment re. a little too high honesty setting 😅

haakonvt (approved):

haakonvt:
🦄 This is breaking, but we have clear and existing alpha warnings so 👍

PR #2756, opened 2026-08-08
refactor: speed up pre-commit hooks

MortGron:
## Description
The two pre-commit hooks from the pre-commit-hooks package have been replaced.

1. The debug statement rule is already handled by ruff
2. The check docstring is first rule is deprecated and "fundamentally flawed" according to https://github.com/pre-commit/pre-commit-hooks#deprecated--replaced-hooks. A related rule from ruff has been added. The formatting of docstrings is to a large extent already handled by the ruff formatter.

## Checklist:
- [ ] Tests added/updated.


cognite/client/_sync_api/entity_matching.py
```diff
@@ -24,6 +25,8 @@
 from cognite.client.utils._async_helpers import run_sync
 from cognite.client.utils.useful_types import SequenceNotStr
 
+T_ContextualizationJob = TypeVar("T_ContextualizationJob", bound=ContextualizationJob)
+
```
haakonvt, line 29: I wonder why this was added? Another bug in the auto-gen script perhaps 😅 If it is unused, feel free to just delete the lines (and keep diff so it wont be auto-generated immediately again)

cognite/client/_api/entity_matching.py
```diff
@@ -235,6 +233,7 @@ async def fit(
             ...     description="AssetMatchingJob1",
             ... )
         """
+        # TODO: Not in service contract
```
haakonvt, line 236: @olacognite could you verify these TODO comments are still needed?

cognite/client/_sync_api/entity_matching.py
MortGron: The same typevar is defined in the async api. In both palces it is unclear to me if they are used for anything internal, but since they are public they might be used by external users I assume.

PR #2757, opened 2026-08-09
chore: enable ruff rule for mutable default arguments

MortGron:
## Description
Enable ruff rule to dissallow mutrable default arguments. Based on discussion in #2695

## Checklist:
- [ ] Tests added/updated.


haakonvt (approved):

thorkildcognite (commented):
🦄

PR #2761, opened 2026-08-10
refactor(to_pandas): single resource classes delegate `to_pandas()` to their list class

haakonvt:
Split from #2724 

Removes duplicated `to_pandas` logic between resource classes and their list counterparts (e.g. `Group`/`GroupList`, `Node`/`NodeList` etc.) by having single resources delegate to `_LIST_CLASS([self]).to_pandas(...)` and squeeze the result, instead of maintaining a second copy of the conversion logic.

Adds `_LIST_CLASS`/`_RESOURCE` back-reference enforcement as a new test in `test_meta.py`: every `CogniteResourceList._RESOURCE` must point to a resource class whose `_LIST_CLASS` points right back, so future additions can't silently drift out of sync. Also adds `test_standalone_to_pandas_allowlist`, documenting the deliberate exceptions that _don't follow this pattern_.

Pure refactor, no behavior change. Paves the way for the pandas v2/v3 fix.

### Tip to reviewer
Almost all of the changes is part of just one commit, cd3b63f7e47c49ecbf3ce3474f5b2f3a0e35e4bd, (57 files changed, 482 insertions(+)). The PR can be reviewed commit-by-commit


cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -455,39 +456,14 @@ def to_pandas(  # type: ignore [override]
         Returns:
             pd.DataFrame: The dataframe.
         """
-        df = super().to_pandas(
-            expand_metadata=False, ignore=ignore, camel_case=camel_case, convert_timestamps=convert_timestamps
+        assert self._LIST_CLASS is not None
```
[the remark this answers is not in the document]
haakonvt: This is for mypy. Test ensures this for all subclasses.

cognite/client/data_classes/data_modeling/instances.py
```diff
@@ -455,39 +456,14 @@ def to_pandas(  # type: ignore [override]
...
+        df = cast("type[DataModelingInstancesList]", self._LIST_CLASS)([self]).to_pandas(
```
audunska, line 460: Why do we need to override this and not just defer to the base `CogniteResource` implementation? Will it not use the `DataModelingInstancesList` implementation if we do that?

audunska (commented):
Nice stuff. Out of curiosity, would it have been an option to do it the other way? I.e., have a default `to_pandas` on `CogniteResource`, but override it in `CogniteResourceList` to iterate over the implementation in `_RESOURCE`.

haakonvt:
> (...) Out of curiosity, would it have been an option to do it the other way? I.e., have a default to_pandas on CogniteResource, but override it in CogniteResourceList to iterate over the implementation in _RESOURCE.

Just for efficiency reasons, repeated calling a function for millions of items in a list isn't exactly cheap in Python 😅 

...or, a much better reason is that a pandas dataframe concatenation from individual dataframes would be a really bad idea (just making all those individual ones)

cognite/client/data_classes/data_modeling/instances.py
haakonvt: We do want to delegate to `DataModelingInstancesList`, but we cannot use the generic `CogniteResource.to_pandas` path directly as the base method always forwards args like `expand_metadata` and `metadata_prefix`, while `DataModelingInstancesList.to_pandas` uses e.g. `expand_properties`/`remove_property_prefix`.

audunska: Whatever you say 😅

audunska (dismissed):

cognite/client/data_classes/data_modeling/instances.py
haakonvt: A lot of garbage in this repo tbh 😅  The one-base-class-fits-all pattern for example...

haakonvt:
🦄

haakonvt:
🦄

audunska (approved):

PR #2762, opened 2026-08-11
feat(SyntheticTimeSeries): only allow Python identifiers as variable names

MortGron:
## Description
If you try to make an expression like `"a-b-1"`, with the variable substitution `{"a-b": "ts_ex_id"}`, it will currently fail with an unclear error.
If you make the expression `"a - a/b"`, with the variable subsutiotn `{"a": "ts_ex_id_1", "a/b": "ts_ex_id_2"}`, it will not fail, but incorrectly be translated to the expression `"ts{externalId:'ts_ex_id_1'} - ts{externalId:'ts_ex_id_1'}/b"`.

The safest thing to do seems like dissallowing variable names that are not Python identifiers, or we might end up with variable names that also contains mathemtical operators.

This PR explicitly fobids variable names that are not Python identifiers with a clear error message.

## Checklist:
- [x] Tests added/updated.


haakonvt:
This seems to me like a breaking change. I'm happy to let it through regardless though - as synthetic TS see very little use, but lets get some more eyes on this (perhaps a slack thread in #topic-sdk)

MortGron:
> This seems to me like a breaking change. I'm happy to let it through regardless though - as synthetic TS see very little use, but lets get some more eyes on this (perhaps a slack thread in #topic-sdk)

I do not have any urgent need for this, so it can wait to next major release. The non-breaking bug fix part of this is in #2766.

PR #2746, opened 2026-08-06
chore(master): release cognite-sdk-python 8.13.0

erlendvollset:
:robot: I have created a release *beep* *boop*
---

### Features

* **credentials:** let users choose the account for interactive login (DM-3512) ([#2730](https://github.com/cognitedata/cognite-sdk-python/issues/2730)) ([bef4f27](https://github.com/cognitedata/cognite-sdk-python/commit/bef4f27a836b2e93410fd2411e59043ee28a5cc0))
* **functions:** add py314 to RunTime ([#2735](https://github.com/cognitedata/cognite-sdk-python/issues/2735)) ([cf66c51](https://github.com/cognitedata/cognite-sdk-python/commit/cf66c51a8b73c243a7d1a8f742fb137463386f67))
* **streams:** rename StreamTemplateWriteSettings to StreamWriteSettings ([#2753](https://github.com/cognitedata/cognite-sdk-python/issues/2753)) ([801ee75](https://github.com/cognitedata/cognite-sdk-python/commit/801ee75c61c49024f79701ea10dacf46706b03fc))
* **transformations:** add external data sources client API ([#2732](https://github.com/cognitedata/cognite-sdk-python/issues/2732)) ([60b6c36](https://github.com/cognitedata/cognite-sdk-python/commit/60b6c36e42c60020198f298ad784580e00c0bdc2))

### Bug Fixes

* **records:** validate property paths instead of splitting bare strings ([#2747](https://github.com/cognitedata/cognite-sdk-python/issues/2747)) ([b90a384](https://github.com/cognitedata/cognite-sdk-python/commit/b90a3847d0ee49e5eaa37b8ae980cf1b6fe4d0eb))

### Documentation

* add all __call__ methods ([#2633](https://github.com/cognitedata/cognite-sdk-python/issues/2633)) ([b99af78](https://github.com/cognitedata/cognite-sdk-python/commit/b99af789b78fdfd905977db0ba2b7b5db20f1692))
* **transformations:** add external data sources guide ([#2734](https://github.com/cognitedata/cognite-sdk-python/issues/2734)) ([629c506](https://github.com/cognitedata/cognite-sdk-python/commit/629c50699aa42ea60978d92dbd748f5419ec1387))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #2764, opened 2026-08-12
ci: use plain version tags so Read the Docs can version releases (DM-3216)

yangru96:
## Description

Set `"include-component-in-tag": false` so release tags become `v8.13.0` instead of `cognite-sdk-python-v8.13.0`.

Read the Docs cannot parse the current tag format as a version, so no release can be activated as a docs version. The project only has `latest` (built from `master`), meaning users always read docs for unreleased code. With plain tags we can activate release versions on RTD and make `stable` the default.

Notes:
- The tag`v8.12.0` has been pushed manually at the commit`d7806bd` as the lookup anchor. Without it Release Please would fall back to the leftover `v7.89.0` and generate a release PR with ~730 commits. The next release PR should contain only the six entries from #2746.
- The release PR branch name also drops the component, so #2746 will be superseded by a new PR and should be closed manually.

Follow-ups: activate versions and set default to `stable` on RTD (needs admin access), then update the README badge/links that are hardcoded to `/en/latest/`.

## Checklist:
- [ ] Tests added/updated.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
+  "include-component-in-tag": false,
   "packages": {
     ".": {
       "release-type": "python",


haakonvt (approved):
LGTM

dumitru-nicolae-marasoiu (approved):

andersfylling:
:unicorn:

andersfylling (approved):

PR #2767, opened 2026-08-13
chore(release-please): hide docs commits from changelog

haakonvt:
Override `changelog-sections` to the same defaults as the "python release-type" strategy uses, but with docs hidden

The `changelog-sections` option, when given, overrides the default. There is no option to just override:
`{ "type": "docs", "section": "Documentation", "hidden": true }`

The list is most likely static for the forseeable future, so simply copying it in here should be just fine.


polomani (approved):

haakonvt:
🦄

PR #2770, opened 2026-08-15
fix(mocks): add config to Cognite client mock

MortGron:
## Description
Currently accessing the config attributes on the mocked Cognite client fails:
image

On any real Cognite client the same attributes are always accessible:
image

## Checklist:
- [x] Tests added/updated.


cognite/client/testing.py
```diff
@@ -184,6 +184,7 @@
 from cognite.client._sync_api.workflows.tasks import SyncWorkflowTaskAPI
 from cognite.client._sync_api.workflows.triggers import SyncWorkflowTriggerAPI
 from cognite.client._sync_api.workflows.versions import SyncWorkflowVersionAPI
+from cognite.client.credentials import Token
```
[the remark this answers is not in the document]
MortGron: From the official [documentation](https://docs.python.org/3/library/unittest.mock.html#auto-speccing), there are multiple ways to do this. 1. adding the default values as class attributes on the `ClientConfig`, 2. the way I have done in this PR, 3. the way you suggest. 

2. the only side effect I can see is the pypi version check, so as long as that is disabled, it should be ok with option 2 in my opinion
3. causes the most fragmantation and code duplication.

MortGron:
@cognitedata/python-sdk-maintainers @cognitedata/starbase could this get a review? It is needed to easily migrate a project I work on from SDK v7 to v8

cognite/client/testing.py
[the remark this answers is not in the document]
haakonvt: What are we missing if we just do `create_autospec(ClientConfig, instance=True, spec_set=True)`?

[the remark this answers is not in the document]
MortGron: @haakonvt for me that does not work. My udnerstanding is that `instance=True` does not affect `create_autospec`'s ability to detect dynamically set attributes. See also https://stackoverflow.com/a/67258631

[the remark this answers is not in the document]
haakonvt: What about:
```py
self.config = create_autospec(
    ClientConfig,
    instance=True,
    foo_dynamic_attr_1=123,
    bar_dynamic_attr_2=456,
)
flip_spec_set_on(self.config)

File: cognite/client/testing.py [hunk 2]
```diff
@@ -203,6 +204,23 @@ def __call__(cls, *args: Any, **kwargs: Any) -> Any:
         return instance
 
 
+def mock_client_config() -> MagicMock:
```
haakonvt: No need to specify parameters that are in init, right?

MortGron: Sorry, I do not understand. If I remove any parameter from this method, they will not be accessible from `client.config`. I tried commenting out `timeout`, and it is no longer accessible.

haakonvt: Discussed on slack - pushed a commit now

haakonvt (dismissed):
🦄  - thanks!

polomani (dismissed):

cognite/client/testing.py
```diff
@@ -202,6 +204,38 @@ def __call__(cls, *args: Any, **kwargs: Any) -> Any:
...
+    Creates a strict mock of ClientConfig that safely bypasses __init__ side-effects.
+
+    Why this setup is non-trivial:
+    - We cannot instantiate a real `ClientConfig` because its __init__ triggers side-effects like PyPI
+    version checks and logger configurations.
+    - We cannot simply use `create_autospec` with `instance=True` because that loses all instance
+    attributes defined in __init__.
+
+    The Workaround: We initialize a loose mock and manually set all instance variables (annoying but safe).
+    After seeding, we flip `_spec_set = True`.
```
audunska, line 218: These are really implementation details. Should they go into the docstring of a public function, or should they go in a comment?

haakonvt: Agree - I've made it "python private" now

audunska (approved):

haakonvt:
🦄

PR #2774, opened 2026-08-18
fix(data_classes): invalidate CogniteResourceList lookups on mutation

### LINKED ISSUE #1188: `CogniteResourceList` lacks correct dunder method implementations
It inherits from `collections.UserList`, but with its added features, e.g. `get` lookup accepting `id` or `external_id`, almost all mutating (magic or not) methods breaks some functionality.

https://github.com/python/cpython/blob/abd6e97020e5773ce2136228e31930a6d9e82dc0/Lib/collections/__init__.py#L1196

jaideeppyne:
## Description

Fixes #1188.

`CogniteResourceList` builds three identifier→item lookup maps as `@cached_property` — `_id_to_item`, `_external_id_to_item`, `_instance_id_to_item` — which `.get(id=.../external_id=.../instance_id=...)` reads. The maps are built lazily on the first `.get(...)` call.

Only `extend()` was written to keep those cached maps in sync. **Every other membership-mutating operation** — `append`, `insert`, `remove`, `pop`, `clear`, `__setitem__`, `__delitem__`, `__iadd__` (`+=`), `__imul__` (`*=`) — is inherited straight from `UserList` and mutates `self.data` **without invalidating the caches**. So once a map is cached:

- `get(id=X)` after `pop`/`remove`/`del` returns the **removed** item (should be `None`);
- `get(id=X)` after `append`/`insert`/`+=` returns **`None`** for an item that **is** present.

This is exactly the gap the existing `# TODO: We inherit a lot from UserList that we don't actually support...` comment flags.

## Fix

Add `_clear_identifier_lookups()` — which drops the three `cached_property` entries from `self.__dict__` so they rebuild on next access — and override each mutating method to call it after delegating to `super()`. The invalidate-then-rebuild approach is simpler and less error-prone than hand-patching the dicts.

## Test

`test_get_is_invalidated_by_mutations` primes the lookup cache with a `get()`, then runs each mutator (`pop`/`remove`/`append`/`insert`/`+=`/`__setitem__`/`del`/`clear`) and asserts `get()` reflects the new membership (removed → `None`, added → found). It fails on the current code (`get(id=2)` returns the popped item) and passes with the fix.

I ran the `TestCogniteResourceList` suite: 22 passed. The one remaining failure (`test_to_pandas`) is a pre-existing pandas-version dtype mismatch unrelated to this change (tracked separately). `ruff` check/format clean; `mypy` clean on the changed file.

### DIFF
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -304,7 +304,7 @@ class CogniteResourceList(UserList, Generic[T_CogniteResource]):
         return item
 
     def __setitem__(self, i: SupportsIndex | slice, item: T_CogniteResource | Iterable[T_CogniteResource]) -> None:
-        super().__setitem__(i, item)  # type: ignore[index, assignment]
+        super().__setitem__(i, item)  # type: ignore[index]
         self._clear_identifier_lookups()
 
     def __delitem__(self, i: SupportsIndex | slice) -> None:
--- a/cognite/client/data_classes/_base.py
+++ b/cognite/client/data_classes/_base.py
@@ -303,17 +303,20 @@ class CogniteResourceList(UserList, Generic[T_CogniteResource]):
         self._clear_identifier_lookups()
         return item
 
-    def __setitem__(self, i: Any, item: Any) -> None:
-        super().__setitem__(i, item)
+    def __setitem__(self, i: SupportsIndex | slice, item: T_CogniteResource | Iterable[T_CogniteResource]) -> None:
+        super().__setitem__(i, item)  # type: ignore[index, assignment]
         self._clear_identifier_lookups()
 
-    def __delitem__(self, i: Any) -> None:
+    def __delitem__(self, i: SupportsIndex | slice) -> None:
         super().__delitem__(i)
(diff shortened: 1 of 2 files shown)


jaideeppyne:
Thanks for the review — addressed all three in 3d5f563:

- `__iadd__` now delegates to `self.extend(other)`, so `+=` enforces the same duplicate-id `ValueError` (and cache maintenance) as `extend()` instead of bypassing it via `UserList`.
- `__setitem__` / `__delitem__` now use `SupportsIndex | slice` (and a precise item type) instead of `Any`.
- Added an assertion that `+=` with a duplicate id raises `ValueError`.

mypy/ruff clean and the mutation test still passes.

jaideeppyne:
Small follow-up in 423372e: dropped an unused `assignment` code from the `# type: ignore[index, assignment]` on `__setitem__`. Since `mypy.ini` sets `warn_unused_ignores = true`, the redundant code tripped `[unused-ignore]`; `# type: ignore[index]` alone is sufficient. mypy and ruff are now clean on the file. Thanks again for the thorough review!

PR #2777, opened 2026-08-19
ci: key venv cache on exact Python version to avoid stale interp…

yangru96:
## Description
Fix Windows CI failures in `test_full`: the venv cache key was keyed on the requested minor Python version (e.g. `3.13`) instead of the exact resolved version. When GitHub bumps the hosted Python patch on the Windows runner (e.g. `3.13.14` -> `3.13.15`), the old cache key still matches and restores a `.venv` whose launcher stub points at the now-removed old interpreter path, causing `poetry install` to fail with an executable-not-found error.

Now the cache key uses `steps.setup-python.outputs.python-version` (the exact resolved version), so a patch bump naturally invalidates the cache instead of restoring a broken venv.

## Checklist:
- [ ] Tests added/updated. — N/A, CI config only
- [ ] Documentation updated. — N/A, no docstring/API changes

image

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -13,7 +13,8 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6
+    - id: setup-python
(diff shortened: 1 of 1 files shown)


haakonvt (approved):
🦄 Nice!

PR #2778, opened 2026-08-19
ci: add delay between pytest reruns to reduce integration test flakiness

yangru96:
## Description
Add a delay between pytest reruns in `test_full` to reduce integration test flakiness. Tests hit a shared CDF project from up to 10 parallel matrix jobs, which can trip per-minute rate limits (429s) or cause transient job failures; retrying immediately doesn't help since the rate-limit window hasn't cleared. `--reruns-delay 20` gives it time before retrying.

## Checklist:
- [ ] Tests added/updated. — N/A, CI config only
- [ ] Documentation updated. — N/A, no docstring/API changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,7 +84,8 @@ jobs:
           COGNITE_BASE_URL: ${{ vars.CDF_BASE_URL_CI }}
           COGNITE_CLIENT_NAME: python-sdk-integration-tests
         # Testpaths are defined in the pytest.ini file:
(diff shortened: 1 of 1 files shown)


audunska (commented):
20s seems like a lot for a CI run. How does this interact with github action timeouts?

erlendvollset:
We also already do exponential backoff with jitter on 429s

yangru96:
Fair points — the client already backs off ~10 times on 429s, so 20s adds nothing and only risks the 15-minute job timeout. Will close this PR then  :)

PR #2780, opened 2026-08-21
fix(pandas): Support pandas v3 timestamp semantics while preserving v2 behavior (DM-3534)

haakonvt:
Enable pandas v3 compatibility by making SDK timestamp resolution version-conditional (`"ns"` on v2, `"ms"` on v3), remove the pandas upper bound, update migration/docs and tests, and cache version/unit helper lookups.


pyproject.toml
```diff
@@ -31,19 +31,19 @@ numpy = [
     "numpy >=1.26; python_version >= '3.12'"
 ]  # poetry 2.4.1 does not correctly limit numpy 2.2 to only python 3.10
 sympy = ["sympy"]
-pandas = ["pandas (>=2.1, <3)"]
+pandas = ["pandas >=2.1"]
```
audunska, line 34: Haven't really reviewed anything yet, but shouldn't we have a strict upper bound still? Like, `>= 2.1, <4`

haakonvt: No, but it sure is up for discussion - _and has been in the past_. 

Both numpy and pandas are very mature and responsible packages that we used to have open upper bounds for. It was only when v3 of pandas came out with what I deemed "a potentially breaking change for us" that we temporarily forced major version == 2 until it was addressed by us (aka as of this PR).

I didn't find the original discussion, but I found this; best summarised as "lock out future major releases just in case":
https://cognitedata.slack.com/archives/CG10VQPFX/p1773826110266419?thread_ts=1773760929.634279&cid=CG10VQPFX

cognite/client/utils/_pandas_helpers.py
```diff
@@ -64,6 +65,30 @@ def pandas_major_version() -> int:
...
+
+
+@cache
+def timestamp_dtype_unit() -> Literal["ns", "ms"]:
+    """The datetime64 resolution to use for all timestamp columns/indices produced by the SDK, see
+    `is_pandas_v2_or_lower` for the reasoning."""
+    return "ns" if is_pandas_v2_or_lower() else "ms"
+
+
+def to_pandas_timestamp(ms: int, tz: str | datetime.timezone | None = None) -> pd.Timestamp:
```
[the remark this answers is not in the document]
haakonvt: not going to do a type widening based on "LLM gut feeling"

tests/tests_unit/test_utils/test_pandas_helpers.py
```diff
@@ -0,0 +1,131 @@
+from __future__ import annotations
+
+from collections.abc import Callable, Iterator
+from typing import Any
+
+import pytest
+from _pytest.monkeypatch import MonkeyPatch
```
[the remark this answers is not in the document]
haakonvt: good idea but already covered by `mark.dsl`

cognite/client/utils/_pandas_helpers.py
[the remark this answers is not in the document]
audunska: "I can feel it in my weights!"

cognite/client/utils/_pandas_helpers.py
```diff
@@ -64,6 +65,30 @@ def pandas_major_version() -> int:
     return int(__version__.split(".")[0])
 
 
+@cache
+def is_pandas_v2_or_lower() -> bool:
+    """Pandas v2 (and lower) always defaulted to nanosecond-precision timestamps, so a lot of user code
+    (implicitly or not) relies on that. Pandas v3 instead infers precision from the input on a case-by-case
```
audunska, line 71: Why would code be fine with ms precision when using pydantic v3, if it's not fine with it when using v2? Just because that code is presumably new code?

cognite/client/utils/_pandas_helpers.py
```diff
@@ -145,7 +170,11 @@ def convert_nullable_int_cols(df: pd.DataFrame) -> pd.DataFrame:
 
 def convert_timestamp_columns_to_datetime(df: pd.DataFrame) -> pd.DataFrame:
```
audunska, line 201: Would be nice to have a docstring on this function describing what it takes as inputs. I can infer that it's cdf output with integer timestamp columns in ms, but would be better to have it explicit.

cognite/client/utils/_pandas_helpers.py
```diff
@@ -383,15 +412,20 @@ def _create_timestamp_index(
     import numpy as np
     import pandas as pd
 
+    unit = timestamp_dtype_unit()
     match timestamps, timezone:
         case list(), None:
-            return pd.to_datetime(timestamps, unit="ms")
+            return pd.to_datetime(timestamps, unit="ms").as_unit(unit)
```
audunska: Should we introduce a helper function for this callchain too, like the `to_pandas_timestamp` helper? Seems to me that would make it clearer.

MIGRATION_GUIDE.md
```diff
@@ -21,6 +21,7 @@ Changes are grouped as follows:
 - The SDK now ships with a new mock for the async client, namely `AsyncCogniteClientMock`. Both it and the previous `CogniteClientMock` are greatly improved and provide better type safety, checking of call signatures and spec_set=True is now enforced for all APIs (even the mocked client itself), through the use of `create_autospec` and bottom-up construction of nested APIs.
 - With the move to an async client, concurrency now works in Pyodide e.g. Jupyter-Lite in the browser. This also means that user interfaces like Streamlit won't freeze while resources from CDF are being fetched!
 - Good to know: File upload speeds are now significantly faster on Windows after `requests` are gone!
+- Classes from the SDK now has a new public load method that short-circuits on "optional" input, e.g. `NodeId.load_if(data.get("instance_id")) -> NodeId | None`.
```
yangru96, line 24: this one doesn't seems to be related to the current PR?

yangru96:
In the file `iam.py`, line 275 it is still using `pd.Timestamp(-1, unit="ms")`. maybe also change it to `to_pandas_timestamp(-1)` to keep consistent?
`df.loc[df[key] == pd.Timestamp(-1, unit="ms"), key] = pd.NaT`

cognite/client/utils/_pandas_helpers.py
haakonvt: It is because in v2, resolution was always nanoseconds, but in v3, it was made adaptive and will infer from the data. Thus, any v3 user should never assume anything regarding the resolution. Since we are using milliseconds everywhere in CDF, I thought this seemed like a good default.

MIGRATION_GUIDE.md
haakonvt: Yes, I piggybacked this missing oneliner, sue me 😂

haakonvt:
> In the file `iam.py`, line 275 it is still using `pd.Timestamp(-1, unit="ms")`. maybe also change it to `to_pandas_timestamp(-1)` to keep consistent? `df.loc[df[key] == pd.Timestamp(-1, unit="ms"), key] = pd.NaT`

I mean, it's just used to create the mask where we replace the values with `pd.NaT`, so not something being returned to the user. Equal timestamps compare equal regardless of resolution and since this is always `"ms"` the comparison will always work. 

...but I'll change it 😅

tests/tests_unit/test_utils/test_pandas_helpers.py
```diff
@@ -129,3 +131,30 @@ def test_changing_the_constant_changes_to_pandas_output(
         assert str(df_v2["last_updated_time"].dtype) == "datetime64[ns]"
         assert str(df_v3["last_updated_time"].dtype) == "datetime64[ms]"
         assert df_v2["last_updated_time"].dtype != df_v3["last_updated_time"].dtype
+
+
+@pytest.mark.dsl
+def test_pandas_timestamp_constructors_only_used_in_pandas_helpers() -> None:
```
audunska, line 137: noice

yangru96 (approved):
LGTM :)

audunska (approved):
LGTM

haakonvt:
🦄

PR #2782, opened 2026-08-22
docs: add version and date to homepage

MortGron:
## Description
It is not obvious which version you are viewing when you go the the documentation homepage. And it is not clear when this version is from.

Before:
image

After:
image

## Checklist:
- [ ] Tests added/updated.


haakonvt (approved):
🦄 nice!

PR #2783, opened 2026-08-23
fix(functions): list schedules with function external id

MortGron:
## Description
Specifying `function_external_id` in `schedules.list` always return an empty list even though schedules exists:

image

Pytest has been updated to v9 to support subtests

## Checklist:
- [x] Tests added/updated.


tests/tests_integration/test_api/test_functions.py
```diff
@@ -152,7 +152,9 @@ def test_call_function_with_oneshot(self, cognite_client: CogniteClient, a_funct
 
 
 class TestFunctionSchedulesAPI:
-    def test_create_retrieve_delete(self, cognite_client: CogniteClient, a_function: Function) -> None:
+    def test_create_retrieve_delete(
+        self, cognite_client: CogniteClient, a_function: Function, subtests: pytest.Subtests
+    ) -> None:
```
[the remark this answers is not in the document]
MortGron: pytest v9 supports subtests

cognite/client/_api/functions/schedules.py
haakonvt, line 1: Let’s start throwing warnings here please 😵‍💫

haakonvt: Let’s use `_get_function_internal_id` + catch error and warn that next major will raise

MortGron: My limited testing of the `list` methods in the SDK suggests they do not throw errors if you filter on an unknown resource, but rather return an empty result. Is your suggestions to start throwing errors in this case?
Or is your suggestion that we should deprecate the use of `function_external_id` in this method completely?

haakonvt: Oh this is just /list?

MortGron: Do you still want `_get_function_internal_id` ?

haakonvt: No need

haakonvt:
@HMEiding could you take a quick look on this one?

PR #2784, opened 2026-08-23
docs: add missing hosted extractor data classes. fix /list filters for Limits- & Metering APIs

MortGron:
## Description
Add missing hosted extractors data classes to the docs

Because hosted extractors has a class called `Prefix` and there is an existing class called `Prefix` in filters, we need to help Sphinx understand which `Prefix` links should point to.

Before: 
image

After:
image

## Checklist:
- [ ] Tests added/updated.


haakonvt (approved):
🦄 thanks!

PR #2788, opened 2026-08-25
chore: Support `content` parameter in `BasicAPIClient._post` method

haakonvt:
In order to support using protobuf to push state time series datapoints (and normal datapoints in the future if anyone wants to do a refactoring job...), we need to support the `content` parameter for our internal basic `_post` method.


erlendvollset (approved):

haakonvt:
🦄

PR #2789, opened 2026-08-25
chore(tests): Make debug notice tests tolerate extra notices

haakonvt:
Make data modeling debug notice integration tests tolerant of new/additional notices.

The backend now returns `UnfilteredContainerScanNotice` in some responses, introduced in ark commit cognitedata/ark#21018. These tests only need to verify that the expected `SortNotBackedByIndexNotice` is present and deserialized correctly, not that it is the only notice returned.

Currently this is breaking CI.


audunska (approved):

haakonvt:
🦄

PR #2790, opened 2026-08-25
feat(state time series): Add insert support for datapoints using protobuf

haakonvt:
# Too long, will split 😅


haakonvt:
Too long PR, will split. First part here: https://github.com/cognitedata/cognite-sdk-python/pull/2794

PR #2792, opened 2026-08-26
chore(Views for Records): change API maturity warning from alpha to GA, drop alpha header

polomani:
## Description
1. Remove alpha header from the record views calls
2. keep the warning (SDK doesn't have to be more mature than the API)

## Checklist:
- [x] Tests added/updated.


haakonvt (approved):
LGTM, please update the title to be slightly more verbose as it ends up in the changelog

polomani:
🦄

PR #2794, opened 2026-08-26
chore: Add write data classes for state time series datapoints  (DM-4078)

haakonvt:


cognite/client/data_classes/datapoints.py
```diff
@@ -112,6 +114,112 @@ class StatusCode(IntEnum):
...
+            datapoints=data["datapoints"],
+        )
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        return {
+            "instanceId" if camel_case else "instance_id": NodeId.load(self.instance_id).dump(
+                camel_case=camel_case, include_instance_type=False
+            ),
+            "datapoints": [StateDatapointWrite.load(dp).dump(camel_case=camel_case) for dp in self.datapoints],
+        }
```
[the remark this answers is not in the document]
haakonvt: @gemini-code-review your first point, load in the SDK ALWAYS expects camelCase, the primary use is to load API responses. In a few select cases (like in this PR) adds snake_case support when we have a documented need.

Your second point, the type annotation is `datapoints: Sequence[StateDatapointWrite | dict[str, Any]]` so I don't see the problem.

Your third point, dump only calls NodeId.load once, that's no issue. Since we don't freeze this (class because likely a user may want to append to the datapoints e.g. in a loop) we have to do a load-dump roundtrip to ensure all the data is valid when dumping. We need to support dict input for datapoints.

haakonvt:
Maybe `chore` could be `feat` in the PR title, idk

audunska:
> Maybe `chore` could be `feat` in the PR title, idk

That would add it to the release-please changelog, which I don't think we want yet.

cognite/client/data_classes/datapoints.py
```diff
@@ -112,6 +114,117 @@ class StatusCode(IntEnum):
...
+        if isinstance(data, cls):
+            return data
+
+        if not isinstance(data, dict):
+            raise TypeError(
+                f"Each state datapoint must be either a 'StateDatapointWrite' or a 'dict', not {type(data)}"
+            )
+        status = data.get("status") or {}
+        # Note: Contrary to normal load methods, we also accept snake_case here. This is because passing dicts with datapoints
+        #       to be inserted is the "old" way of doing things in the SDK, and we want this process to stay "familiar":
```
audunska, line 171: I don't understand the reasoning here. What's different in this dataclass from other dataclasses that don't accept snake case?

haakonvt: @audunska It will make more sense in the next PR that actually uses these classes 😅 

The long story is that...
- the existing `insert`- and `insert_multiple` methods has had dict-format as one of the primary expected/accepted input formats for a _long_ time
- ...thus to keep the new and upcoming `insert_states` method feel familiar, I'd like to keep the dict possibility
- ...but for states, the dict shape now actually differs between snake_case and camelCase due to e.g. numeric_value vs numericValue. 

We could ofc just document that "if you pass a dict you need to pass the exact shape the API expects", but I know our Python users love snake_case, so I added that here for convenience. I think I'll also use snake case in the docstring examples tbh...

haakonvt: From old `insert` method:
https://cognite-sdk-python.readthedocs-hosted.com/en/latest/generated/cognite.client.AsyncCogniteClient.time_series.data.insert.html#cognite.client.AsyncCogniteClient.time_series.data.insert

image

audunska: Ah ok, so it's just that the existing timeseries dtos don't have any names that differ between snake case and camel case. Got it!

tests/tests_unit/test_data_classes/test_datapoints.py
```diff
@@ -117,3 +118,109 @@ def test_identifier_priority(self, dps_lst_cls: type[CogniteResourceList]) -> No
...
+            "datapoints": [{"timestamp": 1000, "numericValue": 0}],
+        }
+
+    def test_dump_snake_case(self, state_dps_insert: StateDatapointsInsert) -> None:
+        assert state_dps_insert.dump(camel_case=False) == {
+            "instance_id": {"space": "my-space", "external_id": "my-ts"},
+            "datapoints": [{"timestamp": 1000, "numeric_value": 0}],
+        }
+
+    def test_load_dump_roundtrip(self, state_dps_insert: StateDatapointsInsert) -> None:
```
audunska: This can be parametrized by `camelCase` on the dump argument, since load accepts both

haakonvt: Coulda shoulda woulda 😅

haakonvt: @audunska updated now!

audunska (approved):

haakonvt:
🦄

PR #2799, opened 2026-08-28
fix(transformations): external data sources endpoint is create-only, not upsert

Khyat-Cognite:
## Summary
- The Transformations "external data sources" endpoint (`POST /transformations/externaldata`) is now create-only on the backend (jetfire-backend) and in the `service-contracts` OpenAPI spec: a request whose `externalId` already exists for the project is rejected with `409` and nothing is created. Previously it was a true upsert (replace-in-place).
- Renamed `TransformationExternalDataSourcesAPI.upsert()` → `create()` and dropped the now-meaningless `upsert_mode: Literal["replace"]` kwarg (it was doc-only, never functional).
- Regenerated the sync API via `scripts/sync_client_codegen`.
- Updated the "upsert model" docstrings on `ExternalDataSourceWrite`/`OneLakeExternalDataSourceWrite`.
- Updated the narrative guide (`docs/source/transformations_external_data_sources.rst`): permissions table, registration section, and rewrote "Rotate credentials" to delete-then-create since there's no update path.
- Renamed `test_upsert_*` tests to `test_create_*`; added a unit test asserting `CogniteDuplicatedError` on a 409/duplicate response, and a live second-create-same-id assertion in the integration test.

This SDK client is `sdk_maturity="alpha"` and was never in a tagged release, so this is not marked as a breaking change.

## Test plan
- [x] `ruff check` / `ruff format --check` on all changed files
- [x] `pytest tests/tests_unit/test_api/test_transformations_external_data.py -v` — 9/9 pass, including the new 409 test
- [x] `pytest tests/tests_unit/test_data_classes/test_transformations/test_externaldata.py -v` — unchanged, 10/10 pass
- [x] Broader `tests/tests_unit -k "external_data or externaldata or transformations"` — 257/257 pass

### DIFF
--- a/cognite/client/_api/transformations/externaldata.py
+++ b/cognite/client/_api/transformations/externaldata.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import AsyncIterator, Sequence
-from typing import TYPE_CHECKING, Literal, overload
+from typing import TYPE_CHECKING, overload
 
 from cognite.client._api_client import APIClient
 from cognite.client._constants import DEFAULT_LIMIT_READ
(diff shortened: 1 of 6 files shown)


haakonvt (approved):
Approved this breaking change as alpha warning (SDK maturity) has been present 👍

Khyat-Cognite:
🦄

PR #2802, opened 2026-08-31
test: address some flaky records and DML integration tests

evertoncolling:
## Description
This PR intents to fix two somewhat flaky tests related to records sync functionality:

- `test_sync_returns_partial_page_without_hanging` failed with `assert 1 == 3` -> It took only the first page from `records.sync(...)` and asserted that page contained all 3 ingested records with `has_next is False`. The service may split the change feed across several pages.

- `test_sync_iterates_until_feed_exhausted` fails less often, with `AssertionError: expected the 3 ingested records to span more than one chunk of size 2` -> `chunk_size` is sent as the API's `limit`, a hard per-page cap, so with 3 records and `limit=2` a single page is impossible unless the feed held fewer than 3 records. The sync change feed is eventually consistent and lags behind `filter` and the `ingested_records` fixture only waits for `filter` visibility before yielding.

## Checklist:
- [x] Tests added/updated.

### DIFF
--- a/tests/tests_integration/test_api/test_data_modeling/test_graphql.py
+++ b/tests/tests_integration/test_api/test_data_modeling/test_graphql.py
@@ -31,6 +31,8 @@ def data_model_for_query_test(cognite_client: CogniteClient, integration_test_sp
 
 
 class TestDataModelingGraphQLAPI:
+    # TODO: refactor to fix flakiness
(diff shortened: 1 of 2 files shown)


haakonvt (dismissed):
🦄  thanks! The records API tests have been really flaky lately 😓

haakonvt (dismissed):
🦄

haakonvt (approved):
🦄

polomani (approved):

PR #2803, opened 2026-09-01
fix(ClientConfig): Make API subversion mutable at runtime + easy to override per API class

haakonvt:
3 things discovered while annoyingly trying to temporarily change API version for state time series tests in need of the "beta" flag.
- `ClientConfig.api_subversion` is typed as `str | None` but is always `str`
- Is set in stone once per API class at init, while we should be able to mutate it
- Fixes the helper function `_alpha_version_header` that could return "beta-alpha" which is kinda weird thing to ask for 😉

### DIFF
--- a/cognite/client/_basic_api_client.py
+++ b/cognite/client/_basic_api_client.py
@@ -187,7 +187,7 @@ class BasicAsyncAPIClient:
     def __init__(self, config: ClientConfig, api_version: str | None, cognite_client: AsyncCogniteClient) -> None:
         self._config = config
         self._api_version = api_version
-        self._api_subversion = config.api_subversion
+        self.__api_subversion_override: str | None = None
         self._cognite_client = cognite_client
         self._init_async_http_clients()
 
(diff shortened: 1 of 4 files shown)


cognite/client/_basic_api_client.py
```diff
@@ -223,9 +238,12 @@ def _select_async_http_client(self, is_retryable: bool) -> AsyncHTTPClientWithRe
...
     def _alpha_version_header(self) -> dict[str, str]:
-        subversion = self._config.api_subversion
-        version = subversion if "alpha" in subversion else subversion + "-alpha"
-        return {"cdf-version": version}
+        sub = self._api_subversion
+        if sub.isdecimal():  # default is something like "20230101" (see __api_subversion__ in _version.py)
+            return {"cdf-version": f"{sub}-alpha"}
+
+        # Maybe the user has set "beta" or something else, whatever the case, we just return "alpha":
+        return {"cdf-version": "alpha"}
```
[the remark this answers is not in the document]
haakonvt: @gemini-code-assist good idea. have updated now

cognite/client/_basic_api_client.py
```diff
@@ -223,9 +238,14 @@ def _select_async_http_client(self, is_retryable: bool) -> AsyncHTTPClientWithRe
...
-        version = subversion if "alpha" in subversion else subversion + "-alpha"
-        return {"cdf-version": version}
+        sub = self._api_subversion
+        if "alpha" in sub:
+            return {"cdf-version": sub}
+        elif sub.isdecimal():  # default is something like "20230101" (see __api_subversion__ in _version.py)
+            return {"cdf-version": f"{sub}-alpha"}
+        else:
+            # Maybe the user has set "beta" or something else, whatever the case, we just return "alpha":
+            return {"cdf-version": "alpha"}
```
polomani, line 248: why? what does it mean? beta now allowed?

polomani (approved):

cognite/client/_basic_api_client.py
polomani: not a breaking change to what it was, just wonder how to set beta

haakonvt: If the user had set `api_subversion` to `"beta"`, which is very valid ofc, this function would return `"beta-alpha"` 😂

haakonvt:
🦄

cognite/client/_basic_api_client.py
haakonvt: (This is just a helper method for the SDK to auto-set alpha header when needed)

PR #2807, opened 2026-09-01
feat(state time series): add retrieve support for raw state datapoints (DM-4094)

haakonvt:
This PR does/adds:
- extract functionality for state dps from the protobuf binary format
- extends `Datapoints` (and `Datapoint`) with `numeric_state(s)` and `string_state(s)`.
- adds `is_state_dps` property to track when retrieving state datapoints
- starts raising good error messages for all cases not yet implemented like state aggregates or any numpy functionality
- comprehensive test for state datapoint retrieval after *state set mutations*

### DIFF
--- a/cognite/client/_api/datapoint_tasks.py
+++ b/cognite/client/_api/datapoint_tasks.py
@@ -847,14 +847,19 @@ class BaseRawTaskOrchestrator(BaseTaskOrchestrator):
             self.dps_data[idx].append(arr)
             if missing_idxs:
                 self.null_timestamps.update(self.ts_data[idx][-1][missing_idxs].tolist())
+
         if self.query.include_status:
             self.status_code[idx].append(DpsUnpackFns.extract_status_code_numpy(dps))
             self.status_symbol[idx].append(DpsUnpackFns.extract_status_symbol_numpy(dps))
 
     def _unpack_and_store_basic(self, idx: tuple[float, ...], dps: DatapointsRaw) -> None:
         self.ts_data[idx].append(DpsUnpackFns.extract_timestamps(dps))
+
         if self.query.ignore_bad_datapoints:
-            self.dps_data[idx].append(DpsUnpackFns.extract_raw_dps(dps))
+            if self.is_state_dps:
+                self.dps_data[idx].append(DpsUnpackFns.extract_raw_num_and_str_state_dps(cast(StateDatapoints, dps)))
+            else:
+                self.dps_data[idx].append(DpsUnpackFns.extract_raw_dps(dps))
         else:
             self.dps_data[idx].append(DpsUnpackFns.extract_nullable_raw_dps(dps))
         if self.query.include_status:
--- a/cognite/client/_api/datapoint_tasks.py
+++ b/cognite/client/_api/datapoint_tasks.py
@@ -823,32 +823,37 @@ class BaseRawTaskOrchestrator(BaseTaskOrchestrator):
 
     def _unpack_and_store(self, idx: tuple[float, ...], dps: DatapointsRaw) -> None:  # type: ignore [override]
         if self.use_numpy:
-            self.ts_data[idx].append(DpsUnpackFns.extract_timestamps_numpy(dps))
-            assert self.raw_dtype_numpy is not None
-            if self.query.ignore_bad_datapoints:
-                self.dps_data[idx].append(DpsUnpackFns.extract_raw_dps_numpy(dps, self.raw_dtype_numpy))
-            else:
-                # After this step, missing values (represented with None) will become NaNs and thus become
-                # indistinguishable from any NaNs that was returned! We need to store these timestamps in a property
-                # to allow our users to inspect them - but maybe even more important, allow the SDK to accurately
-                # use the DatapointsArray to replicate datapoints (exactly).
-                arr, missing_idxs = DpsUnpackFns.extract_nullable_raw_dps_numpy(dps, self.raw_dtype_numpy)
-                self.dps_data[idx].append(arr)
-                if missing_idxs:
-                    self.null_timestamps.update(self.ts_data[idx][-1][missing_idxs].tolist())
-            if self.query.include_status:
-                self.status_code[idx].append(DpsUnpackFns.extract_status_code_numpy(dps))
-                self.status_symbol[idx].append(DpsUnpackFns.extract_status_symbol_numpy(dps))
-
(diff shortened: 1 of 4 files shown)


cognite/client/_api/datapoint_tasks.py
```diff
@@ -793,32 +823,54 @@ def _include_outside_points_in_result(self) -> None:
...
-            if self.query.include_status:
-                self.status_code[idx].append(DpsUnpackFns.extract_status_code_numpy(dps))
-                self.status_symbol[idx].append(DpsUnpackFns.extract_status_symbol_numpy(dps))
+            self._unpack_and_store_numpy(idx, dps)
+        else:
+            self._unpack_and_store_basic(idx, dps)
+
+    def _unpack_and_store_numpy(self, idx: tuple[float, ...], dps: DatapointsRaw) -> None:
+        if self.is_state_dps:
+            raise NotImplementedError(
```
haakonvt, line 832: Note to reviewer: the reason for raising in `_unpack_*` and `_get_result` is that the first is skipped if there are no datapoints. When there are datapoints, we also want to skip as soon as possible (we dont know the datapoints are states before we start receiving them...).

cognite/client/_api/datapoint_tasks.py
```diff
@@ -527,6 +529,7 @@ def __init__(
         self.raw_dtype_numpy: type[np.object_] | type[np.float64] | None = None
         self._is_done = False
         self._final_result: Datapoints | DatapointsArray | None = None
+        self._is_state_dps: bool
```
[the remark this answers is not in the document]
haakonvt: @gemini-code-assist wrong. `is_state_dps` can only be read after `is_done is True`, and `is_done` is gated on `ts_info` being set, which happens atomically with `_is_state_dps` in `_store_ts_info` before any code path can mark a task done (both `test_retrieve_state_datapoints_empty` and the zero-batch aggregate cases in `test_unsupported_state_datapoint_retrievals_raise_not_implemented` already exercise this and pass + a ton of existing tests in `tests/tests_integration/test_api/test_datapoints.py`).

cognite/client/_api/datapoint_tasks.py
```diff
@@ -710,11 +721,20 @@ def offset_next(self) -> Literal[1]:
         return 1  # millisecond
 
     def _create_empty_result(self) -> Datapoints | DatapointsArray:
-        status_cols: dict[str, Any] = {}
+        status_cols: dict[str, Any] = {}  # TODO: Perhaps better to make this "all info" instead of just status
```
polomani, line 724: why is it `Any` here?

cognite/client/_api/datapoint_tasks.py
```diff
@@ -793,32 +823,54 @@ def _include_outside_points_in_result(self) -> None:
...
+        assert self.raw_dtype_numpy is not None
+        if self.query.ignore_bad_datapoints:
+            self.dps_data[idx].append(DpsUnpackFns.extract_raw_dps_numpy(dps, self.raw_dtype_numpy))
         else:
-            self.ts_data[idx].append(DpsUnpackFns.extract_timestamps(dps))
-            if self.query.ignore_bad_datapoints:
+            # After this step, missing values (represented with None) will become NaNs and thus become
+            # indistinguishable from any NaNs that was returned! We need to store these timestamps in a property
+            # to allow our users to inspect them - but maybe even more important, allow the SDK to accurately
+            # use the DatapointsArray to replicate datapoints (exactly).
```
polomani, line 845: can we use some sort of sentinel value for NaNs instead?

cognite/client/utils/_datapoints.py
```diff
@@ -44,14 +45,15 @@
...
 
-DatapointAny = AggregateDatapoint | NumericDatapoint | StringDatapoint
-DatapointsAny = AggregateDatapoints | NumericDatapoints | StringDatapoints
+DatapointAny = AggregateDatapoint | NumericDatapoint | StringDatapoint | StateDatapoint
+DatapointsAny = AggregateDatapoints | NumericDatapoints | StringDatapoints | StateDatapoints
 
-DatapointRaw = NumericDatapoint | StringDatapoint
-DatapointsRaw = NumericDatapoints | StringDatapoints
+DatapointRaw = NumericDatapoint | StringDatapoint | StateDatapoint
+DatapointsRaw = NumericDatapoints | StringDatapoints | StateDatapoints
```
polomani, line 54: adding a new arm is sort of breaking for the user, right? e.g. if they built some branching logic around response types

cognite/client/utils/_datapoints.py
```diff
@@ -122,7 +144,7 @@ def extract_timestamps_numpy(cls, dps: DatapointsAny) -> npt.NDArray[np.int64]:
 
     @classmethod
     def extract_raw_dps(cls, dps: DatapointsRaw) -> list[float | str]:  # Actually: exclusively either one
-        return list(map(cls.raw_dp, dps))
+        return list(map(cls.raw_dp, dps))  # type: ignore [arg-type]
```
polomani, line 147: the types look to be matching, is it a mypy issue?

polomani (approved):

cognite/client/_api/datapoint_tasks.py
haakonvt: It was and still is Any because I was lazy and am lazy 😉 It stores two columns that are immediately loaded into a class. If I were to add type hint here, I would need a cast both places unpacking it because one is list and the other array.

cognite/client/_api/datapoint_tasks.py
haakonvt: How? The array is a proper, contiguous C-array.

haakonvt: Also, this part is just whitespace-shifted

cognite/client/utils/_datapoints.py
haakonvt: Well, the API adds a new response type and I don't think we can do anything about that in the SDK other than being honest about it

haakonvt: Open to suggestions ofc

cognite/client/utils/_datapoints.py
haakonvt: `cls.raw_dp` does not expect state dps, but using cast has overhead (tiny) so I'd rather not

haakonvt (commented):
🦄

PR #2814, opened 2026-09-04
fix(config): Deprecate `cdf_cluster` property; warn on overspecified `base_url`/`cluster` combo (DM-4104)

haakonvt:
Our `ClientConfig` object has a public property `cdf_cluster` that causes confusion, especially since we allow passing both `cluster` _and_ `base_url`. The docs are clear on "just passing one of them" and that if you are to pass just the cluster, it has to conform to the "normal CDF cluster" spec.

However, a user passing both gets into a weird state. `base_url` takes precedence like the docs says, but the `cdf_cluster` property sees that `cluster` was passed and just returns that instead of attempting to guess from the `base_url`.

This PR adds a warning to users passing both (also saying this will become an error in v9), and properly ignores the `cluster`.

This came out of quite a confusing Slack thread:
https://cognitedata.slack.com/archives/CG10VQPFX/p1784109996968779


cognite/client/config.py
```diff
@@ -364,11 +380,28 @@ def load(cls, config: dict[str, Any] | str) -> ClientConfig:
         )
 
     @property
+    @deprecated(
+        "This property is a best-effort helper used internally in the SDK to determine the CDF cluster to make e.g. "
+        "warnings/error messages better, it is never used in API request routing (that is decided by 'base_url'). "
+        "A return value of None from this property simply means the cluster could not be guessed from the base URL, "
+        "and is not an error. It was never meant to be a reliable source of truth, which is impossible due to the SDK "
+        "allowing arbitrary base URLs. Will be removed in in the next major version (v9)."
```
yangru96: here is a small typo that _**in**_ appears twice :)

cognite/client/config.py
```diff
@@ -364,11 +380,28 @@ def load(cls, config: dict[str, Any] | str) -> ClientConfig:
...
+    )
     def cdf_cluster(self) -> str | None:
+        """The CDF cluster, if known: as passed to 'cluster', else guessed from 'base_url'.
```
yangru96, line 391: After the PR, when there are both base_url and cluster, cluster would be ignored but this docstring might still be a bit confusing to users? maybe change it to something like "_Guessed from base_url, or from cluster if only cluster was provided at init._" or similar?

cognite/client/config.py
```diff
@@ -364,11 +380,28 @@ def load(cls, config: dict[str, Any] | str) -> ClientConfig:
...
+
+        .. deprecated:: 8.15.0
```
yangru96: the current mater is already _8.15.0_ and _8.16.0_ is releasing. should this be changed to _8.17.0_ meaning this feature will get deprecated on 8.17.0 because deprecation version should be the release version that actually introduces the deprecation otherwise users would think this feature already got deprecated on version 8.15.0?

cognite/client/config.py
haakonvt: I only put that there so you'd think I didn't use AI to write it 😉

haakonvt:
```suggestion
        "allowing arbitrary base URLs. Will be removed in the next major version (v9)."
```

cognite/client/config.py
```diff
@@ -364,11 +380,28 @@ def load(cls, config: dict[str, Any] | str) -> ClientConfig:
...
+            This property is a best-effort helper used internally in the SDK to determine the CDF cluster to make e.g.
+            warnings/error messages better. It was never meant to be a reliable source of truth, which is impossible
+            due to the SDK allowing arbitrary base URLs. Will be removed in in the next major version (v9).
```
haakonvt:
```suggestion
            due to the SDK allowing arbitrary base URLs. Will be removed in the next major version (v9).
```

cognite/client/config.py
haakonvt:
```suggestion
        .. deprecated:: 8.16.0
```

cognite/client/config.py
haakonvt: Maybe, I think the loud warning is what people are really going to read, so I'm not too worried about this.

yangru96 (approved):

haakonvt:
🦄

PR #2817, opened 2026-09-07
fix: Correctly parse `cdf-is-auto-retryable` header when false

haakonvt:
Fixes the `"cdf-is-auto-retryable"` header parsing so that "false" is no longer treated as truthy. Adds proper tests for missing/true/false.


audunska (approved):

yangru96 (approved):

haakonvt:
🦄

PR #2837, opened 2026-09-09
chore(pandas): Remove pandas v1 compatibility code

haakonvt:
Removing a code path that was for pandas v1 which is now dead as we have minimum version at >= 2.1.


dumitru-nicolae-marasoiu (approved):

audunska (approved):

haakonvt:
🦄

PR #2842, opened 2026-09-14
fix(logging): Stop exposing credentials in debug logs and `str()` of data classes (DM-4118, VULN-112)

haakonvt:
Although we specifically advise against using debug logging in production:

> (...) Thus, it is not recommended in production environments, or where credentials cannot be easily disabled or rotated, or where log data may be accessed by others.

image

...there was room for improvement in the SDK to redact logging known sensitive fields from log output. This PR does exactly that: Each data class declares its own credential fields (`_SENSITIVE_FIELDS`), and redaction reads those declarations instead of guessing from key names. `dump()` is untouched, so requests still send the real values.

Covers request payloads, response bodies (e.g. `sessions.create()` hands back a usable `nonce`) and `headers` plus `str()`/`repr()` of data classes, which leaked with no logging involved at all: `print(my_client_credentials)` was enough. Only runs when a log handler will actually consume the record, so smaller overhead than before for users not using debug logging.

Redaction-related functionality has been moved out of `BasicAsyncAPIClient` and into `cognite/client/utils/_redaction.py`


cognite/client/utils/_redaction.py
```diff
@@ -0,0 +1,121 @@
...
+        if cls._SENSITIVE_TYPES:
+            for type_value in cls._SENSITIVE_TYPES:
+                by_discriminator.setdefault(type_value, set()).update(spellings)
+        else:
+            anywhere.update(spellings)
+
+    return SensitiveFields(
+        anywhere=frozenset(anywhere),
+        by_discriminator={key: frozenset(value) for key, value in by_discriminator.items()},
+    )
```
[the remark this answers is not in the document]
haakonvt: @gemini-code-assist the first two points don't apply, but the third pointed at a real gap.

- Not a race condition: no threads involved, and `functools.cache` is thread-safe.
- Not a circular dependency: the function-level import avoids one. `_base` imports `redact` from `_redaction`; hoisting `CogniteResource` to module level breaks `import cognite.client` outright.
- Incomplete registry: `sensitive_fields()` first does `import cognite.client.data_classes`, which reaches 83 of 85 modules. The two it misses declare nothing, so nothing leaks today.
- `__init_subclass__` wouldn't help here: it also only fires when a module is imported, so a class in an unimported module goes unregistered either way and we'd still need the eager package import.

The real gap is that 83/85: a future credential field in a module the package root doesn't pull in would be invisible to the registry. This is now covered by a new test that snapshots the registry in a fresh subprocess after importing only the package, then walk-imports everything and fails if anything new shows up. I've even verified this by planting a declaration in `extractor_extensions/v1.py`.

cognite/client/data_classes/hosted_extractors/sources.py
```diff
@@ -884,8 +894,13 @@ def _load_authentication(cls, resource: dict[str, Any]) -> Self:
 @dataclass
 class RESTQueryAuthenticationWrite(AuthenticationWrite):
     _type = "query"
+
+    _SENSITIVE_FIELDS: ClassVar[frozenset[str]] = frozenset({"value"})
+    # 'value' only means a credential in an object that reports this type:
```
polomani: what does this comment mean? can't decipher 😓

cognite/client/data_classes/transformations/common.py
```diff
@@ -379,19 +382,8 @@ def as_credential_provider(self) -> OAuthClientCredentials:
     def as_client_credentials(self) -> ClientCredentials:
         return ClientCredentials(client_id=self.client_id, client_secret=self.client_secret)
 
-    def dump(self, camel_case: bool = True) -> dict[str, Any]:
```
polomani, line 382: is that basically a breaking change. or?

haakonvt: No, it inherits this exact dump method

cognite/client/data_classes/hosted_extractors/sources.py
haakonvt:
```suggestion
```

cognite/client/data_classes/_base.py
```diff
@@ -89,12 +90,16 @@ class CogniteResource(ABC):
...
+    _SENSITIVE_FIELDS: ClassVar[frozenset[str]] = frozenset()
+    _SENSITIVE_TYPES: ClassVar[frozenset[str]] = frozenset()
+
     def __eq__(self, other: Any) -> bool:
         return type(self) is type(other) and self.dump() == other.dump()
 
     def __str__(self) -> str:
         item = convert_and_isoformat_time_attrs(self.dump(camel_case=False))
-        return _json.dumps(item, indent=4)
+        return _json.dumps(redact(item), indent=4)
```
trygveu, line 102: **issue: (non-blocking)** `__str__`/`__repr__` are redacted here, but the Jupyter/pandas auto-display path is not: `CogniteResource._repr_html_()` -> `to_pandas()` -> `CogniteResourceList.to_pandas()` (`_base.py:411`, `df = pd.DataFrame(self.dump(...))`) or `base_resource_to_pandas_fallback()` (`_pandas_helpers.py:576`, `dumped = resource.dump(...)`) both call raw `dump()` without `redact()`. So evaluating e.g. `client_credentials` as the last line of a notebook cell (no `print`/`str()` needed) still renders the secret in the HTML table — exactly the leak vector this PR is trying to close, per this class's own docstring about `_repr_html_` being for Jupyter display. Worth wiring `redact()` into these two call sites (or into a shared helper) before merging.

cognite/client/data_classes/transformations/common.py
```diff
@@ -322,7 +322,7 @@ def _load(cls, resource: dict[str, Any]) -> Self:
         )
 
 
-class OidcCredentials:
+class OidcCredentials(CogniteResource):
```
trygveu, line 325: **question: (non-blocking)** Making `OidcCredentials` inherit `CogniteResource` switches its `dump()` from `basic_obj_dump` (keeps `None` fields) to the inherited `basic_instance_dump` (drops any attribute that is `None`, see `_base.py:49-53`). 

Could this break anything downstream?

cognite/client/data_classes/hosted_extractors/sources.py
haakonvt: Meant to explain that the very generic sensitive field name "value" is only stripped when part of an object that we can discriminate based on the `type = "query"` field that has to co-exist.

haakonvt: ...was confusing, so I just removed it.

trygveu (commented):
Thanks for picking this up quickly. I have a question on the html repr and omitting None values.

cognite/client/data_classes/_base.py
haakonvt: @trygveu Yes, I considered this actually and decided to leave it out. The reason is that the DataFrame representation of the object is what most users actually interact with (sad but true). Thus it would be _very breaking_ to change the `to_pandas` mechanics to redact.

However, there is a "sort of middle ground" here, namly to do something only when `_repr_html_` is being called. It would mean that the DataFrame representation of the exact same underlying object would differ based on whether it was displayed directly (as a CogniteResource), or if it was displayed _after_ an explicit conversion to a DataFrame (which most users would do).

I'm not entirely sold that this is worth it, as it only covers the case when a CogntieResource(List) object is displayed directly. What's you take (since you say "blocking issue")?

cognite/client/data_classes/transformations/common.py
haakonvt: `scopes`/`audience` are typed plain string in the API spec, not nullable, so omitting them is the correct representation of "unset".

Re. tests: these two classes became `CogniteResource` subclasses in this PR, which made them part of the automatic test suite, e.g. `test_dump_load_only_required` constructs them with optionals unset and round-trips the dump, so that case is covered (12 tests across both classes).

cognite/client/data_classes/_base.py
trygveu: Ack. My thinking was that the notebooks might dominate, but on second thought, I agree with you. Downgraded to non-blocking and OK enough to ignore

cognite/client/data_classes/transformations/common.py
haakonvt: On the update path, `.source_oidc_credentials.set(...)` replaces the credentials object as a whole, so an omitted audience/scopes is identical to an explicit null here.

polomani (approved):

haakonvt:
🦄

trygveu (approved):

PR #2846, opened 2026-09-16
fix(tests): update expected error message in test_search_node_data_with_invalid_property

erlendvollset:
## Summary
- The Data Modeling search API now returns a more specific error message for unknown search properties: `Property 'x' is not available when searching view 'y'` instead of the previous `Unknown property`.
- Updated `test_search_node_data_with_invalid_property` to assert against the current error wording.

## Test plan
- [x] Ran the integration test against the live API and confirmed it passes with the updated assertion.


haakonvt (approved):
🦄

PR #2847, opened 2026-09-16
fix(files): prevent path traversal via file metadata directory in download()

erlendvollset:
## Summary

`FilesAPI.download(..., keep_directory_structure=True)` derived each file's local directory by slicing one character off the server-returned `FileMetadata.directory` (`metadata.directory[1:]`). A value with two or more leading slashes (e.g. `"//attacker-chosen/path"`) stayed absolute after the slice, so `download_root / directory` discarded the download root entirely — and `mkdir(parents=True, exist_ok=True)` created that path before the existing containment check in `_process_file_download` ever ran.

Fix, in `_get_ids_filepaths_directories`: strip *all* leading slashes, then verify the resolved directory stays inside the download root — before any directory is created.

No separate backslash handling is needed: on Windows `\` is parsed as a separator at join time, so `..\..\etc` is caught by the same `resolve()` + `is_relative_to` check; on POSIX it's an ordinary filename character and can't traverse. Same holds for drive letters and UNC paths, which replace the root on Windows and are therefore rejected. Verified across both path flavours.

## Test plan

- [x] Unit tests in `tests/tests_unit/test_api/test_files.py`: leading-slash normalization, `..`-traversal rejection, `keep_directory_structure=False` bypass, and an end-to-end `download()` asserting no directories are created for malicious metadata
- [x] `poetry run pytest tests/tests_unit/test_api/test_files.py` — 97 passed
- [x] `poetry run mypy cognite/client/_api/files.py` + `ruff check`/`format` — clean
- [x] `poetry run python scripts/run_checks.py` — clean (updates the `_sync_api/files.py` codegen hash)


cognite/client/_api/files.py
```diff
@@ -1324,6 +1324,18 @@ async def download(
...
+    def _sanitize_metadata_directory(raw_directory: str, identifier: int) -> PurePosixPath:
+        # CDF file metadata directories are always absolute, unix-style paths (e.g. "/a/b"), but that
+        # value is server-returned and must be treated as untrusted: reject parent traversal and strip
+        # any (possibly repeated) leading slashes so the result cannot turn into an absolute path when
+        # joined onto the caller-supplied download root.
+        posix_path = PurePosixPath(raw_directory)
+        if ".." in posix_path.parts:
+            raise RuntimeError(f"File with id {identifier} has an invalid directory: '{raw_directory}'")
+        relative_parts = [part for part in posix_path.parts if part not in ("/", "//")]
+        return PurePosixPath(*relative_parts)
```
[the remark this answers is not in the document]
erlendvollset: fixed.

bisera-cognite (dismissed):

bisera-cognite:
This looks good. I ran a test against both the master branch and the fix branch which confirmed the fix is in place.

cognite/client/_api/files.py
```diff
@@ -1332,13 +1349,18 @@ def _get_ids_filepaths_directories(
...
         filepaths, file_directories = [], []
+        resolved_root = directory.resolve()
         for identifier, metadata in id_to_metadata.items():
             if not isinstance(identifier, int):
                 continue
             file_directory = directory
             if metadata.directory and keep_directory_structure:
-                # CDF enforces absolute, unix-style paths (i.e. always stating with '/'). We strip to make it relative:
-                file_directory /= metadata.directory[1:]
+                file_directory = file_directory / FilesAPI._sanitize_metadata_directory(metadata.directory, identifier)
```
amorken: I was wondering about https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to and `PurePath(metadata.directory).relative_to(resolved_root, walk_up=False)` but there's more version caveats than I like in that doc so maybe doing it by hand (uh, by claude?) is simply the right thing to do.

amorken (dismissed):
🦄 
The sanitization function... It doesn't give me warm fuzzies, but then again that's probably a me problem. =)

cognite/client/_api/files.py
erlendvollset: That's better. We test against all supported Python versions and platforms, so any versioning issues should be caught. I'll try tomorrow.

cognite/client/_api/files.py
```diff
@@ -1332,13 +1332,23 @@ def _get_ids_filepaths_directories(
...
             if metadata.directory and keep_directory_structure:
-                # CDF enforces absolute, unix-style paths (i.e. always stating with '/'). We strip to make it relative:
-                file_directory /= metadata.directory[1:]
+                # CDF enforces absolute, unix-style paths (i.e. always starting with '/'). We strip all
+                # leading slashes to make it relative: an absolute right-hand side would make pathlib
+                # discard the download root entirely. The metadata is server-returned and thus untrusted,
+                # so we additionally verify that the result stays inside the download directory - this
+                # must happen before download() creates the directories.
+                file_directory = directory / metadata.directory.lstrip("/")
+                if not file_directory.resolve().is_relative_to(resolved_root):
```
amorken, line 1347: Alright, this is actually much nicer. Thanks!

amorken (approved):

erlendvollset:
🦄

PR #2848, opened 2026-09-17
fix(logging): Redact all credential-carrying headers from logs (DM-4117, VULN-113)

haakonvt:
The SDK only ever talks to Cognite's own API endpoints, and we've redacted every credential scheme Cognite itself has used there; OIDC's `Authorization` header today, and API tokens before those were fully dropped. But `ClientConfig.headers` lets callers attach arbitrary headers of their own, e.g. for a proxy or gateway sitting in front of CDF, and a credential passed that way (`api-key`, `x-service-token`, ...) was logged in plaintext at DEBUG level.

Headers are now redacted based on a case-insensitive match against common credential-related words (auth, token, secret, api-key, password, cookie, etc.), with an allowlist for known false positives (e.g. WWW-Authenticate). Also redacts headers in str(ClientConfig)/notebook repr and in successful-request response headers (e.g. Set-Cookie).


cognite/client/utils/_redaction.py
```diff
@@ -15,8 +15,31 @@
...
 REDACTED = "***"
-SENSITIVE_HEADER_NAMES = frozenset({"authorization", "proxy-authorization"})
-
+# We consider (and redact) any header that has even a partial match with any of these words. It's better
+# to redact too much than too little:
+SENSITIVE_HEADER_REGEX = re.compile(
+    r"auth|token|secret|password|passphrase|cookie|credential|signature|jwt|"
+    r"(?:api|access|private|signing|session)[-_ ]?key",
+    re.IGNORECASE,
+)
```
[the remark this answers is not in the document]
haakonvt: @gemini-code-assist please re-review

bisera-cognite (approved):
This looks good. I ran a test against the two branches(`master` and `redact-all-the-headers`) which confirmed the fix is in place.

tests/tests_unit/test_api_client.py
```diff
@@ -255,6 +256,48 @@ async def test_request_payload_only_redacted_when_debug_logging_enabled(
...
+    ) -> None:
+        # ClientConfig.headers takes arbitrary headers, and users pass credentials through it under
+        # all sorts of names. None of them should end up in a log record:
+        secret = "PLANTED-SECRET-VALUE"
+        api_client = APIClient(
+            ClientConfig(
+                client_name="any",
+                project="test-project",
+                base_url=BASE_URL,
+                headers={"api-key": secret, "x-service-token": secret},
```
audunska, line 275: Maybe add an allowlisted header here to test that it survives?

audunska (approved):
LGTM, one minor optional suggestion

tests/tests_unit/test_api_client.py
haakonvt: Pretty well tested as part of unit tests though?

haakonvt:
🦄

